PR #368, opened 2013-12-09
Hy as a Python syntax  extension

lmatthieu:
Hi,

A test, inspired by Pyxl of using hy as a syntax extension. 
It uses Python support for specifying source code encodings (PEP 263).
## How to use it

All python files with inline Hy code must have the following first line
 # coding: hy

You can mix Python and Hy code with this syntax:

```
@(defn hello [] 
     (print "hello"))
@(hello)

sum = 1 + @(+ 1 1)
```

After installing Hy, you must install the syntax extension with:

```
python setup.py hysyntax
```

This step add a file hy.pth in your python site-package directory (allows hy syntax extension to be loaded on Python startup).

A test file is provided in tests/contrib/hysyntax.

Any questions, bug reports, and other feedback on this code are welcomed !

Matthieu

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -20,3 +20,18 @@
 * Berker Peksag <berker.peksag@gmail.com>
 * Henrique Carvalho Alves <hcarvalhoalves@gmail.com>
 * Joe Hakim Rahme <joehakimrahme@gmail.com>
+* Kenan Bölükbaşı <kenanbolukbasi@gmail.com>
+* Abhishek L <abhishek.lekshmanan@gmail.com>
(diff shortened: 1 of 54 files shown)


brandonwillard:
I was just reminded of this `codecs`-based embedded syntax approach and just noticed this attempt to integrate it into Hy.  @hylang/core, is this something that's still worth considering?  Seems like it could provide a nice form of Hy-in-Python (e.g. a non-string one).

Kodiologist:
It's not something I'd invest effort in, personally, but I guess it could be cool to have. The devil's in the details.

Issue #378, opened 2013-12-16
Hy handles *some-variable* incorrectly


Coaldust:
Proper Lisp code uses `+some-variable+` for non-local constants, and `*some-variable*` for non-local mutables.  See http://google-styleguide.googlecode.com/svn/trunk/lispguide.xml#Global_variables_and_constants or any other Lisp style guide that covers the issue.

Hy converts `*some-variable*` to `SOME_VARIABLE` which is what proper Python code uses for constants.  In other words, this is exactly the opposite of the expected behavior.  This will confuse programmers trying to use Hy code from Python.

I suggest changing Hy to behave as Lisp and Python programmers would expect it, by having `*some-variable*` in Hy become `some_variable` (what proper Python code uses for global mutables; see http://google-styleguide.googlecode.com/svn/trunk/pyguide.html?showone=Naming#Naming ) in Python, and `+some-variable+` in Hy become `SOME_VARIABLE` in Python.  This would prevent unnecessary confusion and bugs.

You may also want to consider making Hy automatically do the right thing with `*some-variable*`, so Hy programmers don't have to write `(global *some-variable*)` manually.  This will not be necessary for `+some-variable+` since the global annotation in Python is only necessary for mutation, not reading (see http://docs.python.org/3/reference/simple_stmts.html#global ).  It would also enforce proper naming.

Being able to view the emitted Python code is invaluable to finding bugs like this, and to Hy programmers making the transition from Python.  I hope it will be maintained.

algernon:
See the discussion in #383: we saw no compelling reason to change the earmuffs. For Hy, earmuffs are some minor sugar, and they do not - and should not - work like in other lisps. We're not other lisps.

agentultra:
Ear muffs are just a convention in CL and don’t mean anything…

PR #383, opened 2013-12-19
Switched * with +

Willyfrog:
`*foo*` converts to foo now and `+bar+` to `BAR`
*THIS BREAKS CURRENT CODE*, you have been warned
 Also added tests for translation of `*foo*` `+bar+` and `questionable?` symbols

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -19,8 +19,11 @@ These rules help make sure code is idiomatic and interface-able in both
 languages.
 
 
-  * Symbols in earmufs will be translated to the uppercased version of that
-    string. For example, `*foo*` will become `FOO`.
+  * Symbols in earmufs will be considered globals.
+
+  * Symbols between + are considered constants and will be translated to 
+    the uppercased version of that string. For example, `+foo+` will 
+    become `FOO`.
 
   * UTF-8 entities will be encoded using
     `punycode <http://en.wikipedia.org/wiki/Punycode>`_ and prefixed with
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -18,20 +18,21 @@ this, it's going to come in handy.
 These rules help make sure code is idiomatic and interface-able in both
 languages.
 
+* UTF-8 entities will be encoded using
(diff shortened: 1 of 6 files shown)


torsten:
Hi All,

Sorry for the unsolicited feedback. I am very interested in hy and was just reading through the issues, I have a suggestion to this one:

Why not do it similarly to Clojure? To represent `*yay?` as a Java symbol they convert it to `_STAR_yay_QMARK_`, could that work for hy, too?

```clojure
user=> (defn *yay? [] (println "YAY"))
#'user/*yay?
user=> *yay?
#<user$_STAR_yay_QMARK_ user$_STAR_yay_QMARK_@48802d16>
```

Willyfrog:
That would be as valid as having an underscore in front. Not against it, although it might be harder to remember if you happen to need to use it (which you shouldn't as is private)

rcarmo:
So this means I'm going to have to do stuff like

```hy
(import [logging [getLogger +debug+ +warn+]])
```
...etc. Looks ugly, but manageable.

Still, why not mangle `*foo*` to something relatively un-Pythonic like `Foo` or `_FOO`?

algernon:
Aight, I'll merge this ~tomorrow, as is (after fixing up the conflicts).

paultag:
Did we ever figure out what existing *foo* code becomes for interop?

algernon:
`*foo*` will have the earmuffs stripped, judging by the code. It shoud likely become a global with earmuffs stripped, if I understand the docs correctly.

On the other hand, looking at this and #378, I'm not so sure we want this. We're Hy, not some-other-lisp. And there's nothing terribly wrong with the earmuffs...

rcarmo:
Well, I’m not a fan of the plus signs, FWIW...

(.log +debug+ “This is going to look somewhat weird”)

paultag:
So *foo* and foo will clash?

I'm still not sure I care enough to fix this, but if we have to, make sure
earmufs are safe from massive breakage.

algernon:
Hrm. Lets skip this, to be honest. I too, fail to see a compelling reason for the change.

I'm closing this, as wontfix.

PR #598, opened 2014-05-13
Introducing the with-redefs macro

xcthulhu:
Ported from clojure ; used for mocking out a library function during testing


paultag:
Really clever, I forgot about this. I've not had enough coffee to think this through yet, but can we get tests passing? Also, thoughts, @hylang/core ?

paultag:
Tests aren't passing still, and nothing has been updated here for a while -- mind re-opening when the tests pass, @xcthulhu ?

Issue #640, opened 2014-08-22
add type annotations


paultag:
they're really neat. let's do more of that.

vodik:
Any further thoughts on this? It be cool to be able to support [`typing.NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple) and the upcoming [data classes](https://www.python.org/dev/peps/pep-0557/)

kirbyfan64:
It would be cool but a bit pointless. You wouldn't be able to type-check any code yet.

We'd probably have to make a lightweight mypy frontend that compiles Hy files to typed_ast (which is identical to the ast module, but with type comment support)...

Kodiologist:
With syntax like `(defn f [[a int] [b int]] (print a b))`, I don't see how you'd distinguish an optional argument from a type hint.

vodik:
Okay, thought a little bit about this, read more details where Python is going, and played with some code. Apologize upfront if my terminology is off.

First is we should be able to get away with always emitting strings for annotations. With [PEP 563](https://www.python.org/dev/peps/pep-0563/) coming, this is effectively going to become python's default behaviour anyways. Tools relying on annotations should already support deferred evaluation.
 
And I think we have two options in how to implement it, one which might be easier than the other:

# Emit annotations in the AST:

This means syntax support, so that something like this:

```clojure
(def foo ^int)
(defn bar ^int [x ^int] ...)
```

Would turn into:

```python
foo: int
def bar(x: int) -> int:
    ...
```

We'd have to teach the compiler to emit `ast.AnnAsign` and set the annotation attribute on other entities like `ast.arg`. And we probably have to add another HyObject: HyAnnotation.

# Manually build `__annotations__`:

The above example could also be implemented as the following Python:

```python
def bar(x):
    ...

__annotations__ = {"foo": "int"}
bar.__annotations__ = {"return": "int", "x": "int"}
```

And now there's no need to change the compiler as we should be able to solve the problem by generating more python instead. However, this could be prove to be slower...

# Possibly do the whole thing with just macros

I think I'd much prefer inline annotations, but its also possible we might be able to implement with macros. Basing this off cojure's core.typed, for example:

```clojure
(ann bar [ int -> int ])
(defn bar [x] ...)
```

But the problem (and I don't know if this can be done cleanly), needs to attach the annotation after the definition. It also doesn't solve annotated assignment (`foo: int`). We could use `(ann foo int)` for that, but could that mean a function `foo` with no arguments?

[...]

PR #656, opened 2014-09-09
Type annotations for fn/lambda

Foxboron:
Beta stage.
=======
Reff #640  

 `optional` and `kwargs` does not work as syntax isn't really decided upon, and we can't annotate return values as it needs macro changes.
Current approach checks for a `HyList` as given var and destructs this. Passing the annotations through `self.compile` pulling out `_expr` from the `Result()` object. This way the types are rather "dynamic" when passed along.

```hy
=> (defn a [[y int]] (print "lol"))
=> a.--annotations--
{'y': <class 'int'>}

=> (defn a [[y "HEYHO LETS GO!"]] (print "lol"))
=> a.--annotations--
{'y': 'HEYHO LETS GO!'}

=> (defn a [[y (lambda [[x "Just wanna get hy"]] (print "Heyho!"))]] (print "lol"))
=> a.--annotations--
{'y': <function <lambda> at 0x7f4939167598>}
```

The current(?) syntax i imagen we could use is (adding with comment below, ninja edit)

```hy
(defn foo [[x int] &optional [[a str] "Hey"]] -> int
  (+ 3 4))
```

CC: @hylang/core 
Comments, code review, rate and hate!

:shipit:

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1894,14 +1894,21 @@ class HyASTCompiler(object):
         arglist = expression.pop(0)
         ret, args, defaults, stararg, kwargs = self._parse_lambda_list(arglist)
 
+
         if PY34:
             # Python 3.4+ requires that args are an ast.arg object, rather
             # than an ast.Name or bare string.
-            args = [ast.arg(arg=ast_str(x),
-                            annotation=None,  # Fix me!
-                            lineno=x.start_line,
-                            col_offset=x.start_column) for x in args]
-
+            ret_args = []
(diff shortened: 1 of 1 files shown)


Tritlo:
+1, I really like the type annotation syntax, with the "ann" decorator.

Foxboron:
@Tritlo The lib is somewhat functional for use :P

paultag:
I'm going to close this until @Foxboron is done prototyping this stuff :)

:+1:

Issue #246, opened 2013-07-13
Implement nonlocal for py3


paultag:
Grumble

Foxboron:
This is solved now \o/

PR #719, opened 2014-12-14
Added `nonlocal`. `global` takes multiple args.

unmerged:
Related to https://github.com/hylang/hy/issues/246

### DIFF
--- a/tests/compilers/test_ast.py
+++ b/tests/compilers/test_ast.py
@@ -220,6 +220,18 @@ def test_ast_bad_global():
     cant_compile("(global (foo))")
 
 
+if PY3:
+    def test_ast_good_nonlocal():
+        "Make sure AST can compile valid nonlocal"
+        can_compile("(nonlocal a)")
+        can_compile("(nonlocal foo bar)")
+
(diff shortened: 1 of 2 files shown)


tests/compilers/test_ast.py
```diff
@@ -211,12 +211,25 @@ def test_ast_bad_assert():
...
 
 
 def test_ast_bad_global():
     "Make sure AST can't compile invalid global"
     cant_compile("(global)")
-    cant_compile("(global foo bar)")
+    cant_compile("(global (foo))")
+
+
+if PY3:
```
paultag: so skipping won't work great if we raise a SyntaxError -- which is why we have the if-python2 macro

farhaven:
The tests seem to run fine here with both Python 2.7 and 3.4. This is based on a git HEAD from today with cfa805c and 41e5175 cherry-picked in.

Foxboron:
All tests are passing with this patch. LGTM.
Then we could also get the #803 merged
:shipit:

paultag:
LGTM; this is awesome! Nice work, @unmerged !

Issue #744, opened 2015-01-03
Move hylang/hy-style-guide to docs.hylang.org


berkerpeksag:
I saw a couple of questions about writing idiomatic Hy code on #hy. So it would be great to move hylang/hy-style-guide to our main repo to make it more visible and up-to-date.

iKevinY:
I agree with migrating the style guide to the main Hy documentation; it seems a little strange to have a separate repository just for the style guide. I can work on this migration if no one has already started doing so.

berkerpeksag:
@iKevinY please go ahead and a PR :)

iKevinY:
Done! Having strange issues with some code block formatting, though.

paultag:
:+1: thanks! :dancer:

theanalyst:
Thanks @iKevinY :)

PR #745, opened 2015-01-03
Migrate hylang/hy-style-guide to Hy Documentation

iKevinY:
Here's what I have so far to address #744. I'm running into a strange issue where lines in code blocks that *should* line up don't, as though some half-width character shenanigans are going on. Does anyone know if this is a Sphinx issue or if I just formatted the code blocks incorrectly? It seems hypocritical to assert that `let` blocks should be vertically aligned, only to have them shown out of alignment a few pixels below that very sentence.

I'll squash everything down once the PR looks good to merge.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -32,6 +32,7 @@ Contents:
 
    quickstart
    tutorial
+   style-guide
    language/index
    contrib/index
    hacking
--- a/docs/style-guide.rst
+++ b/docs/style-guide.rst
@@ -11,11 +11,12 @@ The Hy style guide intends to be a set of ground rules for the Hyve
 to write idiomatic Hy code. Hy derives a lot from Clojure & Common
 Lisp, while always maintaining Python interopability.
 
+
 Prelude
--------
+=======
 
 The Tao of Hy
-~~~~~~~~~~~~~
+-------------
 
 ::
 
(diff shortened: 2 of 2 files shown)


berkerpeksag:
Thanks for the PR!

> Does anyone know if this is a Sphinx issue or if I just formatted the code blocks incorrectly?

It's a docutils/reST feature. Whitespaces are important in reST. So, the markup should be like the following example:

```rst
+ Vertically align ``let`` blocks.

  .. code-block:: clj

     (let [[foo (bar)]
           [qux (baz)]]
        (foo qux))
```

iKevinY:
Oh whoops, I forgot to put the code blocks on the same indentation level as the bulleted list items – changing that right now. I was more concerned with the fact that the square brackets in the lines 1 and 2 of the above code block don't line up properly, though.

paultag:
Hey, looks great! Thanks, @iKevinY and @berkerpeksag !

I love the Harry Potter quote! :D Thanks! You rock!

iKevinY:
All credit for the style guide content goes to @theanalyst!

iKevinY:
Well, I did a little digging and it seems the alignment issue stems from the fact that the "let" is bolded, and the bold text is wider than the normal text. Using Web Inspector to manually set the font weight of "let" to normal causes the lines to align properly. Not sure if there's anything we can do about this.

PR #751, opened 2015-01-14
Fix typos in tutorial docs

shrayasr:
1. It was mistyped as lisP comprehensions instead of lisT comprehensions (Unless it was intended as a pun :wink:)

> ~~2. Function was defined as `optional-arg` but used as `optional_arg`~~

Reverted this change after discussions with @theanalyst and @berkerpeksag

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -337,7 +337,7 @@ which is equivalent to::
   with open("/tmp/data.in") as f:
       print f.read()
 
-And yes, we do have Lisp comprehensions!  In Python you might do::
+And yes, we do have List comprehensions!  In Python you might do::
 
   odds_squared = [
     pow(num, 2)


docs/tutorial.rst
```diff
@@ -413,7 +413,7 @@ Are you familiar with passing in `*args` and `**kwargs` in Python?::
   >>> args = [1 2]
   >>> kwargs = {"keyword2": 3
   ...           "keyword1": 4}
-  >>> optional_arg(*args, **kwargs)
+  >>> optional-arg(*args, **kwargs)
```
theanalyst: python doesn't support optional-arg?
 :)

theanalyst:
the first one list/p is an error to be fixed .. the second one, not so sure

shrayasr:
The second isn't really an **error**. In the before section, the function is defined as `optional-arg` and then later it is used as `optional_arg`. I changed it to the former for consistency sake :)

docs/tutorial.rst
shrayasr: Ah. Then I think it should be `optional_arg` everywhere since Hy supports it, no?

theanalyst: It does, but then dashes turn to underscores anyway.. so both are equivalent.. probably you can leave this as it is for now

shrayasr:
Since Python doesn't support using of `-` in function names, does it make sense to change the example to use `_` in the place of python for consistency sakes?

docs/tutorial.rst
shrayasr: Ah. I see. 

Allright, I'll remove the commit then ?

berkerpeksag: Yes, please revert it and squash your commits. Otherwise LGTM.

shrayasr: All right. Will do.

One more thing, shouldn't it be **Pro tips** and not **Protips** ?

berkerpeksag: No idea :) I'm not a native speaker. Ping @paultag @cwebber.

shrayasr: Allright, thanks :)

shrayasr:
@paultag @cwebber, Shouldn't the last section in the Tutorial be **Pro tips** instead of **Protips** ? cc @theanalyst

berkerpeksag:
Thanks!

shrayasr:
No problem :) Should I add myself to the AUTHORS file? It says so in CONTRIBUTING but doesn't specify if docs changes qualify to add myself there

berkerpeksag:
Oh, sorry, I will add you in a minute :)

shrayasr:
No problem, I will send a PR for that as instructed :smile:

PR #755, opened 2015-01-14
Adapt attribute completion from IPython completer 

icholy:
Recreating #754 because I can't reopen it after squashing.

PR #756, opened 2015-01-14
Fix issues with tab completion.

icholy:
Fixes issue introduced in #755

- [x] freezing
- [x] `defmacro`/`defreader` issue
- [x] inconsistent underscore matching/conversion
- [x] complete macros
- [x] unicode

### DIFF
--- a/hy/completer.py
+++ b/hy/completer.py
@@ -29,6 +29,7 @@ import os
 import re
 import sys
 from contextlib import contextmanager
+from hy._compat import string_types
 
 docomplete = True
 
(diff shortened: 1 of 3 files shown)


hy/completer.py
```diff
@@ -60,8 +60,11 @@
 
 class Completer(object):
     def __init__(self, namespace=None):
-        if namespace and not isinstance(namespace, dict):
```
icholy: A dictionary has a falsy value when it's empty. So we have to check with `is not None`.

hy/completer.py
```diff
@@ -92,16 +95,14 @@ def attr_matches(self, text):
         return matches
 
     def global_matches(self, text):
-        path = PATH
-        if self.namespace:
-            path.append(self.namespace)
```
icholy: This was adding the namespace to the global PATH variable on every completion. 
However, this code never used to be run because the namespace was `None`.

hy/completer.py
icholy: We could also just have the default value for `namespace` be `{}`

icholy:
botched the squash ...

icholy:
@berkerpeksag @Foxboron done. Pls review.

icholy:
Can this get merged?
Seeing my broken code in master is giving me an OCD attack.

Foxboron:
The code looks sane. But i'll pull this and play with it a little before approving it..

paultag:
Let's merge bugfix code with low regression potential quicker :)

hy/completer.py
```diff
@@ -83,29 +87,37 @@ def attr_matches(self, text):
...
+        matches = []
+        for p in self.path:
+            for k in p.keys():
+                k = str(k).replace("_", "-")
```
berkerpeksag: str -> six.text_type. Did you test it with non-ASCII names? https://github.com/algernon/adderall/ has a couple of examples like `flattenrevᵒ`.

Foxboron: Completing unicode names might need some work aswell?

icholy: Unicode completion works for python 3.

Foxboron: What about Python 2 :)?

icholy: When using python 2, matches that contain non-ascii characters get ignored by readline.

hy/completer.py
```diff
@@ -27,6 +27,7 @@
 
 import os
 import re
+import six
```
icholy: Where should I add this dependency?

hy/completer.py
```diff
@@ -83,29 +88,38 @@ def attr_matches(self, text):
...
         return matches
 
     def global_matches(self, text):
-        path = PATH
-        if self.namespace:
-            path.append(self.namespace)
+        matches = []
+        for p in self.path:
+            for k in p.keys():
+                if isinstance(k, str) or isinstance(k, six.text_type):
```
berkerpeksag: If you want to check for both str and unicode and to be commpatible with Python 2 and Python 3, you can use `isinstance(k, six.string_types)`.

berkerpeksag: Oh, by the way, we don't use six :)

berkerpeksag: You can add the following snippet to `hy._compat.py`:

```py
if PY2:
    string_types = basestring,
else:
   string_types = str,
```

hy/completer.py
berkerpeksag: We don't need to add six in my opinion. Just add the following snippet to `hy/_compat.py` and import it `from hy._compat import string_types`.

```py
if PY2:
    string_types = basestring,
else:
    string_types = str,
```

berkerpeksag:
LGTM. Could you please squash the commits? Thanks!

PR #760, opened 2015-01-17
Added flow macros

Tritlo:
I've added flow macros for Hy, both the switch macro, and the guard macro.

### DIFF
--- a/docs/contrib/flow.rst
+++ b/docs/contrib/flow.rst
@@ -2,7 +2,7 @@
 Flow
 ==========
 
-.. versionadded:: 0.10.2
+.. versionadded:: 0.10.1
 
 The ``flow`` macros allow a programmer to direct the flow of his program with
 greater ease.
@@ -11,16 +11,16 @@ greater ease.
 Macros
 ======
 
-.. _guard:
+.. _case:
 .. _switch:
 
-guard
+case
 -----
 
-``guard`` allows you to guard against a condition.
+``case`` allows you to decide based on the value of a variable.
 
 
-Usage: `(guard (cond1) (body1) (cond2) (body2) ...)`
+Usage: `(case variable val1 (body1) val2 (body2) ...)`
 
 Example:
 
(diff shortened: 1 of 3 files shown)


Foxboron:
whats the difference between switch and cond?

Tritlo:
Switch binds to a variable like:

     (switch a
          (= 3 ) (print " a was three")
                    (print "a was not three"))

Though guard is almost exactly like cond, except for the use of [] for each cond, code pair, i.e.

  (cond [(< value 5) (print "value is smaller than 5")]
       [(= value 5) (print "value is equal to 5")]
       [(> value 5) (print "value is greater than 5")]
      [True (print "value is something that it should not be")])

would be

      (guard (< value 5) (print "value is smaller than 5")
                 (= value 5) (print "value is equal to 5")
                 (> value 5) (print "value is greater than 5")
                 True (print "value is something that it should not be"))

kirbyfan64:
IMO, couldn't the macros be in the rST file with all the other macros?

nicolas-p:
Maybe we could have `cond`, `swich` and `case` which all use square brakets to delimit each condition-expression pair, and `cond'`, `swich'` and `case'` in which there are no square brakets and the arguments are implicitely grouped two-by-two.

Square brakets are not necessary in simple examples but when the conditions or expression take several lines, it is good to have them so you can identify when the pairs start and end.

I think both versions are useful.

docs/contrib/flow.rst
```diff
@@ -0,0 +1,60 @@
+==========
+Flow
+==========
+
+.. versionadded:: 0.10.1
+
+The ``flow`` macros allow a programmer to direct the flow of his program with
```
paultag: can we make the `his` gender neuter? :)

paultag:
Very interesting PR series - I like it! LGTM!

paultag:
@hylang/core RFR

Foxboron:
Yepz, LGTM.
:shipit:

PR #761, opened 2015-01-24
Add 'replace_hy_obj' to safely replace Hy objects

larme:
Currently '.replace' method is used to replace hy objects. This is not
safe when we are not sure if the 'obj' in 'obj.replace(other)' is an
instance of HyObject.

In these cases, we can use function 'replace_hy_obj(obj, other)'
instead. This function will try to wrap 'obj' if it's not an instance of
HyObject.

This also means that we need a wrapping function in hy.models'. Hence I
moved the '_wrap_value' function from hy.macros into hy.models. To avoid
circular importing, the wrapper functions are provided individually by
each model type's own file.

### DIFF
--- a/hy/models/__init__.py
+++ b/hy/models/__init__.py
@@ -65,5 +65,5 @@ def replace_hy_obj(obj, other):
     if isinstance(wrapped_obj, HyObject):
         return wrapped_obj.replace(other)
     else:
-        raise TypeError("Can't replace a Hy object "
-                        "with a non wrappable non Hy object")
+        raise TypeError("Don't know how to wrap a %s object to a HyObject"
+                        % (type obj))
--- a/hy/models/__init__.py
+++ b/hy/models/__init__.py
@@ -66,4 +66,4 @@ def replace_hy_obj(obj, other):
         return wrapped_obj.replace(other)
     else:
         raise TypeError("Don't know how to wrap a %s object to a HyObject"
-                        % (type obj))
+                        % type(obj))
--- a/hy/models/__init__.py
+++ b/hy/models/__init__.py
@@ -38,6 +38,7 @@ class HyObject(object):
 
 _wrappers = {}
 
+
 def wrap_value(x):
     """Wrap `x` into the corresponding Hy type.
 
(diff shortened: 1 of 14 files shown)


hy/models/complex.py
```diff
@@ -30,3 +30,5 @@ class HyComplex(HyObject, complex):
     def __new__(cls, number, *args, **kwargs):
         number = complex(number)
         return super(HyComplex, cls).__new__(cls, number)
+
+_wrappers[complex] = HyComplex
```
berkerpeksag: Can we move all these `_wrapper[foo] = ...` stuff into `hy/models/__init__.py`?

larme: We can, but it will create some circular imports because then `hy/models/__init__.py` need to `import hy/models/complex.py` etc.. I'm not sure which way is better.

hy/models/__init__.py
```diff
@@ -35,3 +35,35 @@ def replace(self, other):
...
+
+    if isinstance(obj, HyObject):
+        return obj.replace(other)
+
+    wrapped_obj = wrap_value(obj)
+
+    if isinstance(wrapped_obj, HyObject):
+        return wrapped_obj.replace(other)
+    else:
+        raise TypeError("Can't replace a Hy object "
```
larme: This error message doesn't clearly state what's wrong. Maybe we should change it to something like

> "Can't wrap a <type xxx> object to a HyObject"

paultag:
This approach is just so much better.

RFR @hylang/core - I'll give it a review in a sec

hy/models/__init__.py
paultag: Yeah, might be nice, since this will be sent to the user once in a while :)

hy/models/complex.py
paultag: Yeah, I'm OK leaving them in the modules. @berkerpeksag any other thoughts on this?

paultag:
Yep, LGTM. This will save a bunch of stupid bugs from happening. Really great work, @larme

hy/models/complex.py
berkerpeksag: > Yeah, I'm OK leaving them in the modules.

+1 :)

Foxboron:
Looks good to me. :shipit:

PR #777, opened 2015-03-12
Add shadow functions for comparison operators

### LINKED ISSUE #775: Using < and > as first-class functions
Hello. While playing with Hy I found that I could do :
```hy
=> (list (map + [4 7 5 9] [1 9 4 2]))
[5, 16, 9, 11]
```
but not :
```hy
=> (list (map < [4 7 5 9] [1 9 4 2]))
NameError: name '<' is not defined
```
I was expecting :
```hy
[False, True, False, False]
```

hy 0.10.1 using CPython(default) 3.4.2 on Darwin

tianon:
Fixes #775

### DIFF
--- a/hy/core/shadow.hy
+++ b/hy/core/shadow.hy
@@ -61,4 +61,33 @@
         (reduce operator.truediv args)))))
 
 
-(setv *exports* ['+ '- '* '/])
+(defn comp-op [op args]
+  "Helper for shadow comparison operators"
+  (if (< (len args) 2)
+    (raise (TypeError "Need at least 2 arguments to compare"))
+    (reduce operator.and_
+            (list-comp (op x y)
+                       [(, x y) (zip args (slice args 1))]))))
+(defn < [&rest args]
+  "Shadow < operator for when we need to import / map it against something"
+  (comp-op operator.lt args))
+(defn <= [&rest args]
+  "Shadow <= operator for when we need to import / map it against something"
+  (comp-op operator.le args))
+(defn = [&rest args]
+  "Shadow = operator for when we need to import / map it against something"
+  (comp-op operator.eq args))
+(defn != [&rest args]
+  "Shadow != operator for when we need to import / map it against something"
+  (comp-op operator.ne args))
+(defn >= [&rest args]
+  "Shadow >= operator for when we need to import / map it against something"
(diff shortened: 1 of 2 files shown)


tianon:
This wasn't as easy as I first thought because `(1 < 2) < 3` (which is how the other shadow operator functions operate on their arguments) is _very_ different from `1 < 2 < 3` (which is really parsed as `1 < 2 and 2 < 3`).

hy/core/shadow.hy
```diff
@@ -61,4 +61,32 @@
...
+(defn != [&rest args]
+  "Shadow != operator for when we need to import / map it against something"
+  (comp-op operator.ne args))
+(defn >= [&rest args]
+  "Shadow >= operator for when we need to import / map it against something"
+  (comp-op operator.ge args))
+(defn > [&rest args]
+  "Shadow > operator for when we need to import / map it against something"
+  (comp-op operator.gt args))
+; TODO figure out a way to shadow "is" (and then shadow "is_not" too)
```
tianon: I suppose this would apply to `and` and `or` as well, since I don't believe `and` can be shadowed either.

tests/native_tests/shadow.hy
```diff
@@ -50,3 +50,24 @@
...
+              (catch [TypeError] True)
+              (else (throw AssertionError)))))
+  (for [(, x y) [[< >=]
+                 [<= >]
+                 [= !=]]]
+    (for [args [[1 2]
+                [2 1]
+                [1 1]
+                [2 2]]]
+      (assert (= (apply x args) (not (apply y args)))))))
```
tianon: This should probably include some 3+ arg tests too. :disappointed:

hy/core/shadow.hy
agentultra: They could be, but not the keyword version… you’d need to import operators.and_

tianon: Yeah, but then it wouldn't really be a "shadow" as much as just a handy way to use `and_` hy-style.

tianon:
Updated with a better comment about things we can't shadow ATM and with more appropriately comprehensive tests. :+1:

paultag:
LGTM! RFR @hylang/core

berkerpeksag:
Thanks! :)

PR #781, opened 2015-03-16
keyword-only arguments

zackmdavis:
I'll be honest; I'm kind of ashamed that this patch for #453 is the result of crude reasoning-by-analogy to how `&optional` is implemented and experimental observations tinkering with `inspect.getfullargspec` and `compile( ... , ast.PyCF_ONLY_AST)`, rather than a deep, crystal-clear understanding of the theory and implementation of Hy and Python's AST.

On the other hand, it seems to actually work&mdash;
```
(hy) zmd@ExpectedReturn:~/Code/hy/hy$ hy --spy
hy 0.10.1 using CPython(default) 3.4.0 on Linux
=> (defn example [a &kwonly b [c nil]]
...   (print a b c))
def example(a, *, b, c=None):
    return print(a, b, c)
=> (example 1 :b 2 :c 3)
example(1, b=2, c=3)
1 2 3
=> (example 1 :b 2)
example(1, b=2)
1 2 None
=> (example 1)
example(1)
TypeError: example() missing 1 required keyword-only argument: 'b'
```
What do you think??

(Tests pass locally on Python 3.4.0; I'll leave it to CI to give a verdict for 3.3. Python 2 lacks the underlying PEP 3102 functionality and isn't supported at time of writing, but if desired, maybe we could treat `&kwonly` args as `**kwargs` in that case and raise a [warning](https://docs.python.org/2/library/warnings.html) that the keyword-only-ness won't be enforced under Python 2?)

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -441,6 +441,41 @@ Parameters may have the following keywords in front of them:
         => (zig-zag-sum 1 2 3 4 5 6)
         -3
 
+&kwonly
+    .. versionadded:: 0.12.0
+
+    Parameters that can only be called as keywords. Mandatory
+    keyword-only arguments are declared with the argument's name;
+    optional keyword-only arguments are declared as a two-element list
+    containing the argument name followed by the default value (as
+    with `&optional` above).
+
+    .. code-block:: clj
+
+        => (defn compare [a b &kwonly keyfn [reverse false]]
+        ...  (let [[result (keyfn a b)]]
+        ...    (if (not reverse)
+        ...      result
+        ...      (- result))))
+        => (apply compare ["lisp" "python"]
+        ...        {"keyfn" (fn [x y]
+        ...                   (reduce - (map (fn [s] (ord (first s))) [x y])))})
+        -4
+        => (apply compare ["lisp" "python"]
+        ...        {"keyfn" (fn [x y]
+        ...                   (reduce - (map (fn [s] (ord (first s))) [x y])))
+        ...         "reverse" true})
+        4
+
+    .. code-block:: python
+
+        => (compare "lisp" "python")
+        Traceback (most recent call last):
+          File "<input>", line 1, in <module>
+        TypeError: compare() missing 1 required keyword-only argument: 'keyfn'
+
+    Availability: Python 3.
+
 .. _defn-alias / defun-alias:
 
 defn-alias / defun-alias
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -485,11 +485,13 @@ class HyASTCompiler(object):
 
     def _parse_lambda_list(self, exprs):
         """ Return FunctionDef parameter values from lambda list."""
-        ll_keywords = ("&rest", "&optional", "&key", "&kwargs")
+        ll_keywords = ("&rest", "&optional", "&key", "&kwonly", "&kwargs")
         ret = Result()
         args = []
         defaults = []
         varargs = None
+        kwonlyargs = []
+        kwonlydefaults = []
         kwargs = None
         lambda_keyword = None
 
@@ -505,6 +507,8 @@ class HyASTCompiler(object):
                                           "arguments or one &key argument")
                     lambda_keyword = expr
                 elif expr == "&key":
+                    lambda_keyword = expr
+                elif expr == "&kwonly":
                     lambda_keyword = expr
                 elif expr == "&kwargs":
                     lambda_keyword = expr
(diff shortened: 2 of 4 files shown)


hy/compiler.py
```diff
@@ -544,6 +548,24 @@ def _parse_lambda_list(self, exprs):
                 args.append(k)
                 ret += self.compile(v)
                 defaults.append(ret.force_expr)
+            elif lambda_keyword == "&kwonly":
+                if not PY3:
+                    raise HyTypeError(expr,
+                                      "keyword-only arguments are only "
+                                      "available under Python 3")
+                if isinstance(expr, HyList):
+                    if not len(expr) == 2:
```
berkerpeksag: `if len(expr) != 2:` looks better to me.

docs/language/api.rst
```diff
@@ -436,6 +436,38 @@ Parameters may have the following keywords in front of them:
...
+        ...        {"keyfn" (fn [x y]
+        ...                   (reduce - (map (fn [s] (ord (first s))) [x y])))})
+        -4
+        => (apply compare ["lisp" "python"]
+        ...        {"keyfn" (fn [x y]
+        ...                   (reduce - (map (fn [s] (ord (first s))) [x y])))
+        ...         "reverse" true})
+        4
+
+    .. code-block:: python
```
berkerpeksag: clj

docs/language/api.rst
```diff
@@ -436,6 +436,38 @@ Parameters may have the following keywords in front of them:
         => (zig-zag-sum 1 2 3 4 5 6)
         -3
 
+&kwonly
+    *(Python 3 only.)* Parameters that can only be called as
```
berkerpeksag: We could turn this to a "Availability: Python 3" note like https://docs.python.org/3/library/os.html#os.ctermid

hy/compiler.py
```diff
@@ -544,6 +548,24 @@ def _parse_lambda_list(self, exprs):
                 args.append(k)
                 ret += self.compile(v)
                 defaults.append(ret.force_expr)
+            elif lambda_keyword == "&kwonly":
+                if not PY3:
+                    raise HyTypeError(expr,
```
berkerpeksag: Could you add a Python 2 only test to test this behavior?

berkerpeksag:
Nice to have: Updating tutorial for keyword-only only args would be great, but that can go in a new PR of course.

docs/language/api.rst
```diff
@@ -436,6 +436,38 @@ Parameters may have the following keywords in front of them:
...
+        ...         "reverse" true})
+        4
+
+    .. code-block:: python
+
+        => (compare "lisp" "python")
+        Traceback (most recent call last):
+          File "<input>", line 1, in <module>
+        TypeError: compare() missing 1 required keyword-only argument: 'keyfn'
+
```
berkerpeksag: also needs `.. versionadded:: 0.10.2` or 0.11 I'm not sure :) ping @paultag @Foxboron

zackmdavis:
Fixed up and force-pushed in response to the inequality-notation and availability-note comments. (I'm imagining that we care more about clean commit history than immutable history given that almost certainly no one has based work on this branch.)

docs/language/api.rst
zackmdavis: This was actually intentional; I wanted to show the exception when the mandatory keyword argument was omitted, but Sphinx/Pygments didn't do any syntax highlighting when it was all part of the same Clojure-style block (probably because it got confused by the Python stack trace), whereas if just the part with the stack trace is in a separate Python block, I think it looks OK&mdash;
riddle_of_the_sphinx_syntax_highlighting

hy/compiler.py
zackmdavis: Yes, but it'll have to be a little later after I find more time to think about how to do it properly. I had originally tried something like this (in native_tests/language.hy)&mdash;
``` hy
(defn test-kwonly []
  "NATIVE: test keyword-only arguments"
  (if PY3
    (do
      ;; [redacted for brevity]
    )
    ;; Python 2
    (let [[attempt-to-use-kwonly
           (try
            (setv this-will-fail (fn [&kwonly foo] (inc foo)))
            (catch [e [Exception]] e))]
          [expected-exception-message
           "keyword-only arguments are only available under Python 3"]]
      (assert (isinstance attempt-to-use-kwonly HyTypeError)
              (.format "expected a HyTypeError, got {}"
                       attempt-to-use-kwonly))
      (assert (= (. attempt-to-use-kwonly args [0])
                 expected-exception-message)
              (.format "expected HyTypeError to have message \"{}\""
                       expected-exception-message)))))
```
but then the entire test run would die on the TypeError ("Ran 1 test in 13.094s / FAILED (errors=1)") instead of catching it like I had intended.

zackmdavis: Added an AST test in 9de19504 which covers the Python 2 behavior.

docs/language/api.rst
paultag: should be clear-er now :)

paultag:
What's the status here? Let's get this guy ready! :D

zackmdavis:
@paultag I've rebased this branch off master and added a `.. versionadded:: 0.12.0` directive to the documentation (imagining, [not entirely without evidence](https://github.com/hylang/hy/pull/652#issuecomment-123795945), that that's what the next release will be).

berkerpeksag:
Looks great to me, thanks!

algernon:
LGTM :+1:

PR #782, opened 2015-03-18
Added (last) function to core language

acron0:
Hi,

This is my first PR (hopefully of many). I'm going to be working with Hy in production over the next few months and so will be hopefully be able to feed back any features I work on.

This particular addition is trivially small, but a chance to say 'hello' :cake:

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -61,3 +61,4 @@
 * Shenyang Zhao <dev@zsy.im>
 * Zack M. Davis <code@zackmdavis.net>
 * Adrià Garriga Alonso <adria@monkingme.com>
+* Antony Woods <antony@teamwoods.org>
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -944,7 +944,19 @@ This can be confirmed via Python's built-in ``help`` function::
     times_three(x)
     Multiplies input by three and returns result
     (END)
+	
+last
+-----------
 
+.. versionadded:: 0.10.2
+
+``last`` can be used for accessing the last element of a collection:
+
+.. code-block:: clj
+
+    => (last [2 4 6])
+    6
+	
 
 let
 ---
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -245,7 +245,7 @@
 
 (defn last [coll]
   "Return last item from `coll`"
-  (nth coll (- (len coll) 1)))
+  (get (list coll) -1))
 
 (defn list* [hd &rest tl]
   "Return a dotted list construed from the elements of the argument"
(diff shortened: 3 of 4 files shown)


microamp:
Hello, thanks for the contribution! Hy in production for a few months?

berkerpeksag:
LGTM. Could you add documentation and add your name to AUTHORS?

microamp:
It's because ``len`` would only work with strict collections (e.g. ``list``).

tuturto:
Using ```last``` with generator will consume all the generated values, with no way getting them back. Sounds quite rare case to me when it would be useful. But on the other hand, it would be nice if collections and generators/iterators would work in a similar way.

So, I'm leaning towards that ```last``` should work with iterators and generators too. What are other opinions about this?

Willyfrog:
@tuturto you could `tee` the iterator and consume the duplicated one

microamp:
@tuturto @Willyfrog

That's similar to the approach I took when implementing [`drop-last`](https://github.com/microamp/hy/commit/7f5c8e39d8d34acabe476d35f7a2ef71904b5571). Also, I just remembered that the previous version of `butlast`(fixed [here](https://github.com/hylang/hy/commit/23f31d4ac1ac72ca13005bea218b9c2363fa940e)) wasn't working with generators/iterators because of the `len` issue mentioned above.

Your thought on this, @paultag?

acron0:
Proposed alternative: https://www.refheap.com/98654

Tested with:
`=> (last (take 5 (itertools.count 10)))`
`14L`

nicolas-p:
I recently wrote some code for which I needed `first` and `last`. I had to import `last` from Toolz. So I think it would be a useful addition to Hy.

microamp:
@acron0 

My apologies on getting back to you late.

I'd go with the second implementation (the one with ``(list coll)``). I realised that ``itertools.tee`` is irrelevant here since it doesn't stop the original iterable from being exhausted anyway. It is the case with ``butlast`` too.

Mind doing another commit for this when you can?

By the way, you can do something like ``(get aslist -1)`` (think of it as Python's ``aslist[-1]``) instead of having to reverse the collection (then get the first item).

acron0:
Changes are committed. It's obscenely simple now :P

microamp:
+1 from me

We'll wait for one more vote to get it merged since the implementation is now a bit different from what @berkerpeksag signed off previously.

berkerpeksag:
LGTM.

berkerpeksag:
Thanks @acron0 :)

acron0:
Thank you folks! Great to see such a positive, engaging team.

paultag:
Nice work, y'all!

And, thanks Antony!

On Wed, Apr 8, 2015 at 3:02 AM, Antony <notifications@github.com> wrote:

> Thank you folks! Great to see such a positive, engaging team.
>
> —
> <https://github.com/hylang/hy/pull/782#issuecomment-90824741>.
>



-- 
:wq

PR #785, opened 2015-03-23
Added new macros to `hy/contrib/meth.hy` that allow routes to be decorated.

ntrifunovic:
Added new macros to `hy/contrib/meth.hy` that allow routes to be decorated.

**Usage Examples:**

Using [Flask-HTTTPAuth](https://github.com/miguelgrinberg/Flask-HTTPAuth)
    
    (import [flask_httpauth [HTTPBasicAuth]])
    
    (def auth (HTTPBasicAuth))
    
    (route-decorated index [(. auth login_required)] "/" []
                     "Hello world!")

### DIFF
--- a/hy/contrib/meth.hy
+++ b/hy/contrib/meth.hy
@@ -1,16 +1,36 @@
 ;;; Hy on Meth
 ;;; based on paultag's meth library to access a Flask based application
 
-(defmacro route-with-methods [name path methods params &rest code]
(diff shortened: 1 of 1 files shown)


berkerpeksag:
Thanks for the patch! I'm not sure about the API. Most of the time you will decorated your views in Flask. It would be better to add a decorator option to the existing `*-route` macros.

paultag:
Looks like this is a bit out of date with master, and no update for a while. Closing this until it's ready - re-open it once we're ready again!

PR #793, opened 2015-04-07
Add argument destructuring

### LINKED ISSUE #792: (feature request) function argument destructuring
In clojure, I can do
```clojure
(defn awe [[x y]] (+ x y))
(awe [2 3]) ; => 5
```
But not in hy :( destructuring within a `let` works just fine though.

Why is this useful? A larger example
```clojure
(map
  (fn [[x y]] (, (int x) y)))
  (.items my-dictionary))
```

Which otherwise needs to be one of
```clojure
(map
  (fn [args] (let [[[x y] args]] (, (int x) y))))
  (.items my-dictionary))
```
or
```clojure
(map
  (fn [args] (, (int (. args [0]) (. args[1]))))
  (.items my-dictionary))
```

kirbyfan64:
Fixes #792. I feel like this was too easy and that I did something wrong...but it *seems* fine.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1934,6 +1934,19 @@ class HyASTCompiler(object):
             raise HyTypeError(expression,
                               "First argument to (fn) must be a list")
         ret, args, defaults, stararg, kwargs = self._parse_lambda_list(arglist)
+        for i, arg in enumerate(args):
+            if isinstance(arg, HyList):
+                # Destructuring argument
+                if not arg:
+                    raise HyTypeError(arglist,
+                                      "Cannot destruct empty list")
+                args[i] = var = HySymbol(self.get_anon_var())
+                expression = HyExpression([
+                    HyExpression([
+                        HyString("setv"), arg, var
+                    ])]
+                ) + expression
+                expression = expression.replace(arg[0])
 
         if PY34:
             # Python 3.4+ requires that args are an ast.arg object, rather
--- a/tests/compilers/test_ast.py
+++ b/tests/compilers/test_ast.py
@@ -404,6 +404,12 @@ def test_ast_tuple():
     assert type(code) == ast.Tuple
 
 
+def test_argument_destructuring():
+    """ Ensure argument destructuring compilers. """
+    can_compile("(fn [[a b]] (print a b))")
+    cant_compile("(fn [[]] 0)")
+
+
 def test_lambda_list_keywords_rest():
     """ Ensure we can compile functions with lambda list keywords."""
     can_compile("(fn (x &rest xs) (print xs))")
--- a/tests/native_tests/language.hy
+++ b/tests/native_tests/language.hy
@@ -1153,3 +1153,8 @@
   (assert
    (= (identify-keywords 1 "bloo" :foo)
       ["other" "other" "keyword"])))
+
+(defn test-argument-destr []
+  "Make sure argument destructuring works"
+  (defn f [[a b] [c]] (, a b c))
+  (assert (= (f [1 2] [3]) (, 1 2 3))))


jaredly:
Would it take better advantage of the python runtime to just convert it into the ast generated by this?

```
def foo((a,b)):
  print a+b
```

For reference, here is the ast for the function's `args` in python 2.7

```json
"args": {
  "_type": "<class '_ast.arguments'>", 
  "args": [
    {
      "elts": [
        {
          "_type": "<class '_ast.Name'>", 
          "ctx": {
            "_type": "<class '_ast.Store'>"
          }, 
          "id": "a", 
          "col_offset": 7, 
          "lineno": 2
        }, 
        {
          "_type": "<class '_ast.Name'>", 
          "ctx": {
            "_type": "<class '_ast.Store'>"
          }, 
          "id": "b", 
          "col_offset": 9, 
          "lineno": 2
        }
      ], 
      "_type": "<class '_ast.Tuple'>", 
      "ctx": {
        "_type": "<class '_ast.Store'>"
      }, 
      "lineno": 2, 
      "col_offset": 7
    }
  ], 
  "kwarg": null, 
  "defaults": [], 
  "vararg": null
}, 
```

paultag:
Can someone confirm that syntax works in py 3

kirbyfan64:
@jaredly @paultag That's Python 2-only. It was removed in Python 3. See [PEP 3113](http://legacy.python.org/dev/peps/pep-3113/).

I *could* add a Python 2-only one, but it'd be kind of useless and put little gain, IMO.

jaredly:
oh right. makes sense

tests/compilers/test_ast.py
```diff
@@ -404,6 +404,12 @@ def test_ast_tuple():
     assert type(code) == ast.Tuple
 
 
+def test_argument_destructuring():
+    """ Ensure argument destructuring works. """
+    can_compile("(fn [[a b]] (print a b))")
```
zackmdavis: Testing that something compiles isn't the same as testing that it behaves as intended. (For instance, the `can_compile("(fn [[a b]] (print a b))")` assertion passes on master.) Perhaps it may be worth adding an assertion that the AST object returned by `can_compile` is what we expect, or maybe adding a native test along the lines of&mdash;

```
(defn test-argument-destructuring []
  (defn function-with-destructured-args [a [b c] [d e f]]
    (, a b c d e f))
  (assert (= (, 1 2 3 4 5 6)
             (function-with-destructured-args 1 [2 3] [4 5 6]))))
```

(which errors against master with `NameError: name 'b' is not defined`, but passes against the `destruct-args` branch)?

hy/compiler.py
```diff
@@ -1934,6 +1934,20 @@ def compile_function_def(self, expression):
...
+                    raise HyTypeError(arglist,
+                                      "Cannot destruct empty list")
+                args[i] = var = HySymbol(self.get_anon_var())
+                expression = HyExpression([
+                    HyExpression([
+                        HyString("setv"), arg, var
+                    ])]
+                ) + expression
+                expression = expression.replace(arg[0])
+            i += 1
```
zackmdavis: Forgive me, but I don't understand the purpose of incrementing `i` here at the end of the loop. Won't `i` have that value on the next iteration anyway?

kirbyfan64: @zackmdavis Yeah...that was a brainfart. :)

tests/compilers/test_ast.py
kirbyfan64: @zackmdavis Makes sense. Would it go in `native_tests/language.hy`?

zackmdavis: > Would it go in native_tests/language.hy?

Sure? (said with the intonation of a question, as if to allude to [your experience with](https://github.com/hylang/hy/commits?author=kirbyfan64) how things are done around here outweighing [mine](https://github.com/hylang/hy/commits?author=zackmdavis))

kirbyfan64:
Commits squashed and ready for merging!

berkerpeksag:
Thanks! Can you fix this?

```
  File "/home/travis/build/hylang/hy/tests/native_tests/language.hy", line 1159, in test_argument_destr
    (defn f [[a b] [c]]) (return (, a b c))
NameError: global name 'return' is not defined
```

kirbyfan64:
@berkerpeksag Fixed. I guess that's what I get for not running the tests myself...

kirbyfan64:
Let's try this again...

Commits squashed and ready for merging! :|

berkerpeksag:
LGTM.

kirbyfan64:
Bump...

paultag:
So, LGTM, however, before we merge, let's check in on Python 3.5 - I think there's some advanced unpacking stuff coming down the pipe - it would be interesting if we made this 3.3 -> 3.4 only. Can someone check up on that?

paultag:
Nope I lied. Perhaps 3.6? I think this is good to merge.

paultag:
```
def x(_hy_anon_var_1):
    [x, y] = _hy_anon_var_1
    [x, y]
```

I wonder if that's due to not using as_expr right in the `Result`

paultag:
(hy-spy session)

```
=> (defn x [[x y]] (print x y))
def x(_hy_anon_var_1):
    [x, y] = _hy_anon_var_1
    [x, y]
    return print(x, y)
=> (x [1 2])
x([1, 2])
1 2
```

paultag:
Otherwise LGTM. Let's just explain that and perhaps patch it before merge.

kirbyfan64:
@paultag I lost you...could you try to say all that again?

paultag:
@kirbyfan64 so when you run `hy --spy` and do the following in the REPL:

```
=> (defn x [[x y]] (print x y))
def x(_hy_anon_var_1):
    [x, y] = _hy_anon_var_1
    [x, y]
    return print(x, y)
```

The generated AST has a stray `[x, y]`, which I'm assuming is for expr-asignment on the code-branch in the `fn`. I think if we use the `Result` object, and let the `as_expr` handle the return, that block won't result in the code branch.

The generated code will run fine, but just has a stray instruction there to create a list :)

kirbyfan64:
@paultag 

> The generated AST has a stray `[x, y]`, which I'm assuming is for expr-asignment on the code-branch in the `fn`. I think if we use the `Result` object, and let the `as_expr` handle the return, that block won't result in the code branch.

Sorry, I'm a newbie to the Hy world. The last 10 times I toyed with the compiler, I broke completely unrelated things in the process somehow.

I think it might be this code:

```python
expression = HyExpression([
    HyExpression([
        HyString("setv"), arg, var
    ])]
) + expression
```

It's definitely something here...but I don't know what. Maybe my use of two nested `HyExpression`s is causing it to do that?

kirbyfan64:
@paultag Also, doesn't that happen anyway with assignments:

```
ryan@DevPC-LX:~/bfort$ hy --spy
hy 0.10.1 using CPython(default) 2.7.6 on Linux
=> (defclass X [] [])
class X:
    pass
=> (fn [] (setv a (X)) (setv a.a 1) (setv a.b 2))
def _hy_anon_fn_1():
    a = X()
    a.a = 1L
    a.a
    a.b = 2L
    return a.b
=> 
```

zackmdavis:
@paultag 

> let's check in on Python 3.5 - I think there's some advanced unpacking stuff coming down the pipe [...] Can someone check up on that?

Are you perhaps thinking of [PEP 448 Additional Unpacking Generalizations](https://www.python.org/dev/peps/pep-0448/), which is Accepted for 3.5 but still in the [patch review](http://bugs.python.org/issue2292) stage? It's about allowing the splat and double-splat operators to be used more times in more places, but seems to me to be independent of the kind of destructuring proposed here which doesn't really exist in Python ([anymore](https://www.python.org/dev/peps/pep-3113/)).

kirbyfan64:
Bumpza!

paultag:
OK, I'm happy calling that a setv bug. LGTM!

berkerpeksag:
Thanks!

PR #801, opened 2015-05-06
Added `ap-map-whenl` to contrib.

acron0:
Added `ap-map-whenl` as an alternative to `ap-map-when` which takes the predicate in lambda form.

I am ready to abandon this PR if some on can tell me of another way to do this :)

    => (require hy.contrib.anaphoric)
    => (setv foo 2)
    
    ;; doesn't work
    => (list (ap-map-when (= it foo) (* it 2) [1 2 1 2]))
    TypeError: 'bool' object is not callable

    ;; works
    => (list (ap-map-whenl (= it foo) (* it 2) [1 2 1 2]))
    [1L, 4L, 1L, 4L]

### DIFF
--- a/docs/contrib/anaphoric.rst
+++ b/docs/contrib/anaphoric.rst
@@ -90,6 +90,23 @@ determin when to apply the form.
     => (list (ap-map-when even? (* it 2) [1 2 3 4]))
     [1, 4, 3, 8]
 
+.. _ap-map-whenl:
+
+ap-map-whenl
+===========
+
+Usage: ``(ap-map-whenl predfn rep list)``
+
+As ``ap-map-when``, but takes a predicate in lambda form.
+
+.. code-block:: hy
+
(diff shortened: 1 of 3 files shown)


acron0:
`algernon> (ap-map-when (fn [x] (== x foo)) (* it 2) [1 2 3 4])`

Mind the doors.

paultag:
@acron0 :heart: thanks for your time anyway!

agentultra:
That too. Derp.

Sorry, have only had one coffee. Daughter woke me up super early this morning. :)

agentultra:
When I originally implemented ap-map-when I originally designed it with this behaviour. However we went with its current incarnation for various reasons; primarily that it was neater.

I’d like to avoid defining similar macros for different parameter types if we could. Is it possible for us to do some introspection at macro-expansion time to support both function types?

Cheers

PR #803, opened 2015-05-11
Add documentation for the (nonlocal ...) directive

farhaven:
This only makes sense once #719 gets merged.

Signed-off-by: Gregor Best <gbe@unobtanium.de>

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -1011,6 +1011,8 @@ conditional expression. Some examples:
 nonlocal
 --------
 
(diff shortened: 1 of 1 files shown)


kirbyfan64:
You should mention that this is Python 3-only.

farhaven:
There should probably be a `versionadded` tag as well, but I don't know what version this will end up in. Do I just add the current version there?

berkerpeksag:
I like documentation patches :) You can add 0.11.1. We can update it later. Thanks!

PR #800, opened 2015-05-06
prevent PEP 448 from breaking the world

zackmdavis:
Approximately seven hours ago, some AST changes [landed in Python](https://hg.python.org/cpython/rev/a65f685ba8c0) that had some alarming consequences for Hy's behavior, consequences such as this—
```
$ nosetests
..............E..E....E.........E............E......F....F.E...E.........E..E.......................................F............................F..EE..........FF.......................1
.................F.........................
...F..................F..................................................................................................................................FF..FF.

[failure details redacted ...]

Ran 388 tests in 44.353s

FAILED (errors=11, failures=13)
```

—and this—

```
hy 0.10.1 using CPython(default) 3.5.0a4+ on Linux
=> (defn all-the-args [&rest args &kwargs kwargs] (, args kwargs))
=> (apply all-the-args [1 2 3] {"four" 5})
((), {})
```

This pull request proposes the addition of some version conditionals to `compile_apply_expression` to restore sane behavior with bleeding-edge Python.

Issue #664, opened 2014-09-29
`Let` creates a closure where I unable to change outer variables because of UnboundLocalError


svetlyak40wt:
``` lisp
(defun test []
  (setv foo true)
  (let []
       (if foo
           (setv foo false))))

(test)
```

Compiles into:

``` python
def test():
    foo = True

    def _hy_anon_fn_1():
        if foo:
            foo = False
            _hy_anon_var_1 = foo
        else:
            _hy_anon_var_1 = None
        return _hy_anon_var_1
    return _hy_anon_fn_1()
test()
```

which leads to:

```
Traceback (most recent call last):
  File "/home/art/yandex-certificator/hy/env/bin/hy", line 8, in <module>
    load_entry_point('hy==0.10.0', 'console_scripts', 'hy')()
  File "/home/art/yandex-certificator/hy/hy/cmdline.py", line 305, in hy_main
    sys.exit(cmdline_handler("hy", sys.argv))
  File "/home/art/yandex-certificator/hy/hy/cmdline.py", line 293, in cmdline_handler
    return run_file(options.args[0])
  File "/home/art/yandex-certificator/hy/hy/cmdline.py", line 194, in run_file
    import_file_to_module("__main__", filename)
  File "/home/art/yandex-certificator/hy/hy/importer.py", line 74, in import_file_to_module
    eval(ast_compile(_ast, fpath, "exec"), mod.__dict__)
  File "minimal", line 7, in <module>
    (test)
  File "minimal", line 3, in test
    (let []
  File "minimal", line 4, in _hy_anon_fn_1
    (if foo
UnboundLocalError: local variable 'foo' referenced before assignment
```

I know why this does not work in python, but this behaviour is not expected in lisp like language.

gilch:
We don't have to fix scoping in general just to get the `let` form working, do we? The function in a `let` form is defined once, and immediately called once. In and out.
```Python
def test():
    foo = True
    def _hy_anon_fn_1(foo=foo):
        if foo:
            foo = False
            _hy_anon_var_1 = foo
        else:
            _hy_anon_var_1 = None
        return _hy_anon_var_1, foo
    _hy_anon_var_1, foo = _hy_anon_fn_1()
    return _hy_anon_var_1
```
[...]

Issue #663, opened 2014-09-29
Yield does not work inside nested `let` and `if`


svetlyak40wt:
Such code:

``` lisp
(let [[var 1]]
     (if var
         (yield "Hello")))
```

Compiles into:

``` python
def _hy_anon_fn_1():
    var = 1L
    return (yield u'Hello' if var else None)
_hy_anon_fn_1()
```

And return inside a generator leads to `SyntaxError: 'return' with argument inside generator` error.

rcarmo:
Guys, this is starting to be a bit of a pain... Same issue, same kind of logic. Can we get a fix for this?

```hy
(defn map-coords [seq &optional[width 11] [height 11] [spacing 3]]
    (let [[x-inc (+ width spacing)]
          [y-inc (+ height spacing)]
          [x spacing]
          [y spacing]
          [month-offsets []]]
        (for [d seq]
             (let [[wday (% (+ 1 ( .weekday d)) 7)]]
                (if (= wday 6)
                    (do
                        (setv y spacing)
                        (setv x (+ x x-inc))))
                (print x y)
                (yield {:date d :x x :y y :width width :height height})
                (setv y (+ y y-inc))))))
```

kirbyfan64:
@rcarmo You could always use `setv` as a workaround. Or:

```hy
(defmacro multi-setv [&rest args]
  `(do
    ~@(map (fn [[a b]] `(setv ~a ~b)) (zip (slice args nil nil 2) (slice args 1 nil 2)))))
```

Which would let you do:

```
(multi-setv [x-inc (+ width spacing)
                   y-inc (+ height spacing)
                   ...])
(continue work here)
```

paultag:
or (yield-from (let )) - I had a PR for some of this before

On Wed, Jul 15, 2015 at 3:45 PM, Ryan Gonzalez <notifications@github.com>
wrote:

> @rcarmo <https://github.com/rcarmo> You could always use setv as a
> workaround. Or:
>
> (defmacro multi-setv [&rest args]
>   `(do
>     ~@(map (fn [[a b]] `(setv ~a ~b)) (zip (slice args nil nil 2) (slice args 1 nil 2)))))
>
> Which would let you do:
>
> (multi-setv [x-inc (+ width spacing)
>                    y-inc (+ height spacing)
>                    ...])
> (continue work here)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hylang/hy/issues/663#issuecomment-121724682>.
>



[...]

PR #805, opened 2015-05-20
coroutines with async and await syntax

zackmdavis:
This pull request proposes preliminary support for [PEP 492](https://www.python.org/dev/peps/pep-0492/) coroutine functions in Hy under Python 3.5+, with a new `async-fn` compiler primitive, a `defasync` macro (corresponding to `async def` in Python), and an `await` expression. This draft at time of writing (46115fc9f) has some limitations: as the comment in the included echo server example (a straightforward port from [the example in the PEP](https://www.python.org/dev/peps/pep-0492/#working-example)) notes, `await` doesn't work inside of `let` bindings; the `async for` and `async with` statments are not yet supported; and getting `hy --spy` to work with the new syntax required some (manual, not part of the commit) tinkering to install [astor](https://github.com/berkerpeksag/astor) from master (which includes some supporting work not yet on PyPI) ... but the basic idea seems sound:

```
(hy35) zmd@ExpectedReturn:~/Code/hy35/hy$ hy --spy
hy 0.11.0 using CPython(default) 3.5.0a4+ on Linux
=> (defasync rah [] (print "rah!"))
async def rah():
    return print('rah!')
=> (rah)
rah()
<coroutine object rah at 0x7f9c1095dc50>
=> 
sys:1: RuntimeWarning: coroutine 'rah' was never awaited
(hy35) zmd@ExpectedReturn:~/Code/hy35/hy$ hy eg/asyncio/echo-server.hy &
[1] 8460
(hy35) zmd@ExpectedReturn:~/Code/hy35/hy$ Replhy Server listening on localhost:8000!

(hy35) zmd@ExpectedReturn:~/Code/hy35/hy$ telnet localhost 8000
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
New connection ...
Hello PEP 492 world
Sending b'Hello PE... back!
Hello PEP 492 world
```
Developed against python/cpython@c4766cf15.

### DIFF
--- a/eg/asyncio/echo-server.hy
+++ b/eg/asyncio/echo-server.hy
@@ -1,0 +1,29 @@
+;; adapted from the example in PEP 492
+;; https://www.python.org/dev/peps/pep-0492/#working-example
+
+(import asyncio)
+
+(defasync replhy-server []
+  (print "Replhy Server listening on localhost:8000!")
+  (await (asyncio.start-server handle-connection "localhost" 8000)))
+
+(defasync handle-connection [reader writer]
+  (print "New connection ...")
+
+  (while true
+    ;; XXX TODO: make it possible for this to be `(let [[data (await
+    ;; (.read reader 8192))]]` ... and so on. At time of writing,
+    ;; that's a `SyntaxError: 'await' outside async function`
+    ;; (presumably because the Hy `let` binding gets compiled to an
+    ;; ordinary non-async Python function).
+    (setv data (await (.read reader 8192)))
+    (if-not data
+       (break))
+    (print (.format "Sending {:.10}... back!" (repr data)))
+    (.write writer data)))
+
+(defmain [&rest args]
+  (let [[loop (asyncio.get-event-loop)]]
+    (.run-until-complete loop (replhy-server))
+    (try (.run-forever loop)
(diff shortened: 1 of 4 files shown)


theanalyst:
Awesome! thanks for taking the effort

paultag:
Oh fuck yes. Hella LGTM

@hylang/core RFR - let's get this landed.

zackmdavis:
(rebased and force-pushed because this was by necessity based on the commit in #800 which got rebased before landing, and to add a comment)

@paultag 

> Hella LGTM [...] let's get this landed. 

Thanks! But do note the blocking or arguably-blocking issues mentioned in the initial comment: the lack of `async-for` and `async-with` (this is trivial and I'll make a note to get to it very soon), the dependency on a new astor release for `--spy` to work, and the problem with `let` (which looks pretty hard, especially since we [haven't](https://github.com/hylang/hy/issues/664) [gotten](https://github.com/hylang/hy/issues/663) it right elsewhere, either).

paultag:
@zackmdavis looks like there's been no update on this for a while, closing this until it's ready. I love this feature, asyncio is a fav of mine, so I'd love for this to be re-opened once it's updated

PR #813, opened 2015-06-13
Documents (left-to-right) order of variable assignments in let expressions

ewaldgrusk:
Also adds an example illustrating this behaviour.


farhaven:
Looks good :+1:

berkerpeksag:
Thanks! (also thanks for the review @farhaven :))

paultag:
:+1: thank you!

PR #814, opened 2015-06-15
Make (defmain ...) a clean macro

farhaven:
Now you can do stuff like

```
(defn main []
  (print "Hy!"))
(defmain [&rest foo]
  (main))
```

without infinite recursion.

### DIFF
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -200,16 +200,14 @@
 
 (defmacro defmain [args &rest body]
   "Write a function named \"main\" and do the if __main__ dance"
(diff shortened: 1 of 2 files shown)


berkerpeksag:
LGTM. I think you can delete the `(global --name--)` line.

farhaven:
`--name--` is a global name at that point. Without `(global --name--)`, the `(setv oldname --name--)` causes `--name--` to be looked up in the local scope of the macro function.

Foxboron:
LGTM. :shipit:

PR #816, opened 2015-06-16
Fix anaphoric macro `ap-if`

ewaldgrusk:
The updated macro definition explicitly states all mandatory and all optional parameters. Hence, supplying the wrong number of arguments will give a more helpful error message.

I should probably mention that it is exactly the macro definition given in _On Lisp_ by Paul Graham.

### DIFF
--- a/hy/contrib/anaphoric.hy
+++ b/hy/contrib/anaphoric.hy
@@ -25,8 +25,9 @@
 ;;; These macros make writing functional programs more concise
 
 
-(defmacro ap-if (test-form &rest args)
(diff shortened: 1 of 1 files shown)


theanalyst:
could you add tests for the `else-form` .. tests live in `tests/native_tests/contrib/anaphoric`

paultag:
Yeah, this is wonderful, can we grab a test, @ewaldgrusk ? Otherwise LGTM!

paultag:
@ewaldgrusk This one is a total no-brainer, but can we get this updated against master and a test added? I'm going to close it while we wait -- wish we could have this in master!

PR #834, opened 2015-07-01
Expand doc for defn with &kwargs

farhaven:
svetlyak40wt mentioned on IRC that the documentation for keyword args is not exactly up to speed.


Foxboron:
+1 :shipit:

PR #843, opened 2015-07-22
Remove 2.6 entirely from test runners

paultag:
(But not codepaths, in case someone cares)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: false
 language: python
 python:
   - "pypy"
-  - "2.6"
   - "2.7"
   - "3.3"
   - "3.4"
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py26,py27,pypy,py33,flake8
+envlist = py27,pypy,py33,flake8
 skipsdist = True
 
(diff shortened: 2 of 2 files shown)


kirbyfan64:
So does this mean Hy no longer supports 2.6?

Foxboron:
No, i like Python2.6
















Not at all. LGTM! :shipit:

paultag:
@kirbyfan64 it means I don't really want 2.6 to block test passage anymore -- I didn't pull our codepaths, so it'll continue until it becomes too much of a maintenance burden -- thoughts?

paultag:
(Python 2.6 was officially EOL'd in 2013)

paultag:
Let's wait to make sure @kirbyfan64 doesn't have blocking objections, and if we're good, we'll merge!

kirbyfan64:
@paultag Hahaha, I hate 2.6 just as much as everyone else. :O

I was just curious because I had to add an explicit call to make set literals work on 2.6 so the tests could pass.

paultag:
@kirbyfan64 fuck it, let's drop 2.6 - if someone needs it, we can re-introduce. Fire away with set literals, that counts as a maintenance burden

Issue #844, opened 2015-07-23
Get rid of let


gilch:
We all know `let` is broken. There have been many proposed solutions, but one option that no-one has offered--or perhaps actually thought of, but never dared to say--is to get rid of `let` altogether.

I understand this may be controversial. If at least a couple of core members feel strongly that Hy *must* have a `let`, then feel free to close this issue and I'll not mention it again. But do hear me out first.

Remember the main selling point of Hy: "We're still Python(TM). Most of the internals translate 1:1 to Python internals." Hy is not Clojure, nor Common Lisp, but homoiconic Python. Do we really need `let`?

In the older Lisps, usually `let` is just used to create a local variable. Hy can do this more simply with the `setv` form. Python has no block scope. Creating and calling a new function just for a local or two is pretty inefficient in Python. Shouldn't we encourage users to use `setv` anyway in such cases? This approach also breaks `break` `continue` `yield` and would break `return` if we had it (we should). There may be other problems I haven't thought of.

Now I'm confident that, in time, we could fix all of these issues, perhaps with an alternative approach. But all of these alternatives will expand into something *complicated* especially compared to `setv`. There will likely be overhead or other nasty surprises.

If someone really needs a new scope block (perhaps to control variable lifetime in a macro), then there's nothing stopping them from explicitly doing it the same way as they would in Python. Perhaps with a class, function or dictionary. Then those familiar with Python would not be surprised by their expectations about how `let` is supposed to work and we wouldn't have to port `nonlocal` to Python 2.

kirbyfan64:
> we wouldn't have to port nonlocal to Python 2.

I feel like that'll end up being necessary regardless...

gilch:
Actually there's a really easy way to make that unnecessary: Drop support for Python 2. Not saying we should necessarily, but Python2 is supposed to go away eventually anyway.

Foxboron:
You are in a camp where we all dream of a day without Python2. We have
partially started dropping support for Python26 tho, so progress \o/

cwebber:
One alternative is to provide let to preserve backwards compatibility, but fully document that it's mostly broken and *why* it is unlikely that a real let will never exist, and not use it in any example code.

agentultra:
Or just implement it as a convenience macro for setv'ing a bunch of variables. It's just Python under the hood.

gilch:
[...]

PR #849, opened 2015-07-24
Let setv take pairs of arguments

kirbyfan64:
@algernon Beat you to it. ;)

Ref. #844.

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -334,6 +334,18 @@ For example:
     => (counter [1 2 3 4 5 2 3] 2)
     2
 
+They can be used to assign multiple variables at once:
+
+.. code-block:: hy
+
+    => (setv a 1 b 2)
+    (1L, 2L)
+    => a
+    1L
+    => b
+    2L
+    =>
+
 
 defclass
 --------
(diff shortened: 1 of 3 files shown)


paultag:
I guess omitting braces is fine here, since there's no body following

My only hangup is if there's a consistency issue with the rest of the language - @hylang/core anyone see a need for braces or am I losing it?

If no one else sees a problem, I've got a :+1: LGTM here!

paultag:
P.S., forgot to say - I love the return semantics.

gilch:
This is supposed to be sequential, right? Can you add a test something like 
```
(setv y False
      x True
      y x)
(assert y)
```

gilch:
@paultag I don't think we need braces here.

agentultra:
seconded.

algernon:
Looks good. A test @gilch described would be great, though, I'd merge this much more confidently. Big :+1: otherwise!

kirbyfan64:
@gilch @algernon Test added (to same commit: no squashing needed).

zackmdavis:
squirrel

PR #860, opened 2015-07-26
reader macro #@ for with-decorator

zackmdavis:
The `with-decorator` special form is not the most ergonomic—this commit
introduces a new builtin `#@` reader macro that expands to an invocation
of `with-decorator`. To support this, `reader_macroexpand` is made to
also look in the default `None` namespace, in imitation of how
regular (non-reader) macros defined in hy.core are looked up. The
docstring of `hy.macros.reader` is also edited slightly for accuracy.

This in the matter of issue #856.

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -1463,6 +1463,25 @@ will be 4 (``1+1 + 1+1``).
     8
 
 
+#@
+~~
+
+.. versionadded:: 0.12.0
+
+The :ref:`reader macro<reader-macros>` ``#@`` can be used as a shorthand
+for ``with-decorator``. With ``#@``, the previous example becomes:
+
+.. code-block:: clj
+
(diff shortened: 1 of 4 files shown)


Foxboron:
:+1: 
:shipit: 
Looks great, you even removed some of my *cough*copypasta*cough* from when i implemented it *lalala*
LGTM, great work.

paultag:
LGTM too!

paultag:
@zackmdavis thanks! This is awesome!

gilch:
Wait a second. In Python, the decorators are pure syntactic sugar. What is this doing that the threading macros don't do already?

zackmdavis:
@gilch nothing, but some may find `#@(decorator (defn foo [x] (bar x)))` to be a sweeter form of syntactic sugar than `(->> (fn [x] (bar x)) (decorator) (def foo))`?

PR #861, opened 2015-07-28
Explicitly set missing source on LexExceptions

farhaven:
This repairs things like the following:

    (import [hy.lex [tokenize]])
    (tokenize "(foo))")

Without this patch, the lexer exception thrown by `tokenize` is unprintable because of the missing source reference.


Foxboron:
This also looks sane. +1 :shipit:

zackmdavis:
***+1***

PR #864, opened 2015-07-28
Make `read` read line-wise and don't ignore lexer errors

farhaven:
This allows us to react to things like "(foo))" being passed in instead of
looping until kingdom come. Also makes tokenizing things that are not
expressions easier:

    => (read)
    "foo"
    u'foo'
    =>

### DIFF
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -371,13 +371,13 @@
    Can take a given input buffer to read from"
   (def buff "")
   (while true
-    (def inn (str (.read from-file 1)))
+    (def inn (str (.readline from-file)))
     (if (= inn eof)
       (throw (EOFError "Reached end of file" )))
     (setv buff (+ buff inn))
     (try
       (def parsed (first (tokenize buff)))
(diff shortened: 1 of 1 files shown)


zackmdavis:
***+1*** Yes, this definitely does behave more reasonably for the examples you mention.

(Although the commit message seems slightly inaccurate in that on master, the unbalanced-parens example seems to successfully get the expression and then ignore the parsing error, rather than hanging in an infinite loop like the commit message led me to expect&mdash;
```
=> (read)
(foo))
('foo')
=> 
```
But I do observe the hanging with the string literal example.)

farhaven:
I'd like to get at least one more :+1: before merging this. You are right about that specific example. Something that breaks is this:

```
=> (read)
foo)
```

which returns `f`, `o`, and `o` with successive calls on master and hangs once it tries to parse the `)`. With the patch, we immediately get a `LexException` once the broken form is passed in.

tuturto:
Looks good and works well :+1:

Issue #240, opened 2013-07-10
Remove redundant keywords


joehakimrahme:
There are redundant keywords in the current implementation:
- `setf`/`setv`/`def`
- `lambda`/`fn`
- `for`/`foreach` (?)
- `throw`/`raise`
- `catch`/`except`
- `do`/`prgn`

I believe we should avoid any ambiguity by removing duplication and settling on one of each. It's one of these things that we should worry about now, while we can afford to break the language.

gilch:
This redundancy also bothers me. If there actually is a meaningful difference in how things are compiled/optimized, (`lambda`/`fn`? `def`/`setv`?) then it should at least be **documented**, so we know which one to use when. But even then I'd prefer not to have to choose. Hy should use Clojure forms rather than Common Lisp forms when there are near equivalents to choose between.

In cases the above doesn't apply, I generally prefer the shorter names.

gilch:
@algernon Can we also have this one in the "Grand Language Cleanup" milestone?

algernon:
The "which one to use" is a funny can of worms. But yes, this belongs to the Grand Language Cleanup.

Last time this came up, one idea was to have a namespace that sets up the aliases. So if someone prefers CL-ish names: `(import hy.style.common-lisp)`

gilch:
>The "which one to use" is a funny can of worms.

*sigh*
I think Python3, then Clojure, then Common Lisp, in that order, is a good preference heuristic for which to use. In fact, I think it resolves all of them. Hy really does not need to be compatible with Common Lisp (or Clojure for that matter). Hy is it's own Lisp dialect, and We're still Python(TM).

>Last time this came up, one idea was to have a namespace that sets up the aliases. 

Here's an idea for ya, `import hy.style` will not change the language, but instead print out friendly error messages whenever you attempt to use the wrong Lisp dialect. Like say you start a form with a `progn` symbol and Hy replies, "you meant to use a `do` here, didn't you, dear?" That way people will actually learn Hy, and will use a mutually intelligible version of it ;)

algernon:
Haha, I like that suggestion.

Oh boy, the possibilities! Tie it in with hydiomatic, and funny warning messages.... mmm. That'd be going a bit too far, though. So perhaps `hy.style` shouldn't be part of the core language.

gilch:
Found some more.
`True`/`true`
`False`/`false`
`None`/`nil`/`null`

[...]

PR #880, opened 2015-08-09
Remove some redundant keywords

gilch:
Ref #240 in the Grand Language Cleanup. This is not enough to close the issue, but it helps. The redundant keywords I am aware of in Hy are:
```
raise/throw
except/catch
do/progn
defn/defun
lif/lisp-if and lif-not/lisp-if-not
defn-alias/defun-alias

True/true
False/false
None/nil/null

def/setv
lambda/fn
first/car
rest/cdr
```
This pull request only touches the first group, likely to be the least controversial. The second group may be a bit more difficult to fix. The third group has slightly different behavior/convention, so may be harder to unify.

Removed: `throw`
Rationale: `raise` is a Python keyword, and thus reserved anyway. This reduces the number of reserved words in Hy, while the other choice would not. It is also less confusing to Python programmers.

Removed: `catch`
Rationale: `except` is a Python keyword, same as above.  `catch` was popular, perhaps because it's shorter, but when choosing between Python reserved words and equivalent Clojure words, Hy must choose Python's words, or redundancy in reserved words remains.

Removed: `progn`
Rationale: `do` is shorter, and `progn` was barely ever used in Hy anyway; Clojure forms are more favored by the community than Common Lisp's.

Removed: `defun`
Rationale: as above, `defn` is shorter, used more, and Clojure, rather than CL. (This commit also removes `defun-alias`.)

Removed: `lisp-if` and `lisp-if-not`
Rationale: `lif` and `lif-not` are shorter. A weak reason, but had to break the tie somehow. These aren't used that much anyway.

Removed: `defn-alias` and `defmacro-alias`
Rationale: After removing the above redundancies, these are no longer required in Hy itself. They are also likely to cause similar redundancy issues in the future, in Hy and elsewhere, thus were bad for the health of the Hy ecosystem.

### DIFF
--- a/docs/contrib/loop.rst
+++ b/docs/contrib/loop.rst
@@ -37,7 +37,7 @@ loop
 ``loop`` establishes a recursion point. With ``loop``, ``recur``
 rebinds the variables set in the recursion point and sends code
 execution back to that recursion point. If ``recur`` is used in a
-non-tail position, an exception is thrown.
+non-tail position, an exception is raised.
 
 Usage: `(loop bindings &rest body)`
 
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -289,10 +289,10 @@ conditional expression.
     {1: 2, 3: 6, 9: 18, 5: 10, 7: 14}
 
 
-do / progn
+do
 ----------
 
-``do`` and `progn` are used to evaluate each of their arguments and return the
+``do`` is used to evaluate each of its arguments and return the
 last one. Return values from every other than the last argument are discarded.
 It can be used in ``lambda`` or ``list-comp`` to perform more complex logic as
 shown in one of the following examples.
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -1300,20 +1300,20 @@ or no arguments to re-raise the last ``Exception``.
 try
 ---
 
-The ``try`` form is used to start a ``try`` / ``catch`` block. The form is
+The ``try`` form is used to start a ``try`` / ``except`` block. The form is
 used as follows:
 
 .. code-block:: clj
 
     (try
         (error-prone-function)
-        (catch [e ZeroDivisionError] (print "Division by zero"))
+        (except [e ZeroDivisionError] (print "Division by zero"))
         (else (print "no errors"))
         (finally (print "all done")))
 
-``try`` must contain at least one ``catch`` block, and may optionally include
-an ``else`` or ``finally`` block. If an error is raised with a matching catch
-block during the execution of ``error-prone-function``, that ``catch`` block
+``try`` must contain at least one ``except`` block, and may optionally include
+an ``else`` or ``finally`` block. If an error is raised with a matching except
+block during the execution of ``error-prone-function``, that ``except`` block
 will be executed. If no errors are raised, the ``else`` block is executed. The
 ``finally`` block will be executed last regardless of whether or not an error
 was raised.
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -63,10 +63,10 @@ Compiles down to:
 which to do the attribute dereference. It uses bare symbols as attributes
 to access (in the example, *bar*, *baz*, *frob*), and compiles the contents
 of lists (in the example, ``[(+ 1 2)]``) for indexation. Other arguments
-throw a compilation error.
+raise a compilation error.
 
-Access to unknown attributes throws an :exc:`AttributeError`. Access to
-unknown keys throws an :exc:`IndexError` (on lists and tuples) or a
+Access to unknown attributes raises an :exc:`AttributeError`. Access to
+unknown keys raises an :exc:`IndexError` (on lists and tuples) or a
 :exc:`KeyError` (on dictionaries).
 
 ->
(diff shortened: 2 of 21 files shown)


algernon:
:+1:, except for `defn-alias` and `defmacro-alias`. They're very useful when you're doing - arguably crazy - stuff like these:

```hy
(defmacro-alias [condᵉ conde] [&rest cs]
  (with-gensyms [s c]
    (let [[ncs (__subst-else cs)]]
      `(with-monad logic-m
         (fn [~s]
           (m-plus (map (fn [~c]
                          ((apply all ~c) ~s))
                        [~@ncs])))))))

(defn-alias [firstᵒ firsto] [l a]
  (fresh [d]
         (consᵒ a d l)))
```

Granted, the number of people who do that are few, but when you want to give a function a hy-ish, perhaps unicode name, and still want to allow python people to use it without tearing their hair out, these macros are handy. And fun. I agree they don't have much use, but they aren't completely redundant.

Foxboron:
I'd also like the alias functions to stay. They are usefull even tho the core lib won't be using them much.

kirbyfan64:
Just to say: I like having both `def` and `setv`: the former is primarily used for globals, and the latter for locals.

algernon:
:+1: on keeping the `setv`/`def` pair. I'd also keep the `true`, `false`, `null` and `nil` aliases (because I use them a lot, and shift is far away). `lambda` and `fn` are slightly different in behaviour, if I remember correctly, but that may have changed.

I also seem to remember there being a slight difference between `first`/`car` and `rest`/`cdr`, but I may be imagining things.

gilch:
>... defn-alias and defmacro-alias. They're very useful when you're doing - arguably crazy - stuff ...

...

>I'd also like the alias functions to stay. They are usefull even tho the core lib won't be using them much.

Maybe I should just move them to contrib. It would keep the core more focused, but you could still import them if you want. On the other hand, it's pretty easy to alias a function even without the macro:
```Hy
(def firsto
  (defn firstᵒ [l a]
    (fresh [d]
           (consᵒ a d l))))
```
I feel like there ought to be a similar way to alias a macro after it has been defined, but it doesn't look easy. The `defmacro-alias` form just rewrites the whole thing using a new name. This approach seems less that optimal.

gilch:
>I'd also keep the `true`, `false`, `null` and `nil` aliases (because I use them a lot, and shift is far away). 

See, I knew the other groups would be more controversial. This pull (so far) only targets the first group, and we don't have to agree to fixing the other two before merging this part. 

We can still talk about them though.

As for the second group, I'll admit to using the easier-to-type aliases `true`/`false`/`nil` almost exclusively in my own code, and they're used a lot in Hy proper. But I still think this is extra complexity for marginal benefit.

Even if you're not convinced by the following, can we at least get rid of `null`? It's hardly ever used, and not easier to type than `nil`.

Correct me if I'm wrong about this assumption, but Python keywords have to be reserved in Hy, because Hy compiles to Python and can be imported from Python. They don't have to mean the same thing in Hy as in Python, though this can get confusing. `def` (for example) has a pretty different meaning, but there's no other good meaning for `True` and `False` that we could use instead. `True`/`False`/`None` are reserved in Python3. They cannot be assigned to in Python, and it would be problematic to allow this in Hy. Thus, if we have to choose, the clear choice is the Python words.

I am unconvinced by the argument that `true` and `false` are easier to type. `True` and `False` literals are actually not used as much as you might expect. Sure, Booleans are used all the time, but usually you type in an expression that evaluates to a Boolean rather than forcing one or the other yourself.

The case where Boolean literals get used the most is in testing. If you're testing at the REPL and just need a generic truthy or falsey value, admit it, you're going to use `1` and `0` over even `true` and `false`, because it just does not get any easier to type than that.

Unit tests, on the other hand, also serve as a kind of documentation. It's worth making them readable. Now if we decide that it's idiomatic in Hy to use `1`/`0` for unit tests, then people will be used to it and it will not be considered hard to read. (Many of Hy's unit tests already do this by the way.) On the other hand, `True` and `False` are no less readable, and they even stand out more.

I am also unconvinced by the argument that `true` and `false` are used all the time. It's a simple one-line fix:
```Python
true, false = True, False
```
The first group in the pull (which we do seem to agree on) breaks stuff a lot harder than this. Yes, the Grand Language Cleanup will break stuff. Best get it over with now. If removing `true`/`false` turns out to be a mistake, it will be an easily correctable mistake. If NOT removing `true`/`false` turns out to be a mistake, it will only get harder to fix.

There's a better argument for keeping `nil`, I suppose. This really does get used a lot. `None` could perhaps be used for something else, but I don't know what. But `nil` may cause confusion for Common Lisp programmers expecting it to be the empty list. `None` wouldn't have that connotation. I'd still prefer we be consistent and get rid of `nil`, but I'd be more willing to make an exception here.

paultag:
I'm OK with purging true false null and nil.

algernon:
Ok, lets get this rolling: `defn-alias` / `defmacro-alias` moved to contrib is a reasonable compromise, I believe, and the rest of this PR is good to go. We'll discuss the other stuff in different PRs / issues (like #240).

@gilch, can you do the contrib move? Or I can merge this as-is, up to the prior commit, and do the move myself.

PR #889, opened 2015-08-11
purged null from Hy

### LINKED ISSUE #240: Remove redundant keywords

gilch:
This pull only removes the redundant `null` from Hy. It is not sufficient to close #240, but it is a step in the right direction.

Rationale: We already have `None`, and `null` was barely ever used anyway. If this breaks something, a simple `(def null None)` will suffice to patch it.

The fate of the other redundant keywords remains open.


paultag:
Yes please. Thank you!

zackmdavis:
Nice branch name, too. :shipit:

PR #893, opened 2015-08-12
xi parameter may appear in function position

gilch:
I just realized something.
```Clojure
user=> (def foo #(% 2 3))
#'user/foo
user=> (foo +)
5
user=> (foo -)
-1
user=> (foo /)
2/3
```
The parameter can appear in the function position.
This won't currently work in Hy's xi forms because the first element isn't checked for xi symbols, but I just fixed that with this slight adjustment.


farhaven:
LGTM, will merge once travis passes.

PR #903, opened 2015-08-17
added defmacro!

gilch:
I saw `defmacro/g!` in _Let Over Lambda_ before I saw it in Hy, but it's just a helper macro for the real utility: `defmacro!`, which is like `defmacro/g!`, plus automatic once-only evaluation for
`o!` parameters, which are available as the equivalent `g!` symbol. This pull translates `defmacro!` to Hy.

For those who haven't read the book, a common problem with macros is multiple-evaluation.

``` Hy
=> (defn expensive-get-number [] (print "spam") 14)
=> (defmacro triple-1 [n] `(+ n n n))
=> (triple-1 (expensive-get-number))  ; evals n three times!
spam
spam
spam
42
```

You can avoid this with a gensym:

``` Hy
=> (defmacro/g! triple-2 [n] `(do (setv ~g!n ~n) (+ ~g!n ~g!n ~g!n)))
=> (triple-2 (expensive-get-number))  ; avoid repeats with a gensym
spam
42
```

But this is more concise and clear with `defmacro!`, which does that for you:

``` Hy
=> (defmacro! triple-3 [o!n] `(+ ~g!n ~g!n ~g!n))
=> (triple-3 (expensive-get-number))  ; easier with defmacro!
spam
42
```

PR #958, opened 2015-10-13
Fix issue found by @tuturto in #903 (take 2)

kirbyfan64:
@tuturto @gilch Please review!

### DIFF
--- a/hy/macros.py
+++ b/hy/macros.py
@@ -18,14 +18,20 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
-from inspect import getargspec, formatargspec
 from hy.models import replace_hy_obj, wrap_value
 from hy.models.expression import HyExpression
 from hy.models.string import HyString
 
 from hy.errors import HyTypeError, HyMacroExpansionError
+from hy._compat import PY3, PY34
+
+if PY3:
+    from inspect import getfullargspec as getargspec
+else:
+    from inspect import getargspec
 
 from collections import defaultdict
+import ast
 
 CORE_MACROS = [
     "hy.core.bootstrap",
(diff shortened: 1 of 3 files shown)


hy/macros.py
```diff
@@ -123,13 +129,42 @@ def _import(module, module_name=module_name):
...
-    formatted_args = formatargspec(*argspec)
-    fn_str = 'lambda {}: None'.format(
-        formatted_args.lstrip('(').rstrip(')'))
-
-    empty_fn = eval(fn_str)
+def make_empty_fn_copy(fn):
+    try:
+        argspec = getargspec(fn)
+    except:
+        return lambda *_: None
```
gilch: Depending on how we resolve #959 this might have to be `return lambda *arg, **kwarg: None`. But why is `getargspec` failing in the first place? Can we fix that instead?

gilch:
Checks are failing. Also needs unit tests.

hy/macros.py
kirbyfan64: @gilch `getargspec` won't work on functions Python thinks are builtins. Ultimately, the rest of this is needed anyway because of non-alphanumeric argument names (e.g. `(defn f [@@])`).

However, I probably should look into what the issue is regarding the `getargspec` call itself.

gilch: Why does Python think a Hy macro is a Python builtin? That doesn't seem right.

kirbyfan64:
@gilch Uhh...this is weird...

I can't reproduce the original issue anymore, even without the `try...except` clause. I'm guessing it was something to do with the string eval?

Could you try merging my branch with your `defmacro!` branch and see if the original issue is still there?

gilch:
I've noticed I have to `pip unininstall hy` first, you can't just install over the top or you're not testing what you think you are. That might be why you can't reproduce it.

gilch:
@kirbyfan64: Can you rebase on current master (for testing)?

kirbyfan64:
...and this got killed. Hold on...

kirbyfan64:
@gilch Done.

zackmdavis:
@kirbyfan64 There are some more duplicated commits on this branch. What specific command are you using to rebase?

kirbyfan64:
@zackmdavis `git pull --rebase origin master`...

I think it should still merge cleanly, though. Or at least GitHub says it's even with `master`. I think the duplicated commits are just a GitHub issue with showing rebased commits.

zackmdavis:
> I think the duplicated commits are just a GitHub issue with showing rebased commits.

No, they're really there&mdash;

``` bash
# making sure my local repository has an accurate view of kirbyfan64's repository
zmd@ExpectedReturn:~/Code/hy/hy$ git remote -v | grep kirby
kirbyfan64	git@github.com:kirbyfan64/hy.git (fetch)
kirbyfan64	git@github.com:kirbyfan64/hy.git (push)
zmd@ExpectedReturn:~/Code/hy/hy$ git fetch kirbyfan64 
# check out the tip of this PR's branch
zmd@ExpectedReturn:~/Code/hy/hy$ git checkout kirbyfan64/fix_macro_err2 
HEAD is now at 12146a2... Fix issue found by @tuturto in #903
# the previous two commits have the same SHAs that GitHub's current rendering of the PR shows
zmd@ExpectedReturn:~/Code/hy/hy$ git log --oneline | head -n 3
12146a2 Fix issue found by @tuturto in #903
54e3d26 Add a few tests for various defn corner cases
820c6fc Trying to setv a callable should raise a nice error
# but the actual content of those commits already exists once on master
zmd@ExpectedReturn:~/Code/hy/hy$ git log --oneline --grep "Trying to setv a callable"
820c6fc Trying to setv a callable should raise a nice error
d3520e5 Trying to setv a callable should raise a nice error
```
> `git pull --rebase origin master` ...

Hm, I don't seem to have a good intuition for what `pull --rebase` does with arguments (I've only used it without arguments, when I have new local commits but the upstream tracking branch has new commits, too, and I want to apply my changes on top of those).

To get out of the current duplicate-commits situation, I think you want to `git rebase -i origin/master` and then delete the lines corresponding to the duplicate commits. And likewise `git rebase origin/master` to update pull requests in the future (assuming you've fetched `origin` recently and that `origin` refers to the source repository). The way I think of it is that `git rebase foo` is going to take all the commits between the-latest-common-ancestor-of-`HEAD`-and-`foo` and `HEAD` and apply them one by one on top of `foo` (where `foo` could be anything that Git understands as pointing to a commit, including a local branch name, a remote branch name like `origin/master`, a short SHA, or a special relative name like `HEAD~3`, which means "three commits before the one I'm currently on").

kirbyfan64:
@zackmdavis Oh well. I'll just open another one...

PR #960, opened 2015-10-13
Give an error when &kwonly is used in a macro definition

### LINKED ISSUE #959: &kwonly in defmacro
```
=> (defmacro foo [&kwonly b] `(print ~b))
import hy

@hy.macros.macro('foo')
def _hy_anon_fn_1(*, b):
    return HyExpression((([] + [HySymbol('print')]) + [b]))
=> (foo)

  (foo)
  ^---^
HyMacroExpansionError: expanding `foo': TypeError("_hy_anon_fn_1() missing 1 req
uired keyword-only argument: 'b'",)

=> (foo :b 42)

  (foo :b 42)
  ^---------^
HyMacroExpansionError: expanding `foo': TypeError('_hy_anon_fn_1() takes 0 posit
ional arguments but 2 were given',)

=> (foo ':b 42)

  (foo ':b 42)
  ^----------^
HyMacroExpansionError: expanding `foo': TypeError('_hy_anon_fn_1() takes 0 posit
ional arguments but 2 were given',)

=> (macroexpand '(foo :b 42))
from hy.core.language import macroexpand
from hy import HyExpression, HyInteger, HyKeyword, HySymbol
macroexpand(HyExpression(((([] + [HySymbol('foo')]) + [HyKeyword('\ufdd0:b')]) +
 [HyInteger(42)])))
  File "C:\Users\ME\Code\repos\hy/hy/core/language.hy", line 265, in macroexpand

    (hy.macros.macroexpand form name))
  File "C:\Python34\Scripts\env-hy\lib\site-packages\hy\macros.py", line 155, in
 macroexpand
    tree = macroexpand_1(tree, module_name)
  File "C:\Python34\Scripts\env-hy\lib\site-packages\hy\macros.py", line 194, in
 macroexpand_1
    raise HyMacroExpansionError(tree, msg)
hy.errors.HyMacroExpansionError: <unprintable HyMacroExpansionError object>
```
The keywords appear to work as simple positional arguments. If there's no way to provide a `&kwonly` argument to a macro form, then it should not be allowed in `defmacro`.

Either change the macro system to accept keyword arguments, or signal an error when someone attempts to use `&kwonly` etc. in `defmacro`.

kirbyfan64:
(closes #959)

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -503,6 +503,64 @@ In Hy:
      address (models.TextField)
      notes (models.TextField)])
 
+Macros
+======
+
+One really powerful feature of Hy are macros. They are small functios that are
+used to generate code (or data). When program written in Hy is started, the
+macros are executed and their output is placed in program source. After this,
+the program starts executing normally. Very simple example:
+
+.. code-block:: clj
+
+  => (defmacro hello [person]
+  ...  `(print "Hello there," ~person))
+  => (Hello "Tuukka")
+  Hello there, Tuukka
+
+The thing to notice here is that hello macro doesn't output anything on
+screen. Instead it creates piece of code that is then executed and prints on
+screen. Macro writes a piece of program that looks like this (provided that
+we used "Tuukka" as parameter:
+
+.. code-block:: clj
+
+  (print "Hello there," Tuukka)
+
+We can also manipulate code with macros:
+
+.. code-block:: clj
+
+  => (defmacro rev [code]
+  ...  (let [op (last code) params (list (butlast code))]
+  ...  `(~op ~@params)))
+  => (rev (1 2 3 +))
+  6
+
+The code that was generated with this macro just switched around some the
+elements, so by the time program started executing, it actually red:
+
+.. code-block:: clj
+
+  (+ 1 2 3)
+
+Sometimes it's nice to have a very short name for macro that doesn't take much
+space or use extra parentheses. Reader macros can be pretty useful in these
+situations (and since Hy operates well with unicode, we aren't running out of
+characters that soon):
+
+.. code-block:: clj
+
+  => (defreader ↻ [code]
+  ...  (let [op (last code) params (list (butlast code))]
+  ...  `(~op ~@params)))
+  => #↻(1 2 3 +)
+  6
+
+Macros are useful when one wished to extend the Hy or write their own
+language on top of that. Many features of Hy are macros, like ``when``,
+``cond`` and ``->``.
+
 Hy <-> Python interop
 =====================
 
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2371,6 +2371,9 @@ class HyASTCompiler(object):
             raise HyTypeError(name, ("received a `%s' instead of a symbol "
                                      "for macro name" % type(name).__name__))
         name = HyString(name).replace(name)
+        for kw in ("&kwonly", "&kwargs"):
+            if kw in expression[0]:
+                raise HyTypeError(name, "macros cannot use %s" % kw)
         new_expression = HyExpression([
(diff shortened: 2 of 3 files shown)


kirbyfan64:
BTW, is there a way to test this? I looked in the tests, but it seems there are no tests for anything like this...

gilch:
I wonder if we should reconsider the all the other `&` parameters. Clearly we need `&rest`. I'm not sure about how `&key` works, because it's not documented. What about `&optional`? It should work positionally, but we can't skip over an optional arg with a keyword.

gilch:
I think you could put an invalid macro declaration in a try form and catch the HyTypeError. Something like this:
```Python
try:
    <test code>
    raise Error("HyTypeError expected")
except HyTypeError:
    pass
```
The testing framework might even have a decorator for this kind of thing. I'd have to look at the docs.

gilch:
I checked, and nose does have a [decorator](https://nose.readthedocs.org/en/latest/testing_tools.html) for this: `nose.tools.raises`. We don't seem to be using `nose.tools` in any tests, but I did see the `try` pattern I mentioned above. This should also work as a native Hy test.

kirbyfan64:
@gilch Test added.

tuturto:
```&key```, ```&optional``` and ```&kwargs``` probably don't make sense in the context of macros currently.

The code looks solid to me :+1:

**Comment by gilch** [comment]
@kirbyfan64, can you add checks and tests for ```&key```, ```&optional``` and ```&kwargs```?

**Comment by kirbyfan64** [comment]
@gilch But `&optional` and `&key` work as expected with macros:

```
ryan@DevPC-LX:~$ hy
=> (defmacro x [&optional v] (if v `"v was given" `"v was not given"))
^[[
ryan@DevPC-LX:~$ hy
=> (defmacro x [&optional v] (if v `"v was given" `"v was not given"))
=> (x 1)
u'v was given'
=> (x)
u'v was not given'
=> (defmacro y [&key {"v" nil}] (if v `"v was given" `"v was not given"))
=> (y 1)
u'v was given'
=> (y)
u'v was not given'
=> 
ryan@DevPC-LX:~$ 
```

**Comment by kirbyfan64** [comment]
@gilch But I did add `&kwargs`.

Tell me when to squash the commits!

**Comment by gilch** [comment]
This test won't fail if `defmacro` starts accepting `&kwonly` again. Read it carefully. It only asserts that *if* there's a `HyTypeError`, that it has the stated message. If there's not an error it doesn't complain.
```Hy
(try
    (eval '(defmacro f [&kwonly a b]))
    (except [e HyTypeError]
      (assert (= e.message "macros cannot use &kwonly"))))
```
You need to detect the case where no exception happens, something like this:
```Hy
(try
    (do 
      (eval '(defmacro f [&kwonly a b]))
      (raise (Exception "HyTypeError expected for &kwonly in defmacro")))
    (except [e HyTypeError]
      (assert (= e.message "macros cannot use &kwonly"))))
```

kirbyfan64:
@gilch Fixed. I also squashed the commits.

gilch:
We've got merge conflicts now, can you fix those?

There's a good enough case for keeping `&optional`, I think. I'm still not sure about `&key` though. The only documentation on `&key` says
> The difference here is that since it’s a dictionary, you can’t rely on any specific ordering to the arguments.

Yet we *must* rely on ordering of the arguments in a macro because we *can't* supply them using keywords.

kirbyfan64:
@gilch Fixed.

paultag:
LGTM; RFR @hylang/core

gilch:
What possible use is `&key` in a Hy macro? That still doesn't appear to be disallowed by this PR. Show me a good use case and I'll merge it now, otherwise, can you give an error for `&key` too?

PR #962, opened 2015-10-14
variadic if

### LINKED ISSUE #830: better elif

gilch:
The `if` form now supports elif clauses.
It works like `cond` but without the implicit `do`.

closes #830.

The old `if` is still available as `if*`, but you don't need it since the 2- and 3- argument cases work exactly as before.

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -823,26 +823,47 @@ keyword, the second function would have raised a ``NameError``.
     (set-a 5)
     (print-a)
 
-if / if-not
------------
+if / if* / if-not
+-----------------
 
 .. versionadded:: 0.10.0
    if-not
 
-``if`` is used to conditionally select code to be executed. It has to contain a
-condition block and the block to be executed if the condition block evaluates
-to ``True``. Optionally, it may contain a final block that is executed in case
-the evaluation of the condition is ``False``.
+``if / if* / if-not`` respect Python *truthiness*, that is, a *test* fails if it
+evaluates to a "zero" (including values of ``len`` zero, ``nil``, and
+``false``), and passes otherwise, but values with a ``__bool`` method
+(``__nonzero__`` in Python 2) can overrides this.
 
-``if-not`` is similar, but the second block will be executed when the condition
-fails while the third and final block is executed when the test succeeds -- the
-opposite order of ``if``.
+The ``if`` macro is for conditionally selecting an expression for evaluation.
+The result of the selected expression becomes the result of the entire ``if``
+form. ``if`` can select a group of expressions with the help of a ``do`` block.
+
+``if`` takes any number of alternating *test* and *then* expressions, plus an
+optional *else* expression at the end, which defaults to ``nil``. ``if`` checks
+each *test* in turn, and selects the *then* corresponding to the first passed
+test. ``if`` does not evaluate any expressions following its selection, similar
+to the ``if/elif/else`` control structure from Python. If no tests pass, ``if``
+selects *else*.
+
+The ``if*`` special form is restricted to 2 or 3 arguments, but otherwise works
+exactly like ``if`` (which expands to nested ``if*`` forms), so there is
+generally no reason to use it directly.
+
+``if-not`` is similar to ``if*`` but the second expression will be executed
+when the condition fails while the third and final expression is executed when
+the test succeeds -- the opposite order of ``if*``. The final expression is
+again optional and defaults to ``nil``.
 
 Example usage:
 
 .. code-block:: clj
 
-    (if (money-left? account)
+    (print (if (< n 0.0) "negative"
+               (= n 0.0) "zero"
+               (> n 0.0) "positive"
+               "not a number"))
+
+    (if* (money-left? account)
       (print "let's go shopping")
       (print "let's go and work"))
 
@@ -850,9 +871,6 @@ Example usage:
       (print "let's go and work")
       (print "let's go shopping"))
 
-Python truthiness is respected. ``None``, ``False``, zero of any numeric type,
-an empty sequence, and an empty dictionary are considered ``False``; everything
-else is considered ``True``.
 
 
 lif and lif-not
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -159,20 +159,25 @@
   ret)
 
 
-(defmacro if-not [test not-branch &optional [yes-branch nil]]
+(defmacro if-not [test not-branch &optional yes-branch]
   "Like `if`, but execute the first branch when the test fails"
-  (if (nil? yes-branch)
-    `(if (not ~test) ~not-branch)
-    `(if (not ~test) ~not-branch ~yes-branch)))
+  `(if* (not ~test) ~not-branch ~yes-branch))
 
 
-(defmacro lif [test &rest branches]
+(defmacro lif [&rest args]
   "Like `if`, but anything that is not None/nil is considered true."
-  `(if (is-not ~test nil) ~@branches))
+  (setv n (len args))
+  (if* n
+       (if* (= n 1)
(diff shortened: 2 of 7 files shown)


tuturto:
Looking sweet. Can't wait for this to be available. Can you add docs for this?

gilch:
Just realized this kind of breaks `lif`. The 2- or 3-arg versions work as before, but truthiness is Pythonic for the elif (ellif?) clauses, which would be surprising. I wonder if there are other unintended consequences. I think I'll have to rewrite `lif`, but it doesn't look difficult.

gilch:
Hm. Why is this failing? I ran nosetests before committing "variadic lif" without trouble. What was merged between my two commits? #920 "if-no-waste" (05857544f1cedadf69cb3393ec4d4a42eee64b61) and #950 "f/one-arg-comparators" (b875feccff5943a51df9e81ce7b4906e8acbcea5).

The complaint is `HyMacroExpansionError: expanding `cond': UnboundLocalError("local variable '_hy_anon_var_1' referenced before assignment",)`. That's not supposed to be possible just from writing a macro. Something's wrong with the compiler.

gilch:
I think I found it. The "if-no-waste" is looking for `'if'`, but should look for `if*` now.

gilch:
Are the `-*` functions like `for*` part of the public API, or are the considered implementation details subject to change? It might affect how it's documented. I'm assuming public for now.

gilch:
@tuturto docs added.

tuturto:
looking good, :+1: from me. Could squash and rebase before merge, we have been a bit sloppy about that.

gilch:
@kirbyfan64 docs fixed.
@tuturto squashed and rebased to hylang/master. All checks have passed.

gilch:
@hylang/core variadic if looks ready, but need second ack.

docs/language/api.rst
```diff
@@ -823,26 +823,47 @@ keyword, the second function would have raised a ``NameError``.
...
 .. versionadded:: 0.10.0
    if-not
 
-``if`` is used to conditionally select code to be executed. It has to contain a
-condition block and the block to be executed if the condition block evaluates
-to ``True``. Optionally, it may contain a final block that is executed in case
-the evaluation of the condition is ``False``.
+``if / if* / if-not`` respect Python *truthiness*, that is, a *test* fails if it
+evaluates to a "zero" (including values of ``len`` zero, ``nil``, and
+``false``), and passes otherwise, but values with a ``__nonzero__`` method can
```
zackmdavis: "values with a `__bool__` method (or `__nonzero__` on Python 2)"?

docs/language/api.rst
```diff
@@ -823,26 +823,47 @@ keyword, the second function would have raised a ``NameError``.
...
 
 Example usage:
 
 .. code-block:: clj
 
-    (if (money-left? account)
+    (print (if (< n 0.0) "negative"
+               (= n 0.0) "zero"
+               (> n 0.0) "positive"
+               "not a number"))
```
zackmdavis: (I don't immediately have an alternative example to propose, but this example is a little awkward in that it's not immediately obvious what input would trigger the `"not a number"` branch&mdash;I think it would need to be a custom class for which `__lt__`, `__gt__`, and `__eq__` return False for input `0.0`.)

hy/core/macros.hy
```diff
@@ -153,20 +153,25 @@
   ret)
 
 
-(defmacro if-not [test not-branch &optional [yes-branch nil]]
+(defmacro if-not [test not-branch &optional yes-branch ]
```
zackmdavis: (accidental extra space between `yes-branch` and the close-bracket)

tests/native_tests/language.hy
```diff
@@ -274,6 +274,32 @@
   (assert (= (cond) nil)))
 
 
+(defn test-if []
```
zackmdavis: I see integers, `nil`, and the empty string in this test, but I think we should exercise the paradigmatic conditions `true` and `false` here, too. (For exhaustiveness, we could also test nonempty strings, empty and nonempty collections, floats, and custom classes. This is perhaps pedantic&mdash;I don't _actually_ expect these extra cases to uncover any bugs in the implementation&mdash;but testing paranoia is probably healthy.)

zackmdavis:
***+1*** (I left some quibbling comments, but nothing that can't be addressed in a follow-up commit.)

docs/language/api.rst
gilch: I thought I spelled it out well enough, but if you missed it, that's evidence against my assumption. The value for `n` that would cause the form to print "not a number" is `nan`, which is, by the way, is short for "not a number". Try it at the repl. `Nan` does indeed fail all three comparisons. I thought that those familiar with electronic floating point math would know about this, that's why I compared against `0.0` instead of just `0` as an extra hint.

I could change the code block to show a repl session that includes defining the form in question in a function and passing `nan` as a possible `n` argument, if you think that's preferable. It does make the example longer, though I worry that to someone unfamiliar with the properties of `nan` it might not be any clearer.

docs/language/api.rst
gilch: Good catch here. I think it would be preferable if Hy were consistent across host implementations, but that seems difficult to fix in this case. I'll mention `__bool__` as the default and `__nonzero__` as the special case for Python2.

docs/language/api.rst
zackmdavis: Oh, right!&mdash;I am aware of floating-point NaN, although its existence would seem to have momentarily slipped my mind while I was writing the grandparent comment. The example is fine.

tests/native_tests/language.hy
gilch: I think the point here is to test Hy, not Python, which has its own tests.

PR #973, opened 2015-10-17
docs/tutorial: Add a short (require) example

### LINKED ISSUE #966: Mention require in tutorial
Ref #965

Those familiar with Clojure would think to look for `require`, but anyone coming from Python would have no idea how to import macros. We can't expect them to search through every form in the reference.

martinakcsilla:
Adds a short (require) example, along with a few words on why macros can't be imported.

Closes #966.

Signed-off-by: Csilla Nagyne Martinak <csilla@csillger.hu>

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -557,9 +557,20 @@ characters that soon):
   => #↻(1 2 3 +)
   6
 
-Macros are useful when one wished to extend the Hy or write their own
+Macros are useful when one wishes to extend the Hy or write their own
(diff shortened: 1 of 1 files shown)


tuturto:
Looking good. Could you still update `import` and `require` use double backtick as in: ``import`` so they'll render correctly? Thanks for the doc update, they're always lagging behind.

martinakcsilla:
Updated. Didn't notice they rendered badly, sorry.

PR #983, opened 2015-10-27
Enable automated testing with Python 3.5

felixonmars:


gilch:
should also add to tox?

felixonmars:
py34 is also missing from tox, though. I'll add both then :)

berkerpeksag:
Could you also update trove classifiers? https://github.com/hylang/hy/blob/master/setup.py#L91

A little bit off-topic, but we should not duplicate the testing logic in both .travis.yml and tox.ini. I'd prefer to handle all things in tox.ini and use TOX_ENV in .travis.yml. See https://github.com/benoitc/gunicorn/blob/master/.travis.yml and https://github.com/benoitc/gunicorn/blob/master/tox.ini for example.

felixonmars:
classifiers updated. I think TOX_ENV is good, and I can update this PR if decided :)

berkerpeksag:
Sorry for my late response. Yes, let's use tox in .travis.yml. Please squash your commits when you're done :)

felixonmars:
Updated.

`python: 3.5` was added due to a Travis issue: https://github.com/travis-ci/travis-ci/issues/4794#issuecomment-143758799

`passenv = TERM` was added to workaround a pypy failure under tox, according to https://botbot.me/freenode/pypy/2015-06-06/?msg=41126051&page=2

zackmdavis:
It's unfortunate that these configs make the Travis [build page](https://travis-ci.org/hylang/hy/builds/89031068) make all the individual job listings say "Python: 3.5", which is confusing, although I _think_ Tox is actually running the correct versions: the `TOXENV=pypy` one taking longer and printing the SSL `InsecurePlatformWarning` in the output is telling.

felixonmars:
Updated to make Travis display better table :)

berkerpeksag:
LGTM (just asked a question above)

Thanks @felixonmars :)

felixonmars:
A friendly ping :)

Foxboron:
Wops, sorry. LGTM
:shipit:

PR #989, opened 2015-11-19
Fix issue found by @tuturto in #903 (again!)

kirbyfan64:
Supersedes #958, whose commit history I somehow butchered. @gilch Can you try it?

### DIFF
--- a/hy/macros.py
+++ b/hy/macros.py
@@ -26,9 +26,10 @@ from hy.errors import HyTypeError, HyMacroExpansionError
 from hy._compat import PY3, PY34
 
 if PY3:
-    from inspect import getfullargspec as getargspec
+    from inspect import getfullargspec, getargspec
 else:
     from inspect import getargspec
+    getfullargspec = getargspec
 
 from collections import defaultdict
 import ast
@@ -132,7 +133,7 @@ def load_macros(module_name):
 
 
 def make_empty_fn_copy(fn):
-    argspec = getargspec(fn)
+    argspec = getfullargspec(fn)
 
     none = ast.Name(id='None', ctx=ast.Load())
     nonify = lambda _: none
--- a/hy/macros.py
+++ b/hy/macros.py
@@ -18,14 +18,20 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
-from inspect import getargspec, formatargspec
 from hy.models import replace_hy_obj, wrap_value
 from hy.models.expression import HyExpression
 from hy.models.string import HyString
 
 from hy.errors import HyTypeError, HyMacroExpansionError
+from hy._compat import PY3, PY34
+
+if PY3:
+    from inspect import getfullargspec as getargspec
+else:
+    from inspect import getargspec
 
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


gilch:
Merge conflicts now. @kirbyfan64 have you figured out what's causing the `getargspec` problem?

paultag:
Test cases? Also, what exactly is this doing? And mind merging with master?

gilch:
It's a lot of complexity for marginal benefit over pre- #924. There's six `if`s and a `try/except`. All code paths need unit tests. Testing is maybe not as hard as it sounds, since it takes different Python implementations to reach these anyway, so a lot of the paths could be reached from one test.

I'm still concerned about that `try/except`, since a the generic lambda it returns defeats the whole purpose of #924. Can you be more specific with the caught exception? Can you avoid generating an exception in the first place?

You can test #903 from that branch just by pasting the macro code in the repl.

gilch:
The code to paste in:
```Hy
(defmacro defmacro! [name args &rest body]
  "like defmacro/g! plus args that start with o! will automatically evaluate
   once only and be bound to the equivalent g!"
  (setv os (list-comp s [s args] (.startswith s "o!"))
        gs (list-comp (HySymbol (+ "g!" (cut s 2))) [s os]))
  `(defmacro/g! ~name ~args
     `(do (setv ~@(interleave ~gs ~os))
          ~@~body)))

(defn expensive-get-number [] (print "spam") 14)
(defmacro! triple-3 [o!n] `(+ ~g!n ~g!n ~g!n))
(triple-3 (expensive-get-number))
```

kirbyfan64:
Here's a smaller test case:

```
(defmacro f [@] `nil)
(f)
```

kirbyfan64:
All cleaned up and merged!

Kodiologist:
I've confirmed that the bug still exists, and there are no merge conflicts, but trying to launch Hy (with Python 3.5.2) raises `AttributeError: 'FullArgSpec' object has no attribute 'keywords'` in `macros.py`. Want to update it, Ryan?

kirbyfan64:
Ugh...this isn't going to work. Remember the `unsupported callable` error? It's because, on Python 3, *`getargspec` raises an error on invalid parameter names*.

So now we'd also have to remove support for non-identifier arguments...

Kodiologist:
Do you think that's a limitation that we should report upstream to the Python folks?

kirbyfan64:
@Kodiologist It's an explicit check though. I think it's just an intentional limitation...

Kodiologist:
> Might still run it by python-ideas though.

I think you should. If the idea is shot down, then apparently, we'll have to restrict ourselves to parameters that are legal Python identifiers. If it isn't, then we can open an issue to remind ourselves to fix this bug once Python allows us to.

kirbyfan64:
@Kodiologist  FWIW, wouldn't this fail then on Python versions <3.7?

Kodiologist:
Yeah, we'd have to gate the tests for this by Python version.

gilch:
> So now we'd also have to remove support for non-identifier arguments...

This is just for improved error messages, right? #924 introduced this problem. I don't think its worth restricting Hy like that just for better diagnostic information. We can come up with some other way to debug macros.

gilch:
Closed by #1172

PR #994, opened 2015-12-08
Fix typos

jwilk:
This PR was brought to you by [mwic](http://jwilk.net/software/mwic).


paultag:
@jwilk - you're the best!

PR #996, opened 2015-12-09
Typos in the tutorial

Eleonore9:
I was reading the tutorial and found typos.
I didn't think it needed a feature/fix branch (might be wrong).


paultag:
Yay! Thanks so much!!

Issue #982, opened 2015-10-26
Repl Bug


Arseny-N:
Hi, nice project ! 

But if in the REPL  you type 
```   => (:)  ```
it crashes 
```
Traceback (most recent call last):
  File "/usr/bin/hy", line 9, in <module>
    load_entry_point('hy==0.11.0', 'console_scripts', 'hy')()
  File "/usr/lib/python2.7/site-packages/hy/cmdline.py", line 347, in hy_main
    sys.exit(cmdline_handler("hy", sys.argv))
  File "/usr/lib/python2.7/site-packages/hy/cmdline.py", line 342, in cmdline_handler
    return run_repl(spy=options.spy)
  File "/usr/lib/python2.7/site-packages/hy/cmdline.py", line 240, in run_repl
    os=platform.system()
  File "/usr/lib64/python2.7/code.py", line 243, in interact
    more = self.push(line)
  File "/usr/lib64/python2.7/code.py", line 265, in push
    more = self.runsource(source, self.filename)
  File "/usr/lib/python2.7/site-packages/hy/cmdline.py", line 113, in runsource
    sys.stderr.write(str(e))
  File "/usr/lib/python2.7/site-packages/hy/errors.py", line 104, in __str__
    self.message))
  File "/usr/lib/python2.7/site-packages/clint/textui/colored.py", line 108, in __radd__
    return str(other) + str(self.color_str)
UnicodeEncodeError: 'ascii' codec can't encode character u'\ufdd0' in position 24: ordinal not in range(128)
```
Here is the version
```
$ hy -v
hy 0.11.0
```

**Comment by gilch** [comment]
I can confirm the error on master. It doesn't just throw an exception, but crashes the repl and kicks you back out into the shell. But only with Python2. It works fine using Python3:
```Hy
=> (:)
  File "<input>", line 1, column 1

  (:)
  ^-^
HyTypeError: `\ufdd0:' needs 1 arguments, got 0
```

Scorpil:
BTW, is there any reason to prefix keywords with \uFDD0? Couldn't find any please where it is actually useful.

paultag:
I really need to write up a copy paste explainer.

Yes. However, I want them to be real Objects. ':foo should be the same as
:foo in the resulting code.

Cheers,
  Paul

On Mon, Dec 14, 2015 at 4:02 PM, Andrew <notifications@github.com> wrote:

> BTW, is there any reason to prefix keywords with \uFDD0? Couldn't find any
[...]

PR #999, opened 2015-12-13
Fix cmdline error to string conversion.

Scorpil:
Fixes issue #982.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -66,3 +66,4 @@
 * Matthew Egan Odendahl <github.gilch@xoxy.net>
 * Tim Martin <tim@asymptotic.co.uk>
 * Johnathon Mlady <john@digitalvectorz.com>
+* Andrew Savchyn <dev@scorpil.com>
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -97,7 +97,7 @@ class HyREPL(code.InteractiveConsole):
             if e.source is None:
                 e.source = source
                 e.filename = filename
-            sys.stderr.write(str(e))
+            print(e, file=sys.stderr)
             return False
 
         try:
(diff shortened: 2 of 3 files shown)


paultag:
Hilarious, yeah. Totally.

What an awesome bug.

Right, so this looks good and it totally fixes the issue, so I say we're good. @hylang/core RFR

I mildly prefer `.format` to `%`, but it's OK. Other things in the compiler use modulo, so I'm tempted to punt on that.

Nice work thanks!

paultag:
@Scorpil Also, looks like you're a new contributor - mind opening a new PR adding yourself to AUTHORS?

berkerpeksag:
I'd prefer `print(str(e), file=sys.stderr)` over `sys.stderr.write("%s\n", str(e))`.

berkerpeksag:
Also, I don't fully understand how this solves #982. I think the main problem is in clint and we need to encode `self.message` before passing it to `colored.yellow()`: https://github.com/hylang/hy/blob/master/hy/errors.py#L104 I don't have a development environment right now so I might be horribly wrong :)

Thanks!

paultag:
@berkerpeksag I think you're right, I was also surprised this fixed the bug when I was just looking at the diff -- I your point really resonates with me right now, (as a strong believer of the unicode sandwich)

@berkerpeksag I'll leave this PR up to you!

Scorpil:
@berkerpeksag i've checked it - you're right. self.message there is correct unicode value on our side, so clint supposed to handle it correctly, but because of the bug it doesn't. I've sent PR https://github.com/kennethreitz/clint/pull/153 but it can take a while.

In the meantime, I've updated my fix to work around this bug by sending encoded string instead of unicode.

Foxboron:
This looks great!
LGTM!

berkerpeksag:
Looks great to me, too. Could you please squash your commits?

Thanks!

Scorpil:
Done. Glad to help.

PR #1002, opened 2015-12-17
Fix anaphoric macro `ap-if`.

ewaldgrusk:
I have rebased pull request #816 onto master and added a test case (as requested).

_Edit:_ Force-pushed an amendment that replaces `(except [e Exception] (assert (instance? HyMacroExpansionError e)))` with `(except [HyMacroExpansionError] true)` (as suggested by @kirbyfan64).

### DIFF
--- a/hy/contrib/anaphoric.hy
+++ b/hy/contrib/anaphoric.hy
@@ -25,8 +25,9 @@
 ;;; These macros make writing functional programs more concise
 
 
-(defmacro ap-if (test-form &rest args)
-  `(let [it ~test-form] (if it ~@args)))
+(defmacro ap-if [test-form then-form &optional else-form]
+  `(let [it ~test-form]
+     (if it ~then-form ~else-form)))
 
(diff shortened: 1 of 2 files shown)


agentultra:
Is there a particular reason for limiting the body to if*? Is there an issue attached to this?

ewaldgrusk:
My fault, I haven’t read the part of the documentation that states that »there is generally no reason to use it directly«.

zackmdavis:
I agree that the error message with the explicit arguments is better.

Before&mdash;
```
=> (ap-if true)

  (ap-if true)
  ^----------^
HyTypeError: `if' needs at least 2 arguments, got 1.
```

After&mdash;
```
=> (ap-if 2)

  (ap-if 2)
  ^-------^
HyMacroExpansionError: b"expanding `ap_if': missing 1 required positional argument: 'then_form'"
```
Thanks @ewaldgrusk!

PR #997, opened 2015-12-09
Fix Python 3.5.1 support by converting kwargs and stararg to str as late as possible (thus preserving positional information)

tianon:
Fixes #995

PR #1006, opened 2015-12-23
Backport fix Python 3.5.1 support by converting kwargs and stararg to str as late as possible

tianon:
(cherry picked from commit 9d8e4ddb8098dab39eb3d4f7ef89dc7ccd13a49f; PR #997)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,21 @@
 sudo: false
 language: python
 python:
-  - "pypy"
-  - "2.6"
-  - "2.7"
-  - "3.3"
-  - "3.4"
-cache: pip
-# command to run tests
-script: make travis
+  - 3.5
+env:
+  - TOXENV=py26
+  - TOXENV=py27
+  - TOXENV=py33
+  - TOXENV=py34
+  - TOXENV=py35
+  - TOXENV=pypy
+  - TOXENV=flake8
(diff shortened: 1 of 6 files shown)


tianon:
Will rebase for CI once #1007 settles.

tianon:
It might also make sense to include #983 here so we're testing Python 3.5 via CI for the new branch -- thoughts? (Happy to add those two commits if there's some agreement.)

paultag:
@tianon sounds legit to me, yep. That's the only way we can test the intended patch for this release, the 3.5 support

tianon:
How about #843?  If I add that one also, the 3.5 one applies completely clean without any conflicts (although I'm happy to resolve the conflicts instead if you'd rather).

paultag:
I'm reluctant to drop 2.6 in a micro stable release update

paultag:
I think we should just resolve conflicts for now

tianon:
Ok cool, will do.

tianon:
Whew, I had to also include #800 for complete Python 3.5 support, but that does it. :+1:  (Only `pypy` still running in CI now.)

tianon:
Woot, green! :metal:

paultag:
LGTM! RFR @hylang/core

PR #1008, opened 2015-12-23
Overhaul macros to allow macros to ref the Compiler

paultag:
This allows macros to take a keyword dict containing useful things by
defining a keyword argument. This allows us to pass in new objects
which might be handy to have in macros.

This changeset refactors module_name to become `compiler`, so that we
can pass the compiler itself into the macros as `opts['compiler']`.

This allows the macro to both get the macro name
(`compiler.module_name`), as well as use the compiler to build AST.

In the future, this will enable us to create "super-macros" which return
AST, not HyAST, in order to manually create insane things from userland.

For userland macros (not `defmacro`) the core.language `macroexpand`
will go ahead and make a new compiler for you.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1989,7 +1989,7 @@ class HyASTCompiler(object):
     @builds(HyExpression)
     def compile_expression(self, expression):
         # Perform macro expansions
-        expression = macroexpand(expression, self.module_name)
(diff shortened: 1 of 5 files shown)


olasd:
Looks sensible :+1:

Foxboron:
:shipit: +1

PR #1011, opened 2015-12-31
Update setup.py

2alex2:
Fix Python 3.5 compatibility for Windows.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ if sys.version_info[:2] < (2, 7):
     install_requires.append('argparse>=1.2.1')
     install_requires.append('importlib>=1.0.2')
 if os.name == 'nt':
-    install_requires.append('pyreadline==2.0')
+    install_requires.append('pyreadline>=2.0')
 
 setup(
     name=PKG,


Foxboron:
Looks great! Thank you.
:shipit:

zackmdavis:
Since pyreadline 2.0 isn't good for all the Pythons/OSes that we want to support (as evidenced by this helpful pull request and the PyPI notes describing 2.1 as a Python 3.5 compatibility bugfix), we might want to say `>=2.1` rather than `>=2.0`.

zackmdavis:
The recent [corroborating report](https://github.com/hylang/hy/issues/1022) makes me wonder if my previous comment—or rather, subsequent inaction—was a little too nitpicky. A not-abandoned (yet??) project should _not_ sit on patches that make things work that currently do not work. But I didn't want to merge without the requirement being tightened to `>=2.1`, and I didn't want to unilaterally tighten the requirement without actually testing it on Windows, and I didn't want to actually test it on Windows because that would take precious minutes that I don't necessarily have to spare. But between this report, #1022, the pyreadline release notes, and the limited amount of damage you can do to functionality that is _already_ broken, maybe I _can_ trust myself to _reason_ about the correctness of a one-character change without testing the result empirically? It is with this thought that go ahead and add 888c7efe0 to this branch and merge. (Travis can't help us with this one, all the CI build targets are Linux.)

Thanks, @2alex2! :sparkling_heart:

PR #1013, opened 2016-01-03
In python 3.3+, generator functions always return a value

### LINKED ISSUE #1010: Return value not generated when a function contains `yield from`
When a coroutine function uses `yield from`, the compiler no longer generates a return value for the final expression in the function. For example:

```hy
(defn yield-from-and-return []
  (print "whatever")
  (yield-from (some-coroutine))
  42)

(defn without-yield-from []
  (print "whatever")
  42)
```

expands to:

```python
def yield_from_and_return():
    print('whatever')
    yield from some_coroutine()
    42

def without_yield_from():
    print('whatever')
    return 42
```

timmartin:
This fixes #1010 for Python 3.3+. Unit tests check that the behaviour is unchanged in earlier versions, and I've checked they pass on 2.7,3.4 and PyPy.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2309,7 +2309,10 @@ class HyASTCompiler(object):
             return ret
 
         if body.expr:
-            if body.contains_yield:
+            if body.contains_yield and not PY33:
+                # Prior to PEP 380 (introduced in Python 3.3)
+                # generators may not have a value in a return
+                # statement.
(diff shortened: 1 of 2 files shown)


berkerpeksag:
LGTM, thanks!

paultag:
Also LGTM; thanks @timmartin ! This is great!

Issue #961, opened 2015-10-13
document &key


gilch:
The docs for `defn` seem to say `&key` exists, but there are no examples. If it's a just a misprint and doesn't actually exist, it should be removed. Otherwise, it needs examples.

kirbyfan64:
> there are no examples

Uhhh...

http://docs.hylang.org/en/latest/tutorial.html

Basically:

```hy
(defn func [&key {"a" "b" "c" "d"}] (+ a c))
```

is the same thing as:

```hy
(defn func [&optional [a "b"] [c "d"]] (+ a c))
```

gilch:
Okay, so it's a little bit documented. The tutorial is not a reference. There's nothing under the **&key** heading in the reference for `defn`.

kirbyfan64:
Outside of the realm of documentation: is there a reason `&key` allows anything other than identifiers in the mapping? For instance:

```
ryan@DevPC-LX:/media/ryan/stuff/hy$ hy --spy
hy 0.11.0 using CPython(default) 2.7.6 on Linux
=> (defn func [&key {1 "b"}] (+ 1 2))
def func(1=u'b'):
    return (1L + 2L)
=> (defn func [&key {a "b"}] (+ 1 2))
def func(a=u'b'):
    return (1L + 2L)
=> 
ryan@DevPC-LX:/media/ryan/stuff/hy$ hy --spy
hy 0.11.0 using CPython(default) 2.7.6 on Linux
=> (defn func [&key {a "b"}] (+ 1 2)) ; ok
def func(a=u'b'):
    return (1L + 2L)
=> (defn func [&key {"a" "b"}] (+ 1 2)) ; quotes work but are useless
def func(a=u'b'):
    return (1L + 2L)
=> (defn func [&key {1 "b"}] (+ 1 2)) ; why the hell would I want to do this
def func(1=u'b'):
    return (1L + 2L)
=> (defn func [&key {True "b"}] true) ; uh oh
def func(True=u'b'):
    return True
=> (defn func [&key {[1 2] "b"}] true) ; this doesn't even work
  File "<input>", line 1, column 19

  (defn func [&key {[1 2] "b"}] true) ; this doesn't even work
                    ^---^
HyTypeError: Can't assign or delete a HyList

=> (defn func [&key {# "b"}] true) ; this doesn't work either
def func():
    return True
=> (defn func [&key {` "b"}] true) ; useless again
def func(`=u'b'):
    return True
[...]

PR #1023, opened 2016-01-31
Keyword arguments to functions are required to be strings

timmartin:
See discussion in #961. Currently you can specify any type you like as the parameter "name" for a function when you use `&key` or `&optional`. This patch restricts it to be a `HyString`.


berkerpeksag:
LGTM

Thanks!

kirbyfan64:
:+1:

PR #1027, opened 2016-03-19
Give proper attribution

vog:
Add a link to XKCD to give proper attribution. Before this change, the attribution to XKCD 224 was only present in the alt text of the image, which is almost invisible in most browsers.


berkerpeksag:
Thanks!

PR #1034, opened 2016-03-29
Mention that issue isn't necessarily needed for PR

tuturto:


kirbyfan64:
IMO, I disagree with this. Sometimes, there's a feature that doesn't get into Hy for one reason or another, and the PR ends up getting closed. If you open an issue first:

1. The community has a chance to give input on the idea *before* it's implemented. This avoids 10k commits while the PR is open as everyone changes their mind on semantic issues.
2. You don't end up with wasted work if the PR ends up not being accepted.

tuturto:
I understand where you're coming from and both are very valid points. That's why I wrote "in general", but maybe we should expand that a bit with something along "Generally there is no need to open an issue before submitting a PR, but feel free to open one if you're unsure about the proposed change or it is a major change to Hy as a language (changed syntax for example)."?

Having to open an issue and then wait for feedback from community will slow things down (granted, which sometimes is a good thing) and probably discourages new contributors from opening their first PR.

kirbyfan64:
> That's why I wrote "in general", but maybe we should expand that a bit with something along "Generally there is no need to open an issue before submitting a PR, but feel free to open one if you're unsure about the proposed change or it is a major change to Hy as a language (changed syntax for example)."?

Agreed. Maybe something more specific, like:

> If you are contributing a major change to the Hy language (e.g. changing the behavior of or removing functions or macros), or you're unsure of the proposed change, please open an issue in the issue tracker before submitting the PR. This will allow others to give feedback on your idea, and it will avoid constant changes or wasted work. For other PRs (such as documentation fixes or code cleanup), you can directly open the PR without first opening a corresponding issue.

tuturto:
Agreed, that's much better. I updated the PR.

kirbyfan64:
:+1:

kirbyfan64:
Based on reading the Hy docs, I think it's OK for me to merge this (one ack for documentation stuff), so please don't shoot me...

PR #1037, opened 2016-03-31
add IRC info in readme

levthedev:
this was mentioned in github.com/hylang/hy/issues/1035


paultag:
👍👍👍👍👍👍

tuturto:
@levthedev thanks for writing this. I might be mistaken, but I think this is your first contribution to Hy. If that's the case, you could add your name to AUTHORS file.

We're looking forward for further contributions and thanks again.

levthedev:
Yes, it is! I'll get on that now :)

Issue #1019, opened 2016-01-26
Support for doctest with hy.


rduplain:
[This gist](https://gist.github.com/rduplain/c403ccc4eb9b17d92c27) has my investigation in all things hy + doctest: how to do it, limitations, an idea for `hy.contrib.doctest`, and a starting point for a py.test plugin to collect and run .hy doctests in a project.

A patch to hy could:

* [ ] Support the full Python doctest command-line program `python -m hy.contrib.doctest FILE` to run doctests in a specific module.
* [ ] Add a section to the docs on how to use doctest with hy.

I can put this together. I'm interested in reactions and/or alternative approaches.

gilch:
What would it take to adapt doctest to use a Hy repl instead of the Python one? The main reason you put tests in docstrings is to show examples of intended use. If the examples are in a different language it kind of defeats the purpose. You might as well just use normal unit tests.

rduplain:
The doctest module [parses](https://github.com/python/cpython/blob/d123b24fb4fa27544594e9e452ad582d492ccefd/Lib/doctest.py#L575) and [runs](https://github.com/python/cpython/blob/d123b24fb4fa27544594e9e452ad582d492ccefd/Lib/doctest.py#L1312) examples, so it'd take re-implementing that.

> The main reason you put tests in docstrings is to show examples of intended use.

+1

> If the examples are in a different language it kind of defeats the purpose.

Not if you are calling your code from Python, not hy. It's still relevant to show Python-based examples.

Kodiologist:
This would be really nice to have, so we could automatically check that all our documentation examples are correct as part of running the test suite. But I don't want it badly enough to put in the considerable effort that I suspect will be required. I'm going to leave the issue open in the hope that some kind soul will be interested someday.

Issue #356, opened 2013-12-04
Make `help()` recogonize macros as well


theanalyst:
It would be nice to have `help` recogonize macros as well. Ideas welcome...

Issue #186, opened 2013-05-20
Macros can't use previous definitions from the same module


khinsen:
The following code works when pasted into a REPL, but not in a module:

```
(defn foo [x y]
  (quasiquote (+ (unquote x) (unquote y))))

(defmacro bar [x y]
  (foo x y))

(bar 1 2)
```

The cause is that the definition of `foo` hasn't been compiled yet when `bar` is executed. There is a special hack in `compile_macro` that compiles and executes macro definitions immediately to make them available, but it doesn't handle the code that a macro definition could depend on.

The only solution I see is to process Hy modules as a sequence of top-level forms, each of which is compiled and evaluated before the next one is looked at. Does anyone see a problem with that, other than that it makes things a bit more complicated?

PR #1040, opened 2016-04-11
Compiler docs

jkitchin:
All this pull request does is add one line to the checkargs decorator to set its **doc** attribute on the returned function, and add docstrings to all the functions decorated by @builds. A few docstrings are marked TODO. I used the documentation at http://docs.hylang.org/en/latest/ for this.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -3056,6 +3056,7 @@ class HyASTCompiler(object):
 
     @builds(HyDict)
     def compile_dict(self, m):
+        """Compile a dictionary."""
         keyvalues, ret, _ = self._compile_collect(m)
 
         ret += ast.Dict(lineno=m.start_line,
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2785,6 +2785,33 @@ class HyASTCompiler(object):
     @builds("defclass")
     @checkargs(min=1)
     def compile_class_expression(self, expressions):
+        """New classes are declared with defclass.
+
+        It can takes two optional parameters: a vector defining a possible
+        super classes and another vector containing attributes of the new
+        class as two item vectors.
+
+        (defclass class-name [super-class-1 super-class-2]
+          [attribute value]
+
+          (defn method [self] (print "hello!")))
+
+        Both values and functions can be bound on the new class as shown by
+        the example below:
+
+        => (defclass Cat []
+        ...  [age None
+        ...   colour "white"]
+        ...
+        ...  (defn speak [self] (print "Meow")))
+
+        => (def spot (Cat))
+        => (setv spot.colour "Black")
+        'Black'
+        => (.speak spot)
+        Meow
+
+        """
         def rewire_init(expr):
             new_args = []
             if expr[0] == HySymbol("setv"):
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2540,6 +2540,7 @@ class HyASTCompiler(object):
     @builds("for*")
     @checkargs(min=1)
     def compile_for_expression(self, expression):
+        """TODO: docstring"""
         expression.pop(0)  # for
 
         args = expression.pop(0)
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2111,6 +2111,19 @@ class HyASTCompiler(object):
     @builds("and")
     @builds("or")
     def compile_logical_or_and_and_operator(self, expression):
+        """and is used in logical expressions.
+
+        It takes at least two parameters. If all parameters evaluate to True,
+        the last parameter is returned. In any other case, the first false
+        value will be returned.
+
+        or is used in logical expressions.
+
+        It takes at least two parameters. It will return the first non-false
+        parameter. If no such value exists, the last parameter will be
+        returned.
+
+        """
         ops = {"and": (ast.And, "True"),
                "or": (ast.Or, "None")}
         operator = expression.pop(0)
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2926,6 +2926,10 @@ class HyASTCompiler(object):
     @builds("defreader")
     @checkargs(min=2)
     def compile_reader(self, expression):
+        """defreader defines a reader macro, enabling you to restructure or modify
+        syntax.
+
+        """
         expression.pop(0)
         name = expression.pop(0)
         NOT_READERS = [":", "&"]
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1725,6 +1725,7 @@ class HyASTCompiler(object):
     @builds("with*")
     @checkargs(min=2)
     def compile_with_expression(self, expr):
+        """TODO: what does this do?"""
         expr.pop(0)  # with*
 
         args = expr.pop(0)
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2244,6 +2244,18 @@ class HyASTCompiler(object):
     @builds("not_in")
     @checkargs(min=2)
     def compile_compare_op_expression_coll(self, expression):
+        """Collection operators
+
+        => (setv a nil)
+        => (is a nil)
+        True
+
+        => (in 2 [1 2 3])
+        True
+
+        => (not-in 4 [1 2 3])
+        True
+        """
         return self._compile_compare_op_expression(expression)
 
     @builds("%")
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2995,6 +2995,7 @@ class HyASTCompiler(object):
 
     @builds(HyInteger)
     def compile_integer(self, number):
+        """Compile an Integer."""
         return ast.Num(n=long_type(number),
                        lineno=number.start_line,
                        col_offset=number.start_column)
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2225,6 +2225,9 @@ class HyASTCompiler(object):
     @builds(">=")
     @checkargs(min=1)
     def compile_compare_op_expression(self, expression):
+        """Comparison operators.
+
+        """
         if len(expression) == 2:
             rval = "True"
             if expression[0] == "!=":
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1515,6 +1515,29 @@ class HyASTCompiler(object):
     @builds(".")
     @checkargs(min=1)
     def compile_attribute_access(self, expr):
+        """. is used to perform attribute access on objects.
+
+        It uses a small DSL to allow quick access to attributes and items in a
+        nested data structure.
+
+        For instance,
+
+        (. foo bar baz [(+ 1 2)] frob)
+        Compiles down to:
+
+        foo.bar.baz[1 + 2].frob
+
+        . compiles its first argument (in the example, foo) as the object on
+        which to do the attribute dereference. It uses bare symbols as
+        attributes to access (in the example, bar, baz, frob), and compiles
+        the contents of lists (in the example, [(+ 1 2)]) for indexation.
+        Other arguments raise a compilation error.
+
+        Access to unknown attributes raises an AttributeError. Access to
+        unknown keys raises an IndexError (on lists and tuples) or a KeyError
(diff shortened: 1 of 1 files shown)


kirbyfan64:
I'm not sure about this...the majority of these are duplicated from the main docs, resulting in having to update them in two places, and the `compile_*` functions seem pretty self-explanatory (e.g. `compile_while` compiles a `while` expression).

berkerpeksag:
I agree with @kirbyfan64. Perhaps we can document briefly what `compile_*` functions do in `HyASTCompiler` docstring or in the internals docs at http://docs.hylang.org/en/latest/language/internals.html

jkitchin:
The point of this is to be able to get help from within hy on what the compiler functions do.  See http://kitchingroup.cheme.cmu.edu/blog/2016/04/03/Getting-hylp-in-hy/ for example. None of the functions defined in compiler.py have any docstrings attached to them, so they are not introspectable in an editor to find out what they do, e.g. http://kitchingroup.cheme.cmu.edu/blog/2016/04/06/Another-step-towards-HyDE/.

This also makes it possible to auto-generate documentation, rather than update it in one place that isn't the code, or two places.

kirbyfan64:
I guess that makes sense...but it would first require the ability to even auto-generate the docs (which isn't possible ATM).

But I still think some of the docstrings, like `compile_complex`, are a little overkill, since the function names are rather self-explanatory.

gilch:
I like the idea of auto-generating the reference docs from the docstrings. They're not identical currently, so we'd have to pick the best of the two or merge them somehow. The examples could also be doctests, once we get that working for Hy code. (see #1019.) This way, we can always be sure the examples work as intended, and we can easily access the references from the repl.

The rest of the docs will still have to be updated by hand, but we should be able to automatically test the examples there too.

gilch:
Maybe we should mark some of these as implementation detail subject to change without notice, rather than part of the public API. A simple docstring like "implementation detail: do not use" would suffice.

jkitchin:
All of these are used by users I think, but not in an obvious way to
users. For example:

(cut col 0 1 2)

uses compile_cut_expression. A user would not directly call this, but by
putting a docstring on it, you can access what cut is supposed to do.

e.g.

#+BEGIN_SRC hy
(import hy)
(for [x  (hy.compiler._compile_table.items)]
  (when (= "cut" (get x 0))
    (print (. x [1] __doc__))))
#+END_SRC

#+RESULTS:
#+begin_example
cut can be used to take a subset of a list and create a new list from it.

        The form takes at least one parameter specifying the list to cut. Two
        optional parameters can be used to give the start and end position of
        the subset. If they are not supplied, the default value of None will
        be used instead. The third optional parameter is used to control step
        between the elements.

        cut follows the same rules as its Python counterpart. Negative indices
        are counted starting from the end of the list.


#+end_example

Matthew Egan Odendahl writes:

> Maybe we should mark some of these as implementation detail subject to change without notice, rather than part of the public API. A simple docstring like "implementation detail: do not use" would suffice.
>
> ---
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/hylang/hy/pull/1040#issuecomment-208468852


--
Professor John Kitchin
Doherty Hall A207F
Department of Chemical Engineering
Carnegie Mellon University
Pittsburgh, PA 15213
412-268-7803
@johnkitchin
http://kitchingroup.cheme.cmu.edu

jkitchin:
It isn't that far off though. In this post
http://kitchingroup.cheme.cmu.edu/blog/2016/04/03/Getting-hylp-in-hy/ I
generated documentation for everything I could find at the time (even a
list of the compiler functions, just no documentation for them).

Self explanatory is in the eye of the beholder ;) For people brand-new
to this, they might disagree. At some point I think having examples that
are doctestable would make it worth having even trivial ones.




Ryan Gonzalez writes:

> I guess that makes sense...but it would first require the ability to even auto-generate the docs (which isn't possible ATM).
>
> But I still think some of the docstrings, like `compile_complex`, are a little overkill, since the function names are rather self-explanatory.
>
> ---
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/hylang/hy/pull/1040#issuecomment-208449607


--
Professor John Kitchin
Doherty Hall A207F
Department of Chemical Engineering
Carnegie Mellon University
Pittsburgh, PA 15213
412-268-7803
@johnkitchin
http://kitchingroup.cheme.cmu.edu

kirbyfan64:
@jkitchin Good point about doctest; that would be really useful, since the documentation examples frequently go out of date.

jkitchin:
For example,

These examples didn't work for me today:

http://docs.hylang.org/en/latest/language/api.html#unquote

I gathered that name is a function, and def doesn't redefine it.

=> (def name "Cuddles")
=> (quasiquote (= name (unquote name)))
(u'=' u'name' <function name at 0x102ee3668>)

The docs advertise:
(def name "Cuddles")
(quasiquote (= name (unquote name)))
;=> (u'=' u'name' u'Cuddles')

Ryan Gonzalez writes:

> @jkitchin Good point about doctest; that would be really useful, since the documentation examples frequently go out of date.
>
> ---
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/hylang/hy/pull/1040#issuecomment-208488746


--
Professor John Kitchin
Doherty Hall A207F
Department of Chemical Engineering
Carnegie Mellon University
Pittsburgh, PA 15213
412-268-7803
@johnkitchin
http://kitchingroup.cheme.cmu.edu

kirbyfan64:
@jkitchin That should work outside the REPL. When you call `name`, Hy re-imports it from `hy.core.builtins`, which overrides anything you had assigned to it.

gilch:
We could override Python's help function with a macro that checks if the symbol is a macro or built-in hy special form, and if so fetches the appropriate docstring. If not it can defer to the built-in Python help. This would help with #356 and #892.

jkitchin:
That is more or less what I was aiming at in this code:
https://github.com/jkitchin/hy/blob/hydoc/hy/core/hylp.hy
That code does the language, shadowed functions, keywords and macros.

I didn't get to deferring to built-in Python help yet, but I would like to
get there!

John

-----------------------------------
Professor John Kitchin
Doherty Hall A207F
Department of Chemical Engineering
Carnegie Mellon University
Pittsburgh, PA 15213
412-268-7803
@johnkitchin
http://kitchingroup.cheme.cmu.edu

jkitchin:
Here goes a hydoc macro that covers functions in hy, including macros, and Python functions. I think this should resolve #892 and #356.

```
(defmacro hydoc [sym]
  "Return docstring for SYM.
  Look for hy functions first, then defer to Python."
  (let [code `(do
               (import hy)
               (import pydoc))]
    (.append code
       (cond
        ;; language functions
        [(in sym (. hy core language *exports*))
         `(pydoc.getdoc (. hy core language ~(HySymbol sym)))]

        ;; shadowed functions
        [(in sym (. hy core shadow *exports*))
         `(pydoc.getdoc (. hy core shadow  ~(HySymbol sym)))]

        ;; Macros
        [(in sym (.keys (get hy.macros._hy_macros nil)))
         `(pydoc.getdoc (get hy.macros._hy_macros nil ~sym))]

        ;; Functions in compiler
        [(in sym (list-comp (get x 0) [x (hy.compiler._compile_table.items)]
                            (string? (get x 0))))
         `(pydoc.getdoc (get hy.compiler._compile_table ~sym))]

        ;; Fallback to Python
        [true
         `(try
           (pydoc.getdoc ~(HySymbol sym))
           (except [e NameError] (.format "{} Not found." ~sym)))]))
    code))
```

gilch:
Maybe more like this?
```Hy
(defn hydoc* [sym]
  (import hy)
  (. (or (->> false
            (.get hy.compiler._compile_table sym)
            (.get (get hy.macros._hy_macros nil) sym)
            (.get hy.core.shadow.__dict__ sym)
            (.get hy.core.language.__dict__ sym))
         (eval sym))
     __doc__))

(defmacro hydoc [sym]
  `(hydoc* '~sym))
```

gilch:
You could override Python's interactive help using these too:
```Hy
(defmacro help [item]
  (if (isinstance item HySymbol)
    `(print (hydoc ~item))
    `((get __builtins__ "help") ~item)))
```

jkitchin:
That is pretty nice, I haven't used the threading macros before. Pretty
concise. I think this might be a bit more robust, it also gets comments
if there is no __doc__ attribute


```(defn hydoc* [sym]
  (import hy)
  (import pydoc)
  (pydoc.getdoc (or (->> false
            (.get hy.compiler._compile_table sym)
            (.get (get hy.macros._hy_macros nil) sym)
            (.get hy.core.shadow.__dict__ sym)
            (.get hy.core.language.__dict__ sym))
         (eval sym))))
```

Just curious is the hydoc* an indication it is a function (the * at the
end) to differentiate it from the macro by the similar name?

Matthew Egan Odendahl writes:

> Maybe more like this?
> ```Hy
> (defn hydoc* [sym]
>   (import hy)
>   (. (or (->> false
>             (.get hy.compiler._compile_table sym)
>             (.get (get hy.macros._hy_macros nil) sym)
>             (.get hy.core.shadow.__dict__ sym)
>             (.get hy.core.language.__dict__ sym))
>          (eval sym))
>      __doc__))
>
> (defmacro hydoc [sym]
>   `(hydoc* '~sym))
> ```
>
> ---
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/hylang/hy/pull/1040#issuecomment-208675587


--
Professor John Kitchin
Doherty Hall A207F
Department of Chemical Engineering
Carnegie Mellon University
Pittsburgh, PA 15213
412-268-7803
@johnkitchin
http://kitchingroup.cheme.cmu.edu

gilch:
>Just curious is the hydoc* an indication it is a function (the * at the
end) to differentiate it from the macro by the similar name?

Not exactly. In mathematics, you often talk about some object `g` and some related object `g'` (g *prime*). In Lisp, the apostrophe is already taken to mean quotation, so we use `*` for this purpose. Usually you have a `*` form that is syntactically annoying to use doing the real work, and some nicer interface to it without the `*`. This is often, but not necessarily, a macro. Hy inherits some design philosophy from Clojure, where you see this pattern a lot internally. In Hy, for example we have `for` that uses `for*` and `if` that uses `if*`.

In this case, the macro lets us avoid quoting the symbol when calling the function, but we still have the function when we need it to e.g. map over a list of symbols and return all of their docstrings. This would be more difficult with only the macro.

I'm less confident this is the right approach for Hy (at least in this instance), because of #1041, (Hy's eval doesn't work like Clojure, we might change that #1043.). We wouldn't need that eval if it was all in the macro. You could just unquote it.

gilch:
The "TODO" docstrings in this PR are for:

- [ ] `except` (used in `try`)
- [ ] `with*` (used in `with`)
- [ ] `for*` (used in `for`)
- [ ] `HyList` (a [Hy Model](http://docs.hylang.org/en/latest/language/internals.html))
- [ ] `HySet` (another Hy Model)
- [ ] `dispatch-reader-macro` (used for [reader macros](http://docs.hylang.org/en/latest/language/readermacros.html?highlight=dispatch_reader_macro))
- [ ] `eval-and-compile`
- [ ] `eval-when-compile`

`eval-and-compile` and `eval-when-compile` were never documented. I didn't know what these do either. I don't see them used anywhere but tests. My best guess is that they were [borrowed](https://www.gnu.org/software/emacs/manual/html_node/elisp/Eval-During-Compile.html) from Emacs Lisp. The blame shows @khinsen creating them in #186, which confirms this. That PR, and the related issue #186, could help us document them.

Kodiologist:
It doesn't look like any core members have approved this. While it would be great to have docstrings, I don't want to have the same documentation appear in two different places in the codebase, like @kirbyfan64 was talking about. I think we need to have a way for docstrings to update the manual  (e.g., #1044) or vice versa. Copying and pasting is no good even as a temporary solution.

PR #1043, opened 2016-04-12
Eval should use globals + locals

### LINKED ISSUE #1041: eval can only see locals
Eval works on symbols for modules:

``` Python
=> (import operator)
import operator
=> (eval 'operator)
from hy.importer import hy_eval
from hy import HySymbol
hy_eval(HySymbol('operator'), locals(), '__console__')
<module 'operator' from 'C:\\Python34\\Scripts\\env-hy\\lib\\operator.py'>
```

Functions can return modules:

``` Python
=> ((fn [] operator))
def _hy_anon_fn_1():
    return operator
_hy_anon_fn_1()
<module 'operator' from 'C:\\Python34\\Scripts\\env-hy\\lib\\operator.py'>
```

Functions can return the result of an eval:

``` Python
=> ((fn [] (eval 'first)))
from hy.importer import hy_eval
from hy import HySymbol

def _hy_anon_fn_1():
    return hy_eval(HySymbol('first'), locals(), '__console__')
_hy_anon_fn_1()
<function first at 0x000000000369DC80>
```

But not if it's a module!

``` Python
=> ((fn [] (eval 'operator)))
from hy.importer import hy_eval
from hy import HySymbol

def _hy_anon_fn_1():
    return hy_eval(HySymbol('operator'), locals(), '__console__')
_hy_anon_fn_1()
  File "c:\users\me\code\repos\hy\hy\importer.py", line 134, in hy_eval
NameError: name 'operator' is not defined
```

I am surprised by this. Are you?

kirbyfan64:
Fixes #1041. @gilch Can you confirm this fixes it for you?

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -746,7 +746,21 @@ class HyASTCompiler(object):
         if len(expr) >= 2:
             elist.append(expr[1])
         else:
-            elist.append(HyExpression([HySymbol("locals")]))
+            dct = HySymbol(self.get_anon_var())
+            it = HySymbol(self.get_anon_var())
+
+            # Merge globals and locals, with locals taking precedence.
+            # Equivalent to (genexpr i [dct [(globals) (locals)] i (.items d)]).
+            genexpr_arg = HyList([HySymbol(dct),
+                                  HyList([HyExpression([HySymbol("globals")]),
+                                          HyExpression([HySymbol("locals")])]),
+                                  HySymbol(it),
+                                  HyExpression([HySymbol(".items"), dct])])
+
+            elist.append(HyExpression([HySymbol("dict"),
+                                       HyExpression([HySymbol("genexpr"),
+                                                     it,
+                                                     genexpr_arg])]))
 
         if len(expr) == 3:
             elist.append(expr[2])
--- a/tests/native_tests/language.hy
+++ b/tests/native_tests/language.hy
@@ -987,7 +987,8 @@
   (setv x (quote 42))
   (assert (= x (eval x)))
   (assert (= 27 (eval (+ (quote (*)) (* [(quote 3)] 3)))))
-  (assert (= None (eval (quote (print ""))))))
+  (assert (= None (eval (quote (print "")))))
+  (assert (= sys (eval (quote sys))))) ; issue #1041
 
 
 (defn test-eval-globals []


gilch:
Travis is complaining. Also globals+locals isn't the current environment, since nonlocals don't work e.g.
```Hy
(((fn []
    (setv spam 42)
    (fn [] (eval 'spam)))))
;; NameError
```
But something similar in Clojure would work:
```Clojure
user=> (((fn [] (let [spam 42] (fn [] (eval 'spam))))))
42
```
It also works in Hy without the eval:
```Hy
=> (((fn []
... (setv spam 42)
... (fn [] spam))))
def _hy_anon_fn_2():
    spam = 42

    def _hy_anon_fn_1():
        return spam
    return _hy_anon_fn_1
_hy_anon_fn_2()()
42
```

gilch:
Another potential issue: `eval` should be able to can mutate a global or local with `setv`. It currently can't mutate locals, but it can for globals. If you make a new dict with globals+locals, then eval can't update globals by writing to it either.

Overall, Hy's eval seems to work more like Python's than like Lisp's. I wonder if this is the wrong approach. How does our REPL even function without a proper eval? Can we just use that for eval instead?

agentultra:
A typical Lisp eval, as in Common Lisp, accepts a form as the parameter and computes its value using the *dynamic environment* to provide meaning for names in the form. The user eval is really a front-end for the internal one which takes an environment as a parameter as *macro-expand-1* does.

Maybe there’s a way to map the Python concept of module-level scope to a dynamic environment in Lisp?

kirbyfan64:
@gilch Well *that* sucks...

There's only one true solution I can think of: do it like shown in this PR, but also merge it with the result of a new function, `hy.?.get_freevars`, which would go something like this:

```python
def get_freevars(func):
    closure = func.__closure__
    freevars = func.__code__.co_freevars
    return dict(zip(freevars, (cell.cell_contents for cell in closure)))
```

Now, `fn` could be amended do always do an assignment like this as the first part of the function:

```python
def _hy_anon_fn_1(): # Or whatever...
    _hy_current_function = _hy_anon_fn_1
```

then `eval` could pass `_hy_current_function` to `get_freevars`. The assignment would only be necessary in the case of `eval` being used; otherwise, the compiler wouldn't need to put it.

gilch:
There's still the problem of assignment. What happens if you `(eval '(setv foo 42))`? I'm also not sure if that `__code__` trick will work in non-CPython implementations (pypy? jython?), or at least not the same way. 

I'm not sure how well the `_hy_current_function` will work. I can see that if you can access a closure like that, can you could implement `nonlocal` in Python 2, but what happens in this case?

```Python
def _hy_anon_fn_1():
    _hy_current_function = _hy_anon_fn_1
    foo = 42
    def _hy_anon_fn_2():
        _hy_current_function = _hy_anon_fn_1  # used in _hy_anon_fn_3, so compiler put it in.
        new_hy_eval(HySymbol('foo'),_hy_current_function)  # oops
        def _hy_anon_fn_3(): ...
```

kirbyfan64:
@gilch At this point, I would just deem assignment broken, especially since it's virtually impossible under Python 3 (in Python 2, you could just put `eval ''` at the top).

Back to the topic...

...

...actually, can't we just define `eval` in user space in `hy.core.macros`? It would have code duplication, but it would also be pretty much 100% guaranteed to work, since it's just a macro.

gilch:
That does sound like a better plan, if it works. However, that would mean you can't pass `eval` to higher-order functions, like `map`. Is there a workaround for that? Maybe we need two evals.

kirbyfan64:
@gilch You can't even do that right now; Hy will pass Python's `eval` instead. Maybe Hy's eval should be named `hy-eval` to make a distinction?

kirbyfan64:
Can't figure this out, and sadly don't have the time. :(

PR #1059, opened 2016-05-09
Include original XKCD title text in comic image

kadamwhite:
When I see an XKCD comic, or an adaptation thereof, my first instinct is to look for title text. Not finding any, this proposes adding the original title text back in; if a good Hy-specific adaptation is proposed that'd be even better.

And no, this isn't important :)


paultag:
:+1: LGTM! :D

Issue #739, opened 2014-12-30
Return expression


cwebber:
Heresy?  But maybe we should provide:

``` clojure
(defn try-this [arg]
 (if (= arg "immediately return")
   ;; bail out early
   (return arg))
 ;; a whole bunch of code in-between
 ;; [...]
 ;; The end-of-function return
 (something-something arg blah))
```

gilch:
Can I also vote for the return expression? I think Hy should be able to do anything Python can. I should be able to translate any Python code I see into near-equivalent Hy. Automatically even, with some kind of py2hy (de)compiler. This would be really hard without return. You would have to rewrite any Python code with multiple exits.

I'm fine with having an implicit return at the tail position. I'd prefer not to have to spell it out if I don't have to.

The return statement is Rossum's [recommended way](https://mail.python.org/pipermail/python-3000/2007-July/008663.html) of breaking out of nested loops. This isn't possible with only an implicit return.

If I'm writing the code myself from scratch, I'd prefer using higher-order functions over explicit nested loops anyway, but that doesn't mean I want to translate someone's working algorithm into that style just to insert some Hy macros.

Foxboron:
@hylang/core opinions? Might be a good idea for the cleanup

paultag:
I'm -1 on return, but let's open a bug and discuss. Basically, I'm worried
we'll wind up with a Ruby-style mess, and having return values be confusing
to grok.

On Fri, Jul 24, 2015 at 11:02 AM, Morten Linderud <notifications@github.com>
wrote:

> @hylang/core <https://github.com/orgs/hylang/teams/core> opinions? Might
> be a good idea for the cleanup
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hylang/hy/issues/739#issuecomment-124551856>.
>



-- 
:wq

algernon:
I'm also -1 on return. Mixing explicit and implicit returns would be confusing, in my opinion.

gilch:
I really don't think explicit returns are any more confusing than `break` or `continue` inside loops, or the whole exception system. You aren't proposing we cut those out too? You'll get used to it. As for mixing explicit and implicit, well, Ruby might be a bad example, but I'll point out that Rust also does it this way. Even Python has an implicit return in `lambda` expressions, but explicit `return` in function `def`s.

gilch:
I feel pretty strongly that Hy functions really ought to have multiple exits, because this is something Python already does, and "We're still Python(TM).".

[...]

Issue #691, opened 2014-11-21
py2 `yield-from` fails to compile if the last item in a function is an expression.


pyos:
For example, this code:

``` hy
(defn g []
    (yield-from (f)))
```

gives a

```
SyntaxError: 'return' with argument inside generator (<input>, line 1)
```

Similarly, any function that has a return value (does not end with a statement that is not an expression?) does not work either:

``` hy
=> (defn g []
...   (raise a)
...   (if b (c "a" "b") d)
...   (yield-from e)
...   (h 1 2 3))
Traceback (most recent call last):
...
SyntaxError: 'return' with argument inside generator (<input>, line 1)
```

An example of a function that does compile:

``` hy
=> (defn g []
...   (yield-from (f))
...   (assert true))  ; note that assert does not have a return value
=> (g)
<generator object g at 0x7ff9dbfecc30>
```

Kodiologist:
This one's still around as of 5bf9ecfc5a40821eb15e2ac7f218ae276f81fa60. That said, I'm inclined to think `yield-from` on Python 2 isn't worth supporting, so we should remove it rather than fix it.

tuturto:
While it was really neat thing to support yield-from on Python 2, I think it's too much effort to keep supporting it now. I think we should stop supporting yield-from on Python 2.

PR #1060, opened 2016-05-10
fix yield-from throw

### LINKED ISSUE #692: py2 `yield-from` does not propagate exceptions *into* the inner generators.
I've found a concrete example of the difference in semantics between the `yield-from` macro and [PEP-380](https://www.python.org/dev/peps/pep-0380#formal-semantics). When using `generator.throw` in Python 2, `yield-from` does not rethrow the exception in the inner generator:

``` hy
=> (defn f []
...   (yield 1)
...   (try
...     (yield 2)
...     (catch [e ValueError] (yield 3))))
=> (defn g []
...   (yield-from (f))
...   (assert true))  ; see issue #691
=> (setv x (g))
=> (next x)
1L
=> (next x)
2L
=> (x.throw ValueError)
  File "<input>", line 257, in g  ; note that ValueError didn't even reach `f`
ValueError
```

In Python 3, `ValueError` reaches `f` and is handled there as expected:

``` hy
...
=> (next x)
2
=> (x.throw ValueError)
3
```

gilch:
Closes #692.

### DIFF
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -219,36 +219,36 @@
        (try
          (setv ~g!y (next ~g!i))
          (except [~g!e StopIteration]
-                 (setv ~g!r (. ~g!e value)))
+           (setv ~g!r (next (iter (. ~g!e args)) nil)))
          (else
-           (while 1
+           (while True
              (try
                (setv ~g!s (yield ~g!y))
                (except [~g!e GeneratorExit]
-                       (try
-                         (setv ~g!m (. ~g!i close))
-                         (except [AttributeError] pass)
-                         (else (~g!m)))
-                       (raise ~g!e))
+                 (try
+                   (setv ~g!m (. ~g!i close))
+                   (except [AttributeError] pass)
+                   (else (~g!m)))
+                 (raise ~g!e))
                (except [~g!e BaseException]
-                       (setv ~g!x (sys.exc_info))
-                       (try
-                         (setv ~g!m (. ~g!i throw))
-                         (except [AttributeError] (raise ~g!e))
-                         (else
-                           (try
-                             (setv ~g!y (apply ~g!m ~g!x))
-                             (except [~g!e StopIteration]
-                                     (setv ~g!r (. ~g!e value))
-                                     (break))))))
+                 (setv ~g!x (sys.exc_info))
+                 (try
+                   (setv ~g!m (. ~g!i throw))
+                   (except [AttributeError] (raise ~g!e))
+                   (else
+                     (try
+                       (setv ~g!y (apply ~g!m ~g!x))
+                       (except [~g!e StopIteration]
+                         (setv ~g!r (next (iter (. ~g!e args)) nil))
+                         (break))))))
                (else
                  (try
                    (if (is ~g!s nil)
                      (setv ~g!y (next ~g!i))
                      (setv ~g!y ((. ~g!i send) ~g!s)))
                    (except [~g!e StopIteration]
-                           (setv ~g!r (. ~g!e value))
-                           (break))))))))
+                     (setv ~g!r (next (iter (. ~g!e args)) nil))
+                     (break))))))))
        ~g!r))
   nil)
 
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -214,20 +214,42 @@
 
 (if-python2
   (defmacro/g! yield-from [expr]
-    `(do (import types)
-         (setv ~g!iter (iter ~expr))
-         (setv ~g!return nil)
-         (setv ~g!message nil)
-         (while true
-           (try (if (isinstance ~g!iter types.GeneratorType)
-                  (setv ~g!message (yield (.send ~g!iter ~g!message)))
-                  (setv ~g!message (yield (next ~g!iter))))
-           (except [~g!e StopIteration]
-             (do (setv ~g!return (if (hasattr ~g!e "value")
-                                     (. ~g!e value)
-                                     nil))
-               (break)))))
-           ~g!return))
+    `(do
+       (setv ~g!i (iter ~expr))
+       (try
+         (setv ~g!y (next ~g!i))
+         (except [~g!e StopIteration]
+                 (setv ~g!r (. ~g!e value)))
+         (else
+           (while 1
+             (try
+               (setv ~g!s (yield ~g!y))
+               (except [~g!e GeneratorExit]
+                       (try
+                         (setv ~g!m (. ~g!i close))
+                         (except [AttributeError] pass)
+                         (else (~g!m)))
+                       (raise ~g!e))
+               (except [~g!e BaseException]
+                       (setv ~g!x (sys.exc_info))
+                       (try
+                         (setv ~g!m (. ~g!i throw))
+                         (except [AttributeError] (raise ~g!e))
+                         (else
+                           (try
+                             (setv ~g!y (apply ~g!m ~g!x))
+                             (except [~g!e StopIteration]
+                                     (setv ~g!r (. ~g!e value))
+                                     (break))))))
+               (else
+                 (try
+                   (if (is ~g!s nil)
+                     (setv ~g!y (next ~g!i))
+                     (setv ~g!y ((. ~g!i send) ~g!s)))
+                   (except [~g!e StopIteration]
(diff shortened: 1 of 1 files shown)


zackmdavis:
(Continued from discussion on #692)

It does look to me like this suffers from #691&mdash;

```
(hy2)zmd@ExpectedReturn:~/Code/hy2/hy2$ hy 
=> (defn fail [] (if some-test (yield-from some-presumptive-generator)))
  File "/home/zmd/Code/hy2/hy2/hy/importer.py", line 44, in ast_compile
    return compile(ast, filename, mode, flags)
SyntaxError: 'return' with argument inside generator (<input>, line 1)
```
The problem seems to be that we sometimes translate a Hy expression in tail position to a Python assign-a-temporary-value-then-return-it maneuver, and mixing `return` and `yield` in the same function doesn't work in Python 2; the compiler ostensibly [tries to deal with this](https://github.com/hylang/hy/blob/fb9a97a7/hy/compiler.py#L2320-L2323), but as our BDFL commented on #691, it would seem to be buggy.

(This pull request might still be a good idea, though! Faithfully translating the "Formal semantics" section of PEP 380 probably has _fewer_ bugs than the existing `yield-from` backport, even if there's another bug elsewhere in the `contains_yield` stuff.)

hy/core/macros.hy
```diff
@@ -214,20 +214,42 @@
...
-                                     (. ~g!e value)
-                                     nil))
-               (break)))))
-           ~g!return))
+    `(do
+       (setv ~g!i (iter ~expr))
+       (try
+         (setv ~g!y (next ~g!i))
+         (except [~g!e StopIteration]
+                 (setv ~g!r (. ~g!e value)))
```
pyos: This should be `(setv ~g!r (next (iter (. ~g!e args)) nil))`. StopIteration instances don't have the `value` attribute in Python 2.

hy/core/macros.hy
```diff
@@ -214,20 +214,42 @@
...
+               (except [~g!e BaseException]
+                       (setv ~g!x (sys.exc_info))
+                       (try
+                         (setv ~g!m (. ~g!i throw))
+                         (except [AttributeError] (raise ~g!e))
+                         (else
+                           (try
+                             (setv ~g!y (apply ~g!m ~g!x))
+                             (except [~g!e StopIteration]
+                                     (setv ~g!r (. ~g!e value))
```
pyos: ditto

hy/core/macros.hy
```diff
@@ -214,20 +214,42 @@
...
+                                     (break))))))
+               (else
+                 (try
+                   (if (is ~g!s nil)
+                     (setv ~g!y (next ~g!i))
+                     (setv ~g!y ((. ~g!i send) ~g!s)))
+                   (except [~g!e StopIteration]
+                           (setv ~g!r (. ~g!e value))
```
pyos: same

pyos:
>mixing return and yield in the same function doesn't work in Python 2

And in Python 3, `return x` inside a generator is equivalent to `raise StopIteration(x)` (the argument is accessible as the `value` attribute on the resulting exception object), so even if yield detection works, the compiler still does the wrong thing. Without this, coroutines would be much less useful.

gilch:
@zackmdavis, I think see why `yield` isn't detected: `defmacro/g!` expands to a `let` form. So the `yield` isn't in the function we think it is in. If I redefine `defmacro/g!` in terms of `setv`, the `yield` detection should work. I think I'll add that to the PR. We were considering getting rid of `let` anyway #1056, and that means removing all of its uses from the core macros.

@pyos, yeah, the [PEP-380 "Formal Sematics" section](https://www.python.org/dev/peps/pep-0380/#formal-semantics) that I translated this from specifically said "Python 3 syntax is used in this section." I figured something would break in Python 2, but I didn't uncover it in my limited initial tests.

When deviating from the direct translation though, I'd prefer to have a unit test to ensure your proposed changes are actually doing what we want. (Ideally we'd test all the branches.) I'm not sure how thorough our current `yield-from` unit tests are, but (if it's not covered already) can you also come up with some concrete examples illustrating the need for these changes? As you pointed out, the compiler is also doing the wrong thing in some cases.

pyos:
@gilch,

>I think see why `yield` isn't detected: `defmacro/g!` expands to a `let` form. So the `yield` isn't in the function we think it is in.

Even so, why does the function that `let` creates contain a `return` when it also contains a `yield`? There's still a bug somewhere else...

>When deviating from the direct translation though, I'd prefer to have a unit test to ensure your proposed changes are actually doing what we want.

Well, it looks like there's already a test that fails due to absense of `value` in Python 2 (check the Travis logs). No test ensures that coroutines/generators return a value in `StopIteration`, though; such a test should do something similar to this Python code: http://ideone.com/hqD1py

gilch:
>Even so, why does the function that let creates contain a return when it also contains a yield? The exception is pretty unambiguous.

I double-checked the expansion of `defmacro/g!`. I miscounted the quotes. While it does use `let` internally, it doesn't appear in the expansion at all, so `yield` wouldn't be wrapped in another function. `defmacro/g!` is not the reason for this problem.

gilch:
A `hy --spy` with the new version. Note the erroneous `return` at the end.
```Python
=> (defn f[x] (raise (StopIteration x))(yield))
def f(x):
    raise StopIteration(x)
    yield
=> (defn g[x] (print (yield-from (f x))))
def g(x):
    :i_1235 = iter(f(x))
    try:
        :y_1236 = next(:i_1235)
        _hy_anon_var_1 = :y_1236
    except StopIteration as :e_1237:
        :r_1238 = next(iter(:e_1237.args), None)
        _hy_anon_var_1 = :r_1238
    else:
        while True:
            try:
                :s_1239 = yield :y_1236
                _hy_anon_var_2 = :s_1239
            except GeneratorExit as :e_1237:
                try:
                    :m_1240 = :i_1235.close
                    _hy_anon_var_3 = :m_1240
                except AttributeError:
                    _hy_anon_var_3 = pass
                else:
                    :m_1240()
                raise :e_1237
                _hy_anon_var_2 = None
            except BaseException as :e_1237:
                :x_1241 = sys.exc_info()
                try:
                    :m_1240 = :i_1235.throw
                    _hy_anon_var_4 = :m_1240
                except AttributeError:
                    raise :e_1237
                    _hy_anon_var_4 = None
                else:
                    try:
                        :y_1236 = :m_1240(*:x_1241)
                        _hy_anon_var_5 = :y_1236
                    except StopIteration as :e_1237:
                        :r_1238 = next(iter(:e_1237.args), None)
                        break
                        _hy_anon_var_5 = None
                _hy_anon_var_2 = _hy_anon_var_4
            else:
                try:
                    if (:s_1239 is None):
                        :y_1236 = next(:i_1235)
                        _hy_anon_var_6 = :y_1236
                    else:
                        :y_1236 = :i_1235.send(:s_1239)
                        _hy_anon_var_6 = :y_1236
                    _hy_anon_var_7 = _hy_anon_var_6
                except StopIteration as :e_1237:
                    :r_1238 = next(iter(:e_1237.args), None)
                    break
                    _hy_anon_var_7 = None
    return print(:r_1238)
  File "c:\users\me\code\repos\hy\hy\importer.py", line 44, in ast_compile
    return compile(ast, filename, mode, flags)
SyntaxError: 'return' with argument inside generator (<input>, line 1)
=>
```

Kodiologist:
@gilch How are we doing here?

gilch:
There are conflicts now, but this change is pretty localized, so it shouldn't be hard to resolve. 

This is waiting on #739, or something. Since there's currently no way to suppress that spurious `return` at the end of the Python expansion, this will always fail. We need a way to suppress that implicit return. We haven't done so (so far) because in most cases, you can simply return `None` and it's effectively the same as if you had no return statement, but that doesn't work here, because it's a generator. `return` must not appear in the output at all. I suggested adjusting the compiler to use a `pass` in the tail position for this purpose, but that's not the only approach we've discussed.

Kodiologist:
Since there's been no progress on this PR for about a year, I'm going to close it. Feel free to reopen it once work has resumed.

gilch:
We might drop Python 2 support before that happens. Then this will be obsolete. Like I said before, it's waiting on another problem, (the spurious `return`) which might still be trouble after that. I'll leave this closed for now.

PR #1065, opened 2016-05-15
Update readthedocs URLs

jwilk:


paultag:
LGTM -  my stance on the URL thing is that protocol relative URIs should be discouraged these days, there's no harm in always using TLS for hardcoded URLs :)

jwilk:
Yeah, I don't think scheme-relative URLs should be used for links to external sites.
Also, they wouldn't even work in offline documentation: the browser would interpret `//foo.example.com/` as `file://foo.example.com/`, which makes no sense.

zackmdavis:
Thanks.

PR #1072, opened 2016-05-25
Replace the existing module loader functionality with one that

elfsternberg:
uses path_hooks.  With a working path hook, it's possible to
register our loader with pkgutils, which means that Django can
now use iter_modules and find Django commands written in Hy.

This also uses Python's internal find_code command, which can be hooked
to use any extension we like, so it's no longer necessary to search for
["%s.hy", "%s/__init__.hy"].  Python's importlib does that by default.

The empty files in the test directory are there just to ensure that pkgutil's
iter_modules can now find both .hy and .py files.


hy/importer.py
```diff
@@ -163,63 +172,117 @@ def write_hy_as_pyc(fname):
...
+                    break
+            else:
+                continue    # not a package
+
+        if modname and '.' not in modname:
+            yielded[modname] = 1
+            yield prefix + modname, ispkg
+
+
+# Monkeypatch both path_hooks and iter_importer_modules to make our
```
berkerpeksag: Thanks! I just skimmed the PR quickly, but I think we can avoid monkeypatching iter_importer_modules by implementing an iter_modules method in HyFileFinder.

Digenis:
@berkerpeksag, where did you find documentation besides the 2-3 relevant peps?
Do you know any model implementation for a path hook?

berkerpeksag:
@Digenis good question :) https://docs.python.org/dev/library/importlib.html#examples has some examples, but I don't know if there's a fully working example (I'd suggest to open an issue at bugs.python.org). As for iter_importer_modules, if I remember correctly, it's a semi-official undocumented API and it looks for an iter_modules method implicitly.

elfsternberg:
@berkerpeksag You're probably right; I'll look into that and consider a better implementation after I finish the work I get paid for.  Sigh  Hacking this stuff is so much more fun.

@digenis Unfortunately, no, I don't.  I spent several days deep inside the source code of importlib and pkgutil, hand-tracing various steps through the process in order to discern what it was really doing with respect to the code, before I managed to come up with this solution.  I'm going to get a pretty good blog post out of it sometime soon, I imagine.  The importlib/pkgutil relationship is excessively complicated.

The big lesson here is that having multiple alternative language syntaxes that produce Python ASTs, and having that code present in the same source tree (as both Hy and Py files), is something Python was never, ever meant to support, and coercing it to support such requires a degree of Python black magic most of us never actually want to get into!

elfsternberg:
@berkerpeksag I'm afraid that pkgutils, at least under Python 3.4, uses the singledispatch() decorator in order to direct iter_importer_modules to the correct iterator.  Moving iter_modules into HyFileFinder didn't work.

elfsternberg:
Gahh.  Importlib isn't a very stable piece of code.

elfsternberg:
I'm withdrawing this pull request until I get a better handle on the internals through Py 2.7.

berkerpeksag:
We might drop 2.7 support in a next feature release (even critical software like Gunicorn will drop 2.7) I'd suggest to make it work in Python 3 first then we can discuss 2.7 :)

PR #1076, opened 2016-06-23
Fixed example in "with"

haskellcamargo:
The examples in the documentation are not working. That would generate an error saying that `f` is **not** a global variable. Hy uses two brackets for `with`.


gilch:
The documentation is correct. #898 dropped the extra brackets.

gilch:
@haskellcamargo try cloning the current version of Hy directly from github.

PR #1093, opened 2016-09-04
Use HTTPS for Wikipedia URLs

jwilk:
Found using [urlycue](https://github.com/jwilk/urlycue).


zackmdavis:
thx :sparkling_heart:

Issue #259, opened 2013-07-19
Replace raw-input with input in Python 2


joehakimrahme:
Python2's `input` function is bad. If you want to `eval` the input, you should do so explicitly. This is why Python2 introduced `raw_input` and why Python3 made `input` behave like `raw_input`.

I think Hy should have a single function `input` that **doesn't** evaluate the string it gets. It should automatically translate into `raw_input` if translated to Python2 and `input` if translated to Python 3.

### Cons
I think this might affect the holy Python <-> Hy interoperability. 

### However
I think we do something similar with division in Hy, where `(/ 3 2)` always evaluates to `1.5` regardless of the version of Python it compiles to.

Basically, I'd like to get your opinion before starting coding this one. We could discuss the technical details of the implementation, but first let's focus on deciding if it's a sound design choice.

Foxboron:
I believe this is fixed now!

PR #1095, opened 2016-09-12
Tutorial: mention Python 3 modification on example

weakish:
- `raw_input`
- `Long`

---

 docs/tutorial.rst | 6 ++++++
 1 file changed, 6 insertions(+)

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -151,6 +151,9 @@ Now let's look at the equivalent Hy program:
 
    (simple-conversation)
 
+Note that ``raw_input`` is for Python 2, if you are using Python 3,
+replace ``raw_input`` with ``input``.
+
 If you look at the above program, as long as you remember that the
 first element in each list of the program is the function (or
 macro... we'll get to those later) being called and that the rest are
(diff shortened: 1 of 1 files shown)


docs/tutorial.rst
```diff
@@ -151,6 +151,9 @@ Now let's look at the equivalent Hy program:
 
    (simple-conversation)
 
+Note that ``raw_input`` is for Python 2, if you are using Python 3,
```
kirbyfan64: @hylang/core Come to think of it, maybe Hy should alias `input` to `raw_input` under Python 2, kind of like it does `map`.

gilch: @kirbyfan64 didn't we fix that already #259?

kirbyfan64: @gilch Oops. Then this whole change is unneeded.

docs/tutorial.rst
```diff
@@ -213,6 +216,9 @@ supports the Common Lisp method of quoting:
    => '(1 2 3)
    (1L 2L 3L)
 
+Note the output is in Python 2. In Python 3 it is `(1 2 3)`
+because Python 3 does not support `Long`.
```
kirbyfan64: This is kind of confusing. Maybe, in the earlier sections of the tutorial, it should mention the int <> long difference and how Hy handles it, and then, down here, just refer to that section.

gilch:
I think the docs should use Python3 by default and only mention Python2 as a special case when necessary.

Kodiologist:
@gilch That's not a bad idea, but it's worth its own PR.

At any rate, it seems that consensus is against this PR, at least as it stands.

PR #1097, opened 2016-09-12
doc: cli: mention `--spy` only works in REPL mode.

weakish:
---


berkerpeksag:
Perhaps we can change the first sentence to:

```
Print equivalent Python code before executing it in the REPL.
```

berkerpeksag:
Thanks!

PR #1102, opened 2016-09-18
Generalize `xor` to any number of arguments

Kodiologist:


gilch:
See #872, for why I can't approve this. Is your interpretation `one-hot` or `parity`? Why not call it one of those instead of `xor` to avoid confusion?

Kodiologist:
The parity version of variadic XOR is the only one I've seen in a programming language, and is both commutative and associative. (You said that "`(xor a b c)` is not `(xor (xor a b) c)` in general", but unless I'm missing something, this is indeed true for the parity version.) I don't know why anybody would expect the one-hot version.

gilch:
The parity xor has no possibility for shortcutting behavior. Therefore, it's not a branch control form like `and`, `cond`, `if`, and `or`. It's just a function. So why go the the trouble of returning the last true or last false input instead of a simple boolean? I can't think of a use case for this. Have you seen another language that actually does this? Or did you just want the binary case to work the same as before?

The one-hot interpretation _can_ shortcut, if you see a second true, you're done. It also has a more sensible choice for a non-boolean return values, like the one true found in the true case, but it's less clear to me what it should return in the false case.

When using pure booleans, `xor` has the same truth table as `!=`, and `^`, so `xor` seemed kind of pointless to me in the first place. I'm tempted to remove it altogether. Python doesn't have it, and neither does Clojure or Common Lisp. A reduce using `^` will give you the parity behavior. If they're not already booleans, you can map `bool` over the iterable first. Of course it doesn't return a non-boolean this way, but that seems to be pointless anyway.

> I don't know why anybody would expect the one-hot version.

The parity interpretation does appear to be more common, but I've seen the one-hot interpretation more than once before.

Kodiologist:
> The parity `xor` has no possibility for shortcutting behavior. Therefore, it's not a branch control form like `and`, `cond`, `if`, and `or`. It's just a function. So why go the the trouble of returning the last true or last false input instead of a simple boolean?

Returning the input value instead of a Boolean isn't (just) for flow control; it's so you can do things like `(setv dessert (xor cake ice-cream))`. But the chief reason it strikes me as important is that otherwise, we would have a needless surprising inconsistency of `xor` with the other logical operators.

> Have you seen another language that actually does this?

No, but there aren't many languages that have a built-in logical (as opposed to bitwise) `xor` operator to begin with, perhaps because of the influence of C.

I seem to have overseen the most important reason to prefer the parity version of variadic xor: because it satisfies `(= (xor x1 x2 x3 … xn) (xor (xor … (xor x1 x2) x3) … xn))`. That is, it's what you get when you iterate binary `xor`, so `xor` is generalized to more than two arguments the same way `or` and `and` are.

kirbyfan64:
I'm personally -0.5 on this. It seems like a great idea, but I really don't see this being used that often, and the behavior seems a bit odd to me.

Not -1 because adding this would give serious bragging rights.

Kodiologist:
Since nobody likes this PR except me, I'll put it out of its misery.

Kodiologist:
I was just reading a logic textbook (Wolfgang Rautenberg's *Concise Introduction to Mathematical Logic*) that pointed out that exclusive or is equivalent to addition modulo 2. This makes the parity interpretation yet more natural, since we already treat `(+ a b c ...)` as `a + b + c + …`, and the latter with the addition performed modulo 2 is a parity detector.

PR #1105, opened 2016-09-18
Translate (.a.b.c x) to x.a.b.c().

Kodiologist:
This is useful for, e.g., calling the `Series.str.lower` method in pandas.


tests/native_tests/dot.hy
```diff
@@ -0,0 +1,20 @@
+(defclass X [object] [])
```
kirbyfan64: There are already some tests for dot application [here](https://github.com/Kodiologist/hy/blob/821ed96faf0dd6467307759ec07ab444de29e5e8/tests/native_tests/language.hy#L399); try adding these there instead.

gilch:
Hy already has a good syntax for this. What exactly is wrong with `(.c x.a.b)`? It makes the first argument (`self` or `cls`) explicit, while keeping the function name in the function position where it belongs. 

You can also already avoid the syntactic sugar altogether with `((. x a b c))`, which is easier to manipulate programmatically.

*There should be one-- and preferably only one --obvious way to do it.*

:-1: from me.

Kodiologist:
> What exactly is wrong with `(.c x.a.b)?`

It's fine when it works, but it doesn't work if `x` is any expression other than a identifier. `(.c (foo 1 2).a.b)` doesn't work (and it would be kind of confusing if it did).

> You can also already avoid the syntactic sugar altogether with `((. x a b c))`, which is easier to manipulate programmatically.

That puts the method after rather than before the object. I assume that the whole reason we have `(.a x)` is to avoid that, because of the preference for prefix notation in Lisp.

> There should be one-- and preferably only one --obvious way to do it.

That principle has never been adhered to in Python, and Hy defies it even more blatantly, and for good reason. It's a fine philosophy for machine code, I guess (if you like RISC better than CISC), but human programmers benefit from having a wide variety of tools, just like human languages benefit from having a wide variety of words.

gilch:
>> What exactly is wrong with `(.c x.a.b)?`

>It's fine when it works, but it doesn't work if `x` is any expression other than a identifier. `(.c (foo 1 2).a.b)` doesn't work (and it would be kind of confusing if it did).

That's a better argument and I'd like to know what the rest of the @hylang/core team thinks of it. But that example should be `(.c (. (foo 1 2) a b))`, which does work. `self` is the `b` you got from `a` that you got from `(foo 1 2)`. Clojure would do it basically the same way for a similar Java interop case, wouldn't it? Hy has a perfectly good syntax for this.

>> You can also already avoid the syntactic sugar altogether with `((. x a b c))`, which is easier to manipulate programmatically.

>That puts the method after rather than before the object. I assume that the whole reason we have `(.a x)` is to avoid that, because of the preference for prefix notation in Lisp.

Yes, Lisp strives to avoid special cases to make it easier to manipulate code with code (like the `->` macro, for example). But my example was perfectly valid prefix notation, as always. I'll add an argument to make it more explicit:
```
((. x a b c) an-arg)
;; becomes x.a.b.c(an-arg)
```
Python lets you fetch a bound method as an object, and Hy lets you use an expression in the function position. The above expression fetches a bound method object, i.e. with `self` baked in. Partially applying functions is a common practice in functional languages, so I think this suits Hy just fine.

Python also lets you get the unbound method, and therefore so does Hy: `(.lower (. Series str) x)`, or `(.lower Series.str x)`, if you insist on sugar.

>> There should be one-- and preferably only one --obvious way to do it.

>That principle has never been adhered to in Python, and Hy defies it even more blatantly, and for good reason. It's a fine philosophy for machine code, I guess (if you like RISC better than CISC), but human programmers benefit from having a wide variety of tools, just like human languages benefit from having a wide variety of words.

It *has* been a major influence for Python itself (to Python's great credit), which is very clear when you compare it to a language with the opposite philosophy, like Ruby or Perl, which is widely derided as a "write-only" language. Python may not have followed this principle perfectly, but that's no excuse to make it worse.

I also think you've misunderstood. It is not that there should only be one way to do anything (impossible, since you can always write a longer program to do the same thing in any Turing-complete language), but that it's preferable when exactly one of those ways is more *obvious*. Clojure, Hy's other major influence, also takes simplicity as a pillar of its culture. 

I've been trying to clean up some of Hy's more blatant violations (but I think the dot sugar is here to stay). If you've noticed other violations, do tell, and I'll see what we can do about removing them, or at least making the preferred option more obvious.

The variety of words belong in libraries, not in the grammar, which should be kept very simple, like Clojure.

Kodiologist:
> It *has* been a major influence for Python itself (to Python's great credit), which is very clear when you compare it to a language with the opposite philosophy, like Ruby or Perl, which is widely derided as a "write-only" language.

For what it's worth, I've been a Perl programmer since 2005, and the rigidity of Python was the primary reason I resisted switching. I only switched once I found something even more pathologically eclectic, which is Lisp. So yes, I admit it, I am The Enemy.

If this PR is requested and I still wanted the feature badly enough (I have found myself really aggravated that `(.a.b.c x)` doesn't do the right thing a few times over the past year or so), I could probably still wrap a macro around my files that implements it in Hy itself. But this seemed to me like an obvious bug in the implementation of `(.foo …)` constructs that was better fixed once and for all then left for every new user to trip over.

kirbyfan64:
> So yes, I admit it, I am The Enemy.

QUICK, CALL MUMEI!! (Kudos to the 0 people who will get that reference.)

In all seriousness l, though, I'm +1 to this.Although I'm a BIG fan of "one way to do it", this is a natural extension of the attribute calling syntax.

gilch:
カバネリ

> better fixed once and for all th[a]n left for every new user to trip over.

Also a good point. Sometimes all or none is better than halfway. I'd be okay with none, but I don't think that's happening in this case. Perhaps it is a natural extension.

> it doesn't work if x is any expression other than a identifier.

Your approach doesn't work if `b` is any expression other than an identifier either. Consider what happens when you have to call a method in your access chain: `x.a.get-b().c(an-arg)`. How should this look in Hy?

Clojure can do it with the `..` macro: `((.. x a (get-b) c) an_arg)`. I think that would be a better extension to Hy than this proposal, since Hy doesn't have `..` (and its more powerful `.` can't do it either), but look at this: `((-> x (. a) (.get-b) (. c)) an-arg)`. Hy can pretty much already do it with the threading macro.

> could probably still wrap a macro around my files

Consider a reader macro for this kind of thing. It's only one extra character: `#.a.b.c`.

> the rigidity of Python was the primary reason I resisted

If you don't like Python's way, why such interest in Hy? CljPerl might be what you're looking for. If it's due to Python's greater popularity (or the resulting ecosystem), remember that Python is more popular *because* of its ways.

Kodiologist:
> カバネリ

I beg your pardon?

> Consider what happens when you have to call a method in your access chain: `x.a.get_b().c(an_arg)`. How should this look in Hy?

I would use `(.c (.a.get-b x) an-arg)`, so functions and methods go in prefix order as usual.

> Clojure can do it with the `..` macro: `((.. x a (get-b) c) an-arg)`.

I'm not familiar with Clojure, but that looks like it would accept arbitrary expressions *except* for bare identifiers, which are quoted. That seems confusing.

> If you don't like Python's way, why such interest in Hy?

On the contrary, I think Python is quite a nice language, even if I don't like it quite as much as Perl. To be sure, I would like it a lot less if the Zen of Python were actually adhered to. But that shouldn't be held against the actual language.

Kodiologist:
@hylang Could somebody take a look at this?

hy/compiler.py
```diff
@@ -2017,19 +2017,24 @@ def compile_expression(self, expression):
...
+                fn = attrs.pop()
 
-                # Get the object we want to take an attribute from
+                # Get the object we're calling the method on
+                # (extracted with the attribute access DSL)
                 if len(expression) < 2:
                     raise HyTypeError(expression,
                                       "attribute access requires object")
-                func = self.compile(expression.pop(1))
+                foo = HyExpression(
```
kirbyfan64: Uhhh...`foo` is a really odd name here. What if you just put the whole thing in the call to `self.compile`? e.g.:

```
func = self.compile(HyExpression([HySymbol('.').replace(fn),
                                  expression.pop(1)] + attrs))
```

kirbyfan64:
:+1:

Kodiologist:
Is this ready to merge?

gilch:
@Kodiologist, no it isn't. I'm still not sure how I feel about this, but you've persuaded me enough to withdraw my downvote to an abstention. You'll still need another approval from @hylang/core.

PR #1110, opened 2016-09-20
Make unary + call __pos__

### LINKED ISSUE #1109: (+ x) shouldn't compile to (0 + x)
Python has a unary `+` that invokes the `__pos__` method. This is not the same as the `__add__` method for `0`. Hy should be able to invoke this with `+` instead of requiring a direct call to `__pos__`.

Hy does handle unary `-` correctly though, I checked (`--spy`):
```Hy
=> (- 1)
(- 1)
-1
```
Unary `+`? Nope.
```Hy
=> (+ 1)
(0 + 1)
1
```
What should have happened:
```Hy
=> (+ 1)
(+ 1)
1
```

Kodiologist:
Fixes #1109.

`__pos__` doesn't work on strings, lists, or tuples, so I've removed some tests that required `(= (+ "a") "a")` etc.

I wanted to name the test `test-pos` instead of `test-add-unary`, but if I do that, the test doesn't get run, and I have no idea why.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1910,7 +1910,6 @@ class HyASTCompiler(object):
                              col_offset=child.start_column)
         return ret
 
-    @builds("+")
     @builds("*")
     @builds("/")
     @builds("//")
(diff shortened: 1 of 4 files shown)


Kodiologist:
@hylang/core Could someone take a look? I think it's nothing complicated; just a simple solution to a simple bug.

Kodiologist:
@hylang Could some core members take a look? I think it's nothing complicated; just a simple solution to a simple bug.

zackmdavis:
at a cursory look-over, seems sane to me :+1:

Issue #1111, opened 2016-09-21
Broken on latest Python 3.6 release (3.6.0b1)


tianon:
Looks like the latest 3.6.0b1 release broke Hy. :cry:

I mentioned this on IRC a little while back and @berkerpeksag mentioned it's probably an easy fix, but it's over my head. :innocent:

```console
root@4315e0242dc3:/# python --version
Python 3.6.0b1
```

```console
root@4315e0242dc3:/# pip install hy
Collecting hy
  Downloading hy-0.11.1.tar.gz (43kB)
    100% |████████████████████████████████| 51kB 517kB/s 
Collecting rply>=0.7.0 (from hy)
  Downloading rply-0.7.4-py2.py3-none-any.whl
Collecting astor>=0.5 (from hy)
  Downloading astor-0.5-py2.py3-none-any.whl
Collecting clint>=0.4 (from hy)
  Downloading clint-0.5.1.tar.gz
Collecting appdirs (from rply>=0.7.0->hy)
  Downloading appdirs-1.4.0-py2.py3-none-any.whl
Collecting args (from clint>=0.4->hy)
  Downloading args-0.1.0.tar.gz
Installing collected packages: appdirs, rply, astor, args, clint, hy
  Running setup.py install for args ... done
  Running setup.py install for clint ... done
  Running setup.py install for hy ... done
Successfully installed appdirs-1.4.0 args-0.1.0 astor-0.5 clint-0.5.1 hy-0.11.1 rply-0.7.4
root@4315e0242dc3:/# hy
hy 0.11.1 using CPython(default) 3.6.0b1 on Linux
=> (+ 2 2)
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/hy/compiler.py", line 2304, in hy_compile
    compiler = HyASTCompiler(module_name)
  File "/usr/local/lib/python3.6/site-packages/hy/compiler.py", line 378, in __init__
    load_stdlib()
  File "/usr/local/lib/python3.6/site-packages/hy/compiler.py", line 73, in load_stdlib
    mod = importlib.import_module(module)
  File "/usr/local/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 978, in _gcd_import
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load
  File "<frozen importlib._bootstrap>", line 950, in _find_and_load_unlocked
[...]

PR #1112, opened 2016-09-22
fix Hy on recent Pythons

zackmdavis:
In issue #1111, @tianon reported that Hy didn't work with Python
3.6.0b1: trying to evaluate a simple expression at the REPL blew up with
`TypeError: required field "is_async" missing from comprehension`. This
was due to a grammar
change (https://www.python.org/dev/peps/pep-0530/#grammar-updates) in
the implementation (https://hg.python.org/cpython/rev/cf91a929b81a) of
PEP 530, which we can easily accomodate.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1516,7 +1516,8 @@ class HyASTCompiler(object):
             gen.append(ast.comprehension(
                 target=target,
                 iter=gen_res.force_expr,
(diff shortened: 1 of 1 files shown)


berkerpeksag:
LGTM

gilch:
Can we add a 3.6 build to Travis first?

PR #1113, opened 2016-09-23
nightly Python on Travis CI??

zackmdavis:
From [commentary](https://github.com/hylang/hy/pull/1112#issuecomment-249065602) on #1112. This is just a guess and probably won't work.

PR #1114, opened 2016-09-24
Revert #924

gilch:
reverts #924, which should never have been merged in the first place. Had we merged #903 first, #924 would not have passed our tests. (see discussion in #903).

### DIFF
--- a/hy/macros.py
+++ b/hy/macros.py
@@ -18,7 +18,6 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
-from inspect import getargspec, formatargspec
 from hy.models import replace_hy_obj, wrap_value
 from hy.models.expression import HyExpression
 from hy.models.string import HyString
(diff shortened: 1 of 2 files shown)


paultag:
gah, just flake8 errors; mind cleaning that up?

gilch:
Ah, that is strange--only flake8? @zackmdavis was this because of #1112 / #1113?

paultag:
It may be due to an upgraded flake8; let me send a PR

paultag:
Ah, no it's due to this PR - getargspec is unused :)

gilch:
Took off too much? Maybe a later commit was using that line.

gilch:
@paultag looks like `getargspec` was used by `(macro)`. and only `formatargspec` was unused.

paultag:
Sorry! I"m the worst. you're right!

gilch:
All checks have passed.
@hylang/core. It was too late to auto-revert, so I did it manually. This increases risk of errors. Can I get some eyeballs, approval?

gilch:
This revert isn't enough to fix #903. (I tried rebasing on this branch.) Something is still very broken here, and I suspect it's `getargspec`, but I might not have merged #903 properly.

gilch:
Closed by #1172

PR #1116, opened 2016-09-24
change quickstart install from PyPI to GitHub

### LINKED ISSUE #1106: Remove the "install from PyPI" instruction from quickstart in "latest" docs.
And replace with instructions on how to install from github master.

Even if a new user finds the "latest" docs before the "stable" version, he might end up installing the stable version because the quickstart instructs him to and then be very confused when the docs don't match up.

I believe the invocation should be 
```
$ pip install git+https://github.com/hylang/hy.git
```

gilch:
Closes #1106

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -13,7 +13,7 @@ Quickstart
 1. Create a `Virtual Python Environment
    <https://pypi.python.org/pypi/virtualenv>`_.
 2. Activate your Virtual Python Environment.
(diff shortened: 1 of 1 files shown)


zackmdavis:
I mean, _ideally_ we would finish the [Grand Language Cleanup](https://github.com/hylang/hy/milestone/8) and actually release 0.12 (or 1.0??) on PyPI, but since everyone is presumably too busy with their dayjobs and other projects to actually get this done (I certainly am) and this has been the case for months, not actively misleading users is a good idea.

PR #1121, opened 2016-09-26
clarify documentation of import

gilch:
Add a comment showing the equivalent Python of the long example, and demonstrate how to get `from` and `as` on the same line.


zackmdavis:
***+1***

PR #1124, opened 2016-09-29
Make (for [] ...) work like (for [_ []] ...) 

Kodiologist:
Before, `(for [] ...)` would execute its body once, instead of zero times.

### DIFF
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -112,7 +112,8 @@
     (macro-error args "`for' requires an even number of args.")]
    [(empty? body)
     (macro-error None "`for' requires a body to evaluate")]
-   [(empty? args) `(do ~@body ~@belse)]
+   [(empty? args)
+    (if belse `(do ~@(get belse 0 (slice 1 None))))]
    [(= (len args) 2) `(for* [~@args] (do ~@body) ~@belse)]
    [true
     (let [alist (cut args 0 nil 2)]
--- a/tests/native_tests/language.hy
+++ b/tests/native_tests/language.hy
@@ -157,13 +157,13 @@
     (setv count2 (+ count2 x)))
   (assert (= count1 15))
   (assert (= count2 15))
-  (setv count 0)
+  (setv n 0)
   (for [x [1 2 3 4 5]
         y [1 2 3 4 5]]
-    (setv count (+ count x y))
+    (setv n (+ n x y))
     (else
-      (+= count 1)))
-  (assert (= count 151))
+      (+= n 1)))
+  (assert (= n 151))
   (assert (= (list ((fn [] (for [x [[1] [2 3]] y x] (yield y)))))
              (list-comp y [x [[1] [2 3]] y x])))
   (assert (= (list ((fn [] (for [x [[1] [2 3]] y x z (range 5)] (yield z)))))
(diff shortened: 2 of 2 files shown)


gilch:
Hy's `for` macro appears to have been based upon Clojure's `doseq`, which behaves the same way:
```Clojure
user=> (doseq [x [0 1]
  #_=>         y '[a b]]
  #_=>   (println [x y]))
[0 a]
[0 b]
[1 a]
[1 b]
nil
user=> (doseq []
  #_=>   (println 42))
42
nil
```
```Hy
=> (for [x [0 1]
...      y '[a b]]
...  (print [x y]))
from hy import HyList, HySymbol
for (x, y) in ((x, y) for x in [0, 1] for y in HyList((([] + [HySymbol('a')]) + [HySymbol('b')]))):
    print([x, y])
[0, 'a']
[0, 'b']
[1, 'a']
[1, 'b']
=> (for []
...  (print 42))
print(42)
42
```
Therefore, I'm not willing to call this a bug. I do agree it's confusing, and could be better documented.

You could argue that Hy's `for` should behave more like Python's `for` instead of Clojure's `doseq`, in which case an empty binding vector should be an error, not a no-op. But we already have `for*` for that, see #874. Perhaps we could rename `for` to `doseq` and `for*` to `for`.

In any case, this is not the right solution. 👎 from me.

Kodiologist:
I can't speak for Clojure, but executing the body exactly once makes no sense. How does that follow from the nonempty case, why would a user expect it, and when would it be useful? Raising an error makes more sense.

kirbyfan64:
I think an error should be thrown. Regardless of how this is implemented, people are going to have different ideas of how is should behave, and the Zen of Python *does* say:

> In the face of ambiguity, refuse the temptation to guess.

arrdem:
Having hacked on and used Clojure a bunch I'm comfortable saying that `(for [] ...)` executing exactly once makes no sense. `for` is not `doseq` or a looping construct, it is a macro for generating list comprehensions with conditional control flow. Clojure itself will reject `(for [] ...)`, as it requires at least one binding thus resolving this ambiguity, but if you will allow zero bindings, any execution whatever flies in the face of the design goal which is list comprehension.

gilch:
@arrdem, you're getting confused by the name "for".

Hy's `for` is _not_ named after Clojure's `for`. Hy's `for`'s semantics are the same as Clojure's `doseq`. This is a good reason to rename Hy's `for` to `doseq`, or since Python (and Hy) use iterables where Clojure would use the sequence abstration, rename it to `doiter` instead.

Clojure's `for` is _not_ like Hy's `for`. Clojure's `for` is a _lazy list comprehension_. Hy already has one of these, and it's called `genexpr`. It should _not_ be renamed `for`, because that would confuse Python users.

In conclusion, we should keep `genexpr`'s name, and tell Clojure users to use it in cases where they would have used Clojure's `for`.

We should rename Hy's `for` to `doiter` by analogy with Clojure's `doseq`.

We should then rename Hy's `for*` to `for`, so that Hy's new `for` behaves like Python's `for`.

gilch:
I'm closing this in favor of #1125. We don't have enough core downvotes for an absolute veto, so this could be reopened at a later time but there's been no further arguments either way for months.

Kodiologist:
@zackmdavis @arrdem Do you approve of this PR? The thumb reactions you put on some past comments suggest so, but it's not clear.

Kodiologist:
@gilch So far as I can see, renaming `for` to `doiter` or `doseq` doesn't make the behavior of executing the body exactly once make sense. If Clojure does that, too bad for Clojure. So the issue stands even if `for` is renamed.

PR #1131, opened 2016-10-11
Allow &rest after &optional, like Python

Kodiologist:
I consider this a bug-fix, since `lambda a=None, *xs: 1` is valid Python, but in Hy, you'd have to write that as `(fn [&rest xs &optional a] 1)` without this patch.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -503,19 +503,13 @@ class HyASTCompiler(object):
         for expr in exprs:
 
             if expr in ll_keywords:
-                if expr == "&rest" and lambda_keyword is None:
-                    lambda_keyword = expr
-                elif expr == "&optional":
+                if expr == "&optional":
                     if len(defaults) > 0:
                         raise HyTypeError(expr,
                                           "There can only be &optional "
                                           "arguments or one &key argument")
                     lambda_keyword = expr
-                elif expr == "&key":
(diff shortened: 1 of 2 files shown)


hy/compiler.py
```diff
@@ -503,19 +503,13 @@ def _parse_lambda_list(self, exprs):
...
             if expr in ll_keywords:
-                if expr == "&rest" and lambda_keyword is None:
-                    lambda_keyword = expr
-                elif expr == "&optional":
+                if expr == "&optional":
                     if len(defaults) > 0:
                         raise HyTypeError(expr,
                                           "There can only be &optional "
                                           "arguments or one &key argument")
                     lambda_keyword = expr
```
kirbyfan64: Now I wonder if this whole thing can be reduced to:

```python
if expr in ("&optional", "&rest", "&key", "&kwonly", "&kwargs"):
    if len(defaults) > 0:
        # ...
    lambda_keyword = expr
```

(This isn't necessary for this PR; it's just me wondering out load.)

kirbyfan64:
As a side note, I have no clue how the heck this went unnoticed for so long...

Also, I'm having way too much fun with GitHub's PR reviews. :D

gilch:
Let's retest this one too.

PR #1132, opened 2016-10-11
Don't check the type of arguments to `inc`, `odd?`, etc.

Kodiologist:
This allows them to be used with numeric types that aren't built in, such as NumPy arrays. Because Python uses duck typing, there's generally no way to know in advance whether a given value will accept a given operator. Of course, things like `(inc "hello")` will still raise a `TypeError`, because so does `(+ "hello" 1)`.

### DIFF
--- a/tests/native_tests/core.hy
+++ b/tests/native_tests/core.hy
@@ -33,6 +33,11 @@
 (defn assert-nil [x]
   (assert (is x nil)))
 
+(defn assert-requires-num [f]
+  (for [x ["foo" [] None]]
+    (assert-true (try (do (f x) False)
+                 (except [e [TypeError]] True)))))
+
 (defn test-coll? []
   "NATIVE: testing coll?"
   (assert-true (coll? [1 2 3]))
@@ -66,12 +71,7 @@
   (assert-equal 0 (dec 1))
   (assert-equal -1 (dec 0))
   (assert-equal 0 (dec (dec 2)))
-  (try (do (dec "foo") (assert False))
-       (except [e [TypeError]] (assert (in "not a number" (str e)))))
-  (try (do (dec []) (assert False))
-       (except [e [TypeError]] (assert (in "not a number" (str e)))))
-  (try (do (dec None) (assert False))
-       (except [e [TypeError]] (assert (in "not a number" (str e))))))
+  (assert-requires-num dec))
 
 (defn test-setv []
   "NATIVE: testing setv mutation"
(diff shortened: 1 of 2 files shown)


tests/native_tests/core.hy
```diff
@@ -33,6 +33,11 @@
 (defn assert-nil [x]
   (assert (is x nil)))
 
+(defn assert-requires-num [f]
+  (for [x ["foo" [] None]]
+    (assert-true (try (do (f x) False)
+                 (except [e [TypeError]] True)))))
```
kirbyfan64: Normally, when testing for an exception, the following style is used instead:

```hy
(try (f x)
      (except [TypeError] true)
      (else (assert false)))
```

tests/native_tests/core.hy
```diff
@@ -394,24 +388,13 @@
...
   (assert-false (zero? -2))
   (assert-false (zero? 1))
-  (assert-true (zero? 0))
-  (try (do (zero? "foo") (assert False))
-       (except [e [TypeError]] (assert (in "not a number" (str e)))))
-  (try (do (zero? []) (assert False))
-       (except [e [TypeError]] (assert (in "not a number" (str e)))))
-  (try (do (zero? None) (assert False))
-       (except [e [TypeError]] (assert (in "not a number" (str e))))))
+  (assert-true (zero? 0)))
```
kirbyfan64: Shouldn't this also be `assert-requires-num`?

Kodiologist: That's what I thought, but when I noticed that `(zero? "foo")` returns `False` instead of raising an exception, I remembered that Python has no notion of numeric equality that's separate from general-purpose equality. So `zero?` should accept pretty much anything.

kirbyfan64: @Kodiologist True.

gilch:
Numpy compatibility is a good argument. General Python duck typing is also a pretty good argument. I'm inclined to approve this change. Who put those checks in there in the first place? I'll check the blame...

@rwtolbert, is this just for better error messages? I don't think that's worth it. Was there some other reason?

@Kodiologist, checks are failing, and it's not just Python: nightly. I don't want to merge this while 2.7 and pypy aren't working.

Kodiologist:
Thanks for catching that, Gilch. I'd been ignoring Travis because of the nightly bug.

gilch:
I'm forcing a retest, now that #1133 is merged.

Kodiologist:
@gilch Travis is happy; is this ready to go?

PR #1133, opened 2016-10-11
Test Python 3.6 apart from nightly

kirbyfan64:
Fixes the now-failing Travis @Kodiologist saw at https://github.com/hylang/hy/pull/1113#issuecomment-253047608.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
   include:
     - python: 3.5
       env: TOXENV=py35
-    - python: nightly
+    - python: 3.6
       env: TOXENV=py36
 env:
   - TOXENV=py27
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
   include:
     - python: 3.5
       env: TOXENV=py35
-    - python: 3.6
+    - python: 3.6-dev
       env: TOXENV=py36
 env:
   - TOXENV=py27


gilch:
@zackmdavis, wasn't this working before? #1113

zackmdavis:
@kirbyfan64 @gilch I chose "nightly" in #1113 because (as I recall) the [Travis CI docs](https://docs.travis-ci.com/user/languages/python/) noted that it pointed to the 3.6-dev branch (and there was no separate 3.6 target). It looks like that page has been updated now, and there is a "3.6-dev" target.

gilch:
So it needs to say `3.6-dev`? Because @kirbyfan64's change only says `3.6`.

Also, since `nightly` points to `3.7-dev` now and nightly is failing, does this mean 3.7 will break Hy again?

.travis.yml
```diff
@@ -4,7 +4,7 @@ matrix:
   include:
     - python: 3.5
       env: TOXENV=py35
-    - python: nightly
+    - python: 3.6
```
gilch: @kirbyfan64 this doesn't seem to be working either. Try `3.6-dev` instead.

zackmdavis:
> So it needs to say `3.6-dev`? 

Probably!

> Also, since nightly points to `3.7-dev` now and nightly is failing, does this mean 3.7 will break Hy again?

I would bet against it: the [specific failure](https://travis-ci.org/hylang/hy/jobs/166872848) that @Kodiologist  pointed out in the other issue looks like a mismatch between the version Tox expected (3.6) and what was actually on the nightly branch (3.7.0a0).

gilch:
Since it's such a minor fix and I have access, I just changed it myself with another commit to the `nightly36` branch. As you can see, it's working now, so I hope you don't object. @kirbyfan64, @zackmdavis, can I get an ack and merge?

kirbyfan64:
:+1:

kirbyfan64:
(I was literally going to fix it right now./

Issue #1070, opened 2016-05-19
Change Hy's __version__ string


gilch:
I suggest changing Hy's version number to "0.11.0-github" or something ("-develop", "-bleeding", "-experimental"?), so we can distinguish it from the PyPI version 0.11. Maybe we could automatically append the git hash or something. That would make it easy for us to tell which version we're testing.

Users are often confused by the version differences we've introduced. It would be easier to talk to them if we had better names for these versions and could ask them for the `__version__` string, instead of "did you get this from PyPI or github and when?". It would also be easier for them to tell the online documents apart if they had very distinct version strings.

We must also be sure to take that bit back off for the next PyPI release branch to prevent that confusion again.

zackmdavis:
0.12a0, like Python does it?

berkerpeksag:
+1 for 0.12a0.

gilch:
That's fine with me, but we need to agree on the rules for when exactly we change the version number and how. Do we increment with every pull request? I think Python's versioning rules are here: https://www.python.org/dev/peps/pep-0440/

Our rules may not need to be that complex yet.

Kodiologist:
I'd recommend appending the Git hash or a 7-character prefix. That way, the version number alone tells us exactly what commit a user is running.

Regarding the "0.12a0" style, what would increment the alpha number "a0"? How would one translate from alpha numbers to commits?

gilch:
Although I suggested it in the first place, I don't think we can just hardcode the git hash in the version string, since the hash doesn't exist until after the commit has been made. Perhaps there is some way for Python to generate it programmatically though. Have you seen any other projects that can do this?

Kodiologist:
> I don't think we can just hardcode the git hash in the version string, since the hash doesn't exist until after the commit has been made. Perhaps there is some way for Python to generate it programmatically though.

Right. I figured the way to do it is to call `git log -1 --format=%H` at build-time, putting it in a generated file.

> Have you seen any other projects that can do this?

[Cataclysm DDA](https://github.com/CleverRaven/Cataclysm-DDA) comes to mind. That's C++, though.

tuturto:
Is this closed by #1135, or do we want to have more automation in place at this point?

Kodiologist:
I think the need's covered.

PR #1135, opened 2016-10-13
Generate hy.version using `git describe`

Kodiologist:
Addresses #1070.

Hy's version will now look like `0.11.0` when you're exactly on a version, `0.11.0-293-g34304c4` when you're 293 commits past version 0.11.0 (34304c4 is the abbreviated Git hash), and `0.11.0-293-g34304c4-dirty` when the working tree is dirty.

Future releases should include an already generated version file.


kirbyfan64:
> `0.11.0-293-g34304c4-dirty` when the working tree is dirty

Is this part really necessary? I normally find it quite jarring when you edit a file and now your version ended up changing.

setup.py
```diff
@@ -22,24 +22,19 @@
...
-else:
-    VSRE = r"^__version__ = ['\"]([^'\"]*)['\"]"
-    mo = re.search(VSRE, verstrline, re.M)
-    if mo:
-        __version__ = mo.group(1)
-    else:
-        msg = "if %s.py exists, it is required to be well-formed" % VERSIONFILE
-        raise RuntimeError(msg)
+    __version__ = subprocess.check_output([
+        "git", "describe", "--tags", "--dirty"]).decode('ASCII').strip()
```
kirbyfan64: Pretty sure this won't work outside the original directory , like `python ../setup.py`. You should pass `working_directory=os.path.dirname(__file__)` to `check_output`.

hy/__init__.py
```diff
@@ -19,7 +19,11 @@
 # DEALINGS IN THE SOFTWARE.
 
 
-from hy.version import __version__, __appname__  # NOQA
+__appname__ = 'hy'
+try:
+    from hy.version import __version__
+except ImportError:
+    __version__ = 'unknown'
```
kirbyfan64: This whole "unknown version" thing kind of icks me a bit. Here's a thought instead:

- `version.py` contains something like this:

    ```python
    __version__ = '0.12.0' # Or equivalent!
    try:
       from hy._version import git_hash as _git_hash
    except ImportError:
        pass
    else:
        __version__ += 'g' + _git_hash
    ```

And then `setup.py` just writes `_version.py`. That should get rid of the whole issue with `version.py` being non-existent.

setup.py
```diff
@@ -22,24 +22,19 @@
...
+    with open(VERSIONFILE, "wt") as o:
+        o.write("__version__ = {!r}\n".format(__version__))
```
kirbyfan64: Just to make sure, will this point to an old hash if I do `setup.py develop` and then change the Git hash?

kirbyfan64:
As a side note, I'm personally -0.5, since these things have a tendency to explode in weird ways (like I mentioned, `setup.py develop`), though I wouldn't really be against it if the "bigger" @hylang/core members wanted it in.

setup.py
Kodiologist: Calling setup.py outside of its own directory seems not to have worked in the first place. (Notice how `VERSIONFILE` was defined.) Is this something we want to support?

setup.py
Kodiologist: Yes, the version file is generated when you run setup.py, so you need to rerun setup.py if you want to refresh it.

hy/__init__.py
Kodiologist: Wouldn't that re-create the possibility of somebody thinking (or leading us to think) that they're on 0.11.0 (or whatever) when they're not?

Kodiologist:
>> `0.11.0-293-g34304c4-dirty` when the working tree is dirty
>
> Is this part really necessary? I normally find it quite jarring when you edit a file and now your version ended up changing.

Few things in this world are really necessary, but this seems useful to me so that we have a warning that the user reporting a bug has altered the code somehow. It may also catch a few of us slipping us (not that *I* ever make mistakes).

setup.py
kirbyfan64: @Kodiologist Oh wow, I didn't realize that. I'd prefer supported it personally, since it just adds one extra keyword argument. Though I suppose it's an extremely rare use case...

setup.py
```diff
@@ -22,24 +22,19 @@
...
-    pass  # Okay, there is no version file.
-else:
-    VSRE = r"^__version__ = ['\"]([^'\"]*)['\"]"
-    mo = re.search(VSRE, verstrline, re.M)
-    if mo:
-        __version__ = mo.group(1)
-    else:
-        msg = "if %s.py exists, it is required to be well-formed" % VERSIONFILE
-        raise RuntimeError(msg)
+    __version__ = subprocess.check_output([
```
zackmdavis: Won't this result in version being "unknown" if Git isn't available? (All reasonable _development_ environments will have Git, but we do at least theoretically have actual users who will be installing from PyPI.)

Kodiologist: Probably it depends on exactly what method we decide to use to put a package on PyPI. However, I think this ought to include a build step, as in [this example](https://pythonhosted.org/an_example_pypi_project/setuptools.html), which uses setup.py itself to do the upload.

setup.py
Kodiologist: Done.

kirbyfan64:
In the future, I think I would like to see a commit hook that does this so that `setup.py develop` still works. For now, though, this works.

tuturto:
To me this looks good. Are there any open issues we need to tackle before merging this in? This adds ```git``` as a dependency to setup.py, but I guess that's quite safe bet?

kirbyfan64:
Just to check, the releases, are going to come with `version.py` already generated, right?

tuturto:
That's my understanding based on the code and discussion here. That of course means that the person doing the release have to remember to generate the version.py, otherwise "unknown" is reported as version number. Good news is that users won't notice anything different.

kirbyfan64:
Crazy thought: what if re-generating `version.py` was a commit/receive hook? No one would need to remember to run it, it would automatically be updated, and it would work with `setup.py develop`.

Kodiologist:
I'm not familiar with how Git hooks work, but keep in mind that we still want to generate it at build-time if we want to be able to flag the build as dirty.

tuturto:
We could even have both I think. version.py would be updated every time code changes on the main repository and running setup.py would also generate it. But I would do that as a separate thing and get this in first and test how things work in practice. 

On that note, flagging build as dirty isn't surefire solution. If I want to tinker with internals of Hy, I usually install editable version from github. At that point workarea is clean and I'll get the same version number as the github version has. Only after that I start editing and messing with internals, but version.py won't be updated anymore.

tuturto:
Since this has been approved and there aren't open issues, I'll merge this. We can tackle the git hook in separate issue.

PR #1138, opened 2016-10-24
Give an error on empty attributes

### LINKED ISSUE #1137: "foo".format is legal syntax but does the wrong thing
`(. "foo" format)` works, but `"foo".format` doesn't (`NameError: name '' is not defined`). hy2py shows that `"foo".format` is parsed as two separate statements, `"foo"` and `.format`.

kirbyfan64:
Closes #1137.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2549,6 +2549,19 @@ class HyASTCompiler(object):
     def compile_symbol(self, symbol):
         if "." in symbol:
             glob, local = symbol.rsplit(".", 1)
+
+            if not glob:
+                raise HyTypeError(symbol, 'cannot access attribute on '
+                                          'anything other than a name '
+                                          '(in order to get attributes of'
+                                          'expressions, use '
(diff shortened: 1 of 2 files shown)


Kodiologist:
You're fast!

For completeness, I would add tests for `2.foo`, `"hello".foo`, and `[2].foo`.

Kodiologist:
It looks like the test fails because `2.foo` compiles, which happens because Hy interprets the `2` here as a reference to a variable named `2`. Is that a parser bug?

kirbyfan64:
Aaaaaand that's what happens when I write the tests and don't run them...

@Kodiologist I think that probably would be classified as a bug, since I can't think of a single reason why someone would do something that stupid, and I think this is the only way you can use a number as a variable.

kirbyfan64:
Meanwhile, I just removed the test.

kirbyfan64:
Anyone else want to approve this? :heart:

PR #1140, opened 2016-10-25
Release 0.12.0 preparation

tuturto:
Reopening pull request, because apparently I don't know how to rebase without causing a mess :D Sorry about the hassle.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -33,6 +33,7 @@ hopefully making the language easier to use.
    * old defmulti renamed to defn
    * Lazy sequences added to contrib
    * defmacro! added for once-only evaluation for parameters
+   * comp, constantly and complement added
 
    [ Bug Fixes ]
    * Better error when for doesn't have body
--- a/NEWS
+++ b/NEWS
@@ -43,7 +43,6 @@ hopefully making the language easier to use.
    * Allow specification of global table and module name for (eval ...)
    * General documentation improvements
    * Contrib.walk: Coerce non-list iterables into list form
-   * *args and **kwargs no longer have own args in Python 3.5
    * Flow macros (case and switch)
    * ap-pipe and ap-compose macros
    * #@ reader macro for with-decorator
--- a/NEWS
+++ b/NEWS
@@ -34,6 +34,7 @@ hopefully making the language easier to use.
    * Lazy sequences added to contrib
    * defmacro! added for once-only evaluation for parameters
    * comp, constantly, complement and juxt added
+   * keyword arguments allowed in method calls before the object
 
    [ Bug Fixes ]
    * Better error when for doesn't have body
--- a/NEWS
+++ b/NEWS
@@ -66,6 +66,7 @@ hopefully making the language easier to use.
    * ap-if requires then branch
    * Parameters for numeric operations (inc, dec, odd?, even?, etc.) aren't type checked
    * import_file_to_globals added for use in emacs inferior lisp mode
+   * hy.core.reserved added for querying reserved words
 
 Changes from 0.10.1
 
--- a/NEWS
+++ b/NEWS
@@ -29,6 +29,8 @@ hopefully making the language easier to use.
    * defreader can use strings as macro names
    * as-> macro added
    * require syntax changed and now supports same features as import
+   * defmulti changed to work with dispatching function
+   * old defmulti renamed to defn
 
    [ Bug Fixes ]
    * Better error when for doesn't have body
--- a/NEWS
+++ b/NEWS
@@ -33,7 +33,7 @@ hopefully making the language easier to use.
    * old defmulti renamed to defn
    * Lazy sequences added to contrib
    * defmacro! added for once-only evaluation for parameters
-   * comp, constantly and complement added
+   * comp, constantly, complement and juxt added
 
    [ Bug Fixes ]
    * Better error when for doesn't have body
--- a/NEWS
+++ b/NEWS
@@ -1,30 +1,29 @@
 Changes from 0.11.0
 
 This release brings some quite significant changes on the language and as a
-result very large portion of previously written Hy programs will require
+result very large portions of previously written Hy programs will require
 changes. At the same time, documentation and error messages were improved,
 hopefully making the language easier to use.
 
    [ Language Changes ]
    * New syntax for let, with and defclass
    * defmacro will raise an error on &kwonly, &kwargs and &key arguments
-   * Keyword arguments to functions are required to be strings
-   * slice replaced with cut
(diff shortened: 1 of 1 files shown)


tuturto:
@hylang/core should we include more in this release or is this ready to be cut? I'm inclined to get this out as soon as possible and release next version in 3-4 months from now on. After this has been acked and merged, @paultag can build packages and upload them to PyPi.

Kodiologist:
I would recommend settling the issue of the names for `True` and `False` (#908, #1147) first. That's likely going to be another big ol' breaking change.

I'd be grateful if you could merge #1105, too.

tuturto:
Looks like #908 and #1147 will take some time still. Not sure if it makes sense to wait for them, but lets see how long they'll take. I'll have a second look at #1105.

tuturto:
#1105 was merged, but there isn't much progress in #908 / #1147. Should we rather include it in the next release? Anything else we definitely want in this release? We can always make another release relatively soon.

Kodiologist:
@paultag It would be awesome if you could return to #1147 so maybe we can settle this before 0.12.0.

Kodiologist:
All right! I think we're ready to go now.

gilch:
We should fix the bug introduced by #924 first. That should never have made it into Hy in the first place, and certainly not a release.

NEWS
```diff
@@ -1,3 +1,68 @@
...
+   * global now takes multiple arguments
+   * Nonlocal keyword (Python 3 only)
+   * Set literals (#{1 2 3})
+   * Keyword-only arguments (Python 3 only)
+   * Setv can assign multiple variables at once
+   * Empty form allowed for setv, del and cond
+   * One-argument division, rationals and comparison operators (=, !=, <, >, <=, >=)
+   * partition form for chunking collection to n-sized tuples
+   * defn-alias and demacro-alias moved into hy.contrib.alias
+   * Nil is returned instead of the last form in --init--
```
gilch: Should say "None" now.

NEWS
```diff
@@ -1,3 +1,68 @@
...
+   * for and cond can take a multi-expression body
+   * Hex and octal support for integer literals
+   * Apply now mangles strings and keywords according to Hy mangling rules
+   * Variadic if
+   * defreader can use strings as macro names
+   * as-> macro added
+   * require syntax changed and now supports same features as import
+
```
gilch: The `xi` macro is probably worth a mention.

NEWS
```diff
@@ -1,3 +1,68 @@
...
+result very large portions of previously written Hy programs will require
+changes. At the same time, documentation and error messages were improved,
+hopefully making the language easier to use.
+
+   [ Language Changes ]
+   * New syntax for let, with and defclass
+   * defmacro will raise an error on &kwonly, &kwargs and &key arguments
+   * Keyword argument labels to functions are required to be strings
+   * slice replaced with cut to stop overloading the python built-in
+   * removed throw, catch, progn, defun, lisp-if, lisp-if-not and filterfalse
```
gilch: It should also say `true, false, nil`.

NEWS
tuturto: got it

NEWS
tuturto: added, thanks

Kodiologist:
I thought a bit about how `__version__` will be handled in the case of installing from PyPI or downloading a GitHub tarball and I came up with [these changes](https://github.com/Kodiologist/hy/commit/a98fabecb2145b526ce9e0d441d6dbb90bd4aeb7). I haven't opened a PR because I'm not sure if it's necessary or it will work. Let me know if you have any `__version__` issues in any case.

Kodiologist:
> We should fix the bug introduced by #924 first.

That doesn't seem forthcoming, since both #989 and #1114 (in addition to previous, closed attempts) have run into trouble, and the bug has been around for over a year. In any case, it's a rather obscure bug that, so far as I can tell, only means that when you incorrectly call a macro that has a funny parameter name, you won't get a useful error message. I don't think 0.12.0 should be held back for it.

It would make sense to create an issue for the bug, though, since it doesn't seem to have one.

gilch:
> In any case, it's a rather obscure bug that, so far as I can tell, only means that when you incorrectly call a macro that has a funny parameter name, you won't get a useful error message.

No, not that minor. It actually broke #903, which was working fine before #924. It's too late to automatically revert #924. Another change must have been applied on top of it. It will only get worse until its fixed, but I'm not sure exactly what's broken now.

Kodiologist:
Making an old pull request break isn't a major bug. It need not even be a bug (your pull request may not have been properly adapted to changes in Hy). And even if this were a major bug, we have no fix in sight after over a year and several attempts, so if we wait for a fix, we could be waiting for another year. We're already nearly at the one-year anniversary of 0.11.1; we've waited far too long for a new release. Lastly, 0.11.1 already had the bug, so our new release won't be a regression. It will just continue to contain a bug that so thoroughly escaped notice by users that nobody's created an issue for it.

rwtolbert:
I know I've been very quiet lately, so don't have much say. 

I agree with the general sentiment that we need to get 0.12.0 out.

The general idea of defmacro! is a really good one, so we should push to get it working post 0.12.0. I worked on defmacro/g! but could never get a version of defmacro! working, so please keep at it.

tuturto:
[merge+hack for #903](/tuturto/hy/tree/nugget) has some changes that make ``defmacro!`` to work again. We can even have nicer error messages for macro expansion most of the time.

jakirkham:
It would be nice if we could get the license file included in the next PyPI release.

xref: https://github.com/hylang/hy/pull/1186

Kodiologist:
I don't think we need to settle the question of `flip` for this release, so everything should be ready. Shall we get this out before 2017?

tuturto:
On a quick glance we don't have anything open that couldn't wait for next release (that hopefully won't take another year to make). :+1: from me for cutting a new release.

tuturto:
@hylang/core do we have any open issues/pull requests we want to get into next release?

tuturto:
nobody in @hylang/core has objected or wanted to add new content, so I'll tentatively tag this in the evening and test that PyPi package can be created and looks/feels correct. @paultag can then create package and upload it to PyPi.

tuturto:
Installed hy from git into a fresh virtualenv, copied tests directory and README.md and run nosetests. All tests pass.

So feel free to hit the ``Squash and merge`` magic button and tag the release.

kirbyfan64:
(!!!)[https://github.com/hylang/hy/releases]

(God, I hope I did that correctly...)

rcarmo:
Is it on PyPi yet?

Issue #1047, opened 2016-04-16
Add Clojure's as-> macro


gilch:
We already have `->` and `->>`, which threads through the first and last argument, respectively. Sometimes you need both or need to thread in the middle. Sometimes you can re-order args using Hy's keywords, but this isn't always possible, particularly for vararg functions. The `as->` macro lets you specify the threading point for each form, be it first, last, or in-between, by re-binding the result of each step to a short name you provide.

gilch:
Maybe something like this?
```Hy
(defmacro as->
 "TODO:docmebro"
 [head name &rest rest]
 (get (setv
       ~name ~head
       ~@(interleave (repeat name) rest))
  -1))
```

gilch:
Some example uses from ClojureDocs: https://clojuredocs.org/clojure.core/as-%3E

gilch:
#1048 was merged, perhaps prematurely. There's still a typo in its docstring. #1052 or #1049 might have changed the implementation. It still needs docs and probably a test.

PR #1141, opened 2016-10-25
Add docs and tests for as-> macro

tuturto:
Adds docs and tets for as-> macro. I had trouble running tests locally, but now that I tried to reproduce it to write down the error message they passed. So please check if that happens on your machine too.

relates to #1047

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -192,6 +192,16 @@ point per form via the name instead of always the first or last argument.
 
 .. code-block:: clj
 
+  ;; example how -> and as-> relate
+
+  => (as-> 0 it
+  ...      (inc it)
+  ...      (inc it))
+  2
+
+  => (-> 0 inc inc)
+  2
+
   ;; create data for our cuttlefish database
 
   => (setv data [{:name "hooded cuttlefish"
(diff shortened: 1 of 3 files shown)


docs/language/api.rst
```diff
@@ -179,6 +179,59 @@ other case, the first false value will be returned. Example usage:
...
+  => (as-> (filter (fn [entry] (= (:name entry)
+  ...                           "slender cuttlefish")) data) x
+  ...      (first x)
+  ...      (:discovered x)
+  ...      (:name x))
+  'Sir Joseph Cooke Verco'
+
+.. note::
+
+  In these examples, REPL will report a tupple as result:
```
kirbyfan64: **tuple**? (Also, missing **the** before REPL.)

docs/language/api.rst
```diff
@@ -179,6 +179,59 @@ other case, the first false value will be returned. Example usage:
...
+as->
+----
+
+.. versionadded:: 0.12.0
+
+Expands to sequence of assignments to the provided name, starting with head.
+The previous result is thus available in the subsequent form. Returns the final
+result, and leaves the name bound to it in the local scope. This behaves much
+like the other threading macros, but requires you to specify the threading
+point per form via the name instead of always the first or last arument.
```
kirbyfan64: ...this is kind of confusing. I mean, I'm guessing that there will likely never be a "simple" explanation for something like this. I think maybe at minimum the example should use something more descriptive than `x`; maybe `assigned_name`?

docs/language/api.rst
```diff
@@ -179,6 +179,59 @@ other case, the first false value will be returned. Example usage:
...
+  ...                           "slender cuttlefish")) data) x
+  ...      (first x)
+  ...      (:discovered x)
+  ...      (:name x))
+  'Sir Joseph Cooke Verco'
+
+.. note::
+
+  In these examples, REPL will report a tupple as result: 
+  ('Sepia prashadi', 'Sepia prashadi'), but the actual value returned is just
```
kirbyfan64: Maybe this should be surrounded by double-backticks for formatting? The sentence could also be worded a bit differently, like:

```
In these examples, the REPL will report a tuple (e.g. `('Sepia prashadi', 'Sepia prashadi')`) as the result, but only a single value is actually returned.
```

docs/language/api.rst
tuturto: good catch, tuple is one of those words that I consistently misspell.

docs/language/api.rst
tuturto: true, ```assigned-name``` is much better

docs/language/api.rst
tuturto: Thanks, I'll update this

gilch:
This might be a good time to fix that typo in the last word of the `as->` docstring (or even rewrite the whole thing to make it clearer). Not that we can even access those for macros #892.

tuturto:
Good catch with the typo @gilch. I hadn't fixed it in both locations.

docs/language/api.rst
tuturto: Instead of ``assigned-name`` I could also us ``it`` as it's short and generic enough. I'll try to come up with a better examples still, that would highlight usage of ``as->`` better (as now the example could be easily rewritten with ``->``).

tuturto:
Polished the documentation a bit again. Anything else that needs some cleaning up here or would this soon be good to go?

olasd:
:+1:

kirbyfan64:
:+1:

tuturto:
As this has multiple thumbs up already, I'll merge this one too (even if I'm the originator).

Issue #1118, opened 2016-09-25
[feature] Allow selective macro import


Kodiologist:
This is related to #277, but (I think) substantially less ambitious.

Right now, the only way macros from another file or module can be included is with `(require foo)`, which imports all macros defined in the module `foo`. So there are all the usual namespace-pollution perils of `(import [foo [*]])`: `foo` might have a macro added to it that happens to have the same name as something in the importing module. The risk of breakage is even greater than with `import` because ordinarily imported variables can be overwritten by ordinary assignments, whereas macros take precedence over ordinary assignments. Hy doesn't even have a warning for a macro shadowing a function.

To fill this gap, I propose a feature that I think will be pretty easy to implement. Analogous to `(import [foo [a b]])`, we should allow `(require [foo [a b]])`, which imports the macros `a` and `b` but nothing else (and throws an error if `a` or `b` isn't a macro available for import).

To make the syntax of `require` more consistent with `import`, we might also make `(require [foo [*]])` mean what `(require foo)` currently means, and make `(require foo)` raise a deprecation warning, so that later we can have it import macros named `foo.a`, `foo.b`, etc., like `(import foo)` does for ordinary names.

gilch:
Yeah, I'm inclined to agree that's a problem. 

I'm not sure yet of that solution, but it seems reasonable at a glance. I'd like to look deeper at how Hy's macros are implemented, maybe treating them as living in modules isn't appropriate, and something like Clojure namespaces is better. Someone more familiar with that part of Hy's internals might have more to say about it.

I'm not sure I'd bother with the deprecation warning at this time, since we're breaking everything for the Grand Language Cleanup anyway.

tuturto:
Closed by 14fddbe

PR #1142, opened 2016-10-27
Give `require` the same features as `import`

Kodiologist:
You can now do `(require foo)`, `(require [foo [a b c]])`, `(require [foo [*]])`, and `(require [foo :as bar])`. The first and last forms get you macros named `foo.a`, `foo.b`, etc. or `bar.a`, `bar.b`, etc., respectively. The second form only gets the macros in the list.

Implements #1118 and perhaps partly addresses #277.

N.B. The new meaning of `(require foo)` will cause all existing code that uses macros to break. Simply replace these forms with `(require [foo [*]])` to get your code working again.

There's a bit of a hack involved in the forms `(require foo)` and `(require [foo :as bar])`. When you call `(foo.a ...)` or `(bar.a ...)`, Hy doesn't actually look inside modules. Instead, these `(require ...)` forms give the macros names that have periods in them, which happens to work fine with the way Hy finds and interprets macro calls.

If this all looks good, I'll update the documentation accordingly.

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -1214,7 +1214,7 @@ require
 parameters in all the same formats as ``import``. The ``require`` form itself
 produces no code in the final program: its effect is purely at compile-time, for
 the benefit of macro expansion. Specifically, ``require`` imports each named
-module and then makes available each requested macro in the current module.
+module and then makes each requested macro available in the current module.
 
 The following are all equivalent ways to call a macro named ``foo`` in the module ``mymodule``:
 
@@ -1271,7 +1271,7 @@ repexpr]])``, but a less error-prone approach is to change the definition of
         (require mymodule)
         (mymodule.repexpr ~n (raw-input "Gimme some input: "))))
 
-It's wise to use ``(require mymodule)`` rather than ``(require [mymodule
+It's wise to use ``(require mymodule)`` here rather than ``(require [mymodule
 [repexpr]])`` to avoid accidentally shadowing a function named ``repexpr`` in
 the main program.
 
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -1210,16 +1210,81 @@ alternatively be written using the apostrophe (``'``) symbol.
 require
 -------
 
-``require`` is used to import macros from a given module. It takes at least one
-parameter specifying the module which macros should be imported. Multiple
-modules can be imported with a single ``require``.
+``require`` is used to import macros from one or more given modules. It allows
+parameters in all the same formats as ``import``. The ``require`` form itself
+produces no code in the final program: its effect is purely at compile-time, for
+the benefit of macro expansion. Specifically, ``require`` imports each named
+module and then makes available each requested macro in the current module.
 
-The following example will import macros from ``module-1`` and ``module-2``:
+The following are all equivalent ways to call a macro named ``foo`` in the module ``mymodule``:
 
 .. code-block:: clj
 
-    (require module-1 module-2)
+    (require mymodule)
+    (mymodule.foo 1)
 
+    (require [mymodule :as M])
+    (M.foo 1)
+
+    (require [mymodule [foo]])
+    (foo 1)
+
+    (require [mymodule [*]])
+    (foo 1)
+
+Macros that call macros
+~~~~~~~~~~~~~~~~~~~~~~~
+
+One aspect of ``require`` that may be surprising is what happens when one
+macro's expansion calls another macro. Suppose ``mymodule.hy`` looks like this:
+
+.. code-block:: clj
+
+    (defmacro repexpr [n expr]
+      ; Evaluate the expression n times
+      ; and collect the results in a list.
+      `(list (map (fn [_] ~expr) (range ~n))))
+
+    (defmacro foo [n]
+      `(repexpr ~n (input "Gimme some input: ")))
+
+And then, in your main program, you write:
+
+.. code-block:: clj
+
+    (require [mymodule [foo]])
+
+    (print (mymodule.foo 3))
+
+Running this raises ``NameError: name 'repexpr' is not defined``, even though
+writing ``(print (foo 3))`` in ``mymodule`` works fine. The trouble is that your
+main program doesn't have the macro ``repexpr`` available, since it wasn't
+imported (and imported under exactly that name, as opposed to a qualified name).
+You could do ``(require [mymodule [*]])`` or ``(require [mymodule [foo
+repexpr]])``, but a less error-prone approach is to change the definition of
+``foo`` to require whatever sub-macros it needs:
+
+.. code-block:: clj
+
+    (defmacro foo [n]
+      `(do
+        (require mymodule)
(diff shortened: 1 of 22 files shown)


hy/compiler.py
```diff
@@ -1737,10 +1737,35 @@ def compile_require(self, expression):
         "unimport" it after we've completed `thing' so that we don't pollute
         other envs.
         """
-        expression.pop(0)
-        for entry in expression:
-            __import__(entry)  # Import it fo' them macros.
-            require(entry, self.module_name)
+        for entry in expression[1:]:
+            if isinstance(entry, HySymbol):
+                # e.g., (require foo)
```
kirbyfan64: REALLY minor nitpick, but this would be a little easier to read without the `e.g.`, since I think it's obvious that it's an example.

tests/resources/tlib.py
```diff
@@ -1,7 +1,12 @@
 from hy.macros import macro
-from hy import HyList
+from hy import HyList, HyInteger
 
 
 @macro("qplah")
 def tmac(*tree):
-    return HyList(tree)
+    return HyList((HyInteger(8), ) + tree)
```
kirbyfan64: ???

hy/compiler.py
```diff
@@ -1737,10 +1737,35 @@ def compile_require(self, expression):
...
+                __import__(entry)
+                require(entry, self.module_name, all_names=True,
+                        prefix=entry)
+            elif (isinstance(entry, HyList) and len(entry) == 2
```
kirbyfan64: This branch and the two that follow repeat a lot of code and don't handle stuff like `(require [x [* *]])` correctly.

hy/compiler.py
```diff
@@ -1737,10 +1737,35 @@ def compile_require(self, expression):
         "unimport" it after we've completed `thing' so that we don't pollute
         other envs.
         """
-        expression.pop(0)
-        for entry in expression:
-            __import__(entry)  # Import it fo' them macros.
-            require(entry, self.module_name)
+        for entry in expression[1:]:
```
kirbyfan64: It seems kind of odd that `require` and `import` handle a pretty much identical syntax in completely different ways. Would it somehow be possible to somehow merge the two?

kirbyfan64:
FWIW, I really like the idea behind this change, but this is quite literally going to one of the biggest breaking changes that have been done...

hy/compiler.py
Kodiologist: They use the same syntax but do pretty different things because `import` actually emits code and `require` just changes the state of the compiler. I think trying to unify the implementations wouldn't make the compiler code much nicer.

hy/compiler.py
Kodiologist: (1) Regarding repetition, I don't think there's enough of substance to factor out, because each branch is small but requires small differences; can you specify what kind of refactoring you'd like to see? (2) What should `(require [x [* *]])` do? Import a macro named `*` twice? `*` is a terrible name for a macro.

tests/resources/tlib.py
Kodiologist: So the difference between the two test macros is clearer, and both actually do something, so we can catch a bug in which the macro isn't being called and its argument is returned unaltered.

hy/compiler.py
kirbyfan64: @Kodiologist Probably be an error, though `(import [x [* *]])` works for some weird reason.

tuturto:
This is good change in my opinion. It will be a bit of a pain to fix all of existing code, but it'll be worth it.

Kodiologist:
@kirbyfan64 Are you happy with the changes I've made? Regarding the "e.g." in the comments, zackmdavis thumbed your comment down, and I guess I agree with him, although it's not very important in any case.

kirbyfan64:
@Kodiologist Whoops, forgot about this.

kirbyfan64:
@hylang/core Does this count as two acks now, since @tuturto said he agrees with the changes, or does someone else have to actually hit the "Approve" button?

kirbyfan64:
@Kodiologist Actually, no, I take that back: you forgot to update the docs!

Kodiologist:
Okay, I'll take a stab at the docs now that everybody's happy with the feature itself.

Kodiologist:
Done.

docs/language/api.rst
```diff
@@ -1210,16 +1210,81 @@ alternatively be written using the apostrophe (``'``) symbol.
 require
 -------
 
-``require`` is used to import macros from a given module. It takes at least one
-parameter specifying the module which macros should be imported. Multiple
-modules can be imported with a single ``require``.
+``require`` is used to import macros from one or more given modules. It allows
+parameters in all the same formats as ``import``. The ``require`` form itself
```
tuturto: If I read the code correctly, following doesn't work:
```
(require [foo [bar :as baz]])
```
but similar does work for ```import``` though
```
(import [foo [bar :as baz]])
```

docs/language/api.rst
```diff
@@ -1210,16 +1210,81 @@ alternatively be written using the apostrophe (``'``) symbol.
...
+    (require [mymodule :as M])
+    (M.foo 1)
+
+    (require [mymodule [foo]])
+    (foo 1)
+
+    (require [mymodule [*]])
+    (foo 1)
+
+Macros that call macros
```
tuturto: This is good addition. I got bitten by this multiple times, until I learned it hard way.

docs/language/api.rst
Kodiologist: Embarassingly, I had no idea that `from foo import bar as baz` is legal Python, let alone supported in Hy. I'll add that form for `require`.

Kodiologist: Done.

tuturto:
One implication that I didn't spot, but that was pointed out to me is that if we have a module with several macros that depend on each other, we either need to make sure all needed macros are actually required. This means that client code has to require them explicitly or with *, or that the macros themselves have to require them. This can make code somewhat ugly. What do you think, how big problem this is?

Kodiologist:
You're welcome. :)

> if we have a module with several macros that depend on each other, we either need to make sure all needed macros are actually required. This means that client code has to require them explicitly or with *, or that the macros themselves have to require them. This can make code somewhat ugly. What do you think, how big problem this is?

That's just the same problem as I described in "Macros that call macros", right? I think having the macros themselves require their dependencies is the right thing to do, as I said in the documentation. But you can reduce repetitive code by writing a function like:

    (defn require-me [&rest body]
      `(do (require mymodule) ~@body))

which you'd use like:

    (defmacro foo [n]
      (require-me `(mymodule.repexpr ~n (raw-input "Gimme some input: "))))

And in general, actually, if you're writing some code specifically to be a subroutine for one or more macros, it's more convenient and less error-prone to write it as a function than a macro.

tuturto:
It's just something to be aware of I guess. Since this has two acks and no open issues I'm aware of, I'll merge this into master.

PR #1147, opened 2016-11-03
Replace `true` with `T` and `false` with `F`

Kodiologist:
Per discussion in #908.

`True` and `False` are still allowed. Allowing users to use these as variable names is probably more trouble than it's worth, and I figure this helps to smooth over human translation from Python code to Hy. It might be a good idea to use `T` and `F` consistently in Hy's own code, not to mention the documentation. (This preference could be stated in the style guide.)

I added a bit to the documentation about the correspondences between `True`, `False`, `None`, and their synonyms because this didn't seem to be documented at all.

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -34,6 +34,13 @@ languages.
 Notes on Syntax
 ===============
 
+logical values
+--------------
+
+``True``, ``False``, and ``None`` have the same meaning as in Python.
+In addition, ``True`` may be written as ``T``, ``False`` as ``F``, and
+``None`` as ``nil``.
+
 integers
 --------
 
@@ -392,7 +399,7 @@ Some example usage:
 
 .. code-block:: clj
 
-    => (if true
+    => (if T
     ...  (do (print "Side effects rock!")
     ...      (print "Yeah, really!")))
     Side effects rock!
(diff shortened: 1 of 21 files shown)


olasd:
I'm sorry but I'm :-1: on this. I feel it makes the code completely unreadable, it invents a convention and provides exactly no gain over `true`/`false`.

paultag:
We should honestly get rid of `true` and `false` as well, having "one way" to do it seems right to me

But hey, that whole mess is my fault :)

paultag:
(same with `nil`)

Kodiologist:
@paultag, are you arguing for `T` and `F` or `True` and `False`?

rcarmo:
Huh. `T` and `F`, seriously? .... but.... why?

Kodiologist:
It makes sense to me that a programming language's most fundamental constants should have concise representations.

rcarmo:
Well, it makes sense to me that they be readable. Racket’s #t and #f is hideous, for instance. Clojure does the sensible thing.

And besides, it’s not as if the extra three or four characters take up that much space on a line of code.

arrdem:
This is a PR which may be closed and succeeded. Please discuss the matter on the tracking ticket #908.

PR #1148, opened 2016-11-07
Add tests for the empty keyword

Kodiologist:
I have some macros for using pandas and NumPy that expect `:` to be a keyword instead of an ordinary symbol. These tests will ensure that we don't break this unless we want to.

### DIFF
--- a/tests/native_tests/language.hy
+++ b/tests/native_tests/language.hy
@@ -1047,6 +1047,19 @@
   (assert (= (get {:foo "bar" ":foo" "quux"} :foo) "bar"))
   (assert (= (get {:foo "bar" ":foo" "quux"} ":foo") "quux")))
 
+
+(defn test-empty-keyword []
+  "NATIVE: test that the empty keyword is recognized"
+  (assert (= : :))
+  (assert (keyword? :))
+  (assert (!= : ":"))
+  (assert (= (name :) ""))
+
+  (defn f [&kwargs kwargs]
+    (list (.items kwargs)))
+  (assert (= (f : 3) [(, "" 3)])))
+
+
 (defn test-nested-if []
   "NATIVE: test nested if"
   (for [x (range 10)]


tuturto:
Not quite sure what I would expect ``:`` to be actually. Why do we want it to be an empty keyword instead of an error?

Kodiologist:
Making it a syntax error would prevent me from making `(geti x : 1)` in Hy work like `x.iloc[:, 1]` in Python.

tuturto:
True, that wouldn't work at all. Had to read up a bit about pandas to understand what's actually going on there. Not 100% sold yet though, I'll mull over this a bit.

gilch:
The `:` appears to be a syntax error in both Clojure and Common Lisp. Emacs Lisp allows it though, and both `(keywordp :)` and `(symbolp :)` return `t`.

I think NumPy support is very important for the future of Hy, #541. I'm not sure of the right approach yet, but it might be nice to use `:` in slicing macros.

gilch:
@tuturto, it's not _that_ hard in Python. `(foo : 'bar)` would be written `foo(**{'':HySymbol('bar')})`

tuturto:
Yes, ``:`` looks good and logical in the example, but I'm not sure it has to be keyword. On the other hand, maybe I shouldn't be concerned that users will create some funky code if ``:`` is treated as a keyword? What other implications are there, if we decide to go this path? Or rather decide that ``:`` being treated as an empty keyword is actual feature and not a bug.

Kodiologist:
Presuming our choice is between treating it as a keyword or as a syntax error, the only downside to treating it as a keyword that I can think of is that if you mistype a keyword argument `:foo` as `: foo`, you'll get an error at run time instead of compile time.

tuturto:
We could also treat ``:`` merely as a symbol, instead of a empty keyword. But of course the usual way to write that would be ``':`` instead of ``:``. So, looks good to me I say.

tuturto:
@hylang/core want to have a look and squash/merge this in?

Kodiologist:
The two-core-approvers-other-than-the-submitter rule is a little tough when there's only four of us who are really active. Eight dubloons to anybody who can shanghai some fresh blood into core.

gilch:
> But of course the usual way to write that would be `':` instead of `:`

We mainly wanted to use `:` in macros. And in macros, it doesn't have to have the quote. You could also assign a function to it or some other value if it's a symbol. Keywords always evaluate to themselves though. A symbol seems more flexible, but a leading `:` means "keyword" for everything else. I'm not sure which is better.

gilch:
> The two-core-approvers-other-than-the-submitter rule is a little tough [...]

I still think the rule is important. You can do anything you want in your own fork, but you have to convince a couple of core members that it's a good idea before we accept a pull request. 

> [...] when there's only four of us who are really active. Eight dubloons to anybody who can shanghai some fresh blood into core.

I don't know of anyone we could recruit at the moment. The more people who are interested in Hy, the more people will contribute. If we put more projects written in Hy up on GitHub we might get more attention. _Pullum, et ovo. Ergo, bootstrap._

There are more of us that are less active, but not inactive. We're not all in the same time zone, so our availability varies. If you make a change when we're too busy we might forget about it when we have the time. If you can talk to people on IRC that might help get things moving. Github also has a new Projects tab that might be worth looking into. It might help prioritize things. We could also try to make assignments.

Kodiologist:
> I still think the rule is important.

Agreed.

> If you can talk to people on IRC that might help get things moving.

The IRC channel doesn't seem to be very active, but I guess the chief troublesome thing is that there are no logs, so I can't see what anybody's said when I'm gone. On the other hand, if there were logs, it would promote splitting the already limited community attention into the IRC channel and the mailing list. (Perl 6, which is much bigger, had a particularly bad version of this problem last time I checked, with completely different people on the IRC channel and the mailing list.) I'm inclined to say we should axe the IRC channel and promote heavier use of the mailing list. Mo' fora mo' problems.

> We could also try to make assignments.

That's not a bad idea.

tuturto:
I would be sad to see irc to go. It's excellent place for informal chat about Hy, projects using Hy and even some unrelated things now and then. But I agree that it's not as active as it used to be. I guess it's one of those things that need to have active users, so that it will attract more active users.

PR #1152, opened 2016-11-17
Fix #1151

### LINKED ISSUE #1151: Lvalue of setv is checked too early when using `or`
This is a subtopic of the dreaded #233 that I think should have been fixed by #824. Consider the code:

    (import [random [random]])
    (setv x (or (< (random) .5) (do (setv y 1) y)))

This gets translated to:

    from random import random
    _hy_anon_var_1 = (random() < 0.5)
    if (not x):
        y = 1
        _hy_anon_var_1 = y

so running it raises `NameError: name 'x' is not defined
`.

For what it's worth, back in 0.11.1, the result would've been:

    from random import random
    y = 1
    x = ((random() < 0.5) or y)

which is wrong because we aren't even attempting the short-circuit.

kirbyfan64:
It was caused by improper management of temp_variables when compiling and/or statements.

See, that wasn't so hard. :)

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1807,16 +1807,17 @@ class HyASTCompiler(object):
                                  ctx=ast.Load(),
                                  lineno=root_line,
                                  col_offset=root_column)
+            temp_variables = [name, expr_name]
 
             def make_assign(value, node=None):
                 if node is None:
                     line, column = root_line, root_column
                 else:
                     line, column = node.lineno, node.col_offset
-                return ast.Assign(targets=[ast.Name(id=var,
-                                                    ctx=ast.Store(),
-                                                    lineno=line,
-                                                    col_offset=column)],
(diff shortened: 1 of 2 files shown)


Kodiologist:
Excellent! I think it should be supplemented by tests for both `or` and `and` that try using a variety of Python statements (e.g., `raise`, `for`, `defclass`) in the arguments. Would you like to write those or shall I?

Kodiologist:
The Travis failure is just Flake8 whining about whitespace, and in some files you didn't change. I'm not sure what that's about.

gilch:
Let's force a recheck.

Kodiologist:
More tests: https://gist.github.com/Kodiologist/52da5fd25d558e31f2259e980d1945a1

kirbyfan64:
@Kodiologist Thanks!

PR #1153, opened 2016-11-21
ignore PyCharm .idea/

gilch:
This is a hidden folder that PyCharm (or IntelliJ with equivalent plugin) uses for bookkeeping. An IDE like PyCharm can make some aspects of managing bigger Python projects (like Hy) easier to deal with, but I don't think it belongs in the repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ dist
 .coverage
 build/
 .noseids
+.idea/


gilch:
😕 why is flake8 complaining about .gitignore?

Kodiologist:
It isn't. It's complaining about a bunch of files you didn't change, as in #1152.

berkerpeksag:
I think user-specific values should go into your global .gitignore file. I'd like to keep the scope of .gitignore small and only cover things that are specific to hylang project itself.

Kodiologist:
I agree. The tracked `.gitignore` should be for files related to the project's own toolchain. Individual developers can ignore files that their own tools generate by using `.git/info/exclude`.

PR #1156, opened 2016-11-24
Burninate the synonyms `true`, `false`, and `nil`

Kodiologist:
Per the straw poll in #908, as an alternative to #1147.

Now you must use `True`, `False`, and `None`, as in Python. Or just assign `true` to `True`, etc.; the old synonyms aren't reserved words anymore.

I got rid of `nil?`, too, since we also have `none?`.

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -392,7 +392,7 @@ Some example usage:
 
 .. code-block:: clj
 
-    => (if true
+    => (if True
     ...  (do (print "Side effects rock!")
     ...      (print "Yeah, really!")))
     Side effects rock!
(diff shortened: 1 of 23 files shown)


tuturto:
Looks good, I didn't spot any problems with change (albeit it's lots of change). We can deal with flake8 errors in a separate PR I think.

gilch:
@Kodiologist you missed some in the style guide.

Kodiologist:
@gilch I just searched for `true`, `false`, and `nil` in `docs/style-guide.rst` and I didn't find anything.

gilch:
And in `docs/language/`-- `tutorial.rst`, `api.rst`, `core.rst`, `internals.rst`.

gilch:
My bad, I was on the wrong branch.

PR #1158, opened 2016-11-26
Ignore a new Flake8 whitespace error

### LINKED ISSUE #1157: Fix flake8 warnings regarding requiring two empty lines after class definition
From PyCQA/pycodestyle@1477fce8dfb65fa3d31a9757a477d3b6239b2f8e

"Added check E305 for two blank lines after toplevel method and toplevel class"

Which seems to be reason Travis builds are failing.

Kodiologist:
Fixes #1157.

The new check is E305. Since we're now using the `--ignore` option, we have to list all the checks that are ignored by default, too.

I decided that ignoring E305 was better than changing the whitespace it was complaining about because, in at least some cases in our current codebase, single blank lines are used to indicate that several top-level definitions are associated with each other.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ tox: venv
 	tox
 
 flake:
-	flake8 hy tests --ignore=E121,E123,E126,E226,E24,E704,W503,W504,E305
+	flake8 hy tests --ignore=E121,E123,E126,E226,E24,E704,W503,E305
 
 clear:
 	clear
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ tox: venv
 	tox
 
 flake:
-	flake8 hy tests
+	flake8 hy tests --ignore=E121,E123,E126,E226,E24,E704,W503,W504,E305
 
 clear:
 	clear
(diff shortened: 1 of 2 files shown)


gilch:
👍 I'm not sure this is the right solution long term, but I'll approve this for now, since it's kind of holding up other PRs.

gilch:
Here's the [table for the error codes](https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes), if that helps anyone.

gilch:
Wait, what's `W504`?

Kodiologist:
Good question. It's in the default of `--ignore` and it's mentioned in passing a few times in the Flake8 documentation. Perhaps it's a check that used to exist but has been removed.

gilch:
Did I find the wrong table? Does everything still work if you delete `W504,` then? If so, you should probably leave it out.

gilch:
Maybe we should also bug the flake8 people about this inconsistency. Or maybe the pycodestyle people. They may have neglected to update their docs.

Kodiologist:
That would be my guess.

gilch:
https://github.com/PyCQA/pycodestyle/pull/502

Looks like `W504` isn't ready yet. The meaning could still change though. Best leave it off for now. We can always put it back later if we have to.

Kodiologist:
Makes sense, I guess.

tuturto:
Looks good to me

PR #1163, opened 2016-12-08
Allow hy2py to use standard input instead of a file

### LINKED ISSUE #1160: Allow hy2py to use standard input instead of a file
Then, at the command line, you could say

    echo '(print "hello")' | hy2py

instead of

    hy2py <(echo '(print "hello")')

The construct `<(...)` only works in Zsh, anyway.

Kodiologist:
Fixes #1160.

### DIFF
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -40,7 +40,8 @@ import hy
 from hy.lex import LexException, PrematureEndOfInput, tokenize
 from hy.compiler import hy_compile, HyTypeError
 from hy.importer import (ast_compile, import_buffer_to_module,
-                         import_file_to_ast, import_file_to_hst)
+                         import_file_to_ast, import_file_to_hst,
+                         import_buffer_to_ast, import_buffer_to_hst)
 from hy.completer import completion
 from hy.completer import Completer
 
@@ -366,9 +367,12 @@ def hy2py_main():
     import platform
     module_name = "<STDIN>"
 
-    options = dict(prog="hy2py", usage="%(prog)s [options] FILE",
+    options = dict(prog="hy2py", usage="%(prog)s [options] [FILE]",
                    formatter_class=argparse.RawDescriptionHelpFormatter)
     parser = argparse.ArgumentParser(**options)
+    parser.add_argument("FILE", type=str, nargs='?',
+                        help="Input Hy code (use STDIN if \"-\" or "
+                             "not provided)")
     parser.add_argument("--with-source", "-s", action="store_true",
                         help="Show the parsed source structure")
     parser.add_argument("--with-ast", "-a", action="store_true",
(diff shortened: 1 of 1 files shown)


hy/cmdline.py
```diff
@@ -366,26 +367,30 @@ def hy2py_main():
...
                               "from the AST"))
-    parser.add_argument('args', nargs=argparse.REMAINDER,
-                        help=argparse.SUPPRESS)
 
     options = parser.parse_args(sys.argv[1:])
 
-    if not options.args:
-        parser.exit(1, parser.format_help())
+    stdin_text = None
+    if not options.FILE or options.FILE == '-':
```
kirbyfan64: `if options.FILE is None`

Otherwise, this will end up reading from stdin with the empty string as an argument.

Kodiologist: All right, but let it be known that I discourage naming a Hy program with the empty string.

hy/cmdline.py
```diff
@@ -366,26 +367,30 @@ def hy2py_main():
...
 
-    if not options.args:
-        parser.exit(1, parser.format_help())
+    stdin_text = None
+    if options.FILE is None or options.FILE == '-':
+        stdin_text = sys.stdin.read()
 
     if options.with_source:
-        hst = pretty_error(import_file_to_hst, options.args[0])
+        hst = (pretty_error(import_file_to_hst, options.FILE)
```
kirbyfan64: This is kinda bugging me. I mean, it just feels a bit weird to duplicate such similar calls. What if you did a function partial, like `pretty_error_partial = partial(pretty_error, ...)` and then used that? (I should've said this before, shouldn't I... :/ )

Kodiologist: But `import_file_to_hst` and `import_buffer_to_hst` are different functions.

kirbyfan64: @Kodiologist I meant:

```python
if stdin_text is None:
    pretty_error_partial = partial(import_file_to_hst, options.file)
else:
    pretty_error_partial = partial(import_buffer_to_hst, stdin_text)
```

and then replace all these with `pretty_error_partial(whatever_other_args)`.

Kodiologist: But `pretty_error_partial` would have only one use, on the line starting `hst = ...`, so what's the point? You realize that `_ast = ...` uses another two functions, namely `import_file_to_ast` and `import_buffer_to_ast`, not `import_file_to_hst` and `import_buffer_to_hst`, right? Or maybe I'm just having a brain fart. I'd recommend just pushing the edit you want to make instead of trying to explain it to me.

Kodiologist: @kirbyfan64 Any updates?

kirbyfan64: Oh right, sorry, I just apparently have no ability to read. *facepalm*

PR #1165, opened 2016-12-08
Added Tritlo's Tail recursion optimizations

bowbahdoe:
(See issue #728 )


Tritlo:
I fixed up my pull request #728, and I'd prefer that that would be accepted rather than this one, seeing as it is my code.

Tritlo:
Hmm. @bowbahdoe, do you understand the error that we get from travis?

PR #1167, opened 2016-12-09
Allow keyword args in method calls before the obj

Kodiologist:
Unlike Python, Hy allows the programmer to intermingle positional and keyword arguments. This change removes an exception to that rule for method calls, in which the method callee always had to be the first thing after the method. Thus, `(.split :sep "o" "foo")` now compiles to `"foo".split(sep="o")` instead of `HyKeyword("sep").split("o", "foo")`.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2068,12 +2068,25 @@ class HyASTCompiler(object):
 
                 # Get the object we're calling the method on
                 # (extracted with the attribute access DSL)
-                if len(expression) < 2:
-                    raise HyTypeError(expression,
-                                      "attribute access requires object")
-
+                i = 1
+                if len(expression) != 2:
+                    # If the expression has only one object,
+                    # always use that as the callee.
+                    # Otherwise, hunt for the first thing that
+                    # isn't a keyword argument or its value.
+                    while i < len(expression):
+                        if isinstance(expression[i], HyKeyword):
+                            # Skip the keyword argument and its value.
+                            i += 1
+                        else:
+                            # Use expression[i].
+                            break
+                        i += 1
+                    else:
+                        raise HyTypeError(expression,
+                                          "attribute access requires object")
                 func = self.compile(HyExpression(
-                    [HySymbol(".").replace(fn), expression.pop(1)] +
+                    [HySymbol(".").replace(fn), expression.pop(i)] +
                     attrs))
 
                 # And get the method
--- a/tests/native_tests/language.hy
+++ b/tests/native_tests/language.hy
@@ -397,26 +397,32 @@
 
   (defclass X [object] [])
   (defclass M [object]
-    [meth (fn [self &rest args]
-      (.join " " (+ (, "meth") args)))])
+    [meth (fn [self &rest args &kwargs kwargs]
+      (.join " " (+ (, "meth") args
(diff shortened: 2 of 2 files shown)


kirbyfan64:
I dunno...I find this one a bit confusing. When you see `(.myfunc ...)`, the eye immediately looks right after to see what the method is being called in.

Kodiologist:
As an example from my own code, I'd like to write

    (.sort-values :ascending F (.append
      (.mean
        (getl obs1 : (filt (.startswith it "race_") obs1.columns)))
      (pd.Series :index ["race_hispanic"] (.mean
        (= (ss ($ obs1 race_hispanic) (pd.notnull $)) "Yes")))))

instead of

    (.sort-values
      (.append
        (.mean
          (getl obs1 : (filt (.startswith it "race_") obs1.columns)))
        (pd.Series :index ["race_hispanic"] (.mean
          (= (ss ($ obs1 race_hispanic) (pd.notnull $)) "Yes"))))
      :ascending F)

or something of that ilk. It's annoying to have to strand the keyword argument and its value at the end, long after `.sort-values` .

gilch:
I'm still not sure what I think of this one. The argument for it is not as strong as it looks. Python kind of has the same problem. In most cases you can also pass positional arguments as keyword arguments:
```Python
>>> def foo(bar,baz=False):
	print(bar, baz)

	
>>> foo(baz=True,bar="""
Something really
really
really
long?""")
```
```
Something really
really
really
long? True
```
Hy should already be able to do this.

That said, there are cases when it doesn't work. For example, a builtin written in C might not have named arguments:
```Python
>>> import operator as op
>>> op.sub(b=5,a=7)
  File "<pyshell#63>", line 1, in <module>
    op.sub(b=5,a=7)
TypeError: sub() takes no keyword arguments
```
You can't pass `*args` that way either. So there are some reasons for it. I need to double check how Clojure and Common Lisp do it.

If we do put this in Hy, I think there are some other things that also need to change to match.

Kodiologist:
> That said, there are cases when it doesn't work. For example, a builtin written in C might not have named arguments

Yes, and, aggravatingly, the Python documentation doesn't mention which methods and functions can't accept keyword arguments. Worse, it probably varies among implementations of Python.

> I need to double check how Clojure and Common Lisp do it.

Dunno about Clojure, but CLOS uses multiple dispatch, so there's no one specially designated invocant. In fact, there's no syntactic difference at all between calling a function and a method. I think this is the right thing to do in a Lisp, but trying to make Hy do it is out of the question because it would involve screwing with Python's name resolution rules.

> If we do put this in Hy, I think there are some other things that also need to change to match.

I'm not sure what you mean. What other things need to change?

For what it's worth, my old `kwc` macro that I used for calling functions with keyword arguments had this feature (it treated method calls and plain functions the same way) and I never hit any weird surprises with it.

Kodiologist:
Yep, that works in Hy (although not in Python 2, where it raises `TypeError: unbound method doo() must be called with Foo instance as first argument (got nothing instead)`):

     (defclass Foo [] [
       __init__ (fn [self foo]
         (setv self.foo foo))
       doo (fn [self bar]
         (+ self.foo bar))])

     (setv spam (Foo "spam"))
     (print (.doo Foo :bar "eggs" :self spam))

Without testing it too thoroughly, I would guess that this doesn't use the same method resolution order as `(.doo spam)`.

PR #956, opened 2015-10-12
Fix issue found by @tuturto in #903

kirbyfan64:

PR #1172, opened 2016-12-12
add defmacro! and fix macro expansion error message

tuturto:
Alternative take on #956, #958 and #989. This one is somewhat simpler approach and doesn't produce nice error messages on macro expansion when parameters have invalid names (from point of view of Python). In most of the cases, error messages are nice.


gilch:
closes #903 
closes #989 
closes #1114 

I'm not sure I can be the one to merge this since many of the commits are mine. @hylang/core care to take a look? (Remember to squash and merge.)

> This one is somewhat simpler approach and doesn't produce nice error messages on macro expansion when parameters have invalid names (from point of view of Python). In most of the cases, error messages are nice.

@tuturto, can you show concrete examples of the good and bad error messages produced by this version? Perhaps we can take another approach.

tuturto:
Example of a good error message:

> => (let)
>   File "<input>", line 1, column 1
> 
>   (let)
>   ^---^
> HyMacroExpansionError: b"expanding `let': missing 1 required positional argument: 'variables'"

Example of a bad error message:

> => (defmacro bad [bad-param!] "hello world")
> => (bad)
>   File "<input>", line 1, column 1
> 
>   (bad)
>   ^---^
> HyMacroExpansionError: b"expanding `bad': missing 1 required positional argument: 'bad_param_bang'"

Example of very bad error message:

> => (defmacro very-bad [bad-!-very-bad] "hello world")
> => (very-bad)
>   File "<input>", line 1, column 1
> 
>   (very-bad)
>   ^--------^
> HyMacroExpansionError: b'expanding `very_bad\': TypeError("_hy_anon_fn_1() missing 1 required positional argument: \'bad_!_very_bad\'",)'

So as soon as we get to the terroritory of invalid parameter names, this code starts to break apart. Mangling helps a bit, but it doesn't cover all the cases.

gilch:
The "bad" case doesn't seem that bad. The fact that mangling happens cannot be a secret from Hy users, or they'll be confused by more than this. When exactly mangling occurs ought to be documented better though.

The "very-bad" case has a `TypeError` instead. That's not good, but it still points the user to a missing positional argument during macro expansion. I don't think it's bad enough to hold up this PR.

We already handle UTF-8 entities with punycode. Maybe we could improve the mangling to handle all cases. Perhaps we should encode _all_ invalid Python identifiers into valid Python identifiers. Even though the AST _mostly_ doesn't care as long as it's valid ASCII, this might prevent other problems like this in the future.

Punycode was a convenient choice, since the encoder is in Python's standard libarary. There are, however, other ASCII characters that are not allowed in Python identifiers. Punycoded strings are not especially readable either, since it's not clear at a glance where in the string the special characters are located. We might want to come up with our own encoding scheme.

Kodiologist:
> I'm not sure I can be the one to merge this since many of the commits are mine.

You have my blessing to do that, for what that's worth. Let's keep things moving on the road to 0.12.

kirbyfan64:
Eh, everybody was taking too long.

PR #1177, opened 2016-12-17
no-readline on OS X fix

eblume:
In the case that the user is running OS X / Darwin and does not have `readline` installed in their virtual environment, currently the completer barfs due to the failed import. **Without knowing any knock-on consequences**, I simply bypass the offending statement in the case that the `readline` import failed. This needs reviewing as there is a high risk of unintended consequences in other environments.

I also took the liberty of adding a quickstart to CONTRIBUTING. Let me know if you'd like me to pull that out. I added it as I could not find anything like it in the existing docs, and it seemed like a natural place for it.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -5,7 +5,7 @@ In order to begin contributing code to the hylang project, ensure that you
 have completed the following steps to set up your development environment.
 
 - Clone the project to a local git repository. Let's assume you've got a
-   handle on that.
+  handle on that.
 
 - Create a virtual environment that will contain the python dependencies
   hylang uses. Starting with Python 3.3, the correct way to do this is::
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,3 +1,42 @@
+Development Quickstart
+======================
+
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.rst
```diff
@@ -1,3 +1,42 @@
+Development Quickstart
```
kirbyfan64: We already have something like this [here](http://docs.hylang.org/en/latest/hacking.html). Anything here should probably go over in `docs/hacking.rst`.

Kodiologist:
I believe `readline` should always be installed. It's a standard module. Was your environment perhaps corrupted somehow?

eblume:
I can confirm that something is broken with my `readlines`, so I'll be closing this. I wonder though if the import check should be restructured so as to not swallow an ImportError, which is the current case.

eblume:
In case anyone finds this in the future - in OS X, if you use `pyenv` to install python, you must first install GNU Readline for pyenv to build python with `readline` enabled. I accomplished this with a simple `brew install readline` followed by uninstalling and re-installing my preferred version of Python in pyenv. (In truth I had to jump through a few small hoops to get readline installed correctly - I did something naughty at some point, not sure what.)

Issue #1176, opened 2016-12-16
Add some functional utilities


tuturto:
- compose, used to compose two functions to one
- constant, used to define function returning constant value
- flip, used to define new function with parameter order flipped
- complement (working on understanding this still)

Originating from email exchange with Philip Xu and their excellet Hymn project (https://github.com/pyx/hymn/blob/cacac3cd3b1ddef8879e8488056a9310cd0ab13d/hymn/utils.hy#L64)

gilch:
The threading macros take care of most of these cases I think. Though macros can be hard to use with HOF, I think they'd be more efficient.

I didn't see the complement function in the link you provided. Is it in the same file?

Can we call it `comp` to match [Clojure's naming](https://github.com/clojure/clojure/blob/010864f8ed828f8d261807b7345f1a539c5b20df/src/clj/clojure/core.clj#L2448)?

tuturto:
complement isn't in the file, that's something @pyx mentioned in email. I asked about the details, but I think it's the same as Clojure's complement.

Following Clojure's naming convention sounds reasonable to me.

pyx:
Thank you @tuturto, the complement function is the same as Clojure's, except we don't need to write it as multi-motheds :)

@gilch is right, those functions will be convenient to use with HOFs while in other cases, threading macros are probably more suitable. 

Please note that the constant function in Hymn generates a customized docstring on the fly from the value, but we usually do not want that happen in more general cases (e.g in stdlib), as there is no guarantee that ```.__str__``` / ```.__repr__``` / ```.__unicode__``` of the value will be pure functions.

Last but not least, let me take this opportunity to say thank you to all of you.  Hy is great, and I still remember the jaw-dropping moments when on that lighting talk, Paul showed us importing hy modules in python and stepping through hy in gdb...

pyx:
FYI, I found out the anaphoric macro version is called ap-compose.

gilch:
How is `filp` supposed to work in Hy? Clojure doesn't have it. I didn't see an example in Hymn either. I think in Haskell's case, flip swaps the order of the first two arguments to a function. But in Haskell, all functions are technically of one argument and it has auto-currying, while in Hy, even normally binary functions like `-` can have multiple arguments. I don't think it gets you as far in Hy. The usual way to handle this in Clojure is with `#()`, I think. (Or `xi` in Hy). But that's macros again.

tuturto:
[...]

Issue #1103, opened 2016-09-18
Add more operators to shadow


gilch:
We've currently only got `'#{+ - * / < <= = != >= >}` shadowed. Python has more operators than this, and so does Hy. I don't think there's any reason not to shadow these other binary operators: `'#{** @ // % << >> & ^ |}`.

The unary operator `'~` should also be easy to shadow, but the other unary operators `'#{+ -}` present some ambiguity. I propose importing the `operator.neg` and `operator.pos` functions for these, (as `neg` and `pos`).

There are a number of other operators in the `operator` module. The distinction between operator and function that Python has is unnatural in a Lisp like Hy. We should automatically import the remaining operators, perhaps with some renaming, as was done for `itertools`. We can treat these as a macro or special form when in the function position, but as a function name otherwise.

gilch:
On second thought, if we fix #1109, then the shadowed `'#{+ -}` operators should just work as unary when applied to a single argument. There's no need to import `neg` and `pos`.

Kodiologist:
It looks like we also need to shadow some alphabetic operators, like `not`:

    => (setv f not) (print (f 1))
    NameError: name 'not' is not defined

But not all of them. There's no sense in shadowing `and` because of its short-circuiting magic, which can't be captured by a mere function.

PR #1179, opened 2016-12-19
Add comp, constantly and complement

tuturto:
relates #1176

### DIFF
--- a/docs/language/core.rst
+++ b/docs/language/core.rst
@@ -61,10 +61,9 @@ comp
 
 Usage: ``(comp f g)``
 
-Compose zero or more functions to a new function. The new function will chain
-given functions together: ``((comp g f) x)`` is thus equivalent to
-``(g (f x))``. If ``comp`` is called without parameters, ``identity`` will be
-returned.
+Compose zero or more functions into a new function. The new function will
+chain the given functions together, so ``((comp g f) x)`` is equivalent to
+``(g (f x))``. Called without arguments, ``comp`` returns ``identity``.
 
 .. code-block:: hy
 
@@ -86,8 +85,8 @@ complement
 
 Usage: ``(complement f)``
 
-Returns a new function that inverts truth value of supplied function ``f``
-when called.
+Returns a new function that returns the same thing as ``f``, but logically
+inverted. So, ``((complement f) x)`` is equivalent to ``(not (f x))``.
 
 .. code-block:: hy
 
@@ -153,7 +152,7 @@ constantly
 Usage ``(constantly 42)``
 
 Create a new function that always returns the given value, regardless of
-parameters given to it.
+the arguments given to it.
 
 .. code-block:: hy
 
--- a/docs/language/core.rst
+++ b/docs/language/core.rst
@@ -54,6 +54,47 @@ Returns ``True`` if *x* is iterable and not a string.
    False
 
 
+.. _comp:
+
+comp
+----
+
+Usage: ``(comp f g)``
+
+Compose one or more functions to a new function. The new function will chain
+given functions together: ``((comp f g) foo)`` is thus equivalent to
+``(g (f foo))``.
+
+.. code-block:: hy
+
+   => (def example (comp + str))
+   => (example 1 2 3)
+   "6"
+
+
+.. _complement:
+
+complement
+----------
+
+.. versionadded:: 0.12.0
+
+Usage: ``(complement f)``
+
+Returns a new function that inverts truth value of supplied function ``f``
+when called.
+
+.. code-block:: hy
+
+   => (def inverse (complement identity))
+   => (inverse True)
+   False
+   => (inverse 1)
+   False
+   => (inverse False)
+   True
+
+
 cons
 ----
 
@@ -95,6 +136,30 @@ Checks whether *foo* is a :ref:`cons cell <hycons>`.
 
    => (cons? [1 2 3])
    False
+
+
+.. _constantly:
+
+constantly
+----------
+
+.. versionadded:: 0.12.0
+
+Usage ``(constantly 42)``
+
+Create a new function that always returns the given value, regardless of
+parameters given to it.
+
+.. code-block:: hy
+
+   => (def answer (constantly 42))
+   => (answer)
+   42
+   => (answer 1 2 3)
+   42
+   => (answer 1 :foo 2)
+   42
+
 
 .. _dec-fn:
 
--- a/docs/language/core.rst
+++ b/docs/language/core.rst
@@ -62,12 +62,12 @@ comp
 Usage: ``(comp f g)``
 
 Compose one or more functions to a new function. The new function will chain
-given functions together: ``((comp f g) foo)`` is thus equivalent to
-``(g (f foo))``.
+given functions together: ``((comp g f) x)`` is thus equivalent to
+``(g (f x))``.
 
 .. code-block:: hy
 
-   => (def example (comp + str))
+   => (def example (comp str +
    => (example 1 2 3)
    "6"
 
--- a/docs/language/core.rst
+++ b/docs/language/core.rst
@@ -61,15 +61,20 @@ comp
 
 Usage: ``(comp f g)``
 
-Compose one or more functions to a new function. The new function will chain
+Compose zero or more functions to a new function. The new function will chain
 given functions together: ``((comp g f) x)`` is thus equivalent to
-``(g (f x))``.
+``(g (f x))``. If ``comp`` is called without parameters, ``identity`` will be
+returned.
 
 .. code-block:: hy
 
-   => (def example (comp str +
+   => (def example (comp str +))
    => (example 1 2 3)
    "6"
+
+   => (def simple (comp))
+   => (simple "hello")
+   "hello"
 
 
 .. _complement:
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -54,7 +54,7 @@
     (fn [&rest args &kwargs kwargs]
       (f (apply g args kwargs))))
   (if fs
-    (reduce compose-2 (list (rest fs)) (first fs))
+    (reduce compose-2 fs)
     identity))
 
 (defn complement [f]
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -47,12 +47,13 @@
   "Checks whether item is a collection"
   (and (iterable? coll) (not (string? coll))))
 
-(defn comp [f &rest g]
-  "Compose one or more functions together"
-  (fn [&rest args &kwargs kwargs]
-    (reduce (fn [previous-result func]
-              (apply func [previous-result]))
-            g (apply f args kwargs))))
+(defn comp [f &rest fs]
+  "function composition"
+  (defn compose-2 [f g]
+    "compose 2 functions"
+    (fn [&rest args &kwargs kwargs]
+      (f (apply g args kwargs))))
+  (reduce compose-2 fs f))
 
 (defn complement [f]
   "Create a function that reverses truth value of another function"
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -47,6 +47,18 @@
   "Checks whether item is a collection"
   (and (iterable? coll) (not (string? coll))))
 
+(defn comp [f &rest g]
+  "Compose one or more functions together"
+  (fn [&rest args &kwargs kwargs]
+    (reduce (fn [previous-result func]
+              (apply func [previous-result]))
+            g (apply f args kwargs))))
+
+(defn complement [f]
+  "Create a function that reverses truth value of another function"
+  (fn [&rest args &kwargs kwargs]
+    (not (apply f args kwargs))))
+
 (defn cons [a b]
   "Return a fresh cons cell with car = a and cdr = b"
   (HyCons a b))
@@ -54,6 +66,11 @@
 (defn cons? [c]
   "Check whether c can be used as a cons object"
   (instance? HyCons c))
+
+(defn constantly [value]
+  "Create a function that always returns the same value"
+  (fn [&rest args &kwargs kwargs]
+    value))
 
 (defn keyword? [k]
   "Check whether k is a keyword"
@@ -456,11 +473,12 @@
 
 (def *exports*
   '[*map accumulate butlast calling-module-name chain coll? combinations
-    compress cons cons? count cycle dec distinct disassemble drop drop-last
-    drop-while empty? even? every? first filter flatten float? fraction gensym
-    group-by identity inc input instance? integer integer? integer-char?
-    interleave interpose islice iterable? iterate iterator? keyword keyword?
-    last list* macroexpand macroexpand-1 map merge-with multicombinations name
-    neg? none? nth numeric? odd? partition permutations pos? product range
-    read read-str remove repeat repeatedly rest reduce second some string
-    string? symbol? take take-nth take-while xor tee zero? zip zip-longest])
+    comp complement compress cons cons? constantly count cycle dec distinct
+    disassemble drop drop-last drop-while empty? even? every? first filter
+    flatten float? fraction gensym group-by identity inc input instance?
+    integer integer? integer-char? interleave interpose islice iterable? 
+    iterate iterator? keyword keyword? last list* macroexpand macroexpand-1 
+    map merge-with multicombinations name neg? none? nth numeric? odd? 
+    partition permutations pos? product range read read-str remove repeat 
+    repeatedly rest reduce second some string string? symbol? take take-nth 
+    take-while xor tee zero? zip zip-longest])
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -48,14 +48,17 @@
   (and (iterable? coll) (not (string? coll))))
 
 (defn comp [&rest fs]
-  "function composition"
-  (defn compose-2 [f g]
-    "compose 2 functions"
-    (fn [&rest args &kwargs kwargs]
-      (f (apply g args kwargs))))
-  (if fs
-    (reduce compose-2 fs)
-    identity))
+  "Function composition"
+  (if (not fs) identity
+      (= 1 (len fs)) (first fs)
+      (do (setv rfs (reversed fs)
+                first-f (next rfs)
+                fs (tuple rfs))
+          (fn [&rest args &kwargs kwargs]
+            (setv res (apply first-f args kwargs))
+            (for* [f fs]
+              (setv res (f res)))
+            res))))
 
 (defn complement [f]
   "Create a function that reverses truth value of another function"
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -47,13 +47,15 @@
   "Checks whether item is a collection"
   (and (iterable? coll) (not (string? coll))))
 
-(defn comp [f &rest fs]
+(defn comp [&rest fs]
   "function composition"
   (defn compose-2 [f g]
     "compose 2 functions"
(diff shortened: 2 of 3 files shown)


gilch:
It looks like `comp` is applying functions in the opposite order of Clojure's. It should work the same as Clojure's `comp`.

Kodiologist:
I agree that Clojure's order is the right one, since in mathematics, $(g ∘ f)(x)$ usually means $g(f(x))$.

pyx:
Agree with gilch and Kodiologist, ((comp f  g) x) means (f (g x)).

Regarding tests, compose does obey three laws, provided function ```f```, ```g``` and ```h``` are pure functions, ```id``` is identity function and ```.```(dot) is compose function:
1. left identity
(id . f) x = f x
2. right identity
(f . id) x = f x
3. associative
((f . g) . h) x = (f . (g . h)) x

This is [how I test compose](https://github.com/pyx/hymn/blob/ca18fab4bc6b5a895f3685f5ffad24c9b3314b55/tests/test_utils.hy#L26,L60) in hymn, even thought it looks like overkill for such a simple function. :)

tuturto:
heh, that's rather embarrassing. I'll fix that. Good that it was spotted so soon.

pyx:
@tuturto, sure, my pleasure, if you find that useful, please do whatever you want, including re-licensing, I didn't put a [no-evil clause](https://news.ycombinator.com/item?id=3693108) on it. :D

pyx:
@tuturto 
In case you were talking about not only the tests but also the function itself, please notice that there is a design decision I made that might not be suitable for standard library: I did not catch the no-argument case, since the function was meant to be used only internally, an exception from ```reduce``` is enough for me.

I think there are two sensible options:
1. ```comp``` can be called with any number of arguments, return ```identity``` function when no argument provided.
  In this case, two ```if```s (or one, with 0.12 new ```if``` syntax) before ```reduce``` will do, handling no/one argument cases.
 
2. ```comp``` should be called with at least one arguments, something like
```hy
(defn compose [f &rest fs]
  "function composition"
  (defn compose-2 [f g]
    "compose 2 functions"
      (f (apply g args kwargs))))
  (reduce compose-2 fs f))
```
should do

tuturto:
Thank you @pyx. I added the atleast-one-argument version and fixed docs while I was at it.

gilch:
We should really follow [Clojure's lead](https://github.com/clojure/clojure/blob/7aad2f7dbb3a66019e5cef3726d52d721e9c60df/src/clj/clojure/core.clj#L2530) and choose option 1: return identity.

tuturto:
I find it bit counter-intuitive that ``comp`` without any parameters returns ``identity``, but it doesn't really make sense to deviate from other lisps. I added that there too now.

Kodiologist:
The idea is that the identity function is the identity element of the group of functions under the composition operator. That is, `(comp f identity)` is equal to `f`. That's why it's the right choice for the nullary case.

tuturto:
oh, hm.. If you put it that way it makes sense.

pyx:
May I suggest a minor improvement,
since we have ```if``` taking out the empty case, and ```reduce``` can handle sequence with a single element by returning that sole element, which is what we expect in this case. ```(comp f) == f```
```hy
(reduce compose-2 (list (rest fs)) (first fs))
```
can be
```hy
(reduce compose-2 fs)
```

gilch:
Function calls are kind of expensive in Python I think it would be more efficient to use the imperative style, rather than calling a new function for every pair. Maybe something like this:
```Python
def comp(*fs):
    if not fs:
        return identity
    fs = reversed(fs)
    def composed(*args,**kwargs):
        for f in fs:
            res = f(*args,**kwargs)
            for f in fs:
                res = f(res)
        return res
    return composed
```

Kodiologist:
@gilch I think a benchmark or example showing an improvement should be a minimum requirement for a performance-motivated change.

gilch:
@Kodiologist, that's fair, but I'm starting to think Hy needs a timing macro. The `timeit` module is proving difficult to use in Hy.

Here's a Hy version:
```Hy
=> (defn comp [&rest fs]
... (if (not fs) identity
...     (= 1 (len fs)) (first fs)
...     (do (setv fs (reversed fs))
...         (fn [&rest args &kwargs kwargs]
...           (for* [f fs] (setv res (apply f args kwargs))
...             (for* [f fs] (setv res (f res))))
...           res))))
```

gilch:
```Hy
(defn comp [&rest fs]
  (if (not fs) identity
      (= 1 (len fs)) (first fs)
      (do (setv fs (reversed fs))
            (for* [f fs] (setv res (apply f args kwargs))
              (for* [f fs] (setv res (f res)))) res))))

(import [time [time]])
(setv start (time))

(for [_ (range 100000)]
  ((comp inc inc inc inc) 1))

(print (- (time) start))
(print "iterative comp completed.")


(defn comp [&rest fs]
   "function composition"
   (defn compose-2 [f g]
     "compose 2 functions"
       (f (apply g args kwargs))))
   (if fs
     (reduce compose-2 fs)
     identity))

(setv start (time))

(for [_ (range 100000)]
  ((comp inc inc inc inc) 1))
(print (- (time) start))

(print "functional comp completed.")
```
```
0.6814742088317871
iterative comp completed.
1.6196417808532715
functional comp completed.
```

pyx:
@gilch 
IMHO, I do not understand the nested for loop, it works because in Hy, ```reversed```, like in python3, returns a generator, so the first iteration on the outer loop fetch the first function, then all the work done inside inner loop, then the generator is exhausted, yielding the expected result (each function been called once and only once).

But, then this composed function can only be called once, because the function list (```fs```) is defined in outer scope of the function, try:

```hy
hy 0.11.0+353.gca6fd66 using CPython(default) 3.5.2 on Linux
=> (defn comp [&rest fs]
...   (if (not fs) identity
...       (= 1 (len fs)) (first fs)
...       (do (setv fs (reversed fs))
...           (fn [&rest args &kwargs kwargs]
...             (for* [f fs] (setv res (apply f args kwargs))
...               (for* [f fs] (setv res (f res)))) res))))
=> (def inc4 (comp inc inc inc inc))
=> (inc4 1)
5
=> (inc4 1)
  File "<input>", line 263, in _hy_anon_fn_1
UnboundLocalError: local variable 'res' referenced before assignment
=> 
```

Here is my take:
```hy
(defn comp [&rest fs]
  (if (not fs) identity
      (= 1 (len fs)) (first fs)
      (do (setv rfs (reversed fs)
                first-f (next rfs)
                fs (tuple rfs))
            (setv res (apply first-f args kwargs))
            (for* [f fs]
              (setv res (f res)))
            res))))
```
Notice:
1. I used ```next``` instead of ```first```, to make explicit the intent that we want to advance the generator once.
2. I used ```tuple``` call to make the rest of sequence re-usable, I guess ```tuple``` should have small footprint and better performance.
3. I named the first function ```first-f```, a little bit ugly, but we need a different name or we will have the (in)famous ```UnboundLocalError``` exception, no ```nonlocal``` here, because we still support python2. Oh, lovely python scoping ... :)

Kodiologist:
Yeah, the nested `for` loop is needlessly weird. I'd write your version, @pyx, without the extra variables, like this:

    (defn comp [&rest fs]
      (if (not fs) identity
          (= 1 (len fs)) (first fs)
            (setv res (apply (get fs -1) args kwargs))
            (for [f (reversed fs)]
              (setv res (f res)))
            res)))

This has similar performance to @gilch's version.

pyx:
@Kodiologist 
Generating reversed list each time on-the-fly (inside the inner function) is fine, as we usually do not feed too much into ```comp``, so the performance lost should be exceptable, but this version will call the last (of the argument list) function twice:
```hy
hy 0.11.0+353.gca6fd66 using CPython(default) 3.5.2 on Linux
=> (defn comp [&rest fs]
...   (if (not fs) identity
...       (= 1 (len fs)) (first fs)
...       (fn [&rest args &kwargs kwargs]
...         (setv res (apply (get fs -1) args kwargs))
...         (for [f (reversed fs)]
...           (setv res (f res)))
...         res)))
=> (def inc4 (comp inc inc inc (fn [n] (print "hit!") (inc n))))
=> (inc4 1)
hit!
hit!
6
```

Kodiologist:
Oh, right, whoops, I meant `(cut fs -2 None -1)`, not `(reversed fs)`.

pyx:
I think the most common case of compose, might be that the composed fuction will be re-used a lot, think ```map```, so on a second thought, we might better prepared the function list before entering into the scope of inner (composed) function.

And I wrote a naive time-it macro as mentioned by @gilch, I will open a new issue.

tuturto:
Nice work and quite big difference in the speed already :heart:

pyx:
Now that we have a proper (I hope so) timeit macro https://github.com/hylang/hy/issues/1181
And we have three implementations:
```hy
(defn comp-f [&rest fs]
   "function composition"
   (defn compose-2 [f g]
     "compose 2 functions"
       (f (apply g args kwargs))))
   (if fs
     (reduce compose-2 fs)
     identity))

(defn comp-i1 [&rest fs]
  (if (not fs) identity
      (= 1 (len fs)) (first fs)
        (setv res (apply (get fs -1) args kwargs))
        (for [f (cut fs -2 None -1)]
          (setv res (f res)))
        res)))

(defn comp-i2 [&rest fs]
  (if (not fs) identity
      (= 1 (len fs)) (first fs)
      (do (setv rfs (reversed fs)
                first-f (next rfs)
                fs (tuple rfs))
            (setv res (apply first-f args kwargs))
            (for* [f fs]
              (setv res (f res)))
            res))))
```
It's time to let the data speak:
```hy
=> (defn comp-f [&rest fs]
...    "function composition"
...    (defn compose-2 [f g]
...      "compose 2 functions"
...      (fn [&rest args &kwargs kwargs]
...        (f (apply g args kwargs))))
...    (if fs
...      (reduce compose-2 fs)
...      identity))

=> (defn comp-i1 [&rest fs]
...   (if (not fs) identity
...       (= 1 (len fs)) (first fs)
...       (fn [&rest args &kwargs kwargs]
...         (setv res (apply (get fs -1) args kwargs))
...         (for [f (cut fs -2 None -1)]
...           (setv res (f res)))
...         res)))

=> (defn comp-i2 [&rest fs]
...   (if (not fs) identity
...       (= 1 (len fs)) (first fs)
...       (do (setv rfs (reversed fs)
...                 first-f (next rfs)
...                 fs (tuple rfs))
...           (fn [&rest args &kwargs kwargs]
...             (setv res (apply first-f args kwargs))
...             (for* [f fs]
...               (setv res (f res)))
...             res))))

=> (print ((comp-f inc inc inc inc) 1))
5

=> (print ((comp-i1 inc inc inc inc) 1))
5

=> (print ((comp-i2 inc inc inc inc) 1))
5

=> (time-it ((comp-f inc inc inc inc) 1))
11.021359261001635

=> (time-it ((comp-i1 inc inc inc inc) 1))
8.513512686997274

=> (time-it ((comp-i2 inc inc inc inc) 1))
12.34987382899999

=> (time-it (inc4 1) (setv inc4 (comp-f inc inc inc inc)))
4.459002538002096

=> (time-it (inc4 1) (setv inc4 (comp-i1 inc inc inc inc)))
4.978545720001421

=> (time-it (inc4 1) (setv inc4 (comp-i2 inc inc inc inc)))
3.3503086150012678
```

Calling ```cut``` is actually slower than recurring 4 times. If we try 10 (on my machine), then it starts to gain:
```hy
=> (time-it (inc10 1) (setv inc10 (comp-f inc inc inc inc inc inc inc inc inc inc)))
11.87790252600098
=> (time-it (inc10 1) (setv inc10 (comp-i1 inc inc inc inc inc inc inc inc inc inc)))
7.043803205000586
=> (time-it (inc10 1) (setv inc10 (comp-i2 inc inc inc inc inc inc inc inc inc inc)))
5.294795802998124
```

pyx:
Now that I think ```flip``` might not be a good idea, but may I propose another two handy functions?

pipe - the counterpart of comp
```hy
(defn pipe [&rest fs]
  "docstring goes here"
  (apply comp (reversed fs)))
```

in used
```hy
=> (def inc-str (pipe int inc str))
=> (inc-str "41")
'42'
```

juxt - I saw this one when I read about the "see also" part on clojure-docs, did not read the implementation though

```hy
(defn juxt [f &rest fs]
  "docstring goes here"
  (setv fs (cons f fs))
    (setv res [])
    (for* [f fs]
      (.append res (apply f args kwargs)))
    res))
```

I can imagine ```juxt``` will be super handy in data processing
```hy
=> ((juxt min max sum) (range 1 101))
[1, 100, 5050]

=> (setv [total difference product quotient] ((juxt + - * /) 24 3))
[27, 21, 72, 8.0]

=> (import [operator [itemgetter]])
=> ((juxt (itemgetter 'name) (itemgetter 'age)) {'name 'Joe 'age 42 'height 182})
['Joe', 42]
```

Kodiologist:
@pyx, this PR is for only the three functions named in the title. Those should be in their own PR if you want them.

gilch:
@pyx you could make a new issue first for those.

From the timing data we see that the `cut` version doesn't perform as well. It can be constructed faster, because it defers some computation to execution. This is not a good trade-off in the common case of creating a composede function once, and then calling it many times in a `map` over a long iterable. It's better to precompute it once. So far, I like @pyx's implementation of `comp` best.

You could use `cut` instead of `reversed` when precomputing. I'm not sure which is faster. We could also test for more lengths in the `if`, and hard code the short cases while avoiding the loop altogether. I'm not sure it would help much, but the overhead of a loop and assignment is probably more significant when there are fewer functions.

pyx:
@Kodiologist, fair enough, but usually ```pipe``` is defined in terms of ```comp``` to avoid duplicate code, so that PR will be depending on this one, in this case, it might be less management effort to go with ```comp```.

```juxt```, I can draft a separate PR.

**Comment by pyx** [comment]
@gilch Since I am new here, sometimes I am don't know which way should I go, e.g, I am not sure if creating a new issue regarding yet-to-be merged PR is a proper way of giving feedback.

That's way I did not draft a PR myself in the first place, I know it should come with proper documentation and test cases, to which the requirements I am not quite sure about.  I will try, for ```juxt```, anyway.

**Comment by pyx** [comment]
I have ```juxt``` implemented with documentation and test, should I create a PR now, or wait until this PR get merged first then rebase, cause it will create merge conflict[1] right now, both PRs touch the ```*exports*``` part.

[1] That's why I proposed ```pipe``` and ```juxt``` into this PR instead of a separate one.

**Comment by Kodiologist** [comment]
@pyx Wait until this one is merged.

**Comment by Kodiologist** [comment]
@gilch I sort of regret even bringing up the subject of benchmarks because any discussion about performance is speculation outside of profiling a real program. Anyway, in my opinion, any of the three versions is fine for now. Do you approve this PR?

**Comment by pyx** [comment]
@Kodiologist , okay, I will add ```pipe``` into the mix (just ```juxt``` right now) and send a PR later. Thanks.

**Comment by gilch** [comment]
> @gilch Since I am new here, sometimes I don't know which way should I go, e.g, I am not sure if creating a new issue regarding yet-to-be merged PR is a proper way of giving feedback.

We don't know either. Don't worry too much about it, you're doing fine.

I would approve adding `juxt` to Hy. Clojure has it and I use it. I'd still suggest a new PR. I feel `pipe` needs more discussion. That's why I don't want it added to this PR. You may open a new issue for discussing that, but I'm not inclined to approve it since Clojure doesn't have it. It mostly duplicates the behavior of the threading macros (`->`,`->>`,`as->`). And `comp`, but in reverse. It is a good point that `pipe` can be implemented in terms of `comp`, but we don't have to do it that way, and it might be more efficient if we don't. It might actually be easier to implement `pipe` first and then `comp` in terms of `pipe`.

Merge conflicts in this file could probably be avoided if the exports list had only one element per line. The merge conflicts in this case are trivial to resolve though.

>@gilch I sort of regret even bringing up the subject of benchmarks because any discussion about performance is speculation outside of profiling a real program. Premature optimization is the root of all evil. Anyway, in my opinion, any of the three versions is fine for now. Do you approve this PR?

I normally agree with that maxim, but I feel you're using it out of context. When developing an application, it's unlikely that any given function will be on the critical path, so it's best to profile and see what's actually the bottleneck. Optimizing other parts is just a waste of time and the optimized code can be harder to maintain too.

However, when developing the standard library, we're not just making one application, but the building blocks of _every_ application written in Hy. Therefore it's much more likely the functions we write will be on the critical path of some Hy applications, _especially_ for something as fundamental as function composition! So I do feel it's worth some effort not to be too inefficient here. That said proving optimality is probably an undecidable problem in the general case, so we do have to give up eventually, but you do hit diminishing returns pretty quickly in practice for a function this short.

I `constantly` and `compliment` are pretty trivial, and appear to be correct. I'll approve those so far. I'll also approve `comp` if it's changed to @pyx's version.

**Comment by kirbyfan64** [comment]
I'm not really seeing a use case for `compliment`... what's the difference between that and `(comp not f)`?

**Comment by Kodiologist** [comment]
@kirbyfan64 Good thinking; `comp not` is shorter than `complement`, anyway.

**Comment by pyx** [comment]
@gilch this is my thought on ```pipe```
When composing with, let's say, more than a handful functions, ```pipe``` is easier to work with, case study:
Assuming we are building a file processing pipeline, and have these functions:
```read-file write-file parse-file rename transform-phase-1 transform-phase-2 transform-phase-3 , etc.```
They are all of type 
```haskell
file -> file
```
We can build the pipeline like this:
```hy
(def process
  (pipe
    read-file
    parse-file
    transform-phase-1
    transform-phase-2
    transform-phase-3
    rename
    write-file))
```
It's just more natural to read in this case than ```comp```.
If I am forced to have only one, I will go with ```pipe```.  But, there is only ```compose``` predefined (have a name) in maths, so many function programming languages only define ```compose``` in their standard libraries, it's not because ```pipe``` is less useful, to reduce the surprise factor, so ```compose``` should stay...

If we can have both ```comp``` and ```pipe```, I totally agree that define ```pipe``` imperatively, then define ```comp``` in terms of ```pipe``` looks better, especially in (I suppose) common scenarios like above, ```pipe``` gets used more.  I might actually go as far as define both imperatively, cause these two functions' behaviours are so well known, and the logic is so trivial, that we do not expect them to be changed in the future, that means almost no maintenance burden.

pyx:
@kirbyfan64  @gilch , I totally agree technically ```(compliment f)``` is just ```(comp not f)```, and in other static compiled languages, the compiler may even inline the function, but iirc, in Hy, ```not``` is not a normal function (pun not intended) but a special form, so, cannot e composed:
```hy
hy 0.11.0+357.g29b3702 using CPython(default) 3.5.2 on Linux
=> (def my-even? (comp not odd?))
NameError: name 'not' is not defined
=> not
NameError: name 'not' is not defined
=> (not True)
False
```
If I remember correctly, you have an issue or milestone regarding this and ```in```, etc.  But it does not work as of today.

BTW, happy holidays, everyone.

Kodiologist:
> If I remember correctly, you have an issue or milestone regarding this and `in`, etc. But it does not work as of today.

Yes, #1103. I'm guessing it will be pretty straightforward.

pyx:
PS. my ```juxt``` branch is here https://github.com/hylang/hy/compare/master...pyx:juxt , just waiting for this PR, then rebase, also, if you all agree to add ```pipe```, I can add that too.

kirbyfan64:
> `not` is not a normal function (pun not intended)

I wonder...`not` should probably be shadowed just like operators like `+`< , right?

tuturto:
I updated ``comp`` to @pyx version. Happy holidays people ❤️

PR #1189, opened 2016-12-26
Clean up hy.contrib

### LINKED ISSUE #1162: Remove eg/ and hy/contrib/, or make guidelines for what belongs there
eg/ and hy/contrib/ have a lot of miscellaneous stuff that isn't part of the core language. (I am, admittedly, part of the problem, having approved several pull requests that add to the latter.) So I'm inclined to think they shouldn't be in Hy's core repository. They represent a maintenance burden for the core development team, when they ought to live freely on their own elsewhere and get maintained by their individual authors, like ordinary Python packages. Python's package system is lightweight enough that it's not unreasonable to have a Python package with only 20 lines of content code.
 
If we think one of these directories or both should stay, we should have some guidelines as to what sort of code belongs and what doesn't. (CONTRIBUTING.rst is probably the right place to put such guidelines.) Like, right now, I have a lot of useful bits of code in my generic utility library [Kodhy](https://github.com/Kodiologist/Kodhy), like macros for indexing NumPy and pandas data structures, and I don't know which if any of these would be at home in contrib.

Some of this code might be of interest to Hy core development only for its tests. It can be helpful to have nontrivial programs in a programming language's test suite in addition to unit tests. If we take that philosophy towards any of the eg/ or hy/contrib/ code, it should perhaps be moved to test/.

Kodiologist:
Closes #1162.


kirbyfan64:
As a side note: what if we *leave* Botsbuildbots, but just remove it from the tests? Seems kind of silly that that's tested, but it's still a fun little easter egg. I personally kind of see it as akin to Python's `import antigravity`.

Kodiologist:
@kirbyfan64 Ugh, fine.

paultag:
:+1: happy to see the cleanup!

paultag:
(FWIW; I'm OK with removing `botsbuildbots`, but it was a nice fun easter egg for a while :) )

PR #1194, opened 2016-12-30
Use `and` in hy.core.shadow.comp_op

### LINKED ISSUE #1191: Inconsistent behaviour between operator and shadowed version
Certain operators have behavior inconsistent with their shadowed counterparts. For instance, see the example below using a Sympy symbol (which overloads `__lt__` and `__gt__`).

```
hy 0.11.0+367.gbc4d678 using CPython(default) 3.5.2 on Linux
=> (import sympy)
import sympy
=> sympy.__version__
sympy.__version__
'1.0'
=> (setv x (sympy.Symbol "x"))
x = sympy.Symbol('x')
=> (< 0 x 1)
(0 < x < 1)
  File "/home/simon/.virtualenvs/hy-git/lib/python3.5/site-packages/sympy/core/relational.py", line 195, in __nonzero__
    raise TypeError("cannot determine truth value of Relational")
TypeError: cannot determine truth value of Relational
=> (setv lt <)
from hy.core.shadow import <
lt = <
=> (lt 0 x 1)
lt(0, x, 1)
And(x < 1, x > 0)
```

The reason for this is that Python expands `(0 < x < 1)` to `(0 < x) and (x < 1)`; the Sympy `Relational` class, an instance of which is created by `(0 < x)`, overrides `__bool__` and raises an error therein. On the other hand, `hy.core.shadow.comp_op` uses `operator.and_` which is `&`; as a result `(lt 0 x 1)` in the snippet above expands to something more like `(0 < x) & (x < 1)`. The behavior of `&` can be modified by overriding `__and__`, and the Sympy `Relational` class does this.

In my opinion the latter behavior (invoking the overridable `__and__`) is desirable in this case, though it may not be so in all cases. One way to resolve this would, I suppose, be to use the shadowed versions of the operators everywhere, rather than sometimes using the builtin versions.

*edited to correct a mistake*

swhalen:
Makes shadow operator behavior match normal (Python) operator behavior
when said operators are used in chained comparisons.

Unlike Python's `and`, the usage of `and` here is not short-circuiting.

Resolves #1191

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,10 @@ Changes from 0.12.1
    [ Language Changes ]
    * xor: If exactly one argument is true, return it
 
+   [ Bug Fixes ]
+   * Shadowed comparison operators now use `and` instead of `&`
+     for chained comparisons
+
 Changes from 0.12.0
 
    [ Bug Fixes ]
--- a/hy/core/shadow.hy
+++ b/hy/core/shadow.hy
@@ -65,7 +65,7 @@
   "Helper for shadow comparison operators"
   (if (< (len args) 2)
     (raise (TypeError "Need at least 2 arguments to compare"))
-    (reduce operator.and_
+    (reduce (lambda [x y] (and x y))
             (list-comp (op x y)
                        [(, x y) (zip args (cut args 1))]))))
 (defn < [&rest args]
--- a/tests/native_tests/shadow.hy
+++ b/tests/native_tests/shadow.hy
@@ -50,6 +50,7 @@
 
 (defn test-shadow-compare []
   "NATIVE: test shadow compare"
+
   (for [x [< <= = != >= >]]
      (assert (try
               (x)
@@ -59,6 +60,7 @@
               (x 1)
               (except [TypeError] True)
               (else (raise AssertionError)))))
+
   (for [(, x y) [[< >=]
                  [<= >]
                  [= !=]]]
@@ -67,6 +69,7 @@
                 [1 1]
                 [2 2]]]
       (assert (= (apply x args) (not (apply y args))))))
+
   (let [s-lt <
         s-gt >
         s-le <=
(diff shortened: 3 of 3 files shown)


kirbyfan64:
Thanks! Want to add yourself to the `AUTHORS` file?

Kodiologist:
After thinking about this, I'm actually going to suggest kind of the opposite of this change, which is to stop shadowing comparison operators at all. The problem is that comparison operators can short-circuit, but a pure function can't; you need a macro for that. So it seems to me that `(setv f <)` should just be an error, since we can't possibly make this do the right thing.

Kodiologist:
Eh, on second thought, that can be a separate issue. However, this change needs tests. The test can use a mock class rather than real Sympy objects, so we don't add Sympy as a dependency.

gilch:
> After thinking about this, I'm actually going to suggest kind of the opposite of this change, which is to stop shadowing comparison operators at all.

Another option would be to restrict the arity for the shadowed versions to 2. I don't like the inconsistency, but it avoids the short-circuiting problem without removing the shadowed function for the common case.

gilch:
If we do remove the shadowed versions we could also add functions `increasing?`, `decreasing?`, `non-increasing?`, and `non-decreasing?` to compensate. These would not be expected to short-circuit, but could stop pulling from a lazy generator after the first contradiction, similar to Python's builtin `any` and `all`.

Kodiologist:
Hmm… my inclination right now is to keep on shadowing these operators, and to add shadowing for other possible short-circuiters like `and`. A user who's clever enough to try to use `<` or `and` as if it were a plain function should also be clever enough to see that `(setv f and) (f (g) (h) (i))` can't possibly short-circuit. It's probably worth a mention in the documentation, but overall, I'd rather provide a logical feature with somewhat counterintuitive consequences than withhold the feature and force people to write boilerplate lambdas. Sorry I'm being so wishy-washy.

PR #1195, opened 2016-12-30
Fix invalid escape sequences

jwilk:
Fixes:
```
hy/cmdline.py:175: DeprecationWarning: invalid escape sequence \_
tests/lex/test_lex.py:267: DeprecationWarning: invalid escape sequence \s
tests/compilers/test_compiler.py:45: DeprecationWarning: invalid escape sequence \*
```
when run against Python 3.6 with warnings enabled.

Found using [pydiatra](https://github.com/jwilk/pydiatra).


berkerpeksag:
Thanks!

PR #1204, opened 2017-01-19
Docs: add an example of evaluating a string

Kodiologist:


paultag:
Damn long time coming. LGTM; will merge when it's green, thank you @Kodiologist

PR #1206, opened 2017-01-20
Allow installtion without Git

Kodiologist:
Fixes #1205.

If all's well, this should probably be released as 0.12.1, since #1205 is kind of nasty to leave in a release.

PR #1208, opened 2017-01-24
Allow installation without Git

Kodiologist:
This a new PR for #1206.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,6 @@ import re
 import sys
 import runpy
 import subprocess
-from distutils.spawn import find_executable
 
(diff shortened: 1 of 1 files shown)


gilch:
Why does py34 pass if the branch is supposed to be missing altogether? Is our test setup broken?

Kodiologist:
The log says that `git clone --depth=50 --branch=version-gen-updates https://github.com/hylang/hy.git ` ran successfully in that case. Presumably that job happened to start early enough that the `git clone` ran before I deleted the wrongly pushed branch.

PR #1209, opened 2017-01-24
Fix 1205

Kodiologist:
Attempt 3 of #1206.


Kodiologist:
@paultag @gilch All green here.

Issue #1214, opened 2017-01-30
xor does not return false if None as param


madhat2r:
(xor False None) does not return anything, when I think it should return false.

Kodiologist:
To be precise, it returns `None`, which is the second argument. I believe this is the correct behavior; remember that `None` is false in Boolean context. So it does return false; it just doesn't return `False`. This is the same thing that `or` does.

madhat2r:
I guess you are correct @Kodiologist . I was just reading the docs and it says:

`xor is used in logical expressions to perform exclusive or. It takes two parameters. It returns True if only of the parameters is True. In all other cases False is returned. `

So I took that to mean that it letteraly retunred false in all other cases. But I can agree, it probabbly won't matter much. ;)

Kodiologist:
Well, no wonder you were confused; the documentation is just plain wrong. I'll make a pull request.

Kodiologist:
While I'm at it, let's make `xor` return the true argument instead of `True`, for consistency.

Kodiologist:
Done in #1215.

PR #1215, opened 2017-01-30
Make xor return single true inputs

Kodiologist:
The documentation should now be correct (#1214).

### DIFF
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -481,8 +481,9 @@
 
 (defn xor [a b]
   "Perform exclusive or between two parameters"
-  (or (and a (not b))
-      (and b (not a))))
+  (if (and a b)
+    False
+    (or a b)))
 
 (def *exports*
   '[*map accumulate butlast calling-module-name chain coll? combinations
--- a/tests/native_tests/language.hy
+++ b/tests/native_tests/language.hy
@@ -1019,12 +1019,23 @@
 
 (defn test-xor []
   "NATIVE: test the xor macro"
-  (let [xor-both-true (xor True True)
-        xor-both-false (xor False False)
-        xor-true-false (xor True False)]
-    (assert (= xor-both-true False))
-    (assert (= xor-both-false False))
-    (assert (= xor-true-false True))))
+
+  ; Test each cell of the truth table.
+  (assert (is (xor False  False) False))
+  (assert (is (xor False True)  True))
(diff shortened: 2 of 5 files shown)


hy/core/language.hy
```diff
@@ -481,8 +481,11 @@
 
 (defn xor [a b]
   "Perform exclusive or between two parameters"
-  (or (and a (not b))
-      (and b (not a))))
+  (if
+     (and (not a) (not b))  b
+     (and (not a)      b)   b
+     (and      a  (not b))  a
+                            False))
```
gilch: Couldn't this be simplified to:
```
(if
  (and a (not b)) a
  b)
```
?

Kodiologist: No, that returns true when `a` and `b` are both true.

gilch: Then how about `(if (and a b) False (or a b))`?

gilch: `xor` has the same truth table as `or`, with one exception: `(and a b)`, and you want it to behave like `or` otherwise.

Kodiologist: Okay. It passes the tests, and I can always appreciate a round of golf.

tests/native_tests/language.hy
```diff
@@ -1019,12 +1019,18 @@
...
-    (assert (= xor-true-false True))))
+
+  ; Test each cell of the truth table.
+  (assert (is (xor False  False) False))
+  (assert (is (xor False True)  True))
+  (assert (is (xor True  False) True))
+  (assert (is (xor True  True)  False))
+
+  ; Same thing, but with numbers.
+  (assert (is (xor 0 0) 0))
```
gilch: might be clearer as `(assert (is (xor False 0) 0))`, proving the second argument is returned, rather than the first.

Kodiologist: Added below.

gilch:
Approved, assuming Travis passes. 👍 I have no further objections. This is an improvement, though I'd just as soon remove `xor` altogether.

@hylang/core can we get another ack?

tuturto:
Can we update NEWS too? It'll make next release easier to do.

Kodiologist:
@tuturto Good idea. I've also added an instruction to do that in CONTRIBUTING.rst.

Kodiologist:
@tuturto We good?

tuturto:
it's good, thanks

Issue #1173, opened 2016-12-12
Rethink mangling


gilch:
Python since version 3.0 allows much of Unicode in its identifiers. Hy's punycode mangling therefore serves no purpose but to make Unicode identifiers harder to read and Hy-Python interop more difficult. Maybe we should remove this "feature" altogether.

Some Hy features are already restricted to Python 3 or later. Unicode identifiers could be another such feature. It's probably a minority of users that need Python 2 support at this point anyway (and this will only become more true over time). They'll be able to make do with ASCII.

On the other hand, mangling of ASCII characters can be improved. Hy already converts `-` to `_`, to allow more Lispy names. Hy code would look very different without this, but it does cause some problems.

The other rules are even worse. The earmuff conversion to all caps is of dubious value. It usually indicates a dynamic variable in Lisp, but Hy doesn't have those. (maybe we could add them #1089) I'd like to remove it.

Hy also converts a trailing `!` to trailing `_bang` and a trailing `?` to a leading `is_`. But if these characters appear anywhere else, they don't get converted. The AST mostly doesn't care when the results are not valid Python identifiers, but we have no guarantee this will continue. It's already been an issue for `getargspec` #1172 . There are other ASCII characters that are allowed in Hy symbols (like `+`) but never get converted at all.

Clojure also has to use Java identifiers, but it has a [more consistent approach](https://clojuredocs.org/clojure.core/munge) that we might consider emulating. Java's identifier rules are almost as strict as Python 2.

Kodiologist:
> Python since version 3.0 allows much of Unicode in its identifiers.

By Lisp standards, it's pretty idiosyncratic. For example, `λ` is legal but `⚘` isn't.

Punycode mangling seems to be broken or inactive at the moment, since `'⚘` returns `⚘`, not `hy_w7h` as documented.

I never use earmuffs, so I would support that removal. Lisps are traditionally case-insensitive, but Python and hence Hy is case-sensitive, so names in all caps are just fine.

The conversions of trailing `?` and `!` seem fine to me except for the annoyance you mentioned in #1115, and to be honest, I feel as if we ought to send a bug to the Python people pointing out that the inconsistency. It seems pretty obvious that Python should either have `is_integer` and `is_lower`, or `isinteger` and `islower`, but not one of each.

gilch:
> By Lisp standards, it's pretty idiosyncratic. For example, λ is legal but ⚘ isn't.

[...]

Issue #1115, opened 2016-09-24
clean up rules for `?`


gilch:
Hy converts a trailing `?` to an `is_` prefix. This means that you can use a more lispy style for common predicate methods:

``` Hy
=> (.integer? 1.0)
1.0.is_integer()
True
=> (.integer? 1.5)
1.5.is_integer()
False
```

But due to an unfortunate inconsistency in Python's naming conventions, it doesn't always work:

``` Hy
=> (.lower? "foo")
'foo'.is_lower()
Traceback (most recent call last):
  File "<input>", line 1, in <module>
AttributeError: 'str' object has no attribute 'is_lower'
=> (.islower "foo")
'foo'.islower()
True
```

There's no underscore in `islower`! This appears to be the case for all string predicate methods. It may also be true elsewhere in the Python standard library, and therefore in other Python libraries.

We may be able to adjust the rules to make this more consistent.
### ~~option 1~~

Make a prefixed `?` become a prefix `is`.

``` Hy
=> (.?lower "foo")
'foo'.islower()
True
```

This way, we can leave the suffix case (and any code using it) as-`is_`. But now there's three ways to write `is_integer`, `?-integer`, and `integer?`.
### ~~option 2~~

Make a suffix `?` become a prefix `is` and suffix `-?` become a prefix `is_`.

``` Hy
=> (.integer-? 1.0)
1.0.is_integer()
=> (.lower? "foo")
'foo'.islower()
```

This can break existing Hy code, but we're doing that anyway.
### option 3

Remove the special treatment of `?`. Use the same names in Hy as in Python. Hy becomes more consistent and less magical. Hy names ending in `?` become as difficult to use in Python as Hy names with an embedded `?` are now. But it's still possible with `getattr`.

We could probably come up with more options, but in any case, it needs to be documented. I didn't see the current `?` rule in the docs either.

kirbyfan64:
Wow I didn't even realize this was a thing. Personally, I prefer either option 1 or 2.

gilch:
Yup, you can see all the conversions [here](https://github.com/hylang/hy/blob/4cee7b1e1064298e1f130d2f681e29259285b1de/hy/lex/parser.py#L48). Note also that a `!` suffix becomes a `_bang` suffix. Hy also uses punycode, but that's in the docs.

Automagic _really_ should be documented. 

Another trick--since it happens at the parser level, it is possible to keep the question mark:
```Hy
=> 'query?
from hy import HySymbol
[...]

PR #1218, opened 2017-02-05
Docs: Hy <-> Python interop fix #1061

### LINKED ISSUE #1061: Add a section in the docs about using Hy from Python.
This doesn't seem to be explained very well. My understanding is that you can
- `import hy` in a Python script and then import modules written in Hy. Naturally, Hy must have been installed for this to work.
- create a Python module `foo` with an `__init__.py` that does the `import hy`, and then just `import foo`. (Hy must still have been installed. I think.)
- compile with `hyc` and get a normal `.pyc` file. Import said `.pyc`. Does this still work? Do you have to install Hy too, or does the `.pyc` include that?

I didn't see _any_ of this in the docs.

What if I want to `eval` (`exec`?) a string containing Hy code? Is there a way to do that from Python? Or can you only `import`? Can I launch a Hy repl from Python's? These capabilities are important for supporting doctest #1019. If these capabilities exist, they should be documented. If not, they should be added.

Ealhad:
I separated the existing text in two sections, with some additional explanations and a link to Hy's `import`.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -74,3 +74,4 @@
 * Karan Sharma <karansharma1295@gmail.com>
 * Sergey Sobko <s.sobko@profitware.ru>
 * Philip Xu <pyx@xrefactor.com>
+* Charles de Lacombe <ealhad@mail.com>
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -1049,6 +1049,7 @@ reverses the comparison.
     => (lif-not False "true" "false")
     "false"
 
+.. _import:
 
 import
 ------
--- a/docs/language/index.rst
+++ b/docs/language/index.rst
@@ -8,6 +8,7 @@ Contents:
    :maxdepth: 3
 
    cli
+   interop
    api
    core
    readermacros
--- a/docs/language/interop.rst
+++ b/docs/language/interop.rst
@@ -1,0 +1,150 @@
+
+=====================
+Hy <-> Python interop
+=====================
+
+   “Keep in mind we’re not Clojure. We’re not Common Lisp. We’re Homoiconic
+   Python, with extra bits that make sense.” — Hy Style Guide
+
+Despite being a Lisp, Hy aims to be fully compatible with Python. That means
+every Python module or package can be imported in Hy code, and vice versa.
+
+Using Python from Hy
+====================
+
+Using Python from Hy is nice and easy, you just have to :ref:`import` it.
+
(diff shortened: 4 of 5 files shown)


Kodiologist:
> You can also compile your module with `hyc`. Then you can import it directly (Hy still has to be installed)

This isn't always true. If you edit the example `greetings` module by renaming the parameter `name` to `addressee`, then the resulting `greetings.pyc` no longer requires Hy. Using `name` makes Hy add `from hy.core.language import name` at the top of the file because `name` happens to be the name of a builtin. (Is this a bug or an unavoidable limitation of Hy? It's not obvious to me.)

More generally, I think that fulfilling #1061 calls for adding discussion of lots of gory details. Your additions here don't suffice. @gilch filed that issue, so maybe he can chime in.

It looks like this is your first pull request. Please don't be scared away by how I've received it; we do need more contributors, to both the code and the documentation. Thanks for pitching in.

Ealhad:
First PR indeed, not scared away 😬 

> More generally, I think that fulfilling #1061 calls for adding discussion of lots of gory details. Your additions here don't suffice.

In that case, maybe this should be completely moved away from the tutorial, which would still contain general interop instructions without the “gory details”.

What I have understood so far is that if anywhere in the code, the name of a Hy builtin is found, it is imported.
For example :
```
(def second 42)
```
triggers 
```
from hy.core.language import second
```

Kodiologist:
> In that case, maybe this should be completely moved away from the tutorial, which would still contain general interop instructions without the "gory details".

Yeah, that makes sense to me. We've generally agreed that the tutorial should be fairly short so it isn't too intimidating.

> What I have understood so far is that if anywhere in the code, the name of a Hy builtin is found, it is imported.

I think that's how it works, yeah. I've never read the part of the codebase that does this.

Kodiologist:
Oh, and don't forget to add yourself to AUTHORS.

Kodiologist:
> I'd suggest documenting mangling in a separate PR.

Although, you may not want to do this now since mangling may change soon (#1173, #1115), anyway.

Ealhad:
Thank you for those words! (tssst, I wanted to put a space before my exclamation mark. It’s hard being French when writing in English)
I’ll make the corrections quickly.

    I'd suggest documenting mangling in a separate PR.
Shouldn’t mangling still be mentioned in this section? Maybe a link would suffice.

Kodiologist:
Yeah, a link would make sense.

PR #1221, opened 2017-02-10
Don't put redundant names in _stdlib

Kodiologist:
This prevents pointless imports of hy.core, like importing `map` in Python 3.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -55,6 +55,11 @@ import keyword
 
 from collections import defaultdict
 
+if PY3:
+    import builtins
+else:
+    import __builtin__ as builtins
+
 
 _compile_time_ns = {}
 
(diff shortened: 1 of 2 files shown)


gilch:
There's a potential problem with this approach. If a builtin acquires a new feature between Python versions (via an extra optional argument, say), and we decide to backport it by reimplementing the newer version in Hy.core, this change would prevent that backport from working in the older versions of Python. I don't think we're backporting any builtins currently however. I was worried about `accumulate`, but it's not actually a Python builtin, so it doesn't appear to be affected.

Kodiologist:
> this change would prevent that backport from working in the older versions of Python.

Not so: the code `getattr(mod, e) is not getattr(builtins, e, '')` checks that the builtin really is the same function as the one we'd otherwise import. We do in fact have some backports right now, such as Python 2's `filter` getting replaced by `itertools.filter`, and the tests I added check that the import statement is generated when and only when it's necessary.

PR #1223, opened 2017-02-14
Allow commas and underscores in numeric literals

Kodiologist:
You can use them as thousands separators.

This change differs from [PEP 515](https://www.python.org/dev/peps/pep-0515/) in that not only does it allow commas in addition to underscores, but it's much more liberal about placement. Any number of underscores or commas can be placed anywhere, even at the start.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,7 @@ Changes from 0.12.1
    [ Language Changes ]
    * `let` has been removed. Python's scoping rules do not make a proper
      implementation of it possible. Use `setv` instead.
+   * Commas and underscores are allowed in numeric literals
    * xor: If exactly one argument is true, return it
 
    [ Bug Fixes ]
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -18,7 +18,7 @@ These rules help ensure that Hy code is idiomatic and interfaceable in both
 languages.
 
 
-  * Symbols in earmufs will be translated to the upper-cased version of that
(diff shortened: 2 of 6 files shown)


Kodiologist:
I don't know why those assertions are failing on Travis; they work on my own machine with CPython 2.7.12.

kirbyfan64:
> Any number of underscores or commas can be placed anywhere, **even at the start**.

Wouldn't it then still be an identifier?

Also, looks like the tests are only for failing for Python 2, so maybe you ran yours on 3?

Kodiologist:
> Wouldn't it then still be an identifier?

No. See the tests.

> maybe you ran yours on 3?

Like I said, they pass with CPython 2.7.12.

Kodiologist:
Okay, phew. It looks like my virtualenv was playing tricks on me and running the tests with Python 3 even though just typing `hy` at the command line got me Python 2. And there was indeed a Python 2 bug, but it's fixed now.

kirbyfan64:
I don't know, it now seems kind of odd that this would happen:

```hy
,,,  ; Identifier
,,1  ; Number
,1,  ; Number
,_,  ; Identifier
__,  ; Identifier
___  ; Identifier
__1  ; Number
```

But I guess it's generally a lose-lose scenario... :/

Ehh, get's my +1.

Kodiologist:
Come to think of it, humorously, this makes `[1, 2, 3]` legal Hy.

PR #1224, opened 2017-02-17
Refactor hy.models

Kodiologist:


hy/contrib/multi.hy
```diff
@@ -21,9 +21,7 @@
 ;; DEALINGS IN THE SOFTWARE.
 
 (import [collections [defaultdict]]
-        [hy.models.expression [HyExpression]]
-        [hy.models.list [HyList]]
-        [hy.models.string [HyString]])
+        [hy [HyExpression HyList HyString]])
```
gilch: I'm confused by this. Shouldn't it be `[hy.models [HyExpression HyList HyString]]`? And yet the tests pass. Are we not testing this?

gilch:
The docs would also have to be updated to match the new structure. Also, is this really an improvement? I'm not sure. There was no discussion. Hy really could use some refactoring though.

hy/contrib/multi.hy
kirbyfan64: @gilch Hy re-exports them in the main module: https://github.com/hylang/hy/blob/master/hy/__init__.py

kirbyfan64:
IMO it is: the models are rather short, and placing them in separate files really didn't do much except make it a little trickier to maneuver through.

Kodiologist:
> The docs would also have to be updated to match the new structure.

Oh, whoops, I assumed that none of it was documented. I'll do that now.

Kodiologist:
Done.

gilch:
I'll approve it then. 👍 

Is this still small enough to squash? I've seen some disagreement on when that's appropriate.

Kodiologist:
I made it into three separate commits because I think it's nice to have the reorganization separate from the slightly more meaningful changes.

PR #1226, opened 2017-02-18
Add bytestring literals

### LINKED ISSUE #495: Find a way to ruin our glorious unicode future with regular bytes / ascii for ${THINGS}
FSVO `${THINGS}` such as setuptools, which requires `package_data` to use regular ascii strings as package names. Annoying.

Thanks to @oubiwann for the bug

it'd be good to get something like `b"foo"` working (since we already have `u"foo"` and `r"foo"`)

Damnit, setuptools.

Kodiologist:
Closes #495 (after a mere three years).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,8 @@ Changes from 0.12.1
    [ Language Changes ]
    * `let` has been removed. Python's scoping rules do not make a proper
      implementation of it possible. Use `setv` instead.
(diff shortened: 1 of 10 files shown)


kirbyfan64:
Never realized this had never been added?

Kodiologist:
I thought it would be pretty hard, but it turned out to be easy once I thought of making a new `HyBytes` model instead of trying to making `HyString` accommodate both.

PR #1228, opened 2017-02-19
Make `fn` work like `lambda` and remove `lambda`

### LINKED ISSUE #910: lambda vs fn
One of the subissues from #240.

They're not quite the same, but still too similar to need both. Hy's `lambda` will use a Python `lambda` if it can. When it can't, because there's a statement or multiple lines, it acts exactly the same as `fn`. I don't know about Hy, but Python mostly doesn't care to distinguish lambdas from ordinary function objects. You can even set `lambda`s as methods for classes.

I'm in favor of keeping `lambda` and removing `fn`. If there actually is a case where you need a one-expression function and can't use a Python lambda, then you can force a Hy lambda to be a function simply by adding a docstring.

If we're OK with not reserving Python keywords as discussed in #899, then after dropping `fn` we can rename `lambda` to `fn` to keep the shorter alias. Then Hy would treat `lambda` like any other symbol. This may be problematic for Python interop, but there are workarounds.

Kodiologist:
Closes #910.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,7 @@ Changes from 0.12.1
    * Added bytestring literals, which create `bytes` objects under Python 3
      and `str` objects under Python 2
    * Commas and underscores are allowed in numeric literals
+   * with-decorator: Allow a `setv` form as the form to be decorated
    * xor: If exactly one argument is true, return it
 
    [ Bug Fixes ]
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,7 @@ Changes from 0.12.1
    [ Language Changes ]
    * `let` has been removed. Python's scoping rules do not make a proper
      implementation of it possible. Use `setv` instead.
+   * `lambda` has been removed, but `fn` now does exactly what `lambda` did
    * Added bytestring literals, which create `bytes` objects under Python 3
      and `str` objects under Python 2
    * Commas and underscores are allowed in numeric literals
--- a/docs/extra/anaphoric.rst
+++ b/docs/extra/anaphoric.rst
@@ -236,7 +236,7 @@ xi
 
 Usage ``(xi body ...)``
 
-Returns a function with parameters implicitly determined by the presence in the body of xi parameters. An xi symbol designates the ith parameter (1-based, e.g. x1, x2, x3, etc.), or all remaining parameters for xi itself. This is not a replacement for lambda. The xi forms cannot be nested.
+Returns a function with parameters implicitly determined by the presence in the body of xi parameters. An xi symbol designates the ith parameter (1-based, e.g. x1, x2, x3, etc.), or all remaining parameters for xi itself. This is not a replacement for fn. The xi forms cannot be nested.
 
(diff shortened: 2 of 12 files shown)


tuturto:
First thing to come to mind is that lambdas can't be decorated:

```hy
=> (defn double [func] (fn [x] (* (func x) 2)))
=> ((with-decorator double (fn [x] x)) 5)
  File "<input>", line 1, column 2

  ((with-decorator double (fn [x] x)) 5)
   ^--------------------------------^
HyTypeError: b'Decorated a non-function'
```

But without the changes, example works:

```hy
(defn double [func] (fn [x] (* (func x) 2)))

=> ((with-decorator double (fn [x] x)) 5)
10

=> ((with-decorator double (lambda [x] x)) 5)
  File "<input>", line 1, column 2

  ((with-decorator double (lambda [x] x)) 5)
   ^------------------------------------^
HyTypeError: b'Decorated a non-function'
```

In my opinion, this is big enough drawback not to do this change.

Kodiologist:
That could be implemented, but why would you write `(with-decorator double (fn [x] x))` when you could just write ` (double (fn [x] x))`? What's the point?

tuturto:
Point of the code was to show a simple testcase which will fail if this change is pulled in. Actual content of the example decorator isn't important. Currently user can be confident that result of ``fn`` can be decorated, regardless of how simple or complex it actually is. With the change, they can't anymore.

Kodiologist:
My argument isn't specific to your test case. It applies no matter how complex the decorator or the decorated `(fn ...)` is: `(with-decorator d (fn ...))` can always be written as `(d (fn ...))`, so `with-decorator` is wholly redundant. The reason decorators exist is so you can use them with `def` (`defn` in Hy) and `defclass`, not with expressions. Indeed, the fact that `(with-decorator d (fn ...))` worked was untested and undocumented.

tuturto:
Right, now I follow. Shouldn't test things before second cup of coffee in morning.

Indeed, following works as expected:

```hy
=> (defn double [func] (fn [x] (* (func x) 2)))
=> (with-decorator double (defn test [x] x))
=> (test 2)
4
```

Kodiologist:
Like the commit message says, "it's necessary to allow `defn` to create a lambda instead of a `def`." For example, if you do `(with-decorator d (defn f [x] (+ x 1)))`, the `(defn ...)` gets compiled to something like `f = lambda x: x + 1`, so `with-decorator` needs to be able to handle `ast.Assign`, which in Hy usually appears from a `(setv ...)` form.

Kodiologist:
So technically, supporting `(setv ...)` forms isn't necessary; it's just an immediate consequence of supporting `ast.Assign`.

Kodiologist:
Or, putting it another way, the `(setv ...)` produced by `defn` used to always lead to `ast.FunctionDef` downstream. Now it might be an `ast.Assign`.

Kodiologist:
And that's commit 2,000. [This calls for celebration!](https://www.youtube.com/watch?v=7G4RYtP5RQQ)

PR #1235, opened 2017-02-24
Minor test reorganization

Kodiologist:
This removes two singleton directories.


Kodiologist:
I guess this is a trivial change, so I can merge it.

PR #1238, opened 2017-02-25
Fix #1237 (iterator-handling bug in `partition`)

### LINKED ISSUE #1237: (partition (cycle coll) n) produce wrong result
```hylang
hy 0.12.1+24.g45b7a4a using CPython(default) 2.7.12 on Darwin
=> (list (take 10 (partition (cycle [1 2 3]) 3)))
from hy.core.language import cycle, partition, take
list(take(10L, partition(cycle([1L, 2L, 3L]), 3L)))
[(1L, 3L, 3L), (3L, 3L, 3L), (3L, 3L, 3L), (3L, 3L, 3L), (3L, 3L, 3L), (3L, 3L, 3L), (3L, 3L, 3L), (3L, 3L, 3L), (3L, 3L, 3L), (3L, 3L, 3L)]
=> (list (take 10 (cycle [1 2 3])))
from hy.core.language import cycle, take
list(take(10L, cycle([1L, 2L, 3L])))
[1L, 2L, 3L, 1L, 2L, 3L, 1L, 2L, 3L, 1L]
```

It seems that the `cycle` inside `partition` is not working right.

Kodiologist:


hy/core/language.hy
```diff
@@ -369,12 +369,15 @@
...
-   step (or step n)
-   slices (genexpr (itertools.islice coll start None step) [start (range n)]))
-  (if (is fillvalue -sentinel)
-    (apply zip slices)
-    (apply zip-longest slices {"fillvalue" fillvalue})))
+  (if (< n 1) [] (do
+    (setv
+     step (or step n)
+     coll-clones (tee coll n)
+     slices (genexpr (itertools.islice (get coll-clones start) start None step)
```
gilch: I think this can just be `islice` instead of `itertools.islice` now that we're auto-importing everything from `itertools`. I wrote this before we did that.

hy/core/language.hy
```diff
@@ -369,12 +369,15 @@
   "Chunks coll into n-tuples (pairs by default). The remainder, if any, is not
    included unless a fillvalue is specified. The step defaults to n, but can be
    more to skip elements, or less for a sliding window with overlap."
-  (setv
-   step (or step n)
-   slices (genexpr (itertools.islice coll start None step) [start (range n)]))
-  (if (is fillvalue -sentinel)
-    (apply zip slices)
-    (apply zip-longest slices {"fillvalue" fillvalue})))
+  (if (< n 1) [] (do
```
gilch: Huh. I'm not sure about this part. This kind of feels like a pointless check. It already handles `n == 0` properly without this. The `n < 0` case doesn't really make sense, so it probably should be "undefined" for this function. I know it was the empty list before, but throwing an exception might actually be better. I compared it to Clojure, which also returns `()` for the `n < 0` case, so there's that.

On the other hand, Clojure actually _diverges_ in the `n == 0` case, which is not good behavior all. This leads me to believe that all `n<1` inputs are nonsensical and should be an error.

gilch:
Good work @Kodiologist. This is exactly the approach I was taking.

hy/core/language.hy
gilch: This looks relavant: http://dev.clojure.org/jira/browse/CLJ-1647

Maybe an `(assert (pos? n))` at the top would suffice.

Kodiologist: `n` = 0 used to work, but `tee` raises an exception when it gets an argument of 0. That's why I added the check.

I agree with you that `n` < 0 should just be an error, but come to think of it, shouldn't `n` = 0 return a list of empty tuples?

gilch: A list of what length though? Even if you attempt to eat a finite iterator by taking zero-sized bites, you'll never finish the meal. On the other hand, it's lazy. This actually does work in Clojure:
```Clojure
user=> (take 10 (partition 0 [1 2 3]))
(() () () () () () () () () ())
```

Kodiologist: Yep, I meant an infinitely long iterator of empty tuples, as Clojure seems to do.

gilch: Wait, you said `tee` raises an exception when `n` is `0`? Does it really?
```Python
>>> tee([1,2,3],0)
()
```
Seems to work on Python 3.5, at least.

Kodiologist: Sorry, I misread the error message. `tee` complains about negative `n`, not zero. So shall we have `partition` with `n` = 0 return infinitely many empty tuples, and raise an exception on negative `n`?

gilch: Yes, let's raise an exception on negative n. `tee` already does that for us. (`islice` also ensures that `step` is positive.)

As for `n == 0`, I'm really not certain. If you actually _want_ an infinite sequence of empty tuples, you'd normally just `(repeat (,))`, so `partition` probably doesn't have to be able to do this. And diverging when you thought you had a finite iterator is not nice.

I'm inclined to just remove this check and let it return `[]` when `n == 0`, then see if anybody complains. @hylang/core thoughts? How should this work?

hy/core/language.hy
Kodiologist: Done.

hy/core/language.hy
Kodiologist: Yeah, I guess that makes sense. I've done that.

tuturto: ```[]``` sounds safer option to me. Or even raising a suitable exception (or is that too drastic?) Having a call to ```partition``` where ```n``` is 0 sounds like a bug to me and Hy probably should report that instead of trying to keep the system limping forward.

PR #1241, opened 2017-02-27
Remove `car` and `cdr` in favor of `first` and `rest`

### LINKED ISSUE #909: Remove `car`/`cdr` in favor of `first`/`rest`
One of the subissues from #240. I think we'll make more progress discussing one subissue at a time.

`rest` / `cdr` are not quite the same, but still redundant. `(cdr coll)` is a macro that expands into a slice like `coll[1:]`. `rest` is a partial application of the function `islice` that does about the same thing as `cdr`. As a function, `rest` can be assigned to variables, passed to other functions, etc. As a macro, `cdr` cannot. Slices also don't work on generators so `cdr` doesn't work, but islices do, and so does `rest`. `rest` will do in almost all cases we currently use `cdr`. 

The main problem with replacing all slices with islices, is that islice will always be a generator, but slices return another instance of the same collection type. Most of the time this isn't a problem. Python's libraries are very good about generally accepting iterables where you might have used a collection before. The main exception is strings, which really need to stay as strings.

If you do run into a case like this, you can still use `(cut coll 1)` for the same effect.

The case for dropping `car` in favor of `first` is even stronger. `car` is a macro like `coll[0]`, while `first` is a function like `next(islice(coll, 0, None))`, which should perhaps be simplified to `next(iter(coll))`. Either way, you get the first element. Except `car` doesn't work on generators, and can't be passed as a higher-order function, etc. There is absolutely no reason to keep `car` except for symmetry with `cdr`, which we should also drop.

Kodiologist:
Closes #909.

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -850,16 +850,26 @@ eval-when-compile
 -----------------
 
 
-first / car
------------
+first
+-----
 
-``first`` and ``car`` are macros for accessing the first element of a collection:
+``first`` is a function for accessing the first element of a collection.
 
 .. code-block:: clj
 
     => (first (range 10))
     0
 
+It is implemented as ``(next (iter coll) None)``, so it works with any
+iterable, and if given an empty iterable, it will return ``None`` instead of
+raising an exception.
+
+.. code-block:: clj
+
+    => (first (repeat 10))
+    10
+    => (first [])
+    None
 
 for
 ---
@@ -1402,17 +1412,23 @@ periods in them. In fact, ``mymodule`` and ``M`` aren't defined by these
 to do introspection of the current module's set of defined macros, which isn't
 really supported anyway.
 
-rest / cdr
-----------
+rest
+----
 
-``rest`` and ``cdr`` return the collection passed as an argument without the
-first element:
+``rest`` takes the given collection and returns an iterable of all but the
+first element.
 
 .. code-block:: clj
 
-    => (rest (range 10))
+    => (list (rest (range 10)))
     [1, 2, 3, 4, 5, 6, 7, 8, 9]
 
+Given an empty collection, it returns an empty iterable.
+
+.. code-block:: clj
+
+    => (list (rest []))
+    []
 
 set-comp
 --------
--- a/docs/language/core.rst
+++ b/docs/language/core.rst
@@ -111,10 +111,10 @@ Returns a fresh :ref:`cons cell <hycons>` with car *a* and cdr *b*.
 
    => (setv a (cons 'hd 'tl))
 
-   => (= 'hd (car a))
+   => (= 'hd (get a 0))
    True
 
-   => (= 'tl (cdr a))
+   => (= 'tl (cut a 1))
    True
 
 
--- a/hy/contrib/multi.hy
+++ b/hy/contrib/multi.hy
@@ -87,6 +87,9 @@
        (with-decorator (method-decorator ~name)
          (defn ~name ~params ~@body))))
 
+(defn head-tail [l]
+  (, (get l 0) (cut l 1)))
+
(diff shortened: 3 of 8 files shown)


docs/language/api.rst
```diff
@@ -850,16 +850,26 @@ eval-when-compile
...
-``first`` and ``car`` are macros for accessing the first element of a collection:
+``first`` is a function for accessing the first element of a collection.
 
 .. code-block:: clj
 
     => (first (range 10))
     0
 
+It is implemented as ``(next (iter coll) None)``, so it works with any
+iterable, and if given an empty iterable, and it will return ``None`` instead
```
kirbyfan64: Second *and* is redundant (though I feel like this might be easier to read if the section on empty literals was a separate sentence).

docs/language/api.rst
```diff
@@ -1402,17 +1412,23 @@ periods in them. In fact, ``mymodule`` and ``M`` aren't defined by these
...
 
-rest / cdr
-----------
+rest
+----
 
-``rest`` and ``cdr`` return the collection passed as an argument without the
-first element:
+``rest`` returns the collection passed as an argument without the
+first element, as an iterable:
```
kirbyfan64: How about:

> ``rest`` takes the given collection and returns an iterable of all but the first element.

Kodiologist:
Most `(get x 0)`s could probably be replaced with `first`. I opted for a more literal translation so I didn't have to check that `first` doesn't prematurely eat up an iterator or something. Do you want me to use `first` where I can?

`head-tail` is an orthogonal bit of refactoring to the removal of `car` and `cdr`, but it was an obvious opportunity for refactoring, which I took when I noticed.

Incidentally, I hate iterables. In practice, the laziness is of no benefit, I have to convert them to a list or tuple in order to do most any operation on them, and I get bitten by subtle bugs that arise because iterators drop each element once it's accessed. Python 3 was in the wrong to make `map`, `range`, etc. produce iterables. I'm of half a mind to make Hy's iterable-producing functions (`map`, `rest`, etc.) automagically produce a collection instead when they're given a collection.

kirbyfan64:
IMO iterables can be really useful (try filtering over a collection of ~2000 items and you'll see what I mean), but using them with collections *can* be annoying sometimes. In Python, it's less of an issue because those functions aren't even used too often, but in Hy, which has more of a functional leaning, it gets more annoying.

I kind of wish sometimes that Hy had a set of functions (e.g. `map-c` or something) that specifically return collections instead...

Kodiologist:
I tend to use macros ([e.g.](https://github.com/Kodiologist/Kodhy/blob/1a4a1b9087a48c384ca5975b0f55f5c23645eaaa/kodhy/macros.hy#L40)) that are both anaphoric and always return lists.

You haven't approved this; what else needs to be done?

gilch:
> Incidentally, I hate iterables. In practice, the laziness is of no benefit, I have to convert them to a list or tuple in order to do most any operation on them, and I get bitten by subtle bugs that arise because iterators drop each element once it's accessed.

The solution might be the seq abstraction from Clojure #1242.

kirbyfan64:
Sorry, forgot about this. I guess the whole list vs iterator thing is more suited for another issue.

Davoodeh:
I am a newbie just been playing around with Hy
Just saw this... It just feels off... Why would anybody want to possibly remove `car` / `cdr` from a Lisp?
I tried to can't there be an alias of some sort built-in Hy for this at least?

Kodiologist:
@Davoodeh They were confusingly similar-but-not-identical-to `first` and `rest`; you couldn't use them in higher-order functions because they were macros rather than functions; and the names make no sense in Hy because there are no cons cells. If you really want them back, you can copy and paste the definitions into your own code.

PR #1244, opened 2017-02-28
Fix #1243: `read` raises EOFError on false inputs

### LINKED ISSUE #1243: read-str raises EOFError given "\"\"", "b\"\"", "[]", or "{}"
Or `"#{}"`.

Kodiologist:


Kodiologist:
I found a very similar bug in `eval`, so I added a fix for that. Remember not to squash.

hy/compiler.py
```diff
@@ -2689,18 +2689,20 @@ def hy_compile(tree, module_name, root=ast.Module, get_expr=False):
     body = []
     expr = None
 
-    if tree:
+    if not (isinstance(tree, HyObject) or type(tree) is list):
```
kirbyfan64: Is this really that useful? IMO part of the charm of Python is duck-typing, where someone can pass a custom but API-compatible list subclass, and it "just works". Can't you just do `isinstance(tree, list)`?

Kodiologist: If you passed in an object that is neither exactly `list` nor a `HyObject`, the below conditional in the assignment to `spoof_tree` would do the wrong thing.

Kodiologist: Also, this is an internal function. If it receives an argument that's of a type we didn't expect, that's a bug.

kirbyfan64: Fair enough.

Kodiologist:
Thanks. I'm not going to ask you to undo the merge or reopen the PR, but I think this wasn't a trivial change, so it should've gotten a second review.

kirbyfan64:
Dammit, that was an accident...

GitHub mobile is killing me here...

PR #1245, opened 2017-03-05
Fix bug: quoting a bytestring raises ImportError

Kodiologist:
No NEWS entry is required because bytestrings are already a new feature.


kirbyfan64:
Haha, this is gold. ;)

(FWIW, I'm just gonna merge this, since it's a really minor change: just adding an import.)

Issue #1049, opened 2016-04-16
Do we *really* want setv to return a tuple?


gilch:
Currently `setv` returns a tuple. Sometimes.

``` Hy
=> (print (setv x 7 y 42))
(7, 42)
=> (print (setv x 42))  ; shouldn't it print (42,)?
42
=> (print (setv x 7 x 42)  ; wouldn't you expect (7, 42)?
(42, 42)
```

I propose changing it to just return the value of the last assignment. This is more consistent with the way a single assignment works. If you assign to the same name more than once, it gets lost anyway, which is surprising. This is also more consistent with Common Lisp's `setq` which we've based `setv` on.

``` Lisp
* (setq x 42 x 42)
;; [dire warnings about using setq at toplevel]
42
*
```

It would also simplify my proposed `as->` macro. #1047 .

shellhead:
Is this a duplicate of #249 ?

gilch:
> It seems like the behavior may have changed since this was originally opened. 

@shellhead No, it's the opposite. `setv` has changed to allow multiple pairs from only one pair in the old PyPI version that you still seem to be using. Use the current github version for the new `setv`.

Issue #871, opened 2015-08-02
inconsistent return behavior for setv in repl


gilch:
When testing #870, I noticed something strange.

``` Hy
=> (setv foo (lambda[]nil))
foo = (lambda : None)
=> ; setv effectively returns nil
=> (setv x 1)
x = 1
=> ; unless assigning an attribute?
=> (setv foo.x 1)
foo.x = 1
foo.x
1
=> ; actually it works for variables too, but not directly in the repl
=> (print (setv x 1))
x = 1
print(x)
1
```

Programmers use the repl to quickly test the behavior of functions. From a quick test like this, anyone would assume at first that `setv` always returns `nil`, but, as seen above, it doesn't. This is confusing behavior.

Issue #1089, opened 2016-08-13
Consider adding dynamic variables to Hy


gilch:
Lisp originally used dynamic variables, instead of the lexical variables used in Python. They're still the norm in Emacs Lisp, and useful enough sometimes that they're still available optionally in Common Lisp. For those of you not familiar with Elisp or Common Lisp, they're basically Clojure's thread-local vars.

I also think that a dynamic `let` would be much easier to implement. The dynamic version doesn't need closures, and could be compiled into a `with` statement. After purging the broken lexical `let` from Hy #1056, we could replace it with a dynamic one that has the same semantics as Elisp's `let` (or Clojure's `binding` form).

tuturto:
This sounds intriguing and useful. Recently I had to hack together something that sounds like this, but it's specifically only for functions and requires extra step for calling them: ```(call foo bar)``` instead of just ```(foo bar)```.

But I'm not completely sure if I understand (or even can think of) all possible cases for this. Variables with dynamic scope would only be available inside a ```let``` form that referes to them? Or would they be available somewhere else too? Could one define a function with ```defn```, bind it to dynamically scoped variable and then call it later during program execution? How would the definition part and calling part look like in Hy code? Any idea (rough sketch is enough) what the resulting Python code would look like?

gilch:
First, a quick demonstration of dynamic variables in Emacs Lisp, so we're on the same page:
```Elisp
ELISP> (defun greet ()
         (format "Hello, %s!" the-name))
greet
ELISP> (defvar the-name "World")
the-name
ELISP> (greet)

"Hello, World!"

nil
ELISP> (defun greet-alice ()
         (let ((the-name "Alice"))
           (greet)))
greet-alice
ELISP> (greet-alice)

"Hello, Alice!"

nil
```
This is not just a global, you can shadow an earlier binding with a later one, and it returns to its previous value:
```Elisp
ELISP> (defun greet-multiple ()
         (greet)
         (greet-alice)
         (let ((the-name "Bob"))
           (greet))
         (greet))
           
greet-multiple
ELISP> (greet-multiple)

"Hello, World!"

"Hello, Alice!"

"Hello, Bob!"

"Hello, World!"

nil
ELISP> (let ((the-name "Everyone"))
         (greet-multiple))

"Hello, Everyone!"

"Hello, Alice!"

"Hello, Bob!"

"Hello, Everyone!"

nil
```
[...]

PR #1246, opened 2017-03-06
Homoiconic REPL

### LINKED ISSUE #351: The REPL is not homoiconic enough
Having the REPL echo back in Python notation kinda sucks. I'm entertaining the idea of this being acceptable:

```
=> ["flat" ["lists" ["without" "commas"]]]
["flat" ["lists" ["without" "commas"]]]
=> {:hy_is {:full_of "wat"}}
{:hy_is {:full_of "wat"}}
=> (import os)
=> (list (os.walk "./hy/lex"))
[["./hy/lex" [] ["__init__.py" "__init__.pyc" "exceptions.py" "exceptions.pyc" "lexer.py" "lexer.pyc" "parser.py" "parser.pyc"]]]
```

Here's a PoC covering the simple cases:
https://github.com/hcarvalhoalves/hy/tree/hy-repl

Let's get the ball rolling.

Kodiologist:
This pull request adds a function `hy-repr` and uses it to print out stuff in the REPL (edit: only when you launch Hy with `--hy-repr`).

Closes #351.

#1244 is a prerequesite. Merge that first and I'll rebase this.

An annoying consequence of this change (edit: now fixed, using the method listed second below) is that `setv` at the REPL now prints the value of whatever you assigned. I think this is bad because when I'm doing data analysis on the REPL, I often say things like `(setv x (some-function-that-produces-a-huge-pandas-dataframe))`, and I don't want that printed. I see two ways around this:

- Make `setv` always return `None`. The question of what `setv` should return, particularly in the REPL, is kind of a mess, anyway (#249, #871, #1049), so this change would eliminate another weird corner of the language.
- Make the REPL check whether the last statement of each input is `ast.Assign`, and if so, don't print the value.

As for the documentation examples, I don't think we should bother systematically updating them before we have documentation tests.

### DIFF
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -32,14 +32,16 @@ import code
 import ast
 import sys
 import os
+import importlib
 
 import astor.codegen
 
 import hy
 
 from hy.lex import LexException, PrematureEndOfInput, tokenize
-from hy.compiler import hy_compile, HyTypeError
-from hy.importer import (ast_compile, import_buffer_to_module,
+from hy.lex.parser import hy_symbol_mangle
+from hy.compiler import HyTypeError
+from hy.importer import (hy_eval, import_buffer_to_module,
                          import_file_to_ast, import_file_to_hst,
                          import_buffer_to_ast, import_buffer_to_hst)
 from hy.completer import completion
@@ -73,25 +75,35 @@ builtins.quit = HyQuitter('quit')
 builtins.exit = HyQuitter('exit')
 
 
-def print_python_code(_ast):
-    # astor cannot handle ast.Interactive, so disguise it as a module
-    _ast_for_print = ast.Module()
-    _ast_for_print.body = _ast.body
-    print(astor.codegen.to_source(_ast_for_print))
-
-
 class HyREPL(code.InteractiveConsole):
-    def __init__(self, spy=False, locals=None, filename="<input>"):
+    def __init__(self, spy=False, output_fn=None, locals=None,
+                 filename="<input>"):
+
         self.spy = spy
+
+        if output_fn is None:
+            self.output_fn = repr
+        elif callable(output_fn):
+            self.output_fn = output_fn
+        else:
+            f = hy_symbol_mangle(output_fn)
+            if "." in output_fn:
+                module, f = f.rsplit(".", 1)
+                self.output_fn = getattr(importlib.import_module(module), f)
+            else:
+                self.output_fn = __builtins__[f]
+
         code.InteractiveConsole.__init__(self, locals=locals,
                                          filename=filename)
 
     def runsource(self, source, filename='<input>', symbol='single'):
         global SIMPLE_TRACEBACKS
         try:
-            tokens = tokenize(source)
-        except PrematureEndOfInput:
-            return True
+            try:
+                tokens = tokenize(source)
+            except PrematureEndOfInput:
+                return True
+            tokens = tokenize("(do " + source + "\n)")
         except LexException as e:
             if e.source is None:
                 e.source = source
@@ -100,10 +112,15 @@ class HyREPL(code.InteractiveConsole):
             return False
 
         try:
-            _ast = hy_compile(tokens, "__console__", root=ast.Interactive)
-            if self.spy:
-                print_python_code(_ast)
-            code = ast_compile(_ast, filename, symbol)
+            def ast_callback(main_ast, expr_ast):
+                if self.spy:
+                    # Mush the two AST chunks into a single module for
+                    # conversion into Python.
+                    new_ast = ast.Module(main_ast.body +
+                                         [ast.Expr(expr_ast.body)])
+                    print(astor.to_source(new_ast))
+            value = hy_eval(tokens[0], self.locals, "__console__",
+                            ast_callback)
         except HyTypeError as e:
             if e.source is None:
                 e.source = source
@@ -117,7 +134,12 @@ class HyREPL(code.InteractiveConsole):
             self.showtraceback()
             return False
 
-        self.runcode(code)
+        if value is not None:
+            # Make the last non-None value available to
+            # the user as `_`.
+            self.locals['_'] = value
+            # Print the value.
+            print(self.output_fn(value))
         return False
 
 
@@ -211,7 +233,7 @@ def run_file(filename):
     return 0
 
 
-def run_repl(hr=None, spy=False):
+def run_repl(hr=None, **kwargs):
     import platform
     sys.ps1 = "=> "
     sys.ps2 = "... "
@@ -221,7 +243,7 @@ def run_repl(hr=None, spy=False):
     with completion(Completer(namespace)):
 
         if not hr:
-            hr = HyREPL(spy, namespace)
+            hr = HyREPL(locals=namespace, **kwargs)
 
         hr.interact("{appname} {version} using "
                     "{py}({build}) {pyversion} on {os}".format(
@@ -236,8 +258,8 @@ def run_repl(hr=None, spy=False):
     return 0
 
 
-def run_icommand(source, spy=False):
-    hr = HyREPL(spy)
+def run_icommand(source, **kwargs):
+    hr = HyREPL(**kwargs)
     if os.path.exists(source):
         with open(source, "r") as f:
             source = f.read()
@@ -272,7 +294,9 @@ def cmdline_handler(scriptname, argv):
         help="program passed in as a string, then stay in REPL")
     parser.add_argument("--spy", action="store_true",
                         help="print equivalent Python code before executing")
-
+    parser.add_argument("--repl-output-fn",
+                        help="function for printing REPL output "
+                             "(e.g., hy.contrib.hy-repr.hy-repr)")
     parser.add_argument("-v", "--version", action="version", version=VERSION)
 
     parser.add_argument("--show-tracebacks", action="store_true",
@@ -315,7 +339,8 @@ def cmdline_handler(scriptname, argv):
 
     if options.icommand:
         # User did "hy -i ..."
-        return run_icommand(options.icommand, spy=options.spy)
+        return run_icommand(options.icommand, spy=options.spy,
+                            output_fn=options.repl_output_fn)
 
     if options.args:
         if options.args[0] == "-":
@@ -332,7 +357,7 @@ def cmdline_handler(scriptname, argv):
                 sys.exit(e.errno)
 
     # User did NOTHING!
-    return run_repl(spy=options.spy)
+    return run_repl(spy=options.spy, output_fn=options.repl_output_fn)
 
 
 # entry point for cmd line script "hy"
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -103,7 +103,9 @@ class HyREPL(code.InteractiveConsole):
                 tokens = tokenize(source)
             except PrematureEndOfInput:
                 return True
-            tokens = tokenize("(do " + source + "\n)")
+            do = HyExpression([HySymbol('do')] + tokens)
+            do.start_line = do.end_line = do.start_column = do.end_column = 1
+            do.replace(do)
         except LexException as e:
             if e.source is None:
                 e.source = source
@@ -119,7 +121,7 @@ class HyREPL(code.InteractiveConsole):
                     new_ast = ast.Module(main_ast.body +
                                          [ast.Expr(expr_ast.body)])
                     print(astor.to_source(new_ast))
-            value = hy_eval(tokens[0], self.locals, "__console__",
+            value = hy_eval(do, self.locals, "__console__",
                             ast_callback)
         except HyTypeError as e:
             if e.source is None:
--- a/NEWS
+++ b/NEWS
@@ -22,6 +22,10 @@ Changes from 0.12.1
      returns.
    * `setv` no longer unnecessarily tries to get attributes
 
+   [ Misc. Improvements ]
+   * New contrib module `hy-repr`
+   * Added a command-line option --hy-repr
+
 Changes from 0.12.0
 
    [ Bug Fixes ]
--- a/docs/contrib/hy_repr.rst
+++ b/docs/contrib/hy_repr.rst
@@ -1,0 +1,42 @@
+==================
+Hy representations
+==================
+
+.. versionadded:: 0.13.0
+
+``hy.contrib.hy-repr`` is a module containing a single function.
+To import it, say ``(import [hy.contrib.hy-repr [hy-repr]])``.
+
+.. _hy-repr-fn:
+
+hy-repr
+-------
+
+Usage: ``(hy-repr x)``
+
+This function is Hy's equivalent of Python's built-in ``repr``.
+It returns a string representing the input object in Hy syntax.
+
+.. code-block:: hy
+
+   => (hy-repr [1 2 3])
+   '[1 2 3]'
+   => (repr [1 2 3])
+   '[1, 2, 3]'
+
+If the input object has a method ``__hy-repr__``, it will be called
+instead of doing anything else.
+
+.. code-block:: hy
+
+  => (defclass C [list] [__hy-repr__ (fn [self] "cuddles")])
+  => (hy-repr (C))
+  'cuddles'
+
+When ``hy-repr`` doesn't know how to handle its input, it falls back
+on ``repr``.
+
+Like ``repr`` in Python, ``hy-repr`` can round-trip many kinds of
+values. Round-tripping implies that given an object ``x``,
+``(eval (read-str (hy-repr x)))`` returns ``x``, or at least a value
+that's equal to ``x``.
--- a/docs/contrib/hy_repr.rst
+++ b/docs/contrib/hy_repr.rst
@@ -5,7 +5,13 @@ Hy representations
 .. versionadded:: 0.13.0
 
 ``hy.contrib.hy-repr`` is a module containing a single function.
-To import it, say ``(import [hy.contrib.hy-repr [hy-repr]])``.
+To import it, say::
+
+  (import [hy.contrib.hy-repr [hy-repr]])
+
+To make the Hy REPL use it for output, invoke Hy like so::
+
+  $ hy --repl-output-fn=hy.contrib.hy-repr.hy-repr
 
 .. _hy-repr-fn:
 
--- a/docs/contrib/index.rst
+++ b/docs/contrib/index.rst
@@ -16,3 +16,4 @@ Contents:
    profile
    sequences
    walk
+   hy_repr
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -193,7 +193,6 @@ Hy.  Let's experiment with this in the hy interpreter::
   [1, 2, 3]
   => {"dog" "bark"
   ... "cat" "meow"}
-  ...
   {'dog': 'bark', 'cat': 'meow'}
   => (, 1 2 3)
   (1, 2, 3)
@@ -203,6 +202,20 @@ Hy.  Let's experiment with this in the hy interpreter::
   Fraction(1, 2)
 
 Notice the last two lines: Hy has a fraction literal like Clojure.
+
+If you start Hy like this (a shell alias might be helpful)::
+
+  $ hy --repl-output-fn=hy.contrib.hy-repr.hy-repr
+
+the interactive mode will use :ref:`_hy-repr-fn` instead of Python's
+native ``repr`` function to print out values, so you'll see values in
+Hy syntax rather than Python syntax::
+
+  => [1 2 3]
+  [1 2 3]
+  => {"dog" "bark"
+  ... "cat" "meow"}
+  {"dog" "bark" "cat" "meow"}
 
 If you are familiar with other Lisps, you may be interested that Hy
 supports the Common Lisp method of quoting:
--- a/hy/contrib/hy_repr.hy
+++ b/hy/contrib/hy_repr.hy
@@ -1,0 +1,77 @@
+(import [hy._compat [PY3 str-type bytes-type long-type]])
+(import [hy.models [HyObject HyExpression HySymbol HyKeyword HyInteger HyList HyDict HySet HyString HyBytes]])
+
+(defn hy-repr [obj]
+  (setv seen (set))
+    ; We keep track of objects we've already seen, and avoid
+    ; redisplaying their contents, so a self-referential object
+    ; doesn't send us into an infinite loop.
+  (defn f [x q]
+    ; `x` is the current object being stringified.
+    ; `q` is True if we're inside a single quote, False otherwise.
+    (setv old? (in (id x) seen))
+    (.add seen (id x))
+    (setv t (type x))
+    (defn catted []
+      (if old? "..." (.join " " (list-comp (f it q) [it x]))))
+    (setv prefix "")
+    (if (and (not q) (instance? HyObject x))
+      (setv prefix "'"  q True))
+    (+ prefix (if
+      (hasattr x "__hy_repr__")
+        (.__hy-repr__ x)
+      (is t HyExpression)
+        (if (and x (symbol? (first x)))
+          (if
+            (= (first x) 'quote)
+              (+ "'" (f (second x) True))
+            (= (first x) 'quasiquote)
+              (+ "`" (f (second x) q))
+            (= (first x) 'unquote)
+              (+ "~" (f (second x) q))
+            (= (first x) 'unquote_splice)
+              (+ "~@" (f (second x) q))
+            ; else
+              (+ "(" (catted) ")"))
+          (+ "(" (catted) ")"))
+      (is t tuple)
+        (+ "(," (if x " " "") (catted) ")")
+      (in t [list HyList])
+        (+ "[" (catted) "]")
+      (is t HyDict)
+        (+ "{" (catted) "}")
+      (is t dict)
+        (+
+          "{"
+          (if old? "..." (.join " " (list-comp
+            (+ (f k q) " " (f v q))
+            [[k v] (.items x)])))
+          "}")
+      (in t [set HySet])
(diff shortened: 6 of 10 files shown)


kirbyfan64:
Ehhh...I'm not really a fan of this one...

IMO part of what I love about Hy is that it's just Python, except different. If someone starts using Hy, they likely already know Python. Using the Python syntax in the repr isn't going to be confusing.

More importantly, though, this only half-way works with Python classes. Imagine a Python class that reprs with Python collection syntax. You'd probably end up having to look at something like this:

```
#{MyClass(val={1, 2, 3})
```

which is worse than the current behavior.

In addition:

- The prepending of `(do` is going to make column numbers off. They have to be fixed.
- This changes the behavior of `--spy`: if an exception is thrown, then the AST is never printed. IMO this removes a lot of its usefulness for me...

Both of these could be fixed by separating the parsing and code execution (which would allow you to put `(do` at the AST level).

Yeah...not a big fan... :O

Kodiologist:
Are you saying you would approve this if I made the changes you've requested, or not?

I don't understand your point or your example about "this only half-way works with Python classes". Can you be a bit more explicit?

> The prepending of `(do` is going to make column numbers off.

Can you give an example of how this appears to the user?

gilch:
I'm also 👎 on this. This has come up before #220 #351 #354. Python's `repr`s in general are never going to work in Hy. We can adjust our own libraries, but not the whole Python ecosystem. We could try to auto-convert `repr`s from Python, but I don't think it'll ever be 100%. Often the `repr`s of collection like objects defer to the `repr`s of what they contain. Remember that Hy libraries can also be used from Python. I'd rather have valid Python `repr`s in _all_ cases than some weird hybrid of the two that doesn't work consistently in either Hy or Python. That means we should adjust Hy toward Python, not the other way around.

gilch:
I just thought of a weird compromise, and I'm not sure how I feel about it. If we implemented some kind of Hy-string eval that we could invoke from Python, then we could make valid Python `repr`s that still contain Hy code:
```Hy
=> '(1 2 3)
hy_eval("'(1 2 3)")
```

gilch:
Ah, no. I don't think that's good enough. You'd also need the reverse, Python, eval, since the collection can only represent objects by their Python repr, so:
 ```
=> '(1 2 {3 4})
hy_eval("'(1 2 (eval \"{3: 4}\")")
```
Nested quotes like that could get gnarly quickly. I don't think that idea is workable either.

**Comment by Kodiologist** [comment]
> Python's `repr`s in general are never going to work in Hy.

Yes, that's one reason I think that this change is better than trying to make more Python syntax into valid Hy syntax.

> We could try to auto-convert `repr`s from Python, but I don't think it'll ever be 100%.

Yeah, that sounds like a bad idea.

> Remember that Hy libraries can also be used from Python.

Yes, that's why `__hy-repr__` is different from `__repr__`.

> I'd rather have valid Python `repr`s in *all* cases

We don't already have that, nor is it likely to be possible. For example, at the Python prompt, `x = [1]; y = [2, 3]; y[0] = x; y[1] = x; y` prints `[[1], [1]]` even though the two elements of `y` are the same object. And lambdas get `repr`ed as `<function <lambda> at 0xdeadbeef>` instead of disassembled. `hy-repr`, like `repr`, can only be an approximation.

> That means we should adjust Hy toward Python, not the other way around.

That would seem to require making Python syntax into valid Hy syntax, which we just rejected.

I don't like your compromise because it will add a lot of visual noise to REPL output, which needs to be easily readable to serve its function of convenience.

**Comment by kirbyfan64** [comment]
> Are you saying you would approve this if I made the changes you've requested, or not?

Depends on what everyone else says.

> I don't understand your point or your example about "this only half-way works with Python classes". Can you be a bit more explicit?

### `foo.py`

```python
class Foo:
    def __init__(self):
        self.items = [1, 2, 3]
    def __repr__(self):
        return 'Foo(items=%r)' % self.items
```

### in repl

```
hy 0.12.1+28.g484daaf.dirty using CPython(default) 2.7.6 on Linux
=> (import foo)
=> #{(foo.Foo)}
#{Foo(items=set([1, 2, 3]))}
=> 
```

It's inconsistent. And it'll stay that way unless random library authors add `__hy_repl__` implementations (but they probably won't).

> Can you give an example of how this appears to the user?

```
hy 0.12.1+28.g484daaf.dirty using CPython(default) 2.7.6 on Linux
=> ]  File "<stdin>", line 1, column 5

  (do ])
      ^
LexException: Ran into a RBRACKET where it wasn't expected.



=> 
```

Kodiologist:
> It's inconsistent. And it'll stay that way unless random library authors add `__hy_repl__` implementations (but they probably won't).

It seems that our options are either to allow the inconsistency or to use a generic placeholder syntax for classes with no `__hy_repr__`, which merely names the class. To me the inconsistency seems a better option, since it provides more information.

In your example, the column number and the caret position seem to be correct.

gilch:
>>That means we should adjust Hy toward Python, not the other way around.

>That would seem to require making Python syntax into valid Hy syntax, which we just rejected.

You're not getting my meaning. I meant that things like `'(1 2)` should print out something like `HyExpression((([] + [HyInteger(1)]) + [HyInteger(2)]))`, or `<Hy '(1 2)>` instead of just `'(1 2)`, which makes no sense from Python.

Kodiologist:
`HyExpression((([] + [HyInteger(1)]) + [HyInteger(2)]))` seems like more of a problem than a solution, but `<Hy  '(1 2)>` seems pretty reasonable.

gilch:
Yeah, I only wrote it that way because that's what came out of the AST. That reallycould be simplified in the `repr` to `HyExpression([HyInteger(1), HyInteger(2)])`, or even `hy([hy(1), hy(2)])` if we had a `hy` function that dispatches to the right model based on the data type it receives. The problem with something like `<Hy '(1 2)>` is how to nest them properly. With a dispatch function, it's easy, and it might even work in the Python repl.

Kodiologist:
In any case, I think that what you're proposing is orthogonal to this PR. You're proposing how to make Python representations for Hy models, and this PR is about Hy representations for all sorts of values, including Hy models. The only way to see any of this PR's changes in the Python REPL would be if you did `from hy.core.language import hy_repl; hy_repl(whatever)`.

Kodiologist:
@hylang/core Anybody else want to chime in on this? (Assume I'll be able to fix the bug Ryan mentioned that `--spy` doesn't work when the code raises an exception. I don't anticipate any problems there.) I think it's obviously a big improvement on what we had. But if nobody else likes it, I can submit another PR with just the first of these two commits, adding `hy-repr` but not using it automatically in the REPL.

Kodiologist:
Another idea is to have a command-line option that enables use of `hy-repr` in the REPL. However, making this feature the default seems wise to me because it will make the REPL much less surprising for newbies.

Kodiologist:
@gilch @kirbyfan64 What do you think of using a command-line option for this, as I mentioned above? If you're against the feature being on by default (with an option to disable it), how about off by default (with an option to enable it)?

@tuturto Can you chip in?

kirbyfan64:
I guess I wouldn't mind it if it were disabled by default...

That being said, what if we put a prefix? e.g prefixing the command with something (no clue what yet) would use this repr...or maybe a reader macro?

Kodiologist:
Then you'd have to remember to use the prefix or reader macro or whatever for every command, which again reduces newbie accessibility.

gilch:
> What do you think of using a command-line option for this

> That being said, what if we put a prefix? e.g prefixing the command with something (no clue what yet) would use this repr...or maybe a reader macro?

This is the kind of thing other Lisps would use a dynamic variable for #1089. It would be nice to toggle other REPL features (like --spy) this way.

Still, I do not approve.

Even with a separate `hy-repr` and `repr`, you can still get weird hybrid representations that are neither fully Python nor fully Hy. This is because the `__repr__` implementation of collection and wrapper types often defer to the `repr` of what they contain. The `__hy-repr__` implementations would have to do the same, since `hy-repr` has to default to just `repr` when there is no `__hy-repr__` implementation--which there won't be for the whole Python ecosystem, save for our own little corner.

It's also more work to implement both a `__repr__` and a `__hy-repr__` for every class.

Thus spake @paultag 
>Hy's failures mostly stem from us trying to cover up the fact it's actually Python

We should expect Hy users, even newbies, to know Python, at at least a basic level--just as Clojure(Script) users must understand Java(Script). I do not approve of trying to _hide_ the Python that is in Hy, since this is the root of many of Hy's difficulties. I might even be helpful for newbies if we turned on --spy in the repl by default.

Kodiologist:
> Even with a separate `hy-repr` and `repr`, you can still get weird hybrid representations that are neither fully Python nor fully Hy.

Like I said, there's no reason one couldn't put a generic placeholder in place of an object with no `__hy-repr__`; it just seems that the Python representation is more likely to be useful. Some `__repr__`s themselves aren't Python code, like Pandas data frames.

> I do not approve of trying to *hide* the Python that is in Hy

Spitting out values in a syntax that you can actually use as input is hiding? I mean, you don't think the fact that the *input* syntax isn't Python is hiding, do you?

tuturto:
Long and good discussion. I love seeing people hashing out some hard problem respectfully ❤️ 

I'm having hard time deciding either way though. I would like the idea that I can just copy/paste repl output back to input and have it work. But I'm bit hesitant if it would be confusing to people (old and new to Hy alike) that it sometimes works and sometimes doesn't.

Lets run an experiment: we could bring this in, have it disabled by default, activated by command line switch and see how it behaves in the wild. After some amount of time, we can get back to the matter and see what we learned. At that point we can adjust things or even remove it completely (as if any project would remove a feature, ever) if this turned out to be a bad idea. How does this sound like?

I didn't look into code really. Just got up and had interesting looking notifications from GitHub waiting 😁

Kodiologist:
I don't think that's a bad idea. We can at least mention it in the tutorial to maximize the chance it will actually see use in the wild.

Kodiologist:
Okay, I fixed the `--spy` bug, added a command-line option, added tests, etc. Take a look. Don't squash, since the middle commit is mostly unrelated.

gilch:
> (as if any project would remove a feature, ever)

A serious concern. Once this is in Hy, it will always be easier to just leave it there.

> Like I said, there's no reason one couldn't put a generic placeholder in place of an object with no __hy-repr__; it just seems that the Python representation is more likely to be useful.

I still don't understand how this is supposed to work. Concrete example time. 

Suppose you have some Python objects (`foo`s) from some Python library. They have reprs like `foo(1, duck())`, but no hy-repr. Now you make some Hy objects (`bar`s) that take some `foo`s as arguments. They have reprs like `bar(foo(1, 2), foo(3, bar(foo(4, 5))))`. So far, so good, the `bar` repr defers to the repr of its `foo` arguments, which defer to the `int` and `bar` reprs and the net result is still valid Python, and it will generate an equivalent `bar` object.

But wait, now we have to write a `__hy-repr__`. Now it prints `(bar foo(1, 2) foo(3, bar(foo(4, 5))))`, and that's one of those weird hybrids that's neither Hy nor Python. It's not clear at all to the future user how to parse this. Note that `bar` appears both in the form `(bar ...)` and `bar(...)`. Why? Because `foo` is written in Python and knows nothing about `__hy-repr__`, so of course it has to use `__repr__` from `bar` instead.

So what if we use generic placeholders instead? Now you've got `(bar <foo instance at 0xXXX1> <foo instance at 0xXXX2>)`. The user can totally parse this. But, you've hidden important information. Not good. It's not really valid Hy, but at least you know exactly where the holes are.

Okay, maybe we can auto-wrap the repr with the placeholders. Now you get `(bar <Py foo(1, 2)> <Py foo(3, bar(foo(4, 5)))>)`. It's still not valid Hy, but now we know both where the holes are and what goes in those holes. But look, the Hy repr version only applies to the first layer! We're still getting the form `bar(...)`. What's the point of a Hy repr if we're printing it all off in Python anyway?

Even though this is a contrived example, it is very representative of the common case. It's very typical in Python for the repr to be a constructor call that defers to the reprs of its arguments. There are more Python libraries written in Python than in Hy, so almost all Python objects will have no hy-repr.

Kodiologist:
I've removed lots of features so far in the few months I've been working on Hy, and I have my eyes on more. (Cons cells, your days are numbered!)

Yes, if you want a nested tower of objects to have Hy representations at lower levels, then you need to add Hy representations at higher levels, too. The only way around that would be to hack method lookup so that `repr` calls `__hy-repr__` in certain cases, which to me sounds like much more trouble than it's worth, if Python is even dynamic enough to allow that.

I guess I'm just not frightened enough of Python-in-Hy representations to think we should give up on any attempt at Hy representations. We're providing Hy representations of common classes, and ways for the user to create their own Hy representations. It's not going to work out of the box with every case, but them's the breaks. It still seems a lot better than the status quo, which provides no way to produce Hy representations at all.

gilch:
> (Cons cells, your days are numbered!)

Cons cells are actually pretty useful in macros. I see no reason to remove them. This has also been discussed: #240 

>The only way around that would be to hack method lookup so that repr calls __hy-repr__ in certain cases, which to me sounds like much more trouble than it's worth

Perhaps not the only way. You could also make it standard to write `__repr__`s in Hy that defer to `(.__hy-repr__ self)` in those certain cases, while the Python `__repr__` doesn't have to know about it and just uses `repr(bar)` as normal. Determining those certain cases would be another use for a dynamic variable. I still think it's more trouble than it's worth, but I think that about `hy-repr` altogether.

> I guess I'm just not frightened enough of Python-in-Hy representations to think we should give up on any attempt at Hy representations. We're providing Hy representations of common classes, and ways for the user to create their own Hy representations.

So this is where we differ. I already disagree with providing Hy representations of common classes. They're confusing. I think we should rethink them so they make sense from Python, or at least make it clear what you're looking at without a debugger.

We _should_ give up on any attempt at Hy representations. It's never going to work right, and there is very little to gain. Python reprs are good enough. Any competent Hy user should be able to translate Python reprs into Hy on the fly, in their heads. (This is actually much easier than the reverse, since Hy is much simpler than Python grammatically.)

So what possible use is a Hy repr anyway? It's _not_ easier to understand than the Python repr. Even Python reprs aren't good enough for serialization (and we have pickle for that), so weird hybrid Hy reprs certainly aren't. You can't generally paste a weird hybrid Hy repr it back into the Hy repl, while you actually _can_ with most Python reprs, simply by using Python's `eval()` on it as a string. We could make this easier than it is currently, and perhaps we should, by
* making Hy's model's reprs valid Python.
* making python's `eval()` more available. It seems to be shadowed by Hy's `(eval)`, so you have to get it from builtins.
* adding long bracket strings so we don't have to escape any quotes. #1042.

Kodiologist:
>  Any competent Hy user should be able to translate Python reprs into Hy on the fly, in their heads.

Yes, but I don't have a telepathic link to my computer. I need to delete or add commas, delete colons, change single quotes to double quotes, etc.

Hy representations round-trip when possible, but they can't when somebody didn't write a `__hy-repr__` usable for that purpose, just as in Python with `__repr__`. If you don't think it's worth giving a programming language the ability to write values out in its own literal syntax, because we can't get it to work without configuration every time, then I don't know what to tell you.

gilch:
> Yes, but I don't have a telepathic link to my computer. I need to delete or add commas, delete colons, change single quotes to double quotes, etc.

Or you could just Python eval the string and let the computer worry about that part. This seems a lot more reliable than your approach.

We could even demonstrate this workflow in the Hy tutorial, if you're worried users won't think of it. Unless you're talking about changing what you input instead of what's output, in which case, it's already Hy.

> I don't know what to tell you

I still don't really get your use case. Concrete examples might help.

Kodiologist:
> Or you could just Python eval the string and let the computer worry about that part.

If I go and paste `(py-eval "[1, 2, 3]")` or whatever back into my source file, then I have source code which has Python and Hy intermingled, which is needlessly distracting, in my opinion, when the object in question has a straightforward Hy representation, namely `[1 2 3]`. I would sooner translate the REPL output into Hy with regexes and by hand so my source file is neater.

I think the elephant in the room with my conversations with you is that [I'm more liberal, in a programming sense, than you are](https://plus.google.com/110981030061712822816/posts/KaSKeg4vQtz). That is, I would rather have features that make a programming language more convenient, even if they create traps for the unwary. This is of course the primary reason I'm using Hy rather than Python, because I don't have much patience for Guido van Rossum's idea of what programming-language constructs are safe for me to use. Macros are a good example of a feature that's very powerful but allows you to create completely illegible code with many subtle and nefarious bugs.

kirbyfan64:
...so, as for the use cases, is there a reason someone can't just use `_` to reference the previous calculation? FWIW I've never, ever had to use copy-paste in any half-decent REPL. Hy already has `_` and readline and a cool prompt, and there's an open issue for making the REPL fancier (e.g. bpython/IPython).

Kodiologist:
I don't see how you would use `_` to put the result in a source file.

gilch:
> I think the elephant in the room with my conversations with you is that I'm more liberal, in a programming sense, than you are.

You label yourself a liberal? (extremist?) I think we kind of already knew that, but the article clarifies your stance a little. But, your implication here is that I get the conservative (moderate?) label. I think this is a bit off topic (sorry about that), but I can't just be silent and let that stand, or it'll just create further misunderstanding.

I read the article you linked to. It didn't jive with me at all. The liberal/conservative axis as described is orthogonal to my values (in the programming sense). I looked at the languages example to see where I stand. I don't fit anywhere.

I like the "extremist conservative" Haskell, and the "liberal" Smalltalk for the same reasons: simplicity, power, good tooling, minimal ceremony. They're excellent examples, respectively, of static typing done right, and how good tools and runtime transparency can compensate for not having it. I also think the tables from "hardcore liberal" Lua gives the language a beautiful simplicity. (Of course, "liberal" Python is a better choice than Lua in most cases.) And lists give Lisps in general that beautiful simplicity.

On the other hand, I dislike the "hardcore conservative" Scala, hate "moderate-conservative" C++, and the "hardcore liberal" JavaScript for the same reasons: they're too complex to get anything done. I also really dislike "Moderate-conservative" C#/Java because of the ceremony, and also Go (in the same category) for lack of power: static typing but no generics. The half-baked static typing in these three languages is more trouble than it's worth.

So yeah, I don't fit on your axis. I reject the label of "less liberal than you". Our difference in philosophy is something else, if it's even there.

Kodiologist:
I would characterize that description of your views as moderate. Beauty and simplicity aren't liberal values, and "beautiful simplicity" is particularly illberal, but on the other hand, you find Java and Go too restrictive.

Kodiologist:
@tuturto @kirbyfan64 Is this ready to merge?

gilch:
So, back on topic. @kirbyfan64 already kind of said "no". Expecting a merge is kind of premature. One more "no" from core and I'm closing this. I'm not sure what @tuturto decided, but if you get two "yes"es I'm overruled. Three's a quorum/three's a veto.

>...so, as for the use cases, is there a reason someone can't just use `_` to reference the previous calculation? FWIW I've never, ever had to use copy-paste in any half-decent REPL. Hy already has `_` and readline and a cool prompt, and there's an open issue for making the REPL fancier (e.g. bpython/IPython).

I pretty much agree, but "never" is such a strong word. Maybe @Kodiologist can come up with some convincing concrete examples of why this is useful enough to consider?

I'm still not seeing much use for this. Usually when editing Lisp, you paste forms from your source file _into_ the repl. Emacs has commands for this workflow. Not the other way around.

If you really want to write Lisp to help you edit your code, just use emacs. Or write a Hy macro to edit your code and put _that_ in your source file instead.

Kodiologist:
> Expecting a merge is kind of premature.

@kirbyfan64 said "I guess I wouldn't mind it if it were disabled by default..." and I did what @tuturto proposed. That's why I asked them if it was good to merge.

> If you really want to write Lisp to help you edit your code, just use emacs. Or write a Hy macro to edit your code and put *that* in your source file instead.

I'm not talking about writing code in the source file and then pasting it into the REPL, transforming the code, and pasting the transformed code back into the source file. I use Emacs, so I indeed would do syntactic transformations like that in Emacs Lisp, if I ever wanted them.

I'm taking about doing some kind of computation and then pasting the computation result in a source file. For example, I might find the unique values of a variable and then put them in the source file with an assert to check that no more unique values have crept in once I've gotten a new version of the dataset. I'm a statistician, so I mostly use interactive mode for data analysis. That and debugging Hy itself are all I use interactive mode for, pretty much.

kirbyfan64:
I think I'd rather see if there's at least one other person here who has a really strong opinion on this this first. Right now, my second vote (@tuturto already gave +1) is pretty much what either merges this or puts this on hold, which makes this a *really awkward* situation...

In particular: @paultag do you have any opinion on this?

Kodiologist:
I've done another rebase to add support for quasiquote and unquote and refine the tests a bit.

gilch:
> Lets run an experiment: we could bring this in, have it disabled by default, activated by command line switch and see how it behaves in the wild. After some amount of time, we can get back to the matter and see what we learned. At that point we can adjust things or even remove it completely (as if any project would remove a feature, ever) if this turned out to be a bad idea. How does this sound like?

Here's a better plan. How about @Kodiologist implements this in Hy, and puts it in contrib. That's already where we're putting experimental features. That way it's not so integrated into Hy, and easy to remove. It's also not any kind of default.

How hard is a repl? You read, eval, print, and loop. Easy. You just change the "print" part to use `hy-repr`, if available. You could start it from the normal repl with an import and function call. Or you could start it from the command line using a script that does that for you.

gilch:
> To me all of this seems like needless complexity, because it's hard to see many people wanting to use a REPL output function other than repr or hy-repr, but it's a reasonable compromise.

Needless? Nah. We've talked about making improvements to the repl before. Those efforts could benefit from that kind of modularity. I could imagine uses for hooks in any of the read, eval or print steps. You could add non-Hy commands as shortcuts or to control aspects of the repl itself like IPython has. You could colorize output, or send it to a graphical window or something. Besides `hy-repr` you could create some kind of `graphical-repr` like Racket has. We'd be able to put those types of experiments in contrib also, or even an independent library. Third parties could also try such things.

> Your second paragraph seems to suggest writing an entirely separate implementation of the REPL for the sake of using hy-repr, which seems like a lot of duplicated code. 

A basic repl just isn't that complex, given read/eval/print functions, which we have. If there's additional functionality (readline?) you want to keep then, yeah, adding hooks to our existing repl might be a better approach.

Kodiologist:
Would you approve `hy --repl-output-fn hy.contrib.hy-repr.hy-repr`, then?

tuturto:
> Would you approve ```hy --repl-output-fn hy.contrib.hy-repr.hy-repr```, then?

Personally I like ```--hy-repr``` tons more, but I guess I can live with this. If we add ```--repl-output-fn```, can we still have ```--hy-repr``` shortcut?

Kodiologist:
Wouldn't a shell alias be better for that?

tuturto:
hm.. true. No need for the shortcut.

gilch:
>Would you approve `hy --repl-output-fn hy.contrib.hy-repr.hy-repr`, then?

I would not be at all opposed to modularizing the repl. Once that's done, I wouldn't be particularly opposed to contrib experiments taking advantage of it. That invocation does seem long-winded (though I suppose one can always use aliases or shell scripts for such things).

I wonder what a good interface would be. Perhaps `--read`, `--eval`, and `--print` options would be OK. Or maybe we could just point it to a Hy (or Python) module that exports `Read`, `Eval`, and/or `Print` callables that can override the defaults. Then it would just be 
```
$ hy --repl hy.contrib.hy-repr
```
If you want to mix-and match, say `read` from one module and `print` from another, then you have to write a module that imports each from wherever. Other future repl implementations (GUI? emacs?) could also use the same config module.

gilch:
I can think of another advantage of the config module approach. You could set state in the module through a special repl command in the `Read` callable that swaps out the implementation used by one of the other callables without the need to restart the repl. You could even set it up to import new ones on the fly.

Kodiologist:
Okay. In the spirit of YAGNI, I'll just implement `--repl-output-fn` and move the `hy-repr` function to /contrib for now.

Kodiologist:
@gilch Done.

tuturto:
Also, realized that ```macroexpand``` is really nice now:

```hy
=> (macroexpand '(if (= a 1) "one" (= a 2) "two" "many"))
'(if* (= a 1) "one" (if (= a 2) "two" "many"))
```

Somehow missed this fact while testing earlier.

Kodiologist:
Okay, this should finally be ready to go.

hy/cmdline.py
```diff
@@ -73,25 +75,35 @@ def __call__(self, code=None):
...
         global SIMPLE_TRACEBACKS
         try:
-            tokens = tokenize(source)
-        except PrematureEndOfInput:
-            return True
+            try:
+                tokens = tokenize(source)
+            except PrematureEndOfInput:
+                return True
+            tokens = tokenize("(do " + source + "\n)")
```
kirbyfan64: Last thing, I swear!! :O

Would this be better as:

```python
tokens = tokenize(source)
do = HyExpression([HySymbol('do')] + tokens)
do.start_line = do.end_line = do.start_column = do.end_column = 1
do.replace(do)
```

that way the `(do ...)` doesn't appear in the stack traces/compile errors?

Kodiologist: In the current implementation, `(do ...)` already doesn't appear. (Your example above no longer has `do` in its output.)

Kodiologist: Hmm, but the underline from a `HyMacroExpansionError` seems to be misaligned. I'll take a look at it.

Kodiologist: Done.

kirbyfan64:
@tuturto Do you have any objections since you last approved 8 days ago, or are we good to go?

tuturto:
@kirbyfan64 we're good to go

Kodiologist:
Thanks, folks. I redid the merge without squashing so the changes between commits are differentiated.

PR #1248, opened 2017-03-08
Fix #151 (again!): yield inside with wasn't propagated to Result

### LINKED ISSUE #151: `yield` inside `with` can lead to generation of a spurious `return`
it tried to return it, d'oh.

we need to disable return mangling on generator functions

kirbyfan64:


Kodiologist:
On Python 3, the `return` still isn't suppressed. For example, hy2py compiles

     (defn f []
       (with [(open "README.md")]
         (yield 123)))
     (assert (= (next (f)) 123))

to

    def f():
        with open('README.md'):
            _hy_anon_var_1 = yield 123
        return _hy_anon_var_1
    assert (next(f()) == 123)

Is this on purpose (because Python 3 allows this, whereas Python 2 doesn't), or a bug?

Kodiologist:
Looks like it is on purpose:

            if body.contains_yield and not PY33:
                # Prior to PEP 380 (introduced in Python 3.3)
                # generators may not have a value in a return
                # statement.

In this case, your docstring for `test-context-yield` should say "before Python 3.3".

hy/compiler.py
```diff
@@ -1485,6 +1485,7 @@ def compile_with_expression(self, expr):
                                            optional_vars=thing)]
 
         ret = ctx + the_with
+        ret.contains_yield = ret.contains_yield or body.contains_yield
```
Kodiologist: Although Python has no `or=`, `|=` seems to do the right thing with bools, so this can be written `ret.contains_yield |= body.contains_yield`. You don't have to do that if you think it's too wacky.

kirbyfan64: I'd rather not, since other parts of the Hy source code use the full, more drawn-out `or`.

kirbyfan64:
@Kodiologist Fixed!

vodik:
Thanks!

PR #1253, opened 2017-03-13
F-strings (FIRST VERSION!! DON'T MERGE!!)

kirbyfan64:
- The code still needs some cleanup.
- This requires Astor 0.6, which *hasn't been released yet*. In particular, it depends on https://github.com/berkerpeksag/astor/pull/54.
- All the debug `print`s are still there.
- There aren't any tests yet.
- ...
- **The code is really freaking complicated.**

But, I figured I might as well ask for some review first, right? ;)

### DIFF
--- a/hy/_compat.py
+++ b/hy/_compat.py
@@ -43,6 +43,7 @@ PY3 = sys.version_info[0] >= 3
 PY33 = sys.version_info >= (3, 3)
 PY34 = sys.version_info >= (3, 4)
 PY35 = sys.version_info >= (3, 5)
+PY36 = sys.version_info >= (3, 6)
 
 if PY3:
     str_type = str
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -33,7 +33,7 @@ import ast
 import sys
 import os
 
-import astor.codegen
+import astor
 
 import hy
 
@@ -77,7 +77,7 @@ def print_python_code(_ast):
     # astor cannot handle ast.Interactive, so disguise it as a module
     _ast_for_print = ast.Module()
     _ast_for_print.body = _ast.body
-    print(astor.codegen.to_source(_ast_for_print))
+    print(astor.to_source(_ast_for_print))
 
 
 class HyREPL(code.InteractiveConsole):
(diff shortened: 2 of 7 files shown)


Kodiologist:
Yes, this is complicated. I feel like `FStringParser` could be greatly simplified, but that's easy for me to say since I haven't tried writing it.

I think it's fine to totally disallow `!` and `:` in identifiers inside an fpart, so that `{a!r}` does what you'd probably want.

I'm kind of surprised Python doesn't allow `fb` or `fu` (heh, F U) string prefixes, but if they don't need them, we don't need them, right?

kirbyfan64:
> Yes, this is complicated. I feel like FStringParser could be greatly simplified, but that's easy for me to say since I haven't tried writing it.

That's what I thought until I wrote it...it seems like it should be simple, but there are a lot of corner cases/oddities to handle.

> I think it's fine to totally disallow ! and : in identifiers inside an fpart, so that {a!r} does what you'd probably want.

Only problem there might be if someone uses an f-string in a `gensym`'d macro, since then they couldn't use the `g!xxx` forms...

I've been trying to think of some sort of "middle ground", but I don't have anything thus far.

Kodiologist:
Could we simply allow `!` or `:` to be escaped with a backslash?

We could always change `defmacro!` to use something other than `!` in its special prefix. That's the nice thing about not being 1.0 yet, right?

kirbyfan64:
> Could we simply allow `!` or `:` to be escaped with a backslash?

That's actually a pretty good idea...

Also FWIW, for comparison purposes, CPython's f-string parser is around ~700-800 LOC. I'd say this one isn't too bad!!!

Kodiologist:
True. It helps that we can write our parser in Python instead of C.

Kodiologist:
We can get this into the next release if you finish it by June.

tuturto:
There probably should be a mention of f-strings in the docs somewhere, but I already like where this is headed.

Kodiologist:
This PR seems to be inactive, so I'm closing it. Please reopen it once work resumes.

PR #1260, opened 2017-03-27
Remove copyright and license headers from code

### LINKED ISSUE #1225: Policy for copyright and license notices at the start of each file
[GNU recommends these](https://www.gnu.org/licenses/gpl-howto.html) when you're using the GPL. However, I doubt they're necessary (we have a LICENSE file), and they add a lot of noise to files, especially small files. In any case, if we decide they should stay, they should probably be in every file, and we need a policy for exactly which people and years get credited in each file, whereas right now they're inconsistent.

- Thumbs-up this issue if you think the notices should be removed.
- Thumbs-down this issue if you think they should stay and be used consistently.

@paultag, could you provide your input?

Kodiologist:
Closes #1225. To discuss this topic in general, comment or vote on that issue.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,5 @@
+Copyright (c) 2012, 2013 Paul Tagliamonte <paultag@debian.org>
+
 Permission is hereby granted, free of charge, to any person obtaining a
 copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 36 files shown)


Kodiologist:
Aborting per the votes in #1225.

PR #1261, opened 2017-03-29
Overhaul semantics of binary operators

### LINKED ISSUE #1103: Add more operators to shadow

### LINKED ISSUE #1104: All shadowed operators should have the same arities as their counterparts.
Probably one of them got updated without the other:

``` Hy
=> (+)
0
=> ((lambda [op] (op)) +)
  File "<input>", line 1, in <lambda>
  File "c:\users\me\code\repos\hy/hy/core/shadow.hy", line 27, in +
    (defn + [&rest args]
  File "c:\users\me\code\repos\hy/hy/core/shadow.hy", line 31, in _hy_anon_fn_1
    (raise (TypeError "Need at least 1 argument to add/concatenate"))
TypeError: Need at least 1 argument to add/concatenate
```

We should check them all though.

Kodiologist:
Fixes #1103.

Fixes #1104.


hy/compiler.py
```diff
@@ -1942,42 +1934,65 @@ def compile_maths_expression(self, expression):
         if PY35:
             ops.update({"@": ast.MatMult})
 
-        inv = expression.pop(0)
-        op = ops[inv]
+        op = ops[expression.pop(0)]
+        right_associative = op == ast.Pow
 
+        if right_associative:
+            expression = list(expression[::-1])
```
kirbyfan64: Why is the outer `list` here necessary? `[::-1]` should already be returning a list anyway.

Kodiologist: Fixed.

Kodiologist:
@kirbyfan64, did you forget to approve this or have you simply not finished reading it?

PR #1264, opened 2017-03-30
Add tests for #533

### LINKED ISSUE #533: REPL doesn't print value of catch/do expressions; prints extraneous objects
The version with `def` appears to work fine, but the others don't show the actual return value.

```
=> (def x (try (/ 1 0) (catch [ZeroDivisionError] 4)))      
=> x
4
=> (try (/ 1 0) (catch [ZeroDivisionError] 4))
=> (try (/ 1 0) (catch [ZeroDivisionError] 4 5))
4
=> (try (/ 1 0) (catch [ZeroDivisionError] 4 5 6))
4
5
```

And on another note, why are the other (not last) values being printed?

Kodiologist:
Closes #533.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -21,6 +21,7 @@ Changes from 0.12.1
    * A `yield` inside of a `with` statement will properly suppress implicit
      returns.
    * `setv` no longer unnecessarily tries to get attributes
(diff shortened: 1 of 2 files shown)


Kodiologist:
Since this change only adds tests, I guess it's trivial enough for one review.

PR #1266, opened 2017-04-05
Fix a crash when tokenizing a single quote

### LINKED ISSUE #1252: Single quote crashes hy2py
## x.hy

```
'
```

## `hy2py x.hy`

```
  File "<string>", line 1, in <module>
  File "hy/cmdline.py", line 406, in hy2py_main
    if stdin_text is None
  File "hy/cmdline.py", line 186, in pretty_error
    print(e, file=sys.stderr)
  File "hy/lex/exceptions.py", line 43, in __str__
    source = self.source.split("\n")
AttributeError: 'NoneType' object has no attribute 'split'
```

Kodiologist:
Fixes #1252.

### DIFF
--- a/tests/test_lex.py
+++ b/tests/test_lex.py
@@ -63,6 +63,17 @@ def test_unbalanced_exception():
         pass
 
 
+def test_lex_single_quote_err():
+    "Ensure tokenizing \"' \" throws a LexException that can be stringified"
+    # https://github.com/hylang/hy/issues/1252
+    try:
+        tokenize("' ")
+    except LexException as e:
+        assert "Could not identify the next token" in str(e)
+    else:
+        assert False
+
+
 def test_lex_expression_symbols():
     """ Make sure that expressions produce symbols """
     objs = tokenize("(foo bar)")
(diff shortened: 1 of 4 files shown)


tests/test_lex.py
```diff
@@ -63,6 +63,17 @@ def test_unbalanced_exception():
         pass
 
 
+def test_lex_single_quote_err():
+    "Ensure tokenizing \"' \" throws a LexException that can be stringified"
+    # https://github.com/hylang/hy/issues/1252
+    try:
+        tokenize("' ")
+    except LexException as e:
+        assert str(e)
```
kirbyfan64: `== '...'`? To prevent a weird error message coming up without anyone realizing it...

Kodiologist: The string is currently `"\x1b[33m\x1b[22mLexException: Could not identify the next token.\n\n\x1b[39m\x1b[22m"`. If we check for equality against that, we're testing the use of ANSI escape codes, which doesn't seem right.

kirbyfan64: @Kodiologist ...which is why the `in` operator is also equally as awesome as `==`. ;)

Kodiologist: How would `"LexException: Could not identify the next token." in str(e)` tell us whether `str(e)` contains a weird error message, in addition to the one we expect?

kirbyfan64: @Kodiologist Yeah, but it's better than nothing!

Kodiologist: Okay.

Issue #1268, opened 2017-04-08
`require` doesn't compile to anything, which interacts badly with bytecode


Kodiologist:
Here's an example with Python 2. If you have a Hy program `foo.hy` like this:

    (require [hy.extra.anaphoric [ap-if]])
    (print (eval '(ap-if (+ "a" "b") (+ it "c"))))

then `hy foo.hy` works fine and prints `"abc"`. But if you say `hyc foo.hy` and then `python foo.pyc`, you get `NameError: name 'ap_if' is not defined`.

Here's why this happens. Currently, `require` doesn't contribute anything to the AST: its compilation method returns an empty `Result`. So `foo.pyc` ends up containing the same thing as if you compiled just `(print (eval '(ap-if (+ "a" "b") (+ it "c"))))`. The only reason `hy foo.hy` works is that the macros that Hy loaded during compile-time are still hanging around during run-time when we get to the `eval`.

The most obvious solution to this is to make `require` forms produce a piece of AST that does the same thing at run-time that the form does at compile time, namely, load the requested macros. The annoying thing about this approach is that it would give byte-compiled modules a dependency that they won't need unless they use a macro inside an `eval`, which is not something that most programs will do. That makes me think we should do the opposite, and prevent `(eval '(some-imported-macro ...))` from working at all. Then, if you want to use an imported macro in an `eval`, you'll need a corresponding `require` form in there, too.

kirbyfan64:
As a side note, I'm starting the get the vibe that we need some tests specifically for hyc. This is around the third issue we've had recently involving something that either doesn't work or is just plain broken.

Kodiologist:
We have some in `test_bin.hy`, but they're not much. However, I'm soon going to submit a PR to make Hy read and write bytecode automatically, which will make bytecode issues much more visible in general.

brandonwillard:
I have a mostly work implementation of something like this here: https://github.com/brandonwillard/hy/blob/macro-changes/hy/compiler.py#L1197

I'll put in a PR when I get all the tests passing again (only a few to go, though).

PR #1269, opened 2017-04-08
Automatically read and write bytecode

Kodiologist:
Importing or executing a Hy file now loads the byte-compiled version if it exists and is up to date, and if not, the source is byte-compiled after it's parsed.

This change can speed up Hy a lot. Here are some examples comparing run times of the current master (491b474) to this commit, on my laptop with Python 3.6:

- `nosetests --exclude='test_bin'` goes from 3.8 s to 0.7 s (a 5-fold speedup)
- `hy -c '(print "hello world")` goes from 0.47 s to 0.20 s (a 2-fold speedup)
- Rogue TV's startup goes from 3.6 s to 0.4 s (a 9-fold speedup)

Accompanying changes include:

- `setup.py` now creates and installs bytecode for `hy.core`, `hy.contrib`, and `hy.extra`.
- The `hyc` command under Python 3 now creates bytecode in `__pycache__`, as usual for Python 3, instead of putting the `.pyc` right next to the source file like Python 2 does.

I've removed a test of `hy.extra.anaphoric.a-if` that triggers #1268 when the test file is byte-compiled and then hits some weird `macroexpand` bug or something when I try to work around that—Nose crashes when trying to produce an error message, and I can't seem to replicate the bug without Nose.


Kodiologist:
Dangit, still failing on CPython 3.3 and Pypy 2. I'll take a look at them later.

Kodiologist:
Well, there's not much I can do about Travis's complaint, since I was able to coax PyPy into no longer creating that error for [a nonexistent test](https://github.com/nose-devs/nose/issues/233) on my own machine, with Ubuntu 16.10, but Travis's Python VM runs Ubuntu 12.04. (As has previously been noted, we need to move away from the dated Nose.) I have no real way to debug this. Unless anybody happens to have a machine running Ubuntu 12.04 sitting around, I'll have to disable testing PyPy compilation on Travis, I guess by checking the environment variable `TRAVIS`.

kirbyfan64:
@Kodiologist I've been able to reproduce the same failures using PyPy 2.5.0 (binaries are [here](https://bitbucket.org/pypy/pypy/downloads/pypy-2.5.0-linux64.tar.bz2)). No Ubuntu 12 needed!

kirbyfan64:
(FWIW, I haven't reviewed this yet, but so far I'm super glad this has happened; IME the main slowdowns with Hy were always due to the compilation times anyway.)

hy/importer.py
```diff
@@ -68,34 +71,77 @@ def import_file_to_ast(fpath, module_name):
...
+        if bytecode_mtime >= source_mtime:
+            # It's a cache hit. Load the byte-compiled version.
+            if PY3:
+                # As of Python 3.6, imp.load_compiled still exists, but it's
+                # deprecated. So let's use SourcelessFileLoader instead.
+                from importlib.machinery import SourcelessFileLoader
+                module = (SourcelessFileLoader(module_name, bytecode_path).
+                          load_module(module_name))
+            else:
+                module = imp.load_compiled(module_name, bytecode_path)
```
kirbyfan64: This part should be encased in a `try...except` and treat an `ImportError` as a cache miss. If you switch from one version of Python 2 to another, this will break due to a magic number mismatch, and you have to run `find . -name '*.pyc' -exec rm {} ';'` as a workaround.

hy/importer.py
```diff
@@ -68,34 +71,77 @@ def import_file_to_ast(fpath, module_name):
...
+        # It's a cache miss, so load from source.
+        sys.modules[module_name] = None
+        try:
+            _ast = import_file_to_ast(fpath, module_name)
+            module = imp.new_module(module_name)
+            module.__file__ = fpath
+            code = ast_compile(_ast, fpath, "exec")
+            try:
+                write_code_as_pyc(fpath, code)
+            except IOError:
```
kirbyfan64: This can also fail with `OSError` in some (admittedly rare) cases.

hy/importer.py
```diff
@@ -68,34 +71,77 @@ def import_file_to_ast(fpath, module_name):
...
+                # We failed to save the bytecode, probably because of a
+                # permissions issue. The user only asked to import the file, so
+                # don't bug them about it.
+                pass
+            eval(code, module.__dict__)
+        except (HyTypeError, LexException) as e:
+            if e.source is None:
+                with open(fpath, 'rt') as fp:
```
kirbyfan64: Maybe also wrap this in a `try...except`? In the theoretical case where a source file disappears before the error is shown (don't laugh; I've accidentally un-plugged my external HDD and run into trouble before), this would outright crash, while maybe just printing a warning or just doing nothing would be a bit better.

Kodiologist:
> @Kodiologist I've been able to reproduce the same failures using PyPy 2.5.0 (binaries are here).

Thanks. I see that these binaries don't have the PEP 3147 backport. Perhaps the reason for the Travis failures is that Ubuntu 12.04 (and possibly also Debian at the time) hadn't yet included the backport. All this suggests that I should be setting `DEBIAN_PYPY_2` only if we're using a sufficiently recent build of PyPy. I have `/usr/share/doc/pypy/changelog.Debian.gz` on my machine, but it doesn't seem to indicate at what version the backport was added. So I have no idea what check I should use.

hy/importer.py
```diff
@@ -235,3 +251,47 @@ def find_module(self, fullname, path=None):
...
 sys.meta_path.insert(0, MetaImporter())
 sys.path.insert(0, "")
+
+
+def is_package(module_name):
+    mpath = os.path.join(*module_name.split("."))
+    for path in map(os.path.abspath, sys.path):
+        if os.path.exists(os.path.join(path, mpath, "__init__.hy")):
+            return True
+    return False
```
kirbyfan64: What happens here if you have a package containing Hy files but with an `__init__.py`?

hy/importer.py
```diff
@@ -235,3 +251,47 @@ def find_module(self, fullname, path=None):
...
+    d, f = os.path.split(source_path)
+    return os.path.join(d, re.sub(r"(?:\.[^.]+)?\Z", ".pyc", f))
+
+
+DEBIAN_PYPY_2 = (not PY3 and platform.python_implementation() == 'PyPy'
+                 and 'dfsg' in platform.python_build()[0])
+if PY3:
+    CACHE_TAG = sys.implementation.cache_tag
+elif DEBIAN_PYPY_2:
+    # Trust me, if I knew a less stupid way to guess CACHE_TAG here, I'd do it.
```
kirbyfan64: Won't this outright break if you run Python with `-B` or `PYTHONDONTWRITEBYTECODE=yes`?

kirbyfan64: Actually, nvm; I just saw you byte-compiled everything in `setup.py`. Still, maybe there should be a double-check to make sure the bytecode file's there, with a nicer error?

Kodiologist:
It turns out although Python 3.3 and the backported PyPy don't have `importlib.util.cache_from_source`, they do have `imp.cache_from_source`. That should make things a lot easier.

hy/importer.py
Kodiologist: Would it be better to check the magic number explicitly, the way I check the mtime?

hy/importer.py
Kodiologist: This is the same as master. Let's not futz with it in this PR.

hy/importer.py
Kodiologist: Evidently, everything still works, since `hy.core` is such a package. I'm not sure why, but I don't really know what `module.__package__` does, anyway.

hy/importer.py
Kodiologist: The code is supposed to function fine when bytecode is missing or can't be written. I've tested that in `test_bin.py`.

kirbyfan64:
@Kodiologist Hmm...I think maybe I can try it on a Ubuntu Docker image...Let's see...

Kodiologist:
I was able to install the Ubuntu 12.04 package for PyPy on my own system, but I can't create a virtualenv because of a crash from, suitably enough, `imp.cache_from_source` (`ValueError: Not a .py file: /tmp/pip-build-zd1NUk/six/six-1.10.0.dist-info/DESCRIPTION.rst`).

.travis.yml
```diff
@@ -3,23 +3,15 @@ language: python
...
-      env: TOXENV=py34
     - python: 3.5
-      env: TOXENV=py35
     - python: 3.6
-      env: TOXENV=py36
-    - env: TOKENV=flake8
-install: pip install tox
-script: tox
+install: pip install --allow-all-external -e .
+script: nosetests
```
kirbyfan64: I don't really envy you right now... ;)

(FWIW maybe we should think about using [travis-tox](https://github.com/ryanhiebert/tox-travis))?

Kodiologist: Either that or use Nose directly as I did here. Tox is kind of overkill on Travis because the whole point of it is to juggle multiple environments and versions of Python, which Travis already does.

Kodiologist:
Finally passing. To summarize what I learned, Travis does *not* use Debian PyPy. It uses a specific PyPy nightly, which I was able to get from [here](http://buildbot.pypy.org/nightly/release-2.5.x/). I could then replicate the bug on my own machine, but I didn't find a way to really fix it. Instead, I've just changed `hy.importer` not to compile `__init.hy__` files when running PyPy under Nose. (It works fine when you don't use Nose.)

hy/importer.py
Kodiologist: @kirbyfan64 You still get notified when I reply to one of your review comments without @-mentioning you, right?

kirbyfan64: Uhh...yeah...I guess I got distracted by another comment?

Yeah, checking the magic number would probably work.

Kodiologist: Okay, all should be good now.

NEWS
```diff
@@ -4,6 +4,11 @@ Changes from 0.12.1
    * `let` has been removed. Python's scoping rules do not make a proper
      implementation of it possible. Use `setv` instead.
    * `lambda` has been removed, but `fn` now does exactly what `lambda` did
+   * Importing or executing a Hy file loads the byte-compiled version if it
+     exists and is up to date. This brings big speed boosts, even for
+     one-liners, because Hy no longer needs to recompile its standard library
+     for every startup.
+   * Importing or executing a Hy file automatically byte-compiles it
```
kirbyfan64: Why not just put these both under the same point?

> Importing or executing a Hy file will automatically byte-compile it and use the compiled version ...

or something?

PR #1272, opened 2017-04-13
raw-input is undefined, replace with input. discussed here: https://g…

caustin-usgs:
…ithub.com/hylang/hy/issues/259

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -143,8 +143,8 @@ Now let's look at the equivalent Hy program:
 
    (defn simple-conversation []
       (print "Hello!  I'd like to get to know you.  Tell me about yourself!")
-      (setv name (raw-input "What is your name? "))
-      (setv age (raw-input "What is your age? "))
+      (setv name (input "What is your name? "))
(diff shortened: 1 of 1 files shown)


Kodiologist:
`raw_input` is a Python 2 builtin and hence available in Hy running under Python 2. Changing the tutorial to use Python 3 would make sense, so I'd encourage you to do that if you're interested, but we probably shouldn't switch versions mid-tutorial without warning.

(Actually, in Hy under Python 2, `input` is aliased to `raw_input`, but this is pretty confusing and hence probably not a good idea to spring on the reader of the tutorial at this point.)

caustin-usgs:
@Kodiologist thanks for the clarification, makes sense

caustin-usgs:
will open a new PR for the switch to python 3 syntax

freezas:
@caustin-usgs Have you already made a branch to Python 3 syntax? I'm willing to help a hand or start a new branch if you haven't got the chance to work on this.

freezas:
No problem ;-)

PR #1273, opened 2017-04-13
Drop support for untested versions of Python

### LINKED ISSUE #1271: Drop support for Python 2.6
It was released in 2008, so it's almost 10 years old. Also, we don't test it.

Kodiologist:
Closes #1271.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,7 @@
 Changes from 0.12.1
 
    [ Language Changes ]
+   * Pythons 2.6, 3.0, 3.1, and 3.2 are no longer supported
    * `let` has been removed. Python's scoping rules do not make a proper
      implementation of it possible. Use `setv` instead.
    * `lambda` has been removed, but `fn` now does exactly what `lambda` did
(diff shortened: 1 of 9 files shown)


Kodiologist:
Misclick there; sorry.

kirbyfan64:
I don't know...I'm still against dropping all 3.x releases until the release reaches EOL. FWIW the code hasn't even been simplified or anything. IMO Python 3.2 and PyPy3 *should* be tested (not sure why it wasn't already that way?)...though I'm pinging @paultag since he might feel differently.

2.6 can definitely die in a fire, though. No one's arguing with that.

Kodiologist:
> I don't know...I'm still against dropping all 3.x releases until the release reaches EOL.

What's your definition of end-of-life? I don't see any such announcements regarding 3.2. The first release of the 3.2.x line was in 2011 and the most recent was in 2014. There was supposed to be a 3.2.7 release in February 2016, but it doesn't seem to have ever happened.

kirbyfan64:
HAHAHA I'm an idiot and 3.2 *is* end-of-life: https://docs.python.org/devguide/#status-of-python-branches

Kodiologist:
That page is handy, thanks.

Issue #72, opened 2013-04-07
make nosetests work without the import * hack


paultag:
it'd be really useful to have hytests

kirbyfan64:
Doesn't matter anymore: #1279

PR #1279, opened 2017-04-18
Migrate from Nose to pytest

### LINKED ISSUE #626: ship a py.test plugin for hy
see if this is doable

Kodiologist:
Closes #626. (We're not providing a plugin, but the conftest.py shows how to collect Hy tests with pytest.)


Kodiologist:
Blimey, it actually works!

Kodiologist:
The hack in conftest.py can be removed once pytest-dev/pytest#2374 is in a released version of pytest.

Kodiologist:
Also closes #72, since we're abandoning Nose.

tests/native_tests/core.hy
```diff
@@ -41,7 +41,7 @@
          (except [TypeError] True)
          (else (assert False)))))
 
-(defn test-coll? []
+(defn test-collq []
```
kirbyfan64: *sigh* We *really* need to work on our mangling...

FWIW pytest lets you [override the test prefix](https://docs.pytest.org/en/latest/customize.html#confval-python_functions). Maybe it'd be worth it to do that with something like:

```ini
[pytest]
# *insert comment lamenting state of world, etc.*
python_function = test_* is_test_*
```

Although it's a hack:

- It's not much more hacky than a questionable (haha!) `q` suffix on every single test function.
- It's a bit less weird to read. (random person looking through code: *Why are there random `q`s everywhere?*)
- The test names will make more sense. This is testing `coll?`, not `collq`...

Kodiologist: Okay, changed.

Kodiologist:
Thanks for reviewing and merrging. Please don't squash when a PR is split into logically distinct commits. I've going to redo the merge.

Issue #1016, opened 2016-01-21
Exception when importing hy.core.macros from Python


timmartin:
If I open a Python 3.4 REPL and type `import hy.core.macros`, it throws an exception:

```
>>> import hy.core.macros
Traceback (most recent call last):
  File "/home/timm/dev/hy/hy/macros.py", line 184, in macroexpand_1
    obj = wrap_value(m(*ntree[1:], **opts))
  File "<eval_body>", line 1, in _hy_anon_fn_2
NameError: name 'is_empty' is not defined

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1191, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1161, in _load_backward_compatible
  File "/home/timm/dev/hy/hy/importer.py", line 188, in load_module
    self.path)
  File "/home/timm/dev/hy/hy/importer.py", line 76, in import_file_to_module
    _ast = import_file_to_ast(fpath, module_name)
  File "/home/timm/dev/hy/hy/importer.py", line 68, in import_file_to_ast
    return hy_compile(import_file_to_hst(fpath), module_name)
  File "/home/timm/dev/hy/hy/compiler.py", line 2596, in hy_compile
    result = compiler.compile(tree)
  File "/home/timm/dev/hy/hy/compiler.py", line 433, in compile
    ret = self.compile_atom(_type, tree)
  File "/home/timm/dev/hy/hy/compiler.py", line 425, in compile_atom
    ret = _compile_table[atom_type](self, atom)
  File "/home/timm/dev/hy/hy/compiler.py", line 627, in compile_raw_list
    ret = self._compile_branch(entries)
  File "/home/timm/dev/hy/hy/compiler.py", line 489, in _compile_branch
    return _branch(self.compile(expr) for expr in exprs)
  File "/home/timm/dev/hy/hy/compiler.py", line 316, in _branch
    results = list(results)
  File "/home/timm/dev/hy/hy/compiler.py", line 489, in <genexpr>
    return _branch(self.compile(expr) for expr in exprs)
  File "/home/timm/dev/hy/hy/compiler.py", line 433, in compile
    ret = self.compile_atom(_type, tree)
[...]

PR #1280, opened 2017-04-19
Fix issue when installing on Python 3.4

jacktasia:
When doing `pip install git+https://github.com/hylang/hy.git` on Python 3.4 I get the exact same error as mentioned in #1016. This would not happen for me when installing on Python 3.5, which I assume is because of an underlying change in how [os.walk](https://docs.python.org/3/library/os.html#os.walk) works in 3.5:

> Changed in version 3.5: This function now calls os.scandir() instead of os.listdir(), making it faster by reducing the number of calls to os.stat().

...and that's somehow changing the order of import modules. Although that's purely speculation and honestly I don't know much about the differences between Python 3.4 and 3.5

Anyway, with this PR I was able to successfully install on Python 3.4 (`pip install git+https://github.com/jacktasia/hy.git@fix-install-3.4#egg=hy`)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,11 +53,20 @@ class Install(install):
     def run(self):
         # Import each Hy module to ensure it's compiled.
         import os, importlib
+        postpone_filenames = ('macros.hy')
+        import_modules_later = []
         for dirpath, _, filenames in os.walk("hy"):
             for filename in filenames:
                 if filename.endswith(".hy"):
-                    importlib.import_module(
-                        dirpath.replace("/", ".") + "." + filename[:-len(".hy")])
+                    module_to_import = dirpath.replace(
+                        "/", ".") + "." + filename[:-len(".hy")]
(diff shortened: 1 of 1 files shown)


Kodiologist:
Hmm. I wonder why the existing code works with Python 3.4 on Travis, but not for you. Did you try installing in a clean virtualenv?

jacktasia:
> Did you try installing in a clean virtualenv?

Nope. Although, I just did and it worked fine from `master`. That said, my use case, which is probably rare is using a fresh python install from an AWS AMI and installing on the system python. So what's interesting is if I take a fresh instance it won't work, but it will work if I copy what's in your travis/tox config:

```
git clone https://github.com/hylang/hy.git
cd hy/
pip install --allow-all-external -e .
```

Anyway, it's probably on my end so I'll just close this and let google index just in case someone runs into a similar problem they can see some options.

PR #1284, opened 2017-04-24
In setup.py, handle backslash-separated paths

Kodiologist:


kirbyfan64:
FWIW can't you just call `os.path.normalize` first?

Kodiologist:
No, because that changes forward slashes to backslashes rather than the reverse. Likewise the Python 2 equivalent,`os.path.normpath`.

PR #1286, opened 2017-04-27
Use license headers consistently

### LINKED ISSUE #1225: Policy for copyright and license notices at the start of each file

Kodiologist:
Closes #1225.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -28,6 +28,18 @@ Pull requests are great! We love them; here is a quick guide:
   documentation fixes or code cleanup), you can directly open the PR without
   first opening a corresponding issue.
 
+- Every Python or Hy file in the source tree that is potentially copyrightable
+  should have the following header (but with ``;;`` in place of ``#`` for Hy
+  files)::
+
+      # Copyright [current year] the authors.
+      # This file is part of Hy, which is free software licensed under the Expat
+      # license. See the LICENSE.
+
+  As a rule of thumb, a file can be considered potentially copyrightable if it
+  includes at least 10 lines that contain something other than comments or
+  whitespace. If in doubt, include the header.
+
 - Before you submit a PR, please run the tests and check your code
   against the style guide. You can do both of these things at once::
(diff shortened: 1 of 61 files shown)


Kodiologist:
@tuturto @kirbyfan64 Let's get this moving if possible, because it touches so many files that merge conflicts are inevitable with any other PRs.

tuturto:
Yeah, I'll try to tackle this soon. Having plenty of things on my plate currently. In the meantime, @paultag any comments?

theanalyst:
@paultag are we actually supposed to remove the copyright of the previous years as well? (ofc, IANAL)

Kodiologist:
@kirbyfan64 Poke.

gilch:
Disclaimer: I am not a lawyer. I probably don't know what I'm talking about. Neither are the other authors. This seems to be the reason we're not getting much participation on this issue. It's holding us up though.

Copyright declarations can help avoid confusion about ownership, but are not actually legally required in the United States at the moment. The copyright automatically applies as soon as the work is written down. I don't know about the laws elsewhere.

I don't think it matters if an individual file is conceivably copyrightable. The copyright and license applies to Hy as a whole, rather than its individual files. And the license itself doesn't appear to require headers at all.

The Git repository tracks authorship and dates well enough for legal purposes, but the headers should not contradict this (and they do at the moment). I'm inclined to approve this pull as an improvement to the status quo. If anyone has strong objections, afterwards make another pull I guess.

Kodiologist:
Thanks!

PR #1293, opened 2017-05-16
Get the documentation building again

### LINKED ISSUE #1290: Get the documentation building again
Also:
- Update links in the README etc. to the documentation versions
- Get a real version number into the built documentation

Kodiologist:
Fixes #1290.


Kodiologist:
You can preview the result at http://docs.hylang.org/en/docs-update/

docs/conf.py
```diff
@@ -1,47 +1,29 @@
...
-#needs_sphinx = '1.0'
-
-# Add any Sphinx extension module names here, as strings. They can be extensions
-# coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ['sphinx.ext.todo']
+# Try to get a version number.
+# N.B. Similar code appears in setup.py.
+# We don't use the dirty flag here because Read the Docs may dirty
+# the repository while generating the documentation.
+VERSIONFILE='../hy/version.py'
```
kirbyfan64: Should be something like:

```python
import os
version_file = os.path.join(os.path.dirname(__file__), 'hy', 'version.py')
```

instead, to support running Sphinx from a different directory.

docs/conf.py
```diff
@@ -1,47 +1,29 @@
...
+try:
+    hy_version = (subprocess.check_output
+        (["git", "describe", "--tags"])
+        .decode('ASCII').strip()
+        .replace('-', '+', 1).replace('-', '.'))
```
kirbyfan64: Bikeshedding, but the indentation here is causing physical pain. :O

setup.py
```diff
@@ -14,6 +14,8 @@
 
 os.chdir(os.path.split(os.path.abspath(__file__))[0])
 
+# Try to get and update the version.
+# N.B. Similar code appears in docs/conf.py.
```
kirbyfan64: In `conf.py` you're allowed to import the main modules, so maybe this could just be shoved in `hy/version_gen.py` or similar? (Not a blocking issue by any means, but still something to think about.)

kirbyfan64: Or even just a top-level `version_gen.py`, since it would only be used by `setup.py` and `conf.py` anyway.

docs/conf.py
Kodiologist: But if you ran Sphinx from a different directory, the Makefile would break.

docs/conf.py
Kodiologist: It's legal Python with indents of spaces in a multiple of 4. How would you indent it?

kirbyfan64: I was mostly just referring to that the indentation on the last three lines should line up with the `s` on the first line.

docs/conf.py
kirbyfan64: I was referring to running `sphinx-build` manually. Admittedly something odd to do, but there isn't really any reason why it shouldn't work.

Kodiologist: Before this PR, did running `sphinx-build` manually from a different directory work? What's the exact command?

setup.py
Kodiologist: Done.

docs/conf.py
Kodiologist: Done.

Kodiologist:
Come to think of it, this is a documentation change, so all I need is Ryan's ACK to merge it. So I will.

PR #1294, opened 2017-05-16
Require capitalizing NaN and Inf like so

### LINKED ISSUE #1288: Any capitalization of "nan" or "inf" is interpreted as a float, not a symbol
For example:

    => (setv Nan 5)

      (setv Nan 5)
            ^--^
    HyTypeError: b"Can't assign or delete a HyFloat"

The fact that Hy accepts literals for floating-point infinity and NaN, albeit unintentional, is a nice feature that we might as well keep. But, Hy ought to recognize only one capitalization.

Kodiologist:
Fixed #1288.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -8,6 +8,7 @@ Changes from 0.13.0
    * The compiler now automatically promotes values to Hy model objects
      as necessary, so you can write ``(eval `(+ 1 ~n))`` instead of
      ``(eval `(+ 1 ~(HyInteger n)))``
+   * Literal `Inf`s and `NaN`s must now be capitalized like that
 
    [ Bug Fixes ]
    * Numeric literals are no longer parsed as symbols when followed by a dot
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -52,6 +52,9 @@ digits.
 
     (print 10,000,000,000 10_000_000_000)
 
+Unlike Python, Hy provides literal forms for NaN and infinity: `NaN`, `Inf`,
+and `-Inf`.
+
 string literals
 ---------------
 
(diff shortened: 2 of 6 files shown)


gilch:
I would also prefer one canonical capitalization, but I'm more inclined to use all-lowercase `nan` and `inf`, since that is the Python repr.
```
>>> float('NaN')
nan
>>> float('Inf')
inf
```

Kodiologist:
I noticed that, but I figured that starting with a capital letter is better for symmetry with `False`, `True`, and `None`, and the meaning of `NaN` is clearer when written in CamelCase.

Also, I'm not 100% sure that the Python `repr` is platform-independent.

tuturto:
For what it's worth, I get all lowercase inf and nan when doing ```float('NaN')``` on my Windows 10 machine or Linux machine, on both Python 2.x and 3.x. That said, I kind of like ```NaN``` nicer looking than ```nan```.

I would be inclined to choose ```NaN``` and ```Inf```` over ```nan``` and ```inf````, but I don't feel about this particularly strongly.

gilch:
I'm not sure what to think. I guess my feelings are not that strong either. @kirbyfan64, what say you?

kirbyfan64:
My only question is whether or not it's actually *worthwhile* to add so much extra code just to check the inf and NaN capitalization...

Kodiologist:
I'm all for concise implementation, but surely you don't mean to suggest that we should interpret all of `Nan`, `nan`, `nAN`, etc. as the special floating-point value, in defiance of the case-sensitivity of the rest of the language, do you? Surely a bunch of code is a reasonable price to pay for avoiding that.

Kodiologist:
C'mon, guys, nothing will happen if you don't make any decisions.

kirbyfan64:
@Kodiologist can you rebase this?

PR #1296, opened 2017-05-18
Don't parse 5.attr or :foo.attr as symbols

### LINKED ISSUE #1143: Attributes of literals
TL;DR: the way attributes, integers, and identifiers are parsed is a hot, hot mess.

So, right now, parsing a number or symbol looks like this:

- Lexer sees token and tags it as an identifier.
- Parser grabs token, tries to covert it to an integer, float, or fraction, and, failing that, compiles it as a symbol.

The second step means that there are a lot of corner cases that are currently quite bizarre:

```
ryan@DevPC-LX /media/ryan/stuff/hy attrerr $ ./run --spy
=> (defn f [&rest args &kwargs kw])
def f(*args, **kw):
    pass
=> 

=> 2.2  ; This is a float.
2.2
2.2
=> a.2  ; This is interpreted as getting the attribute '2' on 'a'.
a.2
NameError: name 'a' is not defined
=> 2.a  ; This is interpreted as getting the attribute 'a' on an identifier '2'.
2.a
NameError: name '2' is not defined
=> a/b  ; This is parsed as a whole identifier 'a/b'...
a/b
NameError: name 'a/b' is not defined
=> 1/2  ; ...but this is a fraction.
from hy.core.language import fraction
fraction(1L, 2L)
Fraction(1, 2)
=> (f :x? 1)  ; Here, the ? isn't converted to is_ like usual.
f(x?=1L)
=> (f :*x* 1)  ; The two stars stay uncoverted.
f(*x*=1L)
=> 
ryan@DevPC-LX /media/ryan/stuff/hy attrerr $ 
```

Kodiologist:
Fixes #1143.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ tox: venv
 	tox
 
 flake:
-	flake8 hy tests --ignore=E121,E123,E126,E226,E24,E704,W503,E305
+	flake8 hy tests --ignore=E121,E123,E126,E226,E24,E704,W503,E302,E305,E701
 
 clear:
 	clear
--- a/hy/lex/parser.py
+++ b/hy/lex/parser.py
@@ -288,6 +288,24 @@ def t_partial_string(p):
 def t_identifier(p):
     obj = p[0].value
 
+    val = symbol_like(obj)
+    if val is not None:
+        return val
+
+    if "." in obj and symbol_like(obj.split(".", 1)[0]) is not None:
+        # E.g., `5.attr` or `:foo.attr`
+        raise LexException(
+            'Cannot access attribute on anything other than a name (in '
+            'order to get attributes of expressions, use '
+            '`(. <expression> <attr>)` or `(.<attr> <expression>)`)',
+            p[0].source_pos.lineno, p[0].source_pos.colno)
+
+    return HySymbol(".".join(hy_symbol_mangle(x) for x in obj.split(".")))
+
+
+def symbol_like(obj):
+    "Try to interpret `obj` as a number or keyword."
+
     try:
         return HyInteger(obj)
     except ValueError:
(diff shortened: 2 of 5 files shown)


hy/lex/parser.py
```diff
@@ -287,6 +287,24 @@ def t_partial_string(p):
...
+    val = symbol_like(obj)
+    if val is not None:
+        return val
+
+    if "." in obj and symbol_like(obj.split(".", 1)[0]) is not None:
+        # E.g., `5.attr` or `:foo.attr`
+        raise LexException(
+            'Cannot access attribute on anything other than a name (in '
+            'order to get attributes of expressions, use '
+            '`(. <expression><attr>)` or `(.<attr> <expression>)`)',
```
kirbyfan64: Missing space in `expression><attr`.

Makefile
```diff
@@ -46,7 +46,7 @@ tox: venv
 	tox
 
 flake:
-	flake8 hy tests --ignore=E121,E123,E126,E226,E24,E704,W503,E305
+	flake8 hy tests --ignore=E121,E123,E126,E226,E24,E704,W503,E302,E305,E701
```
kirbyfan64: If these new error ignores are solely for the single-line `def`s and close-together `with`s in the tests, maybe it'd be worth it running `flake8` on just that file separately, or using `# noqa` comments (which is what I think would be better).

hy/lex/parser.py
Kodiologist: Fixed.

Makefile
Kodiologist: That would be a huge swarm of `noqa` comments, but besides, I don't think these style rules are useful. I've been bitten by the newline one, in particular, dozens of times in writing Hy patches, and in each case the organization of the code only became less clear when I added more space, because I was using the lack of space to visually group related constructs. Likewise I don't see the value in banning `with ...: ...` or `if ...: ...` or whatever. I agree that in *most* cases you want the newline after the colon, but sometimes you don't.

Kodiologist:
@kirbyfan64 Is this good to merge?

Kodiologist:
@kirbyfan64 Poke.

kirbyfan64:
@Kodiologist FYI this has conflicts.

Kodiologist:
Yep, I'll take care of it and merge. Thanks.

PR #1297, opened 2017-05-18
Improvements to `try`

### LINKED ISSUE #798: "else" does not return a value from a try form
I would expect this:

```
(defn x []
  (try
    (some-call)
    (catch [e Exception] 1)
    (else 2)))
```

To compile to this:

```
def x():
    try:
        _hy_anon_var_1 = some_call()
    except Exception as e:
        _hy_anon_var_1 = 1L
    else:
        _hy_anon_var_1 = 2L
    return _hy_anon_var_1
```

But it compiles to this:

```
def x():
    try:
        _hy_anon_var_1 = some_call()
    except Exception as e:
        _hy_anon_var_1 = 1L
    else:
        2L
    return _hy_anon_var_1
```

Note that "else" does not produce a return value for the function. Same result for s/else/finally.

Kodiologist:
Fixes #798.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -793,30 +793,29 @@ class HyASTCompiler(object):
 
         body = body.stmts
 
-        orelse = []
-        finalbody = []
-        handlers = []
+        if not all(expr):
+            raise HyTypeError(expr, "Empty list not allowed in `try'")
         handler_results = Result()
-
-        for e in expr:
-            if not len(e):
-                raise HyTypeError(e, "Empty list not allowed in `try'")
-
-            if e[0] == HySymbol("except"):
-                handler_results += self._compile_catch_expression(e, name)
-                handlers.append(handler_results.stmts.pop())
-            elif e[0] == HySymbol("else"):
-                orelse = self.try_except_helper(e, HySymbol("else"), orelse)
-            elif e[0] == HySymbol("finally"):
-                finalbody = self.try_except_helper(e, HySymbol("finally"),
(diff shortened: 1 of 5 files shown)


hy/compiler.py
```diff
@@ -791,55 +786,57 @@ def compile_try_expression(self, expr):
...
+        body += body.expr_as_stmt() if orelse else ast.Assign(
+            targets=[name],
+            value=body.force_expr,
+            lineno=expr.start_line,
+            col_offset=expr.start_column)
+
+        if not (handlers or finalbody):
+            # (try) or (try BODY)
+            # As is, this would be illegal under Python >= 3.3.
+            # So forget the `try` and just use the BODY.
```
kirbyfan64: IMO this should be an error instead, rather than silently omitting the error handling.

Kodiologist: My own thinking is that this behavior, albeit possibly surprising, will be useful for situations such as the programmer commenting out all the `except` clauses, or accumulating `except` clauses programatically. In general, it seems good for Hy to provide base cases of this kind that Python doesn't. I'm not really sure, though.

kirbyfan64: True, but I personally feel like the surprising worst-case behavior is a lot more confusing than the benefits from the best-case behavior...

Kodiologist: I don't feel strongly on this subject, so I'll remove this feature.

Kodiologist: Done.

PR #1298, opened 2017-05-19
Switch from setf to setv according to NEWS

yuvallanger:
[Current master branch NEWS file says](https://github.com/hylang/hy/blob/a27d737e1c6273dcb4c377d123392fc1951b3150/NEWS#L413):

> * Removes setf in favour of setv

### DIFF
--- a/docs/language/core.rst
+++ b/docs/language/core.rst
@@ -705,7 +705,7 @@ Returns ``True`` if *x* is ``None``.
    => (none? 0)
    False
 
-   => (setf x None)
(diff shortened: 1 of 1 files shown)


Kodiologist:
Good catch; it's embarrassing how long that's hung around. Could you open this PR against `master` instead of `docs-update`? `docs-update` is for a separate PR, namely #1293.

yuvallanger:
@Kodiologist Switched to master base.

PR #1301, opened 2017-05-30
Implement Python 2 `exec`

### LINKED ISSUE #749: exec not defined on hy2

Kodiologist:
Fixes #749.

### DIFF
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -117,6 +117,24 @@
    range range
    zip zip))
 
+(if-python2
+  (defn exec [$code &optional $globals $locals]
+    "Execute Python code.
+
+    The parameter names contain weird characters to discourage calling this
+    function with keyword arguments, which isn't supported by Python 3's
+    `exec`."
+    (if
+      (none? $globals) (do
+        (setv frame (._getframe sys (int 1)))
+        (try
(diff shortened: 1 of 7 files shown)


kirbyfan64:
My 2¢:

This should be `exec*`, and `exec` itself could be defined conditionally in the core modules, that way semantics are identical for both.

Kodiologist:
@kirbyfan64 Done.

hy/core/language.hy
```diff
@@ -15,7 +15,7 @@
 (if-python2
   (import [StringIO [StringIO]])
   (import [io [StringIO]]))
-(import [hy._compat [long-type]]) ; long for python2, int for python3
+(import [hy._compat [PY3 long-type]])
```
kirbyfan64: `PY3` seems like a randomly added import here...

Kodiologist: Evidently that was some kind of leftover I forgot about. Fixed.

Kodiologist:
@tuturto Poke.

PR #1303, opened 2017-06-10
cond accept single-argument branches

### LINKED ISSUE #1084: cond should accept single-argument branches
In Common Lisp, the result of the test itself is the result of the `cond`, if an alternative is not provided:

``` Lisp
* (cond (:foo))

:FOO
```

In Hy it's an error.

``` Hy
=> (cond [:foo])

  (cond [:foo])
        ^----^
HyMacroExpansionError: b'cond branches need at least two items: a test and one or more code branches'
```

It needn't be.

neil-lindquist:
Modified cond to support single argument branches as requested in #1084.  This is my first contribution, so I'd appreciate feedback.

Closes #1084.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -77,3 +77,4 @@
 * Charles de Lacombe <ealhad@mail.com>
 * John Patterson <john@johnppatterson.com>
 * Kai Lüke <kailueke@riseup.net>
+* Neil Lindquist <archer1mail@gmail.com
--- a/NEWS
+++ b/NEWS
@@ -32,6 +32,7 @@ Changes from 0.12.1
      `(try 1 (except [ValueError] 2) (else 3))` returns `3`.
    * xor: If exactly one argument is true, return it
    * hy.core.reserved is now hy.extra.reserved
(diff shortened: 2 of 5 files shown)


kirbyfan64:
Great job!! :+1

neil-lindquist:
I described the single argument version of cond and did a macro expansion example, like was there.

Kodiologist:
I'm afraid there's a bug, which is that the branch is evaluated twice:

    $ hy
    hy 0.12.1+104.g30d456c using CPython(default) 3.6.1 on Linux
    => (setv x 0)
    => (cond [(do (+= x 1) True)])
    True
    => x
    2

You should be able to fix this with `or` or `gensym`. Don't forget to add a test for it, too.

Kodiologist:
Great work, thanks a lot, and welcome aboard. I'm going to reorganize the commits a bit and merge this.

Kodiologist:
I removed a few trailing spaces, too. Try not to commit those; `git diff` is helpful for finding them.

PR #1304, opened 2017-06-12
Fix Tuukka print example

0atman:
It was missing quotes around "Tuukka". Output from hy to confirm everything's good:

```clojure
hy unknown using CPython(default) 3.6.1 on Linux
=> (print "Hello there," "Tuukka")
Hello there, Tuukka
```


kirbyfan64:
Thanks!!

Issue #1289, opened 2017-05-14
Tests are failing on Windows


gilch:
I'm not getting these on Linux, so Travis didn't notice. 

```
============================= test session starts =============================
platform win32 -- Python 3.6.1, pytest-3.0.7, py-1.4.33, pluggy-0.4.0
rootdir: C:\Users\ME\hytest\hy-master, inifile: setup.cfg
collected 153 items

tests\test_bin.py ......F................F....
tests\test_lex.py .................................
tests\test_models.py ..............
tests\compilers\test_ast.py .................................................................
tests\compilers\test_compiler.py ...
tests\importer\test_importer.py .....
tests\importer\test_pyc.py .
tests\macros\test_macro_processor.py ...
tests\macros\test_sharp_macros.py .

================================== FAILURES ===================================
_________________ test_bin_hy_stdin_error_underline_alignment _________________

    def test_bin_hy_stdin_error_underline_alignment():
        _, err = run_cmd("hy", "(defmacro mabcdefghi [x] x)\n(mabcdefghi)")
>       assert "\n  (mabcdefghi)\n  ^----------^" in err
E       assert '\n  (mabcdefghi)\n  ^----------^' in 'hy unknown using CPython(v3.6.1:69c0db5) 3.6.1 on Windows\r\n  File "<input>", line 1, column 1\r\n\r\n  (mabcdefghi)...or: b"expanding `mabcdefghi\': missing 1 required positional argument: \'x\'"\r\n\r\n\r\n\r\nnow exiting HyREPL...\r\n'

tests\test_bin.py:125: AssertionError
__________________________ test_bin_hy_byte_compile ___________________________

    def test_bin_hy_byte_compile():
    
        modname = "tests.resources.bin.bytecompile"
        fpath = modname.replace(".", "/") + ".hy"
    
        for can_byte_compile in [True, False]:
            for cmd in ["hy " + fpath,
                        "hy -m " + modname,
                        "hy -c '(import {})'".format(modname)]:
    
                rm(get_bytecode_path(fpath))
    
                if not can_byte_compile:
                    # Keep Hy from being able to byte-compile the module by
                    # creating a directory at the target location.
                    os.mkdir(get_bytecode_path(fpath))
    
                # Whether or not we can byte-compile the module, we should be able
                # to run it.
>               output, _ = run_cmd(cmd)

tests\test_bin.py:278: 
[...]

PR #1307, opened 2017-06-17
universal newlines in test

gilch:
Change run_cmd for test_bin.py to use universal newlines (text mode) for Windows compatibility.

Is there some reason this popen has to be done in binary mode? Windows does line endings differently, which makes a test fail on Windows. #1289. The other failing test shown in that issue isn't fixed by this though.

### DIFF
--- a/tests/test_bin.py
+++ b/tests/test_bin.py
@@ -23,9 +23,10 @@ def run_cmd(cmd, stdin_data=None, expect=0):
                          stdin=subprocess.PIPE,
                          stdout=subprocess.PIPE,
                          stderr=subprocess.PIPE,
+                         universal_newlines=True,
                          shell=True)
     if stdin_data is not None:
-        p.stdin.write(stdin_data.encode('ASCII'))
+        p.stdin.write(stdin_data)
         p.stdin.flush()
         p.stdin.close()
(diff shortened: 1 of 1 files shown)


gilch:
> Can you make the first line of the commit message more specific, like "Use universal newlines in test_bin.py"?

Done.

> Pushing to hylang/hy makes Travis test your branch twice.

Noted. I wonder if we can turn off that Travis "feature" though, and only test the pr's and master (or other important branches).

I've discovered git on Windows is harder than it should be. Cygwin's git doesn't use the "recommended" `core.autocrlf` default, and instead pretends it's on a Unix. Of course, Git on Ubuntu on Windows (WSL) has the same setting--I thought it might be contributing to the error, so I downloaded GitHub Desktop and cloned the hylang repository using that. (No change in the test though.) Of course, Cygwin's git reported every single text file had changed after that, which made the command line version pretty useless for operating on that repository. The GUI is pretty limited by comparison, so I just pushed the branch. GitHub Desktop used to have its own command line git installation I could access through the menus. I would have used that but, the new Electron update seems to have removed that feature.

gilch:
I've got two approvals and no further complaints, so I'll just merge this.

Issue #1274, opened 2017-04-14
How do you pronounce `.`, the attribute access DSL (or any other symbol for that matter)?


yuvallanger:
Is it pronounced "attribute access DSL" [1]?

Could you add a pronunciation guide to each of the symbols?

[1]: <https://github.com/hylang/hy/blob/18acfe6495e83dccc9c0f3ddf105dc7ccef5a86a/docs/language/api.rst#built-ins>

kirbyfan64:
Usually I just call it "attribute dot thingy" or something...

Kodiologist:
An interesting question. For Hy, as for many programming languages, nearly all the important communication occurs in writing, so there's no natural tendency to come to an agreement about pronunciation.

Kodiologist:
In any case, this doesn't seem to really be in scope for Hy's documentation, so I should close this issue.

Issue #1230, opened 2017-02-23
Reader macros aren't


Kodiologist:
For a long time, Hy has had a feature we call "reader macros", but they're not actually reader macros, just regular macros with an unusual syntax for calling them. A reader macro is a macro that is called at read-time, and hence can change how raw characters of the source are parsed. It's provided with raw text as arguments. An ordinary macro, on the other hand, is called after the text has already been parsed into s-expressions. It's provided with s-expressions as arguments.

Real reader macros would be nice to have, but they would be hard to implement (see #722, which is over 3 years old and probably won't get closed soon). In the meantime, I think we're long overdue for not misleading users about our so-called reader macros. I see two options:

- Remove the feature entirely, so we can use some of the same syntax for real reader macros once we have them, and so-called reader macros won't suddenly be reinterpreted as real reader macros after one upgrades Hy.
- Call them something else, like "sharp macros", since they're called with a character after a sharp sign (`#`). `defreader` would then be `defsharp` or something.

gilch:
Something like Clojure's tagged literals #1046 might be easier to implement. Those aren't true reader macros in the Common Lisp sense either. But Hy is more like Clojure than Common Lisp.

Kodiologist:
Definitely, I think tags of that sort wouldn't be hard, since the macro gets run at the same time as ordinary macros.

gilch:
See also #1042. I don't know that Perl-style heredocs would work well in a Lisp, but maybe something like Lua's long brackets could work. If we had the better string syntax, then we could apply a tag to invoke an arbitrary reader/interpreter/compiler on a string containing any language we like. I think this would get us most of the power of reader macros, but perhaps there's more I haven't thought of.

Kodiologist:
I think I'm going to make a PR to just rename them to sharp macros. The extension to symbols longer than 1 character can be done later if desired.

Issue #1117, opened 2016-09-25
Add `|` quotes and `\` escapes to Hy symbols and keywords.


gilch:
Common Lisp lets you use arbitrary strings as symbols. If you include certain characters you have to escape them, either by quoting the whole string in `||`, or with a `\` per character. 

``` Lisp
* (symbolp '|Am I a symbol?|)

T
* '|Am I a symbol?|

|Am I a symbol?|
* (symbolp 'Am\ I?)

T
* 'Am\ I?

|AM I?|
```

Common Lisp's reader uppercases symbols by default, but valid symbols can have lower case characters as demonstrated above. Hy's reader does similar shenanigans with `*`, `-`, `?`, and `!`, yet does allow them in symbols.

``` Hy
=> '*A-symbol!*
'A_SYMBOL_bang'
=> (HySymbol "*A-symbol!*")
'*A-symbol!*'
```

I propose we add the Common lisp style symbol quotes and escape to Hy, so if you _really_ want the symbol to contain a `-` or end in `?` (etc.), you can, but most of the time it will get mangled to something Python can use more easily as now. This could make Hy easier to work with when these details matter.

``` Hy
=> '|*A-symbol!*|
'*A-symbol!*'
```

kirbyfan64:
Wouldn't it be better to have something like `'"text here"`, i.e. a single quote following the text in double-quotes. The syntax is less weird and easier to read.

gilch:
I don't think so. That particular syntax isn't going to work, since it already means `(quote "text here")` to the reader. We already have the `HySymbol` function to convert strings into quoted symbols. That's not the point. The point is, you should have a read syntax for not-quoted symbols that contain special characters. `('"my func" '"my var")` wouldn't do anything; it's just a list of symbols. If we don't quote them, `("my func" "my var")` is now just a list of strings. But `(|my func| |my var|)` is _executable_ code in Common Lisp--it applies `|my func|` to `|my var|`.

We could perhaps do a reader macro on a string like Clojure does for regex, but I didn't pull this syntax out of my nose. It is how Common Lisp does it and Scheme too. I think we need a better reason than "kirby doesn't like it" to break established convention with some other syntax, or we're just making things harder on Hy's users who may already be familiar with it.

> The syntax is less weird and easier to read.

It's perfectly readable.
Technically the `\` escape alone is enough. We don't need the `|` also, but I think quotation marks like
```
|I'm a symbol!|
```
looks a lot nicer than escapes like
```
[...]

PR #1310, opened 2017-06-21
Tag macros

### LINKED ISSUE #1046: Add Clojure tagged literals or similar
Our reader macros have capabilities similar to Clojure's [tagged literals](http://clojure.org/reference/reader#_tagged_literals), but Hy's are limited to a single character after the dispatch character `#`. In Clojure, it can be an arbitrary symbol after the `#`. One-character macros make sense for something that gets used all the time, or for DSLs where something has to be repeated a lot. But for general user uses, it could be nicer to use a more descriptive name.

I'm not sure how hard it would be to adapt Hy's reader macros to read a symbol instead of a character.

gilch:
Closes #1046 

(See also #1230.)

This also changes the "sharp macro" term to "tag macro", to better match with Clojure. (And besides, the ASCII Number Sign `#` isn't the same as the Unicode `♯` Music Sharp Sign. This was misleading--They're similar but note that the horizontal lines of `#` wouldn't show up well on the musical staff.)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,9 @@
+Changes from 0.13.0
+
+   [ Language Changes ]
+   * Single-character "sharp macros" changed to "tag macros", which can have
+     longer names
+
 Changes from 0.12.1
 
    [ Language Changes ]
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -785,19 +785,15 @@ defsharp
 .. versionadded:: 0.13.0
 
 ``defsharp`` defines a sharp macro. A sharp macro is a unary macro that has the
-same semantics as an ordinary macro defined with ``defmacro``, but can be
-called without parentheses and with less whitespace. The name of a sharp macro
-must be exactly one character long. It is called with the syntax ``#cFORM``,
-where ``#`` is a literal sharp sign (hence the term "sharp macro"), ``c`` is
-the name of the macro, and ``FORM`` is any form. Whitspace is forbidden between
-``#`` and ``c``. Whitespace is allowed between ``c`` and ``FORM``, but not
-required.
+same semantics as an ordinary macro defined with ``defmacro``. It is called with
+the syntax ``#tag FORM``, where ``tag`` is the name of the macro, and ``FORM``
+is any form. The ``tag`` is often only one character, but it can be any symbol.
 
 .. code-block:: clj
 
     => (defsharp ♣ [expr] `[~expr ~expr])
     <function <lambda> at 0x7f76d0271158>
-    => #♣5
+    => #♣ 5
     [5, 5]
     => (setv x 0)
     => #♣(+= x 1)
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -779,19 +779,19 @@ For example,
     42
 
 
-defsharp
+deftag
 --------
 
 .. versionadded:: 0.13.0
 
-``defsharp`` defines a sharp macro. A sharp macro is a unary macro that has the
+``deftag`` defines a tag macro. A tag macro is a unary macro that has the
 same semantics as an ordinary macro defined with ``defmacro``. It is called with
 the syntax ``#tag FORM``, where ``tag`` is the name of the macro, and ``FORM``
 is any form. The ``tag`` is often only one character, but it can be any symbol.
 
 .. code-block:: clj
 
-    => (defsharp ♣ [expr] `[~expr ~expr])
+    => (deftag ♣ [expr] `[~expr ~expr])
     <function <lambda> at 0x7f76d0271158>
     => #♣ 5
     [5, 5]
@@ -801,13 +801,13 @@ is any form. The ``tag`` is often only one character, but it can be any symbol.
     => x
     2
 
-In this example, if you used ``(defmacro ♣ ...)`` instead of ``(defsharp
+In this example, if you used ``(defmacro ♣ ...)`` instead of ``(deftag
 ♣ ...)``, you would call the macro as ``(♣ 5)`` or ``(♣ (+= x 1))``.
 
-The syntax for calling sharp macros is similar to that of reader macros a la
-Common Lisp's ``SET-MACRO-CHARACTER``. In fact, before Hy 0.13.0, sharp macros
+The syntax for calling tag macros is similar to that of reader macros a la
+Common Lisp's ``SET-MACRO-CHARACTER``. In fact, before Hy 0.13.0, tag macros
 were called "reader macros", and defined with ``defreader`` rather than
-``defsharp``. True reader macros are not (yet) implemented in Hy.
+``deftag``. True reader macros are not (yet) implemented in Hy.
 
 del
 ---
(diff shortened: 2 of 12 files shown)


Kodiologist:
> This was misleading

Do you also write the name of the programming language C# as "C♯"? :P

gilch:
These history rewrites are trying my limited git skills. I think I can do this with an interactive rebase, but separating hunks from a single commit like that seems too hard. Maybe I can copy/paste and squash in a new commit. How would you approach this?

> Do you also write the name of the programming language C# as "C♯"?

[Officially, yes](https://en.wikipedia.org/wiki/C_Sharp_(programming_language)#Name). But Unicode is hard to type, and everyone knows the substitution. Hy, on the other hand, actually does support Unicode symbols. Regardless, the stronger argument is really that "tag" is closer to what Clojure calls its version.

> is a bit vague; you mean something like "when `FORM` starts with an identifier character", right?

No, I meant what I said. Something like `#foo5` means the tag symbol is `foo5`, but `#foo 5` means the tag symbol is `foo` and the tagged form is `5`, which is an integer, not a symbol ("identifier"). If you think my proposed version is not clear, OK, let's work on it, but your proposed version is still not right either.

Kodiologist:
> How would you approach this?

Rewind to the first commit (in an interactive rebase, with `git reset --hard`, or even with just `git checkout HEAD~`, although then you have to reconstruct the branch when you're done). Amend it to make the same substantive change to the documentation that you made in the second commit, but say "sharp macro" instead of "tag macro". Now reapply the second commit, resolving the merge conflict by just ignore the first commit's version. It should now be the case that renaming the feature is all that the second commit does.

It may actually be easiest, overall, if you put the name-change commit first. This will oblige you to change less stuff as you add tests and so on,  whether you make your changes by adding new commits or amending. (My own PRs often have some kind of cleanup in the first commit or commits and the change that the PR is really about in the later commits.)

> Something like `#foo5` means the tag symbol is `foo5`, but `#foo 5` means the tag symbol is foo and the tagged form is `5`, which is an integer, not a symbol ("identifier")

Actually, in Hy, `5` is an identifier at the lexical level; see e.g. `t_identifier` in `parsers.hy`. That's how we got #1143. But even in Python, `5` is a legal identifier character. It's just that, in Python as in other C-like languages, an identifier can't solely consist of digits.

Kodiologist:
So to clarify, in Hy's case, I use the term "identifier" to describe a lexical entity, whereas a "symbol" is a Hy model object of type `HySymbol`.

gilch:
> So to clarify, in Hy's case, I use the term "identifier" to describe a lexical entity

Ah, I see what you meant. I've also used that internal terminology in the first commit. But I don't think that is appropriate for the documentation, which is intended not just for Hy's developers, but for Hy's users, who can't be expected to be familiar with Hy's implementation details.

So, we could instead explicitly say that brackets `(`, `[`, `{`, quotes `'`, `"`, or white space (including `;` comments), will terminate a tag. Or we could just use more examples illustrating these.

Kodiologist:
I don't think that inferring what "identifier character" means requires understanding the internal distinction between identifiers and symbols. You can write "symbol character" if you prefer. `5` is a symbol character, too. But listing out all the characters that aren't identifier characters is brittle.

Speaking of characters, I notice that you use the term "hashtag character" in the documentation. A hashtag is some sort of special token on Twitter that consists of a hash followed by an identifier. I think you mean "hash character".

gilch:
> But listing out all the characters that aren't identifier characters is brittle.

Agreed, that's brittle. So are the examples, honestly, but at least we could potentially test those automatically. #1044. I'm inclined to remove the offending sentence altogether. This is documentation, not specification. The existing examples should suffice. Users can resolve this question pretty quickly in the repl.

> you use the term "hashtag character"

Heh, haven't you heard? That what all the kids are calling them nowadays. (And hashtags are used for a lot more than Twitter now.) "Pound", "hash", "number sign", "octothorp", and not to be confused with "sharp". This thing has too many names. Yeah, "hashtag" is arguably not correct for the character itself. I think we should just call it `#` in the docs and let the users worry about how to pronounce it.

Kodiologist:
> I'm inclined to remove the offending sentence altogether.

Sounds fine to me.

> Heh, haven't you heard? That what all the kids are calling them nowadays.

But only because they aren't familiar with the terms "hash" or "tag" and so they think the word "hashtag" refers to the funny character rather than the whole construct. Right?

> I think we should just call it `#` in the docs and let the users worry about how to pronounce it.

That's fine, too. I did close #1274 for more or less the same reason.

Kodiologist:
Good work!

Your test macro `#foo` should do something less trivial than returning its argument unaltered, so we test that the macro is actually run. (Optionally, make such a change to the other test macros in the file, too.)

It's also worth testing:
- Calling a single-character tag macro with whitespace after the tag
- Calling a multi-character tag macro with whitespace after the tag
- Defining and calling a macro whose name is subject to mangling
- Calling a tag macro (that's defined in the test file, not the built-in `#@`) on a non-literal expression

gilch:
I just noticed that you can define the empty string as the name of a sharp macro on the current master (0.13.0), but there's no way to call it. This seems like a good time to fix that.

The ability to define sharp macros with a string name is currently necessary to define certain characters, like `#.`. With this update, arbitrary strings are allowed as tag names, but they can't be called if empty or if they contain whitespace (or a bracket or quote or semicolon). 

#1117 could probably fix this, since any string would have an associated symbol. For now, would it be better to disallow strings to define tags, or just empty strings?

Kodiologist:
Keep allowing strings to define tags. You can disallow uncallable tag macros in this PR if you like, but I wouldn't require that to approve this.

gilch:
OK, I think I got it working properly.

Kodiologist:
Looking good. Don't forget about the case of mangling (bullet point 3 above).

gilch:
Okay, I added a tests for mangling. I had to add a mangling call to the compiler to make it pass. I amended the first commit with that, since I figured that's where you'd want it. This would probably work better with #1117.

Kodiologist:
Great. By the way, you can change the commit message for "Change sharp macros to tag macros" to say "rename" instead of "change" now that you've separated the functional changes out of it.

One other thing I noticed. On master, there's a pretty error message for using an unrecognized sharp macro:

    => #a 1

      #a 1
      ^^
    HyTypeError: b"`a' is not a defined sharp macro."

Not so with your changes:

    => #a 1
      File "/home/hippo/Desktop/hyenv/hy/hy/macros.py", line 220, in tag_macroexpand
        tag_macro = _hy_tag[None][tag]
    KeyError: 'a'

    During handling of the above exception, another exception occurred:

      [ ... ]
      File "/home/hippo/Desktop/hyenv/hy/hy/macros.py", line 224, in tag_macroexpand
        "`{0}' is not a defined tag macro.".format(tag)
    hy.errors.HyTypeError: <exception str() failed>

    During handling of the above exception, another exception occurred:

      [ ... ]
      File "/home/hippo/Desktop/hyenv/hy/hy/errors.py", line 46, in __str__
        line = self.expression.start_line
    AttributeError: 'HyString' object has no attribute 'start_line'

Could you look into this?

gilch:
That should fix it. I needed to use `.replace()` on the mangled symbol to transfer the metadata.

Kodiologist:
Thanks a lot. I'd be chuffed if you would review some of my open PRs as well (particularly ones that already have a review, so I can merge them).

PR #1312, opened 2017-06-21
Remove the Python 2 yield-from macro

### LINKED ISSUE #691: py2 `yield-from` fails to compile if the last item in a function is an expression.

### LINKED ISSUE #692: py2 `yield-from` does not propagate exceptions *into* the inner generators.

Kodiologist:
Backports of Python 3 features to Hy under Python 2 aren't worth the effort of supporting. It's Hy time for people to upgrade.

- Closes #691.
- Closes #692.

### DIFF
--- a/tests/native_tests/native_macros.hy
+++ b/tests/native_tests/native_macros.hy
@@ -306,31 +306,6 @@
   (assert (= (lif-not 0 "false" "true") "true")))
 
 
-(defn test-yield-from []
-  "NATIVE: testing yield from"
-  (defn yield-from-test []
-    (for* [i (range 3)]
-      (yield i))
-    (yield-from [1 2 3]))
-  (assert (= (list (yield-from-test)) [0 1 2 1 2 3])))
-
-(defn test-yield-from-exception-handling []
-  "NATIVE: Ensure exception handling in yield from works right"
-  (defn yield-from-subgenerator-test []
-    (yield 1)
-    (yield 2)
-    (yield 3)
-    (assert 0))
-  (defn yield-from-test []
-    (for* [i (range 3)]
-       (yield i))
-    (try
-     (yield-from (yield-from-subgenerator-test))
-     (except [e AssertionError]
-       (yield 4))))
-  (assert (= (list (yield-from-test)) [0 1 2 1 2 3 4])))
-
-
 (defn test-defmain []
   "NATIVE: make sure defmain is clean"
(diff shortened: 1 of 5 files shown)


kirbyfan64:
FWIW I don't think *every* compatible feature should be removed, but it didn't work too well anyway.

tests/native_tests/native_macros.hy
```diff
@@ -298,31 +298,6 @@
   (assert (= (lif-not 0 "false" "true") "true")))
 
 
-(defn test-yield-from []
```
tuturto: Weren't these two tests run for both Python 2 and Python 3? Would it make sense to keep them around for Python 3 still?

Kodiologist: Good point. I've restored them.

Kodiologist:
@tuturto Is this ready to go?

Issue #919, opened 2015-08-23
Consider adding Common Lisp's Sharpsign-dot (#.) reader macro


gilch:
Pushing the eval of certain forms back to readtime has various uses. A big one is to swap out code based on various conditions, like the `#if` and related preprocessor directives from C.

``` Hy
(#.(if (< *version* 2)  old-func  new-func) arg)
```

This would probably fit well with #740.

Another is to substitute for literals that don't have a literal notation. Compare:

``` Hy
=> (int "1a" 32)  ; base 32 number
int('1a', 32)
42
=> #.(int "1a" 32)  ; expands directly into 42 at readtime, like a literal
42
42
=> (+ "A line that is much too long to fit on one line "
... "doesn't actually have to fit on one line.")
('A line that is much too long to fit on one line ' + "doesn't actually have to fi
t on one line.")
"A line that is much too long to fit on one line doesn't actually have to fit on
 one line."
=> #.(+ "A line that is much too long to fit on one line "
... "doesn't actually have to fit on one line.")
"A line that is much too long to fit on one line doesn't actually have to fit on
 one line."
"A line that is much too long to fit on one line doesn't actually have to fit on
 one line."
```

Because evaluation happens at read time, it's safe to use in loops and such, without a performance hit for recomputing it every time.

I thought it would be a simple implementation in Hy:

``` Hy
#@((hy.macros.reader ".") (fn [expr] (eval expr)))
```

And this does seem to work for the above examples, but I'm running into `HyModel` issues again:

``` Hy
=> #.(lambda [] nil)
```

[long stack trace including `hy.errors.HyCompileError: Internal Compiler Bug \U0001f631\u2937 TypeError: Don't know how to wrap a <class 'function'> object to a HyObject`]

But this works fine in Common Lisp:

``` Lisp
* #.(lambda () nil)

#<FUNCTION (LAMBDA ()) {1002E222DB}>
```

Is there a simple change in the macro definition that would fix this? Can we perhaps make everything a HyObject by default if it doesn't already have a model?

Kodiologist:
Is `eval-when-compile`, which we now have, equivalent to what you asked for here?

gilch:
Maybe. It's not documented, so I'm not sure. And many of my use cases would need a shorter alias.

Kodiologist:
I was wrong; `eval-when-compile` doesn't expand to anything, so it doesn't do this. It's only for changing the state of the compiler, presumably to affect macro expansions.

[...]

PR #1314, opened 2017-06-22
Auto-promote values to HyObjects in the compiler

### LINKED ISSUE #1154: &optional and (apply ...) in a macro cause HyCompileError
The following:

    (defn f [&rest args]
      (print "f:" args))

    (defmacro mac [&optional x]
     `(apply f [~x]))

    (mac)

yields:

    ⤷ Exception: Unknown type: `<class 'NoneType'>'
    Compilation traceback:
    No traceback available. 😟

But there is no crash if you do any of the following:

- Call `mac` with an argument (even `None`)
- Write the body of `mac` as `(f ~x)`
- Change the parameter list of `mac` to `[&optional [x 'None]]` (but if you exclude the single quote, the crash still happens)

`(macroexpand '(mac))` yields `('apply' 'f' [None])`, and `(type (get (macroexpand '(mac)) 2 0))` is `NoneType`, which I believe means that `None` isn't getting wrapped into a Hy model object for some reason.

### LINKED ISSUE #1174: eval doesn't use hy.models._wrappers
I came across this issue trying to narrow down a compiler error I was getting from a macro.
Why should the following piece of code throw an error?

```
=> (import numpy)


=> (= "." (get (name numpy) 0)))
False
=> (eval '(= "." (get (name numpy) 0)))
False
=> (eval `(= "." ~(get (name numpy) 0)))
  File "/Users/dg/anaconda3/lib/python3.5/site-packages/hy/importer.py", line 110, in hy_eval
    hytree.replace(foo)
  File "/Users/dg/anaconda3/lib/python3.5/site-packages/hy/models/list.py", line 31, in replace
    x.replace(other)
TypeError: replace() takes at least 2 arguments (1 given)
```

Kodiologist:
- Fixes #1174.
- Fixes #1154.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,9 @@ Changes from 0.13.0
      longer names
    * Periods are no longer allowed in keywords
    * `eval` is now a function instead of a special form
+   * The compiler now automatically promotes values to Hy model objects
+     as necessary, so you can write ``(eval `(+ 1 ~n))`` instead of
+     ``(eval `(+ 1 ~(HyInteger n)))``
 
    [ Bug Fixes ]
    * Numeric literals are no longer parsed as symbols when followed by a dot
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,7 @@ Changes from 0.13.0
    * Single-character "sharp macros" changed to "tag macros", which can have
      longer names
    * Periods are no longer allowed in keywords
+   * `eval` is now a function instead of a special form
 
    [ Bug Fixes ]
    * Numeric literals are no longer parsed as symbols when followed by a dot
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -863,27 +863,6 @@ doto
   => collection
   [2 1]
 
-eval
-----
-
-``eval`` evaluates a quoted expression and returns the value. The optional
-second and third arguments specify the dictionary of globals to use and the
-module name. The globals dictionary defaults to ``(local)`` and the module name
-defaults to the name of the current module.
-
-.. code-block:: clj
-
-   => (eval '(print "Hello World"))
-   "Hello World"
-
-If you want to evaluate a string, use ``read-str`` to convert it to a
-form first:
-
-.. code-block:: clj
-
-   => (eval (read-str "(+ 1 1)"))
-   2
-
 
 eval-and-compile
 ----------------
--- a/docs/language/core.rst
+++ b/docs/language/core.rst
@@ -230,6 +230,30 @@ Returns ``True`` if *coll* is empty. Equivalent to ``(= 0 (len coll))``.
    False
 
 
+.. _eval-fn:
+
+eval
+----
+
+``eval`` evaluates a quoted expression and returns the value. The optional
+second and third arguments specify the dictionary of globals to use and the
+module name. The globals dictionary defaults to ``(local)`` and the module name
+defaults to the name of the current module.
+
+.. code-block:: clj
+
+   => (eval '(print "Hello World"))
+   "Hello World"
+
+If you want to evaluate a string, use ``read-str`` to convert it to a
+form first:
+
+.. code-block:: clj
+
+   => (eval (read-str "(+ 1 1)"))
+   2
+
+
 .. _every?-fn:
 
 every?
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -15,7 +15,7 @@ import astor.codegen
 
 import hy
 
-from hy.lex import LexException, PrematureEndOfInput, tokenize
+from hy.lex import LexException, PrematureEndOfInput
 from hy.lex.parser import hy_symbol_mangle
 from hy.compiler import HyTypeError
 from hy.importer import (hy_eval, import_buffer_to_module,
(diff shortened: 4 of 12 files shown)


gilch:
We've got conflicting files now.

Auto-promotion sure seems like it would make some things easier.

How would this affect #919? Tag macros are still at compile time, but I think Clojure's tagged literals are at read time, like Common Lisp's `#.`. I'd also like easier access to Python's `eval` in Hy (we already have `exec`). Currently you have to import it from the `builtins` module. We could either rename Hy's eval (to `hy-eval`) or rename Python's eval (to `py-eval`). I'm imagining in-line Python--something like
```
=> (#py"lambda x: x + 1" 41)
42
```
--would be possible with no performance impact. Currently, even with a macro, the (py)eval has to happen at runtime.

gilch:
If we're generally autoboxing unquotes, shouldn't this work?
```Python
=> (disassemble `(+ ~(+ 1 1) 40) True)
from hy.core.language import disassemble
from hy import HyExpression, HyInteger, HySymbol
disassemble(HyExpression(((([] + [HySymbol('+')]) + [(1 + 1)]) + [HyInteger(40)])), True)
  File "c:\users\me\documents\github\autobox-hy\hy\importer.py", line 184, in hy_eval
  File "C:\Users\ME\Documents\GitHub\autobox-hy/hy/core/language.hy", line 78, in disassemble
    (fake-source-positions tree)
  File "C:\Users\ME\Documents\GitHub\autobox-hy/hy/core/language.hy", line 182, in fake_source_positions
    (fake-source-positions subtree)))
  File "C:\Users\ME\Documents\GitHub\autobox-hy/hy/core/language.hy", line 185, in fake_source_positions
    (setattr tree attr 1))))
AttributeError: 'int' object has no attribute 'start_line'
```
It works if I box it manually though.
```Python
=> (disassemble `(+ ~(HyInteger (+ 1 1)) 40) True)
from hy.core.language import disassemble
from hy import HyExpression, HyInteger, HySymbol
disassemble(HyExpression(((([] + [HySymbol('+')]) + [HyInteger((1 + 1))]) + [HyInteger(40)])), True)
'(2 + 40)'
```
Is the compiler really the right place to put this?

Kodiologist:
I think #919 and inline Python (which I, too, would like to see) are orthogonal to this PR. And yeah, I think it would make sense to have `builtins.eval` exported from `hy.core.language` as `py-eval`, although that too would probably be better in a separate PR.

I'll look into the merge conflicts and `disassemble`. The latter is presumably running into problems because it does its own faking of source positions, which isn't playing nicely with how I do that in the compiler.

> Is the compiler really the right place to put this?

I'm pretty sure that the answer is yes. It's the last chance to do it, so doing it then should mean that we only have to implement it in one place in Hy.

Kodiologist:
@gilch Done.

Kodiologist:
Aaaand it's broken on Python 2. I'm working on it now.

Kodiologist:
It was an easy fix; I'd forgotten about the trailing `L`s of long literals.

gilch:
`hy/docs/language/internals.rst` may need some clarification because of these changes. That file was not modified though. This pr digs pretty deep into some of Hy's internals, I want more time to review it.

Kodiologist:
I'm inclined to think that trying to document the implementation was a mistake in the first place. But I don't see anything on that page that's been made obsolete by these changes.

gilch:
`hy-eval` needs a good docstring explaining its use and parameters now that we're using it as a function. (I think special forms and macros need this too, but that's another issue #356)
***
> document the implementation was a mistake

No, the Hy model system is important enough for macro writing that it should be considered part of the public API. See also, [this recent stackoverflow question](https://stackoverflow.com/questions/44333906/generating-python-code-with-hy-macros). This PR certainly would have helped.

It needs to be documented. And I don't think this PR goes far enough to change that.

For example, this works in Clojure
```Clojure
user=> (eval `(/ 2))
1/2
user=> (eval `(/ ~2))
1/2
user=> (eval `(~/ 2))
1/2
user=> (eval `(~/ ~2))
```
There's no distinction between value and model, like Hy has currently. This PR helps--
```Python
=> (eval `(/ 2))  ; works now
from hy.core.language import eval
from hy import HyExpression, HyInteger, HySymbol
eval(HyExpression((([] + [HySymbol('/')]) + [HyInteger(2)])))
0.5
=> (eval `(/ ~2))  ; requires this PR for autobox
from hy.core.language import eval
eval(HyExpression((([] + [HySymbol('/')]) + [2])))
0.5
=> (eval `(~/ 2))  ; Boom.
from hy.core.language import eval
from hy.core.shadow import /
from hy import HyExpression, HyInteger
eval(HyExpression((([] + [/]) + [HyInteger(2)])))
  File "c:\users\me\documents\github\autobox-hy\hy\importer.py", line 185, in hy_eval
  File "c:\users\me\documents\github\autobox-hy\hy\importer.py", line 159, in hy_eval
    replace_hy_obj(hytree, foo)
  File "c:\users\me\documents\github\autobox-hy\hy\models.py", line 50, in replace_hy_obj
    return obj.replace(other)
  File "c:\users\me\documents\github\autobox-hy\hy\models.py", line 178, in replace
    replace_hy_obj(x, other)
  File "c:\users\me\documents\github\autobox-hy\hy\models.py", line 58, in replace_hy_obj
    % type(obj))
TypeError: Don't know how to wrap a <class 'function'> object to a HyObject
```
--but only certain basic values can be autoboxed. Even understanding the error message requires knowledge of these "internals": There's no Hy model for functions. And custom data types will have the same problem. I don't see an easy way to fix this, so it seems to be beyond the scope of this PR.

> But I don't see anything on that page that's been made obsolete by these changes.

I suppose it can wait then.

Kodiologist:
> `hy-eval` needs a good docstring explaining its use and parameters now that we're using it as a function

Done.

>> document the implementation was a mistake
>
> No, the Hy model system is important enough for macro writing that it should be considered part of the public API.

I agree. I'm sorry I was imprecise. What I don't think we should be documenting is the stuff under "Hy Internal Theory".

Kodiologist:
Confusingly, GitHub doesn't seem to display the commits in parentage order, like Git itself does.

PR #1327, opened 2017-07-19
Implement #qC string literals

### LINKED ISSUE #1287: Alternative syntax for string literals
Along the lines of #1042, or at least @gilch's reply to himself, I'd like to give Hy another syntax for string literals. Currently we only support double-quoted strings, which is inconvenient when the string itself contains lots of double quotes, or you're writing Hy code as part of a shell command and hence are running out of quote characters fast. We could adopt Python's triple-quoted strings (with double quotes), but six characters of delimiters (seven if you want a raw string) is pretty bulky for a string literal, and this doesn't help at all if you're embedding the code in a double-quoted string. Here are three proposals. @kirbyfan64, @tuturto, what do you think?

**Proposal A**

Permit delimiting string literals with `«` and `»`. This would work identically to double quotes, except now it's `»` rather than `"` that needs to be escaped.

**Proposal B**

Permit beginning string literals with `#` followed by a backtick and ending them with another backtick. Again, this works just like double quotes.

**Proposal C**

I propose something inspired by Perl's `q` operator, but a little simpler. A literal of this form begins with `#qC`, where `C` is any character that isn't whitespace or `;`. It ends with another instance of `C`, unless `C` belongs to the Unicode General Categories Ps (Punctuation, open) or Pi (Punctuation, initial quote), in which case the character with the code point after `C` terminates the literal.

These literals are always parsed as raw strings, and interpreted as Unicode strings (`unicode` on Python 2, `str` on Python 3) rather than bytestrings.

Here are some examples:

- `#q!Hello, world.!`
- `#q(Hello, world.)`
- `#q"Hello, world."`
- `#q'Hello, world.'`
- `#q("It's okay." he said.)`
- `#q( ;; This is a string, not a comment. )`
- `#q{Curlies}`
- `#qXHow about letters?X`
- `#q«ASCII is for “small minds”.»`
- `#q)This is allowed, but stupid.)`
- `#q(The backslash in \n is preserved)`.

And also:

    #q(Literal newlines
    are allowed, and preserved.)

They produce an ordinary `HyString`.

Kodiologist:
Closes #1287.

There's no documentation or NEWS update yet.

~~Note that matching delimiters are supported, but not inner pairs of matching delimiters (e.g., `#q(())` is parsed the same as `"(" )`). This is because the lexer is regex-based, and Python doesn't implement recursive regexes.~~ Matching delimiters are supported.

~~@gilch, I didn't reserve a character for indicating multi-character delimiters since you hadn't seemed to settle on `[`,`=`, or `<`.~~ Multi-character delimiters are indicated with `<`.

### DIFF
--- a/hy/lex/lexer.py
+++ b/hy/lex/lexer.py
@@ -1,8 +1,11 @@
+# -*- encoding: utf-8 -*-
 # Copyright 2017 the authors.
 # This file is part of Hy, which is free software licensed under the Expat
 # license. See the LICENSE.
 
+import re
 from rply import LexerGenerator
+from rply.lexergenerator import Rule, Match
 
 
 lg = LexerGenerator()
@@ -13,6 +16,12 @@ lg = LexerGenerator()
 end_quote = r'(?![\s\)\]\}])'
 
 identifier = r'[^()\[\]{}\'"\s;]+'
+
+hashstring_paired_delims = (
+  # Unicode General Category "Pi" with the matching closing mark
+  (u"«", u"»"), (u"‘", u"’"), (u"‛", u"’"), (u"“", u"”"), (u"‹", u"›"), (u"⸂", u"⸃"), (u"⸄", u"⸅"), (u"⸉", u"⸊"), (u"⸌", u"⸍"), (u"⸜", u"⸝"), (u"⸠", u"⸡"),  # noqa
+  # BidiBrackets.txt
+  (u"(", u")"), (u"[", u"]"), (u"{", u"}"), (u"༺", u"༻"), (u"༼", u"༽"), (u"᚛", u"᚜"), (u"⁅", u"⁆"), (u"⁽", u"⁾"), (u"₍", u"₎"), (u"⌈", u"⌉"), (u"⌊", u"⌋"), (u"〈", u"〉"), (u"❨", u"❩"), (u"❪", u"❫"), (u"❬", u"❭"), (u"❮", u"❯"), (u"❰", u"❱"), (u"❲", u"❳"), (u"❴", u"❵"), (u"⟅", u"⟆"), (u"⟦", u"⟧"), (u"⟨", u"⟩"), (u"⟪", u"⟫"), (u"⟬", u"⟭"), (u"⟮", u"⟯"), (u"⦃", u"⦄"), (u"⦅", u"⦆"), (u"⦇", u"⦈"), (u"⦉", u"⦊"), (u"⦋", u"⦌"), (u"⦍", u"⦐"), (u"⦏", u"⦎"), (u"⦑", u"⦒"), (u"⦓", u"⦔"), (u"⦕", u"⦖"), (u"⦗", u"⦘"), (u"⧘", u"⧙"), (u"⧚", u"⧛"), (u"⧼", u"⧽"), (u"⸢", u"⸣"), (u"⸤", u"⸥"), (u"⸦", u"⸧"), (u"⸨", u"⸩"), (u"〈", u"〉"), (u"《", u"》"), (u"「", u"」"), (u"『", u"』"), (u"【", u"】"), (u"〔", u"〕"), (u"〖", u"〗"), (u"〘", u"〙"), (u"〚", u"〛"), (u"﹙", u"﹚"), (u"﹛", u"﹜"), (u"﹝", u"﹞"), (u"（", u"）"), (u"［", u"］"), (u"｛", u"｝"), (u"｟", u"｠"), (u"｢", u"｣"))  # noqa
 
 lg.add('LPAREN', r'\(')
 lg.add('RPAREN', r'\)')
@@ -27,6 +36,46 @@ lg.add('UNQUOTESPLICE', r'~@%s' % end_quote)
 lg.add('UNQUOTE', r'~%s' % end_quote)
 lg.add('HASHBANG', r'#!.*[^\r\n]')
 lg.add('HASHSTARS', r'#\*+')
+
+class BalancedQStringRule(Rule):
+    # We use this class instead of a regex so we can match nested
+    # matching delimiters. (Python's `re` doesn't support recursive
+    # regexes.)
+    name = 'QSTRING'
+
+    def __init__(self, opener, closer):
+        self.opener, self.closer = opener, closer
+
+    def matches(self, s, pos):
+        # Match `#q{ ... }`, where the curly braces stand for the
+        # delimiters and `...` can contain nested pairs of matching
+        # delimiters.
+        start = pos
+        if not s[pos:pos + 3] == '#q' + self.opener:
+            return None
+        pos += 3
+        depth = 1
+        while True:
+            if pos >= len(s):
+                return None
+            elif s[pos] == self.opener:
+                depth += 1
+            elif s[pos] == self.closer:
+                depth -= 1
+                if depth == 0:
+                    break
+            pos += 1
+        return Match(start, pos + 1)
+
+# q-string, balanced style: #q{ ... }
+for opener, closer in hashstring_paired_delims:
+    lg.rules.append(BalancedQStringRule(opener, closer))
+# q-string, pointy style: #q<foo> ... foo
+lg.add('QSTRING', r'#q<([^>]+)>(?:.|\n)*?\1')
+# q-string, plain style: #qX ... X
+lg.add('QSTRING', u'#q([^<{}])(?:.|\\n)*?\\1'.format(''.join(
+    re.escape(opener) for opener, _ in hashstring_paired_delims)))
+
 lg.add('HASHOTHER', r'#%s' % identifier)
 
 # A regexp which matches incomplete strings, used to support
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ tox: venv
 	tox
 
 flake:
-	flake8 hy tests --ignore=E121,E123,E126,E226,E24,E704,W503,E302,E305,E701
+	flake8 hy tests --ignore=E121,E123,E126,E203,E226,E24,E704,W503,E302,E305,E701
 
 clear:
 	clear
--- a/hy/lex/parser.py
+++ b/hy/lex/parser.py
@@ -2,6 +2,7 @@
 # This file is part of Hy, which is free software licensed under the Expat
 # license. See the LICENSE.
 
+import re
 from functools import wraps
 from ast import literal_eval
 
@@ -297,6 +298,24 @@ def t_string(p):
 def t_partial_string(p):
     # Any unterminated string requires more input
     raise PrematureEndOfInput("Premature end of input")
+
+
+@pg.production("string : QSTRING")
+@set_boundaries
+def t_qstring(p):
+    # Remove the leading "#q".
+    s = p[0].value[2:]
+    # Remove the delimiters.
+    if s.startswith('<'):
+        m = re.match('<([^>]+)', s)
+        s = s[1 + len(m.group(1)) + 1 : -len(m.group(1))]
+    else:
+        s = s[1:-1]
+    # Remove any leading newline.
+    if s.startswith('\n'):
+        s = s[1:]
+    # The rest is interpreted as a raw Unicode string.
+    return HyString(s)
 
 
 @pg.production("identifier : IDENTIFIER")
--- a/tests/compilers/test_ast.py
+++ b/tests/compilers/test_ast.py
@@ -1,3 +1,4 @@
+# -*- encoding: utf-8 -*-
 # Copyright 2017 the authors.
 # This file is part of Hy, which is free software licensed under the Expat
 # license. See the LICENSE.
@@ -44,6 +45,10 @@ def cant_compile(expr):
         assert isinstance(e.exception, HyTypeError)
         assert e.traceback
         return e
+
+
+def s(x):
+    return can_compile(x).body[0].value.s
 
 
 def test_ast_bad_type():
@@ -473,12 +478,54 @@ def test_ast_unicode_strings():
 
 
 def test_ast_unicode_vs_bytes():
-    def f(x): return can_compile(x).body[0].value.s
-    assert f('"hello"') == u"hello"
-    assert type(f('"hello"')) is (str if PY3 else unicode)  # noqa
-    assert f('b"hello"') == (eval('b"hello"') if PY3 else "hello")
-    assert type(f('b"hello"')) == (bytes if PY3 else str)
-    assert f('b"\\xa0"') == (bytes([160]) if PY3 else chr(160))
+    assert s('"hello"') == u"hello"
+    assert type(s('"hello"')) is (str if PY3 else unicode)  # noqa
+    assert s('b"hello"') == (eval('b"hello"') if PY3 else "hello")
+    assert type(s('b"hello"')) == (bytes if PY3 else str)
+    assert s('b"\\xa0"') == (bytes([160]) if PY3 else chr(160))
+
+
+def test_ast_qstring_plain():
+    assert s('#q" hello world "') == ' hello world '
+    assert s(r'#q" h\ello wo\r\ld "') == r' h\ello wo\r\ld '
+    assert s('#qXliteral\nnewlineX') == 'literal\nnewline'
+    assert s(r'#qXbackslash\nthingX') == r'backslash\nthing'
+    assert s('#qXtrailing backslash\\X') == 'trailing backslash\\'
+    assert s('#q\\backslash delims\\') == 'backslash delims'
+    assert s(r'#qX ; not a comment X') == ' ; not a comment '
+    assert s(r'''#qX "ain't" X''') == ''' "ain't" '''
+    assert s(r"#q'single quotes'") == 'single quotes'
+    assert s(r"#q😈Unicode😈") == 'Unicode'
+
+
+def test_ast_qstring_balanced():
+    assert s(r'#q[bracketed]') == 'bracketed'
+    assert s(r'#q[inner [ open ] brackets]') == 'inner [ open ] brackets'
+    assert (s(r'#q[more  [[ inner ] ]  [brackets]]') ==
+            'more  [[ inner ] ]  [brackets]')
+    assert s(r'#q[ other () {} \  stuff ]') == r' other () {} \  stuff '
+    assert s(r'#q(parens)') == 'parens'
+    assert s(r'#q{braces}') == 'braces'
+    assert s(r'#q⁅funny brackets⁆') == 'funny brackets'
+    assert s(r'#q⸨funny brackets⸩') == 'funny brackets'
+    assert s(r'#q“funny quotes”') == 'funny quotes'
+    assert s(r'#q«funny quotes»') == 'funny quotes'
+    assert s(r'#q«nested «funny» quotes»') == 'nested «funny» quotes'
+
+
+def test_ast_qstring_pointy():
+    assert s(r'#q<my delim>pointy stringmy delim') == 'pointy string'
+    assert s(r'#q<foozle>aa foozli bb foozle') == 'aa foozli bb '
+    assert s(r'#q<(>unbalanced(') == 'unbalanced'
+    assert s(r'#q<(]1💯@[)} {a!>hello world(]1💯@[)} {a!') == 'hello world'
+    assert (s(r'''#q<X>
+Remove the leading newline, please.
+X''') == 'Remove the leading newline, please.\n')
+    assert (s(r'''#q<X>
+
+
+Only one leading newline should be removed.
+X''') == '\n\nOnly one leading newline should be removed.\n')
 
 
 def test_compile_error():


gilch:
I wasn't that picky, but I don't see how `=` could possibly work. I don't have strong feelings about which one, but let's go with `[`.

Kodiologist:
So the opening syntax would be something like `q[foo[`. You said that the closing syntax could just be `foo`, but shouldn't it be `]foo]` so that text editors don't think there are unmatched opening brackets?

gilch:
Good point. So do opening like `q[foo[` and closing like `]foo]`. I'd also like to strip the first newline, if present, like Lua does.

gilch:
If we did the opening like `q[foo]` then the closing could be just `foo` without the unmatched brackets, but I don't have a strong preference here either at the moment.

gilch:
> This is because the lexer is regex-based, and Python doesn't implement recursive regexes.

Python does have a [regex](https://pypi.python.org/pypi/regex) library on PyPI that can do recursive regexes like Perl. It has the same license as Python. I wonder if we can just drop that in without rewriting rply.

Kodiologist:
It works! (https://github.com/Kodiologist/hy/tree/hashstrings-recursive-regex) The catch is that `regex` contains C code and [has ambivalent support for PyPy](https://bitbucket.org/mrabarnett/mrab-regex/issues/240/unable-to-build-the-project-from-source-on#comment-36794787). So I don't think we should do this. Even if we weren't supporting PyPy, recursive matching of delimiters probably isn't worth adding a C dependency to a project that otherwise uses only pure Python.

Kodiologist:
Maybe I could achieve this instead with a `Rule` subclass that uses procedural code instead of a regex in its `matches` method. Do you think it's worth it?

gilch:
> Do you think it's worth it?

Probably yes. This would make it so easy to quote code from any language with a balanced delimiter pair, which is most of them. (Even **[J](https://en.wikipedia.org/wiki/J_(programming_language))** balances [parentheses!](http://www.jsoftware.com/help/primer/parentheses.htm)). I was even considering gating the feature (e.g. like we do for `nonlocal`) so only PyPy wouldn't support it. But if we can do it everywhere, that's better.

Is a rule subclass like that part of rply's public interface? If not, this is a harder call, since it would make updating rply more difficult if its implementation ever changes.

gilch:
This is making me reconsider the choice of `q[foo[` with `]foo]`. If we can do balanced delimiters, maybe it's better not to reserve `[`, since it would probably be a common choice for short strings. I think we should do `q<foo>` with `foo` instead. (And strip one initial newline, if present.)

Kodiologist:
> Is a rule subclass like that part of rply's public interface?

It isn't, but I'm not too worried about that, because I'll only be making assumptions about a small amount of the code, and the author of rply seems pretty responsive.

> I think we should do `q<foo>` with `foo` instead. (And strip one initial newline, if present.)

Sounds good to me.

hy/lex/lexer.py
```diff
@@ -14,6 +15,12 @@
 
 identifier = r'[^()\[\]{}\'"\s;]+'
 
+hashstring_paired_delims = (
+  # Unicode General Category "Pi" with the matching closing mark
+  ("«", "»"), ("‘", "’"), ("‛", "’"), ("“", "”"), ("‹", "›"), ("⸂", "⸃"), ("⸄", "⸅"), ("⸉", "⸊"), ("⸌", "⸍"), ("⸜", "⸝"), ("⸠", "⸡"),  # noqa
+  # BidiBrackets.txt
+  ("(", ")"), ("[", "]"), ("{", "}"), ("༺", "༻"), ("༼", "༽"), ("᚛", "᚜"), ("⁅", "⁆"), ("⁽", "⁾"), ("₍", "₎"), ("⌈", "⌉"), ("⌊", "⌋"), ("〈", "〉"), ("❨", "❩"), ("❪", "❫"), ("❬", "❭"), ("❮", "❯"), ("❰", "❱"), ("❲", "❳"), ("❴", "❵"), ("⟅", "⟆"), ("⟦", "⟧"), ("⟨", "⟩"), ("⟪", "⟫"), ("⟬", "⟭"), ("⟮", "⟯"), ("⦃", "⦄"), ("⦅", "⦆"), ("⦇", "⦈"), ("⦉", "⦊"), ("⦋", "⦌"), ("⦍", "⦐"), ("⦏", "⦎"), ("⦑", "⦒"), ("⦓", "⦔"), ("⦕", "⦖"), ("⦗", "⦘"), ("⧘", "⧙"), ("⧚", "⧛"), ("⧼", "⧽"), ("⸢", "⸣"), ("⸤", "⸥"), ("⸦", "⸧"), ("⸨", "⸩"), ("〈", "〉"), ("《", "》"), ("「", "」"), ("『", "』"), ("【", "】"), ("〔", "〕"), ("〖", "〗"), ("〘", "〙"), ("〚", "〛"), ("﹙", "﹚"), ("﹛", "﹜"), ("﹝", "﹞"), ("（", "）"), ("［", "］"), ("｛", "｝"), ("｟", "｠"), ("｢", "｣"))  # noqa
```
gilch: There's no need to list all these out. https://docs.python.org/3.6/library/unicodedata.html

Kodiologist: `unicodedata` can tell which characters are `Pi`, but not identify the matching end-quote characters, nor does it seem to have BidiBrackets.txt.

Kodiologist:
Okay, `q#<foo>` and balanced delimiters are in there now.

gilch:
I just realized, this will prevent us from using an other tag macro symbols that start with `q`.

The obvious fix is to make them use the same rules as other tag macros, so `#q X` with `X`, and `#q <foo>`. But `#q[`, still works.

Kodiologist:
(Kek, I can't keep my own syntax straight.)

Kodiologist:
> I just realized, this will prevent us from using an other tag macro symbols that start with `q`.

Quite so.

> The obvious fix is to make them use the same rules as other tag macros, so `#q X` with `X`, and `#q <foo>`. But `#q[`, still works.

I would rather not allow or require this whitespace. It looks weird and leads to a construct like `#q X` having non-significant whitespace (where all runs of whitespace are treated the same) on the left of the delimiter (`X`) and significant whitespace (where each whitespace character is preserved literally) on the right. If we must have tag macros whose names begin with `q`, how about removing the plain style of q-string (`#qXfoobarX`) and leaving the balanced style (`#q(foobar)`) and pointy style (`#q<X>foobarX`)?

gilch:
Removing the plain style is fine with me. It seems kind of redundant given the balanced style.

The balanced style works fine with `#q[`, `#q(`, and `#q{`, since this does not violate the rules for tag macros.

The remaining styles are still a problem though. Naked symbols are currently allowed to start with `<` and higher Unicode, including your other brackets. #1117 would help with a lot of our symbol concerns, since naked symbols could contain arbitrary strings with escapes.

Do we need to support so many bracket types for the balanced style? Users could be confused about which characters are allowed in tags. Are the three bracket types enough? If not, would the addition of  `«` and `»` to the other brackets suffice? We could add just one more character, `«`, to what ends an identifier. That wouldn't be too hard to remember.

Is requiring the whitespace so bad for the pointy style? It's clear what's in the delimiter from the `<>`, so it's also clear the separating whitespace isn't. It would usually be used for multi-line strings anyway, so the slight extra length is not a big deal.

Kodiologist:
Even if the only change from what I have now is to remove the plain style, the only way you could get your tag macro shadowed is if its first character is `q` and its second character is `<` or one of the bracket or quote delimiter characters, which would make a strange name for a macro.

`«` (and the others) ought to stay identifier characters so long as they have no special meaning outside of `#qC`. My proposals A and B in #1287 were to use a fixed character for quoting rather than a pick-your-own-quotes operator, but they didn't seem very popular.

Kodiologist:
@kirbyfan64 @tuturto Can you guys weigh in on what you would accept?

tuturto:
Out of all options, I like balanced style best.

Kodiologist:
@tuturto Yeah, but see the back-and-forth between me and Matthew above. Would you accept dropping the plain style but still not requiring or allowing whitespace after the `q`?

tuturto:
Plain style can go. I don't have strong opinion about whitespace, but I'm actually slightly leaning towards requiring it. But for me either (whitespace, no whitespace) is ok.

Kodiologist:
Okay, thanks. @kirbyfan64?

Kodiologist:
Looks like we're Ryanless for this PR. @gilch, would you accept a spaceless balanced style with all the delimiters allowed here? You're proposing a different PR that will prevent calling tag macros whose names begin with `_` (#1354), so perhaps you'll also allow restriction of weirder prefixes like `q⁅` and `q⸌`. If not, I'll take your compromise in which `#q(`, `#q[`, `#q{`, and `#q«` are the only forms of the balanced style.

gilch:
> You're proposing a different PR that will prevent calling tag macros whose names begin with `_` (#1354)

I did notice that, and originally thought about implementing it differently. But Clojure also does it that way. I checked. But I'd rather not add any more exceptions on what's allowed in tags.

(I think it would be nice if Hy's parser could read EDN. This would make it easier to use a nice exchange format with the Clojure ecosystem. With the addition of the `#_` discard syntax, I think it can.)

I've also become very worried about how *any* new string syntax will interact with our tooling. Hy is currently similar enough to Clojure that we can use Clojure editors on Hy with fairly good results. It's too bad Clojure doesn't have something like this already. If we add something new, it's very important that vim-hy and hy-mode (with Parinfer!) can support it. So I'd rather keep the new syntax simple. The Balanced Style would work in most cases, but for arbitrary strings, we need some kind of custom delimiters. (Can we escape a terminating bracket?) I think we need to discuss it more, and try out some of these proposals with the editors before we put it on master.

Kodiologist:
Surely the editor modes exist to support the programming language rather than the reverse, and hence we should be modifying the modes to suit the language rather than modifying the language to suit the modes. And it's nice if Clojure tools also work for Hy, I guess (I've never used one), but it doesn't make much sense to hold back the development of the language to maintain this quasi-compatibility with Clojure. Clojure compatibility has never been a goal for Hy.

> The Balanced Style would work in most cases, but for arbitrary strings, we need some kind of custom delimiters.

Isn't that what the pointy style is for? We were talking about dropping the plain style, but keeping the pointy style as well as the balanced style, so far as I understood.

> I think we need to discuss it more

What would you like to discuss?

gilch:
> Surely the editor modes exist to support the programming language rather than the reverse,

I'm not so sure. New languages have a chicken/egg problem. A new language is not very useful without an ecosystem to support it--libraries, tooling, docs, learning material. So no-one uses it. But no-one wants to write libraries, tooling, docs, learning material for a language no-one uses. Chicken/egg.

The general solution to a chicken/egg problem is bootstrapping. Hy can use Python libraries. Hy can use Clojure editors. Hy is already useful enough to attract some users, who might start writing stuff in/for Hy.

But if you want to use Python libraries, why bother learning Hy? Just use Python. Besides borrowing an ecosystem, we also need to add something to it, or no-one will bother with us. What's our selling point? What's our killer app? Not Python's libraries. Python obviously has easy access to them them, but so does Julia--which is already more popular than Hy. Julia is expression-oriented. Julia even has macros. (And actually, so [does](https://github.com/delapuente/mcpy) [Python](https://pypi.python.org/pypi/MacroPy/1.0.3)).

So that leaves structural editing. It's the big reason why anyone would use these weird S-expression things. It's hard to implement or use structural editing in other languages because they don't have the regular syntax. Hy is attractive to those who already know another Lisp. They'll expect a good editor, but don't expect them to write it. It's on us here at Hy Society.

> Isn't that what the pointy style is for? 

Yes. I forgot you wanted them all to be raw strings.

> What would you like to discuss?

@kirbyfan64 was still leaning towards Lua Style instead of Pointy Style. (Is that still your preference?) This wouldn't require a space after the `#q` either, but would remove the `#q[` possibility from the Balanced Style, which is the only reason I went with the Pointy Style. But now that whitespace is an issue with Pointy Style, I think Lua Style might be better. @tuturto what do you think about using Lua Style?

I'm also wondering if we should strip a leading indent somehow, like how Python processes its docstrings. I'm not sure how this should work though. It might be more trouble than it's worth, but I thought I'd throw that out there for discussion.

kirbyfan64:
Note to self: I need to stop mass-marking-as-read GitHub issues... (It was going to bite me eventually anyway...)

I'd still prefer the Lua style. After working with a crapton of different languages with different ways of having multi-line not-strings, the Lua style is the only one that I really don't dread using.

FWIW automatically removing indent would just be a [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) call, but I'm not sure everyone would *always* want it, even though it's certainly the more common scenario.
 call

Kodiologist:
What exactly do we mean by the Lua style? Like, openers of the form`q[===[` for some number of `=` signs, or with any string for the `===` other than an opening square bracket?

@gilch I doubt that ParEdit, Parinfer, etc. are what have attracted most people to Hy so far, and I doubt that most existing Hy programmers use them. Nor are they the most obvious motivation for a heavily parenthesized syntax.

@kirbyfan64 Are you saying you wouldn't support allowing any of the other styles, or merely that the Lua style should be supported among others?

Kodiologist:
Regarding the leading indent, I don't care too much. I'd lean against it, because it's easy for the programmer to remove but annoying to re-add if you have to, but I'd accept this either way.

gilch:
> What exactly do we mean by the Lua style? 

By "Lua Style", I meant what we had discussed previously, opening like `#q[foo[` (possibly with newline) and closing like `]foo]`, where `foo` could be any string not containing `[]`. (Commonly `#q[[` or `#q[=[`, I expect.)

I like the custom delimiters. Besides being shorter when you have to nest code containing its own custom delimiters, editors could also be configured to highlight contents of these strings differently, e.g. a `#q[Python[` string could by highlighted like Python code and a `#q[HTML[` string like HTML, etc. But Hy's parser will just call it a `HyString` regardless of the `foo`. The parser could even add that as metadata to the HyString object, which the compiler would simply ignore. But other static analysis tools, or even macros could use it, since they'd have access to the Hy model.

Alternatively, we could get closer to Lua's style and drop the `#q` dispatch with a more extreme change to Hy's grammar. Use a regex like `\[=+\[` for the start of a string. Unlike Lua, we'd still need `[[` to parse as a list in a list, but otherwise they'd work exactly like Lua's long brackets. This does create a strange case where `[=[]=]` is a string but `[= []=]` is parsed as `HyList([HySymbol('='), HyList(), HySymbol('=')])`. We'd have to restring the `foo` to some number of `=` in this style, to avoid other strange cases like that. I  really don't like this as much, and we don't get the cool features of arbitrary delimiters in macros, editors, analysis, etc.

> I'd still prefer the Lua style. After working with a crapton of different languages with different ways of having multi-line not-strings, the Lua style is the only one that I really don't dread using.

@kirbyfan64 can you elaborate more on why you like Lua's style, and what you don't like about the other languages? We can't have *exactly* Lua's style in Hy without doing weird things to our list syntax. So I want to know what it is about it that you consider important.

> Regarding the leading indent, I don't care too much. I'd lean against it, because it's easy for the programmer to remove but annoying to re-add if you have to, but I'd accept this either way.

> FWIW automatically removing indent would just be a textwrap.dedent call, but I'm not sure everyone would always want it, even though it's certainly the more common scenario.
call

This is fair. And another tag macro could make that `dedent` call, like `#dedent #q(...`. *Explicit is better than implicit.* It might be nice if there were a way to define a different string prefix that did this, like `#Q` which strips indents, but since `#q` would be baked into the grammer, I don't see how this would be possible.

Actually, this gives me another idea. Forget the `#q` dispatch, and use `#[` instead. We're not currently using that for anything, and neither does Clojure. So strings would open as `#[foo[` (possibly with newline) and close as `]foo]`, where `foo` is any string not containing `[]`. And have the parser put the `foo` in the HyString model as metadata.

@Kodiologist would this suffice for your command-line use case? Or do you need some kind of non-`[` delimiter too?

gilch:
We could still use the balanced style by special-casing `#[(`, `#[{`, and maybe `#[«`, if we *need* it. These would close as `)]`, `}]`, and `»]`, respectively. This is only one `]` longer than the current proposed balanced style. 

But wouldn't the `#[foo[` style good enough by itself? That would simplify things.

Kodiologist:
It's a fine replacement for the pointy style, where flexibility is the chief concern, but somewhat verbose, so it's not a great replacement for the balanced style or the plain style. Even the shortest form, `#[[ … ]]`, is a character longer than the already weighty `#q( … )`. Hence my proposals A and B back in #1287.

gilch:
I guess the original Proposal B, ```#` ```/``` ` ``` would be even more concise than the `#q` Balanced Style. I'm warming up to B, but I don't think `#[[...]]` is that bad, since you'd only need it in the rare cases when `"` isn't good enough. I'm still not okay with requiring Unicode as in Proposal A.

But I'd also like to hear what @kirbyfan64 and @tuturto think about the discussion so far.

Kodiologist:
> I'm still not okay with requiring Unicode as in Proposal A.

It's not required; we'll have the Lua style too. :)

But B is probably just as good because in the case that you need to use a backtick, you can escape it or use the Lua style.

gilch:
> It's not required; we'll have the Lua style too. :)

That's actually a *really* good point. Given the `#[foo[` style, I think I'd be okay with a `«` and `»` balanced style that worked just like `"`/`"` otherwise. So `«` and `»` are allowed inside if balanced, escapes (including `\»`) would work inside unless you use `r«`.

We could even support the other prefixes--`b«` and `f«` in the balanced style.

Kodiologist:
Right, proposals A and B both need escaping and the ability to use the `r` prefix (and we might as well throw in the other prefixes while we're at it) because you can't customize the delimiters. In the case of B I imagine the prefix would go before the `#`.

gilch:
In the case of B, I'd actually put the prefix between the hash and the opening backtick. This would be simpler and more consistent in Hy's grammar.

kirbyfan64:
Hold on I need to re-read this thing. I just got super lost...

Kodiologist:
Don't worry, I've gotten myself utterly confused, too.

kirbyfan64:
Yeah, this is a mess. I still have no clue what the hell "pointy style" is. I think we just need to start from the top. How about we start from the top...

My understanding is that the current state of the code is that is looks like `#q{delim}STUFF delim`, right?

Here's my belief:

Why is the Lua syntax awesome?

For starters, it's dead-simple if you're not using custom delimiters (which is the case most of the time). e.g. `#q[[abc]]`. On the other hand, the current syntax *always* requires a delimiter, e.g. `#q{EOF} stuff EOF`. My stance (partly from using heredocs for a long time) is that:

- 99% of the time the delimiter will either be *EOF* or some weird character, where everyone uses a different one
- There are too many choices. Not only can you pick any delimiter, but you can also pick what delimits the delimiter (e.g. `#qAdelimiterA` is valid). Really, I don't see any use for using a custom character after the `q`, since the delimiter can be whatever you want anyway.

In addition, it plays nicely with Python's *one way to do it* philosophy: it allows just enough flexibility, but most of the time it'll end up being roughly the same thing.

kirbyfan64:
Also, I'm not sure what this has to do with list syntax. Isn't everything handled in the lexer anyway?

Kodiologist:
>  I still have no clue what the hell "pointy style" is.

It looks like `#q<FOOEY>my stringFOOEY`.

> My understanding is that the current state of the code is that is looks like `#q{delim}STUFF delim`, right?

No. The current state of the code allows three styles, none of which allow `#q{delim}STUFF delim`:

1. The plain style: `qXmy stringX`.
2. The balanced style: `q{my string {still part of my string}}`.
3. The aforementioned pointy style, `#q<FOOEY>my stringFOOEY`. It's called the "pointy" style because you have to use ASCII angle brackets.

> Also, I'm not sure what this has to do with list syntax.

Matthew wants a syntax that does not restrict what tag macros you can call, compared to what's already the case. `#[` currently isn't a valid tag macro call, whereas `#«` is, because `[` lexes as the start of a list.

Kodiologist:
> `#[` currently isn't a valid tag macro call, whereas `#«` is, because `[` lexes as the start of a list.

Actually, that's not quite true. The real reason `#[` can't parse as a tag macro call is because a tag macro call can only contain identifier characters, and `[` isn't an identifier character (the regex that defines `identifier` is in fact `[^()\[\]{}\'"\s;]+`).

gilch:
@kirbyfan64 

What @Kodiologist said.

>  no clue what the hell "pointy style" is

Like `#q <END>` string `END`. But @Kodiologist didn't like the space, so he proposed `#q<END>`, which interferes with tag macros.

> what this has to do with list syntax

That's only if we don't have the `#q` as the dispatch and use `[=` instead, which looks like a list and a symbol. So I proposed using `#[` instead.
***
> How about we start from the top

So my current best proposal is both

1. `#[` dispatch

`#[foo[` <remove 1 newline, if present> <string> `]foo]` where `foo` is any string that does not contain `[]`. This means both `#[[...]]` and `#[=[...]=]` are valid. Like Lua, escapes are disallowed, since these are raw strings, and it will also strip one newline from the start, if present.
```
#[[
spam]]
```
is just like `"spam"`.

But unlike Lua, you have to start with `#`, or it's a list. And also unlike Lua, you're allowed to use a delimiter besides some number of `=`. This could be parsed into the Hy model as metadata for use with macros. It could also be a hint to editors to use some other highlighting mode when quoting some other language, e.g. `#[Python[...]Python]` would be highlighted like Python. But most of the time you'd use `#[[` or some number of `=` when nesting like `#[===[`, if you don't have a better idea for the delimiter. And we could say so in the Style Guide.

And

2.  The `«` and `»` balanced style

This is just Proposal A from #1287. It works just like the normal double-quote style, but since it's paired, you can have `«` and `»` inside of it as long as it's balanced (or escaped). (And `"`, of course). It would have the normal string prefixes, like `r«` for a raw string.

It's really too bad double quotes aren't paired as a historical artifact from typewriters, or we'd already be doing this in Python. But given the `#[foo[` style and normal double quotes, Unicode is still not required for developing Hy, so I don't mind adding this as an option.

Kodiologist:
I endorse this proposal and will implement it if it gets sufficient traction.

Kodiologist:
Although, I think that allowing unescaped nested guillemets is probably a bad idea because (1) it complicates the lexer, (2) it will complicate syntax coloring (which will no longer be possible with non-recursive regexes alone), and (3) you're much less likely to need nested guillemets than you would nested parentheses or square brackets or whatever, and if you do, there's still the Lua style.

kirbyfan64:
Same. (TBH I still don't really grasp the use case for proposal 2, but I don't feel like figuring out anything else...)

Kodiologist:
@gilch It looks like at least the three of us (you, me, and Ryan) agree on your proposal. Can you just comment on the question of nested guillemets? Then I'll write it up and make a new PR for it.

gilch:
We seem to agree on the Lua strings, at least.

>  (1) it complicates the lexer,

Not too much though. And it could get better in the future. I wonder what the chances are of that `regex` library replacing Python's standard-library `re` someday. We might also consider a pull request for rply to make the lexer handle this kind of thing better.

> (2) it will complicate syntax coloring (which will no longer be possible with non-recursive regexes alone),

A *very important* concern. We do *not* want to make tooling for Hy difficult to implement. Yes, we'd need the equivalent power of a pushdown automaton. But PDAs are old technology. This is a *solved problem*. Both Vimscript and Elisp are Turing Complete. They can handle it even if their regexes can't.

Python has been called "executable pseudocode". We could publish the Python algorithm to match these strings in our docs under a public domain license for future tool writers to copy in whatever language they prefer. We could publish the equivalent PCRE recursive "regex" too.

Both Perl and Ruby have balanced-style strings with the same issue. How much of a problem *is* it for them, really? How does their tooling handle it? Any general-purpose editor or syntax highlighting library that can handle those properly could also handle Hy with an appropriate script. If you're using an editor that can't, you could escape them anyway. (And consider getting a better editor.)

>(3) you're much less likely to need nested guillemets

True, but I think paired string delimiters imply this feature. Not having it would be *weird*. There might be use for it in i18n. It does seem like a lot of problems for a feature we'd rarely use. I'd also be okay with *only* the Lua strings though.

Guillements are an interesting choice. If we're using Unicode anyway we could have used the English-style 66/99 quotes with `“` `U+201C` and `”` `U+201D`. But they might be harder to distinguish than the guillements in some fonts. And guillements are probably easier to type on most systems. Unfortunately, some languages (like German) like to use them backwards, »like this«. The other way is more common, but we could support both, by allowing guillement strings to start with `»` too. In that case, the nested guillements would also need to be backwards, or escaped.

I'd like to think about this some more.

gilch:
I think I've got a good alternative. Implementing #1117 would allow us to create HySymbols from arbitrary strings using `|` `<string>` `|`. This is a single delimiter like `"` is, so we could avoid the need for balancing logic. We could use a tag macro to convert it back to a string at compile time. So, something like `#q |foo|`, or even `#q|foo|` could work, depending on how we set up the lexing for `|`-quoted symbols. This is still one character shorter than `#[[foo]]`. Also, a HySymbol is-a Python string, and will work as one in most contexts, so you often wouldn't even need the tag macro. Just use `|foo|`. We could also use a different tag macro to convert it differently, like `#b` for a bytestring or `#f` for a format string, etc.

Kodiologist:
> Both Perl and Ruby have balanced-style strings with the same issue. How much of a problem *is* it for them, really? How does their tooling handle it?

I can't speak for Ruby, but the complexity of Perl's quoting forms tends to be behind the corner cases that Perl syntax highlighters have trouble with. That's probably the hardest Perl feature to highlight, except perhaps the magic variables with weird names like `$'` (and modern Perl should generally `use English;` so you can use the longer names instead, anyway).

> If we're using Unicode anyway we could have used the English-style 66/99 quotes with `“` `U+201C` and `”` `U+201D`. But they might be harder to distinguish than the guillements in some fonts.

Agreed, and hard to distinguish from ASCII double quotes, too.

> Unfortunately, some languages (like German) like to use them backwards, `»like this«`. The other way is more common, but we could support both, by allowing guillement strings to start with `»` too.

Oh dear, that sounds like a recipe for insanity. If annoying or confusing people who use guillemets in the opposite direction in their native language is a concern, it's probably better to choose different characters, like `‹` and `›` or `⁅` and `⁆`.

My first objection to quoting with vertical brackets would be, how does the parser tell whether `(foo | x |)` is a function of one argument, the symbol named `" x "`, or three arguments, the first and third of which are the `|` function (the shadow version of the bitwise OR operator from Python)?

gilch:
`(foo | x |)` would *clearly* be parsed as `HyExpression([ HySymbol('foo'), HySymbol(' x ')])` under #1117.

This does mean we can no longer write `|` as our bitwise or. We could certainly spell it out as `bit-or` instead though. We did that with `=` and `setv`. (We'd want to spell out `bit-and` too for consistency.) `\|` could also work, but I don't think it's as pretty.

kirbyfan64:
FWIW is there a reason we can't just use PLY? It works much better with things like this.

Kodiologist:
I'd rather not rename a bunch of operators when we could just use other syntax. So, if there are no objections, I'm going to implement Lua style and a guillemet style supporting nesting. We've rather drawn this out, so it will be nice to conclude it.

gilch:
> I'd rather not rename a bunch of operators when we could just use other syntax.

We'd have to do that for #1117 anyway, which is important for a lot of other issues. I'd rather not have three kinds of string literals when two will do. A change in grammar is a much bigger deal than renaming some core operators. And it's confusing that bitwise-not is the same symbol as unquote, so I want to change that one anyway. For example,
```
`(~(foo))
```
Does the above immediately call `foo` and unquote its result, or put `(foo)` in the expansion and bitwise-negate it? Obvoiusly the repl can tell you. But `(-(foo))` would negate it.

Freeing up `&` from bitwise-and would allow us to reserve it in function calls. So we could use the shorter Clojure style `(foo [& args]...)` instead of `(foo [&rest args])`.

And if we free up `|` from bitwise-or, we can have both arbitrary-string symbols and a concise non-double quote string syntax with only one grammar change instead of two. This would make the guillemet style redundant.

We can leave the other bitwise operators alone.

gilch:
So for this PR, it's fine if it's just the `#[foo[` Lua style. While a little more verbose that we might like in some cases, that's good enough until we implement #1117 with a `#q` tag macro to convert the symbol name to a string in another PR. But if you want to do #1117 here too, that's fine.

Kodiologist:
> A change in grammar is a much bigger deal than renaming some core operators.

The grammar has to change in any case in order to implement a new form of symbol quoting or string literal or whatever.

Wouldn't it be better to quote with some syntax that doesn't require changing an operator from the Python name, like `\ … \` or `#( … )` or `` #` … ` `` or `« … »`?

And seeing as people are going to use funny characters in strings more often than in symbols, shouldn't we have the default for the syntax be a string rather than a symbol?

Kodiologist:
Related to that last point, easier entry of symbol with weird names doesn't actually need a new quoting syntax—it would suffice to add another prefix to string literals that makes the result a symbol instead of a plain string.

Kodiologist:
> FWIW is there a reason we can't just use PLY?

I don't know, not having used it, but rewriting the lexer and parser is out of the scope of this PR.

gilch:
> FWIW is there a reason we can't just use PLY?
> rewriting the lexer and parser is out of the scope of this PR.

@kirbyfan64 PLY has a compatible BSD license, so I don't know of a reason. In what way is it better though? Can it do this kind of nested parsing we've discussed for the balanced styles any better without adding a new regex engine dependency? If so, and if we settle on a balanced style, it might be worth it. But @Kodiologist pointed out that *any* balanced style would complicate our tooling.

gilch:
> The grammar has to change in any case in order to implement a new form of symbol quoting or string literal or whatever.

But it would be a less complex grammar with only the two string literals and the arbitrary symbol syntax, than with three string literal types and the arbitrary symbol syntax.

> Wouldn't it be better to quote with some syntax that doesn't require changing an operator from the Python name, like \ … \ or #( … ) or #` … ` or « … »?

I actually think the `|` and `\` syntax is better for symbols. Freeing up those four names from the operators would help with other issues. But if the others also feel very strongly that we must not rename our bitwise-or, then `||` symbols are not an option. And we'd have to come up with some other syntax for #1117. It's not like Hy has infix notation. Bitwise operators aren't even used that much in Python. But if you *are* prone to bit-bashing, you'd have to use parentheses and indentation to make the expression readable. So I don't think slightly longer names are a big deal.

I'd rather keep `\` as an escape character for symbols even if we don't use the `|` quoting, like in Emacs Lisp. So `\...\` is out.

`#(...)` means `xi` in Clojure. Using it for balanced strings might confuse the Clojure users, but I think it's actually not an unreasonable choice. It's only one character longer than `""`, without introducing Unicode syntax. It also has a nice parallel with `#[[]]`. If we wanted a prefix version of `xi` for the Clojure people, something like `#%(f %1 %2 ... %&)` would work about as well instead. We also considered it for genexprs #867, which isn't `xi` either. They could use a slightly longer tag, but it wasn't clear what when we only had one character sharp macros. Maybe `#for` would work. Balanced strings do have the aforementioned problem with tooling, but if we *are* doing paired delimiters, I want them balanced. I'm not sure if I like this better than the guillements or not (or both?), but I'm leaning towards `#(...)`. It's fine to allow the users to use Unicode, but I'd rather not add any to Hy itself.

We've already discussed A and B, but one of them is paired and the other isn't.

> And seeing as people are going to use funny characters in strings more often than in symbols, shouldn't we have the default for the syntax be a string rather than a symbol?

No, because we want to be able to use funny characters in tag macros. We've already got the super-short `""` notation for strings with an overhead of just two characters. We'd almost always just use that. And we'll be adding `#[[]]` too for crazy raw strings, which are probably not short.

It feels like we're quibbling over saving one character. Compare the overhead of Lua `#[[...]]` (5) to the overhead of Plain `#qX...X` (4), which you thought was good enough before (if barely?). Note that tagged symbol `#q|...|` is also 4.

We could perhaps have `r|foo bar|` be the same as `r"foo bar"`. But that would be ambiguous with `#tag-r|foo bar|`. Is that the tag `tag-r` applied to symbol `|foo bar|` or is that tag `tag-` applied to string `r"foo bar"`. If we always assumed a separation before the`|` then it's unambiguous. And we could do the `#q|foo bar|` tagged symbol strings.

I'm still not getting the use case where neither `""` nor `#[[]]` is good enough, and neither was @kirbyfan64. Unicode is hard to type on most systems, so it's not like we're saving keystrokes. And an Emacs binding could insert `#[[` just as easily as `«`. Guillemonts sure are prettier, but you can also have Emacs abbreviate things with Unicode for your reading pleasure, like `λ` for `fn` or `xᵢ` for `xi`. Why not `«` for `#[[` and `»` for `]]`? Can we do that last one to the Lua-Style strings without also doing it to nested lists? If not, something like `#[;[ ];]` should be unambiguous. This way you can read Unicode, but still be compatible with ASCII tools and files.

So, compared to Lua Style, it's not easier to read (on Emacs). It's not easier to write (unless you have a foreign keyboard, and even they require AltGr). It's not easier to implement tooling for. Many terminals can't print Unicode, so it's not very good for the command line either, and Unicode is still hard to type. I could maybe see `#(...)` being better here, but barely. It's all of `[]` shorter.

Kodiologist:
For what it's worth, I use the bitwise operators much more often for vectorized logic with NumPy or Pandas objects than for actual bitwise logic.

> It feels like we're quibbling over saving one character.

Yeah, we've been bikeshedding mercilessly about this since the beginning. Somebody's gotta give in order for this to end. So, I give. Would you accept a PR that adds the Lua style (without nesting of the customized delimiter) and doesn't do anything else?

gilch:
> Yeah, we've been bikeshedding mercilessly about this since the beginning.

I didn't see the discussion as trivial. The grammar is something we should try to keep simple and understandable. Changes here should be carefully thought out in the context of the whole language instead of blindly accepting the first thing that seems like a good idea. I felt like the options have been improving because of the discussion. I saw good points that I hadn't thought of on my own.

> Would you accept a PR that adds the Lua style (without nesting of the customized delimiter) and doesn't do anything else?

A single change could probably get approved faster than two.

Which version exactly are you proposing? I'd accept the version like `#[foo[` (remove one newline, if present) string body `]foo]`, where `foo` is any string not containing `[` nor `]`, including the empty string, and the `foo` is available as metadata in the `HyString` model.

This is not a balanced style, so a pure FSM regex engine could highlight it, e.g. `#[foo[ #[foo[ bar ]foo] baz ]foo]` would end at the *first* matching, `]foo]`, and be equivalent to the string `" #foo[ bar "`. The remaining ` baz ]foo]` wouldn't get highlighted and might be a syntax error depending on context. No need for a PDA. If you want to nest these, you have to pick a different `foo`, usually just by adding another `=`.

Kodiologist:
> Which version exactly are you proposing? I'd accept the version like `#[foo[` (remove one newline, if present) string body `]foo]`, where `foo` is any string not containing `[` nor `]`, including the empty string, and the `foo` is available as metadata in the `HyString` model.

Yeah, that one.

Issue #1242, opened 2017-02-27
Add Clojure's seq abstraction


gilch:
We're running into occasional annoying problems because the ubiquitous Python iterators are mutable. Clojure doesn't have this problem because it uses the immutable seq abstraction instead. A seq can still have the lazyness of some iterables, but getting the next item won't change what the first seq means.

I've been playing around with a Python implementation of seq. Here's a rough attempt. This could easily be translated to Hy.
```Python
from collections.abc import Iterable

class LazyCons:
    __slots__ = ['car', 'cdr']
    def __init__(self, car, cdr):
        self.car = lambda: car
        def lazy_cdr():
            res = cdr()
            self.cdr = lambda: res
            return res
        self.cdr = lazy_cdr
        
class Seq(LazyCons, Iterable):
    __slots__ = ()
    def __init__(self, iterable):
        it = iter(iterable)
        car = next(it)
        super().__init__(car, lambda: Seq(it))
    def __iter__(self):
        def iterator():
            seq = self
            while 1:
                yield seq.car()
                seq = seq.cdr()
        return iterator()
```
And a function to demonstrate the laziness.
```Python
def verboseRange(*stop_start_step):
    for i in range(*stop_start_step):
        print('yielding', i)
        yield i
```
And the demonstration:
```
>>> spam = Seq(iter(verboseRange(5)))  # note iter() call
yielding 0
>>> spam.car()
0
>>> spam.car()
0
>>> spam.cdr()
yielding 1
<__main__.Seq object at 0x0000019C4D773D68>
>>> spam.cdr().car()
1
>>> list(spam)
yielding 2
yielding 3
yielding 4
[0, 1, 2, 3, 4]
>>> spam.car()
0
```
That `iter()` call isn't required, but it demonstrates converting a mutable iterator into an immutable seq (provided nothing else has a reference to that iterator).

Some more ideas: 

Clojure often converts collections to seq's automatically. Rather than add that to a lot of Hy functions, we might instead have a short reader macro to do it. Although, `(seq xs)` is not that long to begin with.

We could also change the `__repr__` of a `Seq` to print off up to the first 20 items or something. This would make seq's a lot more friendly in the repl, and lazy Python iterables too especially, when combined with the reader macro.

Kodiologist:
How does this proposal compare to @tuturto's `hy.contrib.sequences`?

gilch:
[...]

Issue #1125, opened 2016-10-02
Rename `for` to `doiter` and then `for*` to `for`.


gilch:
See #1124, #874. This step would make Hy more consistent with both Python and Clojure, and make Hy less confusing.

gilch:
We also need to update the docs with this. In particular, mention in the documentation of `for` (formerly `for*`), that Clojure users that want a list comprehension should use `genexpr` instead.

allison-casey:
Obsolete since the unification of for with comprehensions in #1626

PR #1328, opened 2017-07-20
add destructure to contrib

### LINKED ISSUE #1326: implement something like Clojure's destructure macro, at least for mappings
See https://clojure.org/guides/destructuring.

Python already has `a, b, *foo = bar` syntax in assignments. Hy hasn't even implemented all of this yet. #1127.

Clojure's version seems a lot more powerful. It can keep the whole thing with `:as`, and destructure mappings, including `:or` defaults, and use `:keys`/`:strs`/`:syms` to avoid duplicating names.

Python can't destructure maps like that, last I checked.

But Clojure's destructuring can't do `a, *b, c = foo` assignments like Python can. The `&` part has to be the rest of the seq, so there's nothing left for `c`.

There should preferably be only one obvious way to do it, and there's some overlap. But Python can't destructure maps at all right now. (And that might change someday.)

So, rather implementing it everywhere Clojure does in Hy, I propose just implementing (in contrib) something like the "undocumented" `clojure.core/destructure` macro that backs all of these, or at least the associative destructuring part for mappings. (That way we can deprecate it if Python ever gets around to destructuring mappings.)

I'm not sure about the sequental part, but if we do that too, the `&` should be named `:&` instead, since we're already using `&` for Python's bitwise and. Depending on #1127, we might be able to use tuples instead of vectors, possibly with `#*`.

Each version has features the other doesn't (Clojure's `:as`, and Python's `*a, c =`). Maybe we could use `[]` for the Clojure version and `(,)` for the Python version, but then there's not one obvious way to do it. Maybe we could just add the `:as` feature to `(,)` bindings in the macro somehow?

gilch:
Closes #1326.

It should probably have tests and better documentation before the merge.
But I thought I'd take comments first.
It does have some docstrings now and I've tried it at the repl.

### DIFF
--- a/hy/contrib/destructure.hy
+++ b/hy/contrib/destructure.hy
@@ -1,0 +1,93 @@
+;;; Hy destructuring bind
+;; Copyright 2017 the authors.
+;; This file is part of Hy, which is free software licensed under the Expat
+;; license. See the LICENSE.
+(import [hy.models [*]])
+
+(defmacro =: [&rest pairs]
+  "Assignment with destructuring for both mappings and iterables."
+  `(setv ~@(chain.from-iterable (genexpr (destructure binds expr) [(, binds expr) (partition pairs)]))))
+
+(defn destructure [binds expr]
+  "
+  Destructuring bind.
+  The binds may be nested.
+  :as and :& are magic in [] binds. See dest-list.
+  :as, :or and :from are magic in {} binds. See des-dict.
+  "
+  (if (isinstance binds HySymbol) [binds expr]
+      (isinstance binds HyDict) (dest-dict binds expr)
+      (isinstance binds HyList) (dest-list binds expr)
+      (raise (SyntaxError (+ "Malformed destructure. Unknown binding form:\n"
+                             (repr binds))))))
+
+(defn dest-dict [binds expr]
+  "
+  Destructuring bind for mappings.
+  The binding forms may be nested.
+  Targets from binds look up their value.
+  For example, try ``(dest-dict {x :a  y :b} {:a 1  :b 2})``
+  Use the ``:from [foo :bar \"baz\"]`` option to
+  bind a symbol, keyword, or string to the same name.
+  For example, try ``(dest-dict {:from [:a :b c]} {:a 1  :b 2 'c 3})``
+  Use the ``:as foo`` option to bind the whole mapping to ``foo``.
+  Use the ``:or {foo 42}`` option to to bind ``foo`` to ``42``
+  if ``foo`` is requested, but not present in expr.
+  "
+  (setv ddict (gensym 'ddict)
+        ;; First, assign expr to a gensym to avoid multiple evaluation.
+        ret [ddict expr]
+        default (dict-comp k [v]  ; optional, so wrapped for splicing.
+                           [(, k v) (partition
+                                     ;; find :or's HyDict, if present.
+                                     (next
+                                      (genexpr (second x)
+                                               [x (partition binds)]
+                                               (= (first x) ':or))
+                                      {}))]))
+  (for [(, target lookup) (partition binds)]
+    (if (= :or target) (continue)
+        (= :as target) (.extend ret [lookup ddict])
+
+        (= :from target)
+        (.extend ret (chain.from-iterable
+                      (genexpr [(symbolfy key)  ; implied target
+                                `(.get ~ddict
+                                       '~key  ; lookup
+                                       ;; append :or default, if present.
+                                       ~@(.get default (symbolfy key) []))]
+                               [key lookup])))
+
+        (.extend ret (destructure  ; recursion
+                      target
+                      `(.get ~ddict
+                             '~lookup
+                             ;; append :or default, if applicable
+                             ~@(if (isinstance target HySymbol)
+                                 (.get default target [])
+                                 []))))))
+  ret)
+
+(defn symbolfy [s]
+  (-> s name HySymbol)) ;; TODO: mangling?
+
+(defn dest-list [binds expr]
+  "
+  Destructuring bind for sequences.
+  The binding forms may be nested.
+  Targets from binds are assigned in order from the expr.
+  Use ``:& bar`` option in binds to bind the remaining iterable to ``bar``
+  Use ``:as foo`` option in binds to bind the whole iterable to ``foo``.
+  For example, try ``(dest-list [a b [c :& d :as q] :as full] [1 2 [3 4 5]])``
+  "
+  (setv dlist (gensym 'dlist)
+        ret [dlist expr]
+        ibinds (iter binds)
+        i 0)
+  (for [n ibinds]
+    (if (= :as n) (.extend ret [(next ibinds) dlist])
+        (= :& n) (.extend ret `[~(next ibinds) (cut ~dlist ~i)])
+        (do (.extend ret (destructure n `(get ~dlist ~i)))
+            (+= i 1))))
+  ret)
+
--- a/hy/contrib/destructure.hy
+++ b/hy/contrib/destructure.hy
@@ -6,7 +6,8 @@
 
 (defmacro =: [&rest pairs]
   "Assignment with destructuring for both mappings and iterables."
-  `(setv ~@(chain.from-iterable (genexpr (destructure binds expr) [(, binds expr) (partition pairs)]))))
+  `(setv ~@(chain.from-iterable (genexpr (destructure binds expr)
+                                         [(, binds expr) (partition pairs)]))))
 
 (defn destructure [binds expr]
   "
@@ -15,11 +16,20 @@
   :as and :& are magic in [] binds. See dest-list.
   :as, :or and :from are magic in {} binds. See des-dict.
   "
-  (if (isinstance binds HySymbol) [binds expr]
-      (isinstance binds HyDict) (dest-dict binds expr)
-      (isinstance binds HyList) (dest-list binds expr)
+  (defn test [x] (isinstance binds x))
+  (if (test HySymbol) [binds expr]
+      (test HyDict) (dest-dict binds expr)
+      (test HyList) (dest-list binds expr)
       (raise (SyntaxError (+ "Malformed destructure. Unknown binding form:\n"
                              (repr binds))))))
+
+(defn _check [seen magic target]
+  (if (= magic target)
+    (if (in magic seen)
+      (raise (SyntaxError (.format "Duplicate :{0} in destructure."
+                                   (name magic))))
+      (do (.add seen magic)
+          True))))
 
 (defn dest-dict [binds expr]
   "
@@ -37,6 +47,8 @@
   (setv ddict (gensym 'ddict)
         ;; First, assign expr to a gensym to avoid multiple evaluation.
         ret [ddict expr]
+        append ret.extend
+        seen #{}
         default (dict-comp k [v]  ; optional, so wrapped for splicing.
                            [(, k v) (partition
                                      ;; find :or's HyDict, if present.
@@ -45,31 +57,20 @@
                                                [x (partition binds)]
                                                (= (first x) ':or))
                                       {}))]))
+  (defn expand-lookup [target key]
+    [target `(.get ~ddict '~key ~@(if (isinstance target HySymbol)
+                                    (.get default target [])
+                                    []))])
   (for [(, target lookup) (partition binds)]
-    (if (= :or target) (continue)
-        (= :as target) (.extend ret [lookup ddict])
-
-        (= :from target)
-        (.extend ret (chain.from-iterable
-                      (genexpr [(symbolfy key)  ; implied target
-                                `(.get ~ddict
-                                       '~key  ; lookup
-                                       ;; append :or default, if present.
-                                       ~@(.get default (symbolfy key) []))]
-                               [key lookup])))
-
-        (.extend ret (destructure  ; recursion
-                      target
-                      `(.get ~ddict
-                             '~lookup
-                             ;; append :or default, if applicable
-                             ~@(if (isinstance target HySymbol)
-                                 (.get default target [])
-                                 []))))))
+    (defn test [x] (_check seen x target))
+    (if (test ':or) (continue)
+        (test ':as) (append [lookup ddict])
+        (test ':from) (append
+                       (chain.from-iterable
+                        (genexpr (expand-lookup (-> key name HySymbol) key)
+                                 [key lookup])))
+        (append (destructure #* (expand-lookup target lookup)))))
   ret)
-
-(defn symbolfy [s]
-  (-> s name HySymbol)) ;; TODO: mangling?
 
 (defn dest-list [binds expr]
   "
@@ -82,12 +83,15 @@
   "
   (setv dlist (gensym 'dlist)
         ret [dlist expr]
+        append ret.extend
         ibinds (iter binds)
+        seen #{}
         i 0)
-  (for [n ibinds]
-    (if (= :as n) (.extend ret [(next ibinds) dlist])
-        (= :& n) (.extend ret `[~(next ibinds) (cut ~dlist ~i)])
-        (do (.extend ret (destructure n `(get ~dlist ~i)))
+  (for [target ibinds]
+    (defn test [x] (_check seen x target))
+    (if (test ':as) (append [(next ibinds) dlist])
+        (test ':&) (append `[~(next ibinds) (cut ~dlist ~i)])
+        (do (append (destructure target `(get ~dlist ~i)))
             (+= i 1))))
   ret)
 
--- a/hy/contrib/destructure.hy
+++ b/hy/contrib/destructure.hy
@@ -112,9 +112,10 @@
                                       (,)))]))
   (unless (is gsyms None) (.append gsyms ddict))
   (defn expand-lookup [target key]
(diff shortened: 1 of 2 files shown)


Kodiologist:
This looks cool. I'll hunt for bugs.

Kodiologist:
### Design

It would make sense to make `=:` support any even number of arguments, with the same meaning as for `setv`.

What is the macro `dest` for?

You can't destructure an iterable that can't be subscripted (e.g. `(=: [a b c] (genexpr i [i (range 3)]))` doesn't work). Is this a bug or a necessary limitation?

Neither of `(=: {:from ["hello?"]} {"hello?" 5})` nor `(=: {:from ["is_hello"]} {"is_hello" 5})` binds to `hello?`. It's not obvious which should work, but it seems clear that at least one should work.

It's confusing that keys for `dest-dict` come *after* the variable, particularly since the magical destructuring keywords (`:as`, `:from`, and `:&`) come before the variable. That is, `{a :a  b :b  :as all}` should instead be `{:a a  :b b  :as all}`. If you want to make sure that `:as`, `:from`, and `:&` can still be used as regular keywords, try denoting the magic words with `&` instead of `:`, so `:as` would be `&as`, `:from` would be `&from`, and `:&` would be `&rest` or something.

### Implementation

`Exception` should probably be `TypeError`.

`(with-gensyms [x]) x` works, but I don't think it's supposed to; that is, `x` is supposed to be out of scope outside the `with-gensyms`. `with-gensyms` is a holdover from when we had `let` and expected to make it work someday. `(setv ddict (gensym))` does basically the same thing.

Expressions like `` `[~ddict ~expr]`` can be written like `[ddict expr]` now that we have #1314.

`for*` can be `for`. (There's no reason to use `for*` outside Hy core.)

### Testing

`(= [0 1 2 3 4 5] all)` is a Yoda condition

Avoid `all` as a variable name, since it's a Python builtin.

gilch:
I realized I put this up before I implemented `:or` for mapping defaults. I'm not quite sure how to implement that. What should it expand to?

#1326 was pretty underspecified. We're not following Clojure's version exactly, just trying to give Hy similar capabilities. But I have noticed some differences with Clojure's version.

Clojure uses `:keys` `:syms` and `:strs` (always with a vector of symbols) instead of the `:from` with the exact key type. And this `:from` probably works on anything `name` does, not just strings, symbols, and keywords, though this isn't in the tests. I'm not sure which it better, but maybe I haven't thought of something.

Clojure only allows one `:as` in vector bindings (it seems to stop looking after that). This PR allows more than one, and it seems to work sensibly, but this isn't in the tests.

Unlike Python's unpacking assignments, the iterable destructure will ignore any remaining arguments if you don't provide enough binding names. If you try this directly in a Python assignment (e.g. `a, b = [1, 2, 3]` that would be an error: `ValueError: too many values to unpack`. But Clojure also ignores the remainder like this.

If you try an associative destructuring on a non-mapping iterable, Clojure will build a map from it, with every other value as a key (like HyDict). This is important for how Clojure functions parse kwargs. The PR doesn't implement that, but I don't think it wouldn't be hard to add. Python already has kwargs for functions and Hy uses it, so I'm not sure if there's a need. But this capability might be useful for some other reason I haven't thought of.

Kodiologist:
> I realized I put this up before I implemented `:or` for mapping defaults. I'm not quite sure how to implement that. What should it expand to?

I would presume something like `` `(.get ~ddict ~v ~default)``.

gilch:
> What is the macro `dest` for?

I thought it had to be a macro to work, so I tried implementing it that way first. But when I added `=:` to wrap `setv` it turned out to be easier to implement as a function. I originally thought Clojure's `destructure` was a macro, but no, it's an ordinary function. I don't see the need for `dest` anymore, so I'll remove it.

> You can't destructure an iterable that can't be subscripted (e.g. (=: [a b c] (genexpr i [i (range 3)])) doesn't work). Is this a bug or a necessary limitation?

Underspecificaiton again. Python's iterables are not as nice to work with as Clojure's seqs. #1242. But there are ways to do this. If they were seqs we could simply use `nth` instead of `get`. But if you use an iterator for the `expr`, then `nth` will mutate it.

If we anticipate this an assign `(iter expr)` to the gensym, we can just use `first`, but then it gets exhausted before `:as` gets to it. If we went that way, we couldn't support `:as`.

The usual Python workaround for this kind of thing is to realize the whole iterator into a list. (This could also be done at the gensym assignment step.) And indeed, this is how Python's unpacking assignments work. But that's bad for infinite iterators. You might want to bind the whole thing with `:as`, and only take a few elements off the front. Maybe we could use `tee` to split the iterator.

gilch:
> It's confusing that keys for `dest-dict` come *after* the variable, particularly since the magical destructuring keywords (`:as`, `:from`, and `:&`) come before the variable. That is, `{a :a  b :b  :as all}` should instead be `{:a a  :b b  :as all}`. If you want to make sure that `:as`, `:from`, and `:&` can still be used as regular keywords, try denoting the magic words with `&` instead of `:`, so `:as` would be `&as`, `:from` would be `&from`, and `:&` would be `&rest` or something.

You get used to it. This is the order Clojure uses, and I think it's better than the reverse. 

This has probably been debated elsewhere, but off the top of my head, I remember ES6 did it the other way and I wished it worked like Clojure. The assigned names are always on the left in Python. So they should be on the left in the destructure too. Nesting seems to work better this way. The magic keywords are unambiguous in this position, as you pointed out.

gilch:
> `Exception` should probably be `TypeError`.

I meant to change that to something else, but wasn't sure what. `Exception` is too vague. I also considered `SyntaxError`.

> `(setv ddict (gensym))` does basically the same thing.

Not quite, it would have to be `(setv ddict (gensym ddict))` to get the more meaningful symbol for debugging. I thought of doing it that way, but it duplicated `ddict`.

> `(with-gensyms [x]) x` works, but I don't think it's supposed to; that is, `x` is supposed to be out of scope outside the `with-gensyms`. `with-gensyms` is a holdover from when we had `let` and expected to make it work someday. 

I originally tried `defmacro/g!`, but switched to `with-gensyms` when I made it a function. If that's how `with-gensyms` should work we should probably remove or redesign it. It's a separate issue though.

Maybe remove `with-gensyms` and replace with something like `(gensyms foo bar baz)` that expands to `(setv foo (gensym foo)  bar (gensym bar)  baz (gensym baz))`,

> `for*` can be `for`. (There's no reason to use `for*` outside Hy core.)

See #1125.

Kodiologist:
>  it would have to be `(setv ddict (gensym ddict))` to get the more meaningful symbol for debugging

That's what I mean by "basically".



>> `for*` can be `for`. (There's no reason to use `for*` outside Hy core.)
>
> See #1125.

I don't see what you're trying to say. `for` will still work in all these cases if the change you requested there were made.

gilch:
I swapped the yoda condition and changed `all` to `full`.

gilch:
> Neither of `(=: {:from ["hello?"]} {"hello?" 5})` nor `(=: {:from ["is_hello"]} {"is_hello" 5})` binds to `hello?`. It's not obvious which should work, but it seems clear that at least one should work.

It does work:
```
=> (=: {:from ["hello?"]} {"hello?" 5})
from hy import HyString
:ddict_1256 = {'hello?': 5, }
hello? = :ddict_1256[HyString('hello?')]
None
=> (get (vars) "hello?")
vars()['hello?']
5
```
And you wouldn't have to use the get/vars syntax with #1117. This seems like a manifestations of our complaints in #1168 and #1173. The `:from` option is working as designed.

gilch:
* `=:` can take multiple pairs.
* Removed `dest`.
* Changed `Exception` to `SyntaxError`.
* Removed `with-gensyms`.
* Remove quasiquotes where it made sense. But that's not all of them.
* Used `for` instead.

Kodiologist:
I'm sorry, I was imprecise. By `hello?`, I mean the symbol which is written in Hy source as `hello?`, and whose name is actually `is_hello`. Neither `(=: {:from ["hello?"]} {"hello?" 5})` nor `(=: {:from ["is_hello"]} {"is_hello" 5})` bind to this symbol. In particular, the latter binds the symbol whose name is `is-hello`. So `"is_hello"` is partly and incorrectly unmangled. Surely this isn't how you meant to design things.

Kodiologist:
And in addition to the unmangling being only partial, it's going the wrong direction. You could justifiably mangle in this context, but not unmangle.

gilch:
* added `:or` defaults, with tests.
* `:&` now uses a `cut` instead of `islice`, since destructure uses `get` anyway.

gilch:
Added error checking to prohibit duplicate keywords.

gilch:
So a mangle might have to wait on #1336. I might be able to make it work then.

Both keywords and symbols work properly in `:from` now. It's just strings that cause issues. This is easy to work around by writing both the symbol and the string to look up without using `:from` in problematic cases.

Clojure uses `:syms`/`:keys`/`:strs` instead of `:from`. I'm not sure if it's better. But if we had `:strs`, then we could start with a vector of symbols, and convert them to strings instead of the other way around. It might be easier to type this way, since you don't have to repeat `"` around every token. I've streamlined the code enough that it shouldn't be too hard to implement.

Options:
1. Live with it for now, and maybe fix it later depending on #1336 #1173 etc. Warn about the problem in the docs, and explain the workaround.
2. Add `:strs` that works like Clojure, and maybe prohibit strings in `:from`
3. Get rid of `:from` altogether, and just do it like Clojure for all three types.

I'm leaning toward option 1.

The other big issue remaining is how to destructure consumable iterables, be they plain iterators or some other kind of Python generator. This would be straightforward to handle with something like Clojure's immutable seq. #1242. There are three big problems with these iterables Python. First, they're mutable, so just looking at them consumes them. And second, they may be infinite, so realizing the whole thing and saving it into a list may not work either (but this is what Python's multiple assignment does).  The third problem is that they may be nested. If they weren't, you could easily work around this by realizing it into a list, with a `take` call too if it's infinite, on the *right* side.

Options:
1. Live with the restriction for now, and maybe fix it later depending on #1242. Nested mutable infinite generators are a bad idea anyway. It's not like JSON has these.
2. Add a `:pre` magic (to both `dest-list` and `dest-dict`) that takes a custom function to preprocess the input before attempting to destructure it. For example, `:pre list`, `:pre tuple` or even `:pre (comp tuple (partial take 5))`. These could be placed at any nesting depth. You convert to some other structure, like from Pyrisistent. Or do other processing like `:pre (comp list partition)`, or some complex `dict-comp` or `list-comp` in a lambda. Thus it seems like general preprocessors like this would be very powerful and it wouldn't even be that hard to implement. But it could be annoying to retype them for a nested structure, though functions could be assigned an alias in advance instead of in-line. It might also be annoying to have to count symbols for `take`'s number though.
3. Add some kind of `:tee foo` magic just to `dest-list`. Maybe it calls `(setv (, t0 t1) (tee (iter the-iterable)))` on the iterable, and then `(tuple (take x (iter t0)))` on one of them before destructuring the tuple, where x is the number it would need to unpack. Then `foo` will have `t1`. Given `:& bar`, `bar` will have the partly consumed `t0`. Tees seem like a poor substitute for seqs. They're still mutable, and consuming the iterator they're based on makes them skip too. I don't like them for this reason. `:as` is a gotcha waiting to happen. Implementing this also seems kind of involved.
4. Add logic to the *expansion* in `dest-list` to autodetect when plain `get` isn't going to work. (Maybe just a getattr check to see if the operator is supported). In that case, just call `list` on it before putting it in the gensym. This is fairly similar to how Python does things. But this diverges if the iterable is infinite. And the added check happens after the expansion at *runtime*, so it's a bit slower even when it doesn't need to construct a list.

I'm leaning toward option 2.

Kodiologist:
For the issue of strings in `:from`, I think the right thing to do is just avoid the unmangling (which is currently being done with `name`, I think). Then `(=: {:from ["is_hello"]} {"is_hello" 5})` will bind to the symbol usually written `hello?`.

For the issue of iterators, how about something like what I just pushed?

Kodiologist:
Hmm, wait, I see the problem with this. `:as` can't bind something to exactly what's passed in if it's an iterator. However, why not make `:as` listify? Python's `*foo` on the left-hand side does the same thing.

gilch:
I didn't expect that to work. I looked into it, and I noticed an "assert True" when I did a hy2py on the tests. This was due to a misplaced bracket, since one-arg equality is always true #1332. I amended the test to make it work, and now your new commit breaks it. This never should have passed.
```Python
=>  (=: [a b :& the-rest :as full] "abcdefg")
:orig_1259 = 'abcdefg'
:dlist_1260 = list(:orig_1259)
a = :dlist_1260[0]
b = :dlist_1260[1]
the_rest = :dlist_1260[2:]
full = :orig_1259
None
=> (= (, a b the-rest) (, "a" "b" "cdefg"))
((a, b, the_rest) == ('a', 'b', 'cdefg'))
False
=> the-rest
the_rest
['c', 'd', 'e', 'f', 'g']
```
The nice thing about `cut` is that it returns the same type. I'd like to keep that feature, but here it got listified. To be fair, Python works similarly:
```
>>> a, *b = "abcdefg"
>>> b
['b', 'c', 'd', 'e', 'f', 'g']
```
And actually so does Clojure, which converts it to a seq:
```Clojure
user=> (let [[a & b] "abcdefg"] (println b))
(b c d e f g)
nil
```

Also, `:&` should be able to work on an infinite generator. This approach still doesn't allow it.

gilch:
I removed the last commit. But I pushed a `destructure-iterables` branch that still has it, if someone wants to continue from there.

gilch:
Option 5:

Add another destructure type, `dest-iter`. This supports `:&` but not `:as`. Write it with `()` instead of `[]`. This is similar to the difference between a genexpr and a list-comp in Python, so it's easy to remember. It will work fine on any iterable, but `[]` requires random access, as now.

gilch:
I've removed `:from` and replaced it with Clojure's approach. I also added a test for generated lookup keys like `(frozenset [0 1])`, since it wasn't evaluating these properly before. That should take care of the mangling issue.

gilch:
I added a `dict=:` macro for destructuring directly into a dict. (Using symbol keys, to avoid mangling issues.) This returns a value, unlike the `setv`-backed `=:`, which always returns `None`.

I also added a `vals@` function which looks up multiple things at once.

Kodiologist:
Are you still working on this?

gilch:
Yes, but I'd like to wait for #1338 and rebase. (And also #1349, but that one's done.) Those will simplify some things in here.

Kodiologist:
Oh, that makes sense.

Kodiologist:
We have #1338 now.

Kodiologist:
It looks like `dest-list` and `dest-iter` share a lot of code. Do you think refactoring would help?

gilch:
Major refactoring. I was able to pull a lot of the shared code from all three `dest-` functions out into a decorator. It also inspired #1393. I wanted to make those dispatch `if`s less verbose with a macro, but thought we might already have something like that in anaphoric. I didn't see one, but Clojure had it, so I implemented one based on that. We might want to move it to core.

Kodiologist:
It doesn't seem that you're buying much by implementing `ifp` for internal use instead of using `if` or `cond`.

Is `vals@` a leftover internal function, or is it intended for public consumption?

gilch:
Anything at the toplevel without the underscore prefix is supposed to public. I'm not sure how you'd use the `dest-` functions instead of primary dispatcher, `destructure`, though. They have to recuse to it anyway. Maybe if they took the dispatcher as an argument, users could add more binding form types.

My motivation for this design is that Clojure has destructuring built in to most of its binding forms, and a function to help implement this for custom macros. I'm not sure how best to do this in Hy. The `destructure` function could be useful in other macros. But Hy's functions only have sequential unpacking for arguments built in, like Python2 did. Python's unpacking assignments also work.

The `=:` macro should work anywhere a `setv` does. `dict=:` is more contained, so you can use it without clobbering locals. It's also useful for the `(.update (vals foo) (dict=: ...))` trick, which lets you unpack things into namespaces other than the local one.

The `vals@` function will convert any `dict` to a `list` so Python's unpacking assignments can work on it (and Hy's argument unpacking too.) It was inspired by [Ruby's `.values_at` `Hash` method](https://ruby-doc.org/core-2.4.1/Hash.html#method-i-values_at), which is used to "unpack" a Ruby `Hash`.

You could combine `vals@` with `dict=:` to destructure into a list (and one of the tests does this), but this seems verbose, since you're using the symbols twice. I'm not sure how a more direct `list=:` could work, or if we should try.

> It doesn't seem that you're buying much by implementing `ifp` for internal use

If that were all we ever used it for, yeah, but it's also public, and useful in general, since reusing a predicate is a very common pattern in a branching form. I had originally implemented a simpler private macro to do this, but it seemed redundant when we could have the more general one from Clojure.

But I don't think it belongs here, which is why I suggested moving it to core (in a separate PR #1393). We could also put it in a separate contrib file if that's too controversial for the moment. But until that happens, I have to have a local implementation to test any code that uses it.

Kodiologist:
I think making a public `ifp` is perfectly reasonable, but it seems to belong in a different PR and a different file (either core or a different `contrib` module), since it's not about destructuring.

Kodiologist:
So, do you want to remove `ifp` from this PR?

Kodiologist:
I'm closing this for inactivity. Please reopen it when you'd like eyeballs again.

PR #1333, opened 2017-07-21
Hide ads on Read the Docs

Kodiologist:


gilch:
Wasn't there an Admin [opt-out](https://docs.readthedocs.io/en/latest/ethical-advertising.html#opting-out)? Do we even want to?

kirbyfan64:
...is this even allowed? I would think those who don't like (e.g. me) would already have an ad blocker, and I'm not sure RTD would be too happy with the idea...

(TBH I whitelisted RTD anyway, since the ads are pretty non-intrusive, and they're providing a great service.)

Kodiologist:
> Wasn't there an Admin opt-out?

That's the setting the commit message refers to.

> ...is this even allowed?

I have no idea. If they don't like it, we can move.

I don't think advertising in our own documentation is something we should tolerate. If you guys don't want this PR, then let's move.

gilch:
I'm not particularly offended by the ads in this case. Moving things creates dead links on other pages we can't control. It's not something we should do lightly or often, but sooner is probably better than later. (Maybe suppose we could make a redirect doc saying "this doc has moved".)

The community ads are for other open source projects, aren't they? We could also request an ad for Hy.

Where would you move to?

tuturto:
:-1: from me. If we use readthedocs, we should respect their policies.

gilch:
I think that's enough for a veto for this particular change. We can discuss other options though.

Kodiologist:
> Where would you move to?

In terms of the domain, let's use hylang.org, but as the true domain rather than just a redirect. Then we should never have to change domains again.

In terms of hosting, we can use the same VPS I use for Arfer.net, or perhaps GitHub, depending on what's possible with GitHub Pages or whatever.

gilch:
Many GitHub projects do use GitHub Pages for their docs. That seems like a reasonable option. Last I checked, they only do static pages to save on hosting costs, but that doesn't prevent client-side-only JavaScript from working, so some kind of search function would probably be possible.

Kodiologist:
Yeah, all we need is static pages. One can always use Google site search.

kirbyfan64:
GitHub Pages is pretty flexible. I host my own site there, and since client-side JS is as powerful as ever, you can do pretty much anything with it.

PR #1338, opened 2017-07-23
Initialize the return variable of `with`

### LINKED ISSUE #1320: `with` suppressing an exception can break temporary variable management
Consider:

    (defclass E [] []
      (defn __enter__ [self])
      (defn __exit__ [self exc-type exc-value traceback]
        True)) ; Prevent the exception from being propagated

    (defn f []
      (with [(E)]
        (/ 1 0)))

    (f)

This crashes with `UnboundLocalError: local variable '_hy_anon_var_1' referenced before assignment.`, because it compiles to:

    class E:

        def __enter__(self):
            pass

        def __exit__(self, exc_type, exc_value, traceback):
            return True

    def f():
        with E():
            _hy_anon_var_1 = (1 / 0)
        return _hy_anon_var_1
    f()

The exception breaks out of the `with` statement before `_hy_anon_var_1` is assigned, but we still try to `return _hy_anon_var_1`.

Kodiologist:
Fixes #1320.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1452,25 +1452,22 @@ class HyASTCompiler(object):
             raise HyTypeError(expr,
                               "with expects a list, received `{0}'".format(
                                   type(args).__name__))
-        if len(args) < 1:
-            raise HyTypeError(expr, "with needs [[arg (expr)]] or [[(expr)]]]")
-
-        args.reverse()
-        ctx = self.compile(args.pop(0))
+        if len(args) not in (1, 2):
+            raise HyTypeError(expr, "with needs [arg (expr)] or [(expr)]")
 
         thing = None
-        if args != []:
+        if len(args) == 2:
             thing = self._storeize(args[0], self.compile(args.pop(0)))
+        ctx = self.compile(args.pop(0))
 
         body = self._compile_branch(expr)
 
+        # Store the result of the body in a tempvar
         var = self.get_anon_var()
         name = ast.Name(id=ast_str(var), arg=ast_str(var),
                         ctx=ast.Store(),
                         lineno=expr.start_line,
                         col_offset=expr.start_column)
-
-        # Store the result of the body in a tempvar
         body += ast.Assign(targets=[name],
                            value=body.force_expr,
                            lineno=expr.start_line,
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1472,6 +1472,16 @@ class HyASTCompiler(object):
                            value=body.force_expr,
                            lineno=expr.start_line,
                            col_offset=expr.start_column)
+        # Initialize the tempvar to None in case the `with` exits
+        # early with an exception.
+        initial_assign = ast.Assign(targets=[name],
+                                    value=ast.Name(
+                                        id=ast_str("None"),
+                                        ctx=ast.Load(),
+                                        lineno=expr.start_line,
+                                        col_offset=expr.start_column),
+                                    lineno=expr.start_line,
+                                    col_offset=expr.start_column)
 
         the_with = ast.With(context_expr=ctx.force_expr,
                             lineno=expr.start_line,
@@ -1483,7 +1493,7 @@ class HyASTCompiler(object):
             the_with.items = [ast.withitem(context_expr=ctx.force_expr,
                                            optional_vars=thing)]
 
-        ret = ctx + the_with
+        ret = Result(stmts = [initial_assign]) + ctx + the_with
         ret.contains_yield = ret.contains_yield or body.contains_yield
         # And make our expression context our temp variable
         expr_name = ast.Name(id=ast_str(var), arg=ast_str(var),
@@ -1491,7 +1501,10 @@ class HyASTCompiler(object):
                              lineno=expr.start_line,
                              col_offset=expr.start_column)
 
-        ret += Result(expr=expr_name, temp_variables=[expr_name, name])
+        ret += Result(expr=expr_name)
+        # We don't give the Result any temp_vars because we don't want
+        # Result.rename to touch `name`. Otherwise, initial_assign will
+        # clobber any preexisting value of the renamed-to variable.
 
         return ret
 
--- a/tests/native_tests/with_test.hy
+++ b/tests/native_tests/with_test.hy
@@ -34,12 +34,12 @@
         (assert (= t2 2))
         (assert (= t3 3))))
 
-  (defn test-quince-with []
-    "NATIVE: test four withs, one with no args"
-    (with [t1 (WithTest 1)
-           t2 (WithTest 2)
-           t3 (WithTest 3)
-           _ (WithTest 4)]
-          (assert (= t1 1))
-          (assert (= t2 2))
-          (assert (= t3 3))))
+(defn test-quince-with []
+  "NATIVE: test four withs, one with no args"
+  (with [t1 (WithTest 1)
+         t2 (WithTest 2)
+         t3 (WithTest 3)
+         _ (WithTest 4)]
+        (assert (= t1 1))
+        (assert (= t2 2))
+        (assert (= t3 3))))
--- a/tests/native_tests/with_test.hy
+++ b/tests/native_tests/with_test.hy
@@ -43,3 +43,28 @@
         (assert (= t1 1))
         (assert (= t2 2))
         (assert (= t3 3))))
+
+(defclass SuppressZDE [object]
+  (defn --enter-- [self])
+  (defn --exit-- [self exc-type exc-value traceback]
+    (and (not (none? exc-type)) (issubclass exc-type ZeroDivisionError))))
+
+(defn test-exception-suppressing-with []
+  ; https://github.com/hylang/hy/issues/1320
+
+  (setv x (with [(SuppressZDE)] 5))
+  (assert (= x 5))
+
+  (setv y (with [(SuppressZDE)] (/ 1 0)))
+  (assert (none? y))
+
+  (setv z (with [(SuppressZDE)] (/ 1 0) 5))
+  (assert (none? z))
+
+  (defn f [] (with [(SuppressZDE)] (/ 1 0)))
+  (assert (none? (f)))
+
+  (setv w 7  l [])
+  (setv w (with [(SuppressZDE)] (.append l w) (/ 1 0) 5))
+  (assert (none? w))
+  (assert (= l [7])))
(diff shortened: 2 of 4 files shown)


hy/compiler.py
```diff
@@ -366,6 +366,17 @@ def is_unpack(kind, x):
             and x[0] == "unpack_" + kind)
 
 
+# Provide asty.Foo(x, ...) as shorthand for
+# ast.Foo(..., lineno=x.start_line, col_offset=x.start_column)
+class Asty(object):
```
kirbyfan64: ...uhh, is this really necessary for this PR? I know it makes things a bit shorter, but it's also only used a few times and seems rather off-topic...

Kodiologist: It has the potential to be used many other times throughout this file. Probably I'll apply it more broadly in a later PR. But I found myself already wanting it, so I saw no reason to wait.

gilch:
I'd really like to get #1320 fixed. @kirbyfan64, after @Kodiologist's answer are you okay with that `Asty` thing in this PR or not?

I'd certainly consider pure refactoring PRs that do nothing but simplify things like compiler. Technical debt is a thing. If you think the design change merits separate discussion, we could postpone `Asty` and use the old style here instead.

kirbyfan64:
> If you think the design change merits separate discussion, we could postpone `Asty` and use the old style here instead.

Yeah, I'd really prefer it this way.

Kodiologist:
Aww, boo. Fine.

Kodiologist:
Done.

tests/native_tests/with_test.hy
```diff
@@ -34,12 +34,37 @@
...
+        (assert (= t2 2))
+        (assert (= t3 3))))
+
+(defclass SuppressZDE [object]
+  (defn --enter-- [self])
+  (defn --exit-- [self exc-type exc-value traceback]
+    (and (not (none? exc-type)) (issubclass exc-type ZeroDivisionError))))
+
+(defn test-exception-suppressing-with []
+  ; https://github.com/hylang/hy/issues/1320
```
gilch: Lisp convention has `;;` for a comment on its own line. Single `;` is only for margin comments--coments following code on the same line. Lisp editors may expect this. DTRT with it.

https://stackoverflow.com/questions/6365334/lisp-commenting-convention#6365579
https://github.com/bbatsov/clojure-style-guide#comments

Kodiologist: I'm familiar with this convention, but I've never seen the point of it.

gilch: As I said, Lisp editors expect this convention to be followed and get confused when you don't. Look what emacs does to it automatically.
```Hy
(defn test-exception-suppressing-with []
                                ; https://github.com/hylang/hy/issues/1320

  (setv x (with [(SuppressZDE)] 5))
```

Kodiologist: Real talk. I sincerely appreciate how you've resumed work on Hy. You're working on a lot of cool PRs and you're often helpful. But, I don't think it's a good use of my time, or your time, to fuss over spacing and semicolons, and I don't think any hard-and-fast rules about code formatting should be enforced. Rules that are useful most of the time are occasionally going to be less than useful, as the `noqa`s sprinkled throughout the codebase and the weird thing I had to do with `if` in #1345 can attest. And some rules have no discernible value at all, like this one about one versus two semicolons. This is why, although I too use Emacs, I don't use auto-indent.

It's not hard for me to add a semicolon here, and one versus two semicolons makes no difference to me. But I don't want to spend time editing every PR for style issues that are more about tradition and fussy editors than about human legibility.

Kodiologist:
@gilch Is this ready to go?

tests/native_tests/with_test.hy
gilch: > But, I don't think it's a good use of my time, or your time, to fuss over spacing and semicolons,
...
> This is why, although I too use Emacs, I don't use auto-indent for any language.

Turn on Emacs aggressive indent mode and you won't have to waste my time fussing over it.

> and I don't think any hard-and-fast rules about code formatting should be enforced. Rules that are useful most of the time are occasionally going to be less than useful
...
> But I don't want to spend time editing every PR for style issues that are more about tradition and fussy editors than about human legibility.

Lisp must be indented correctly to be legible. That particular rule is absolute. The machine doesn't care, but if humans have to maintain it, indent it properly. Code in a team project is read more than it is written. Don't create more work for everyone that follows.

We really should have an automated Travis check for proper indentation of Hy. But since Hy is young, we haven't gotten around to writing it.

> And some rules have no discernible value at all, like this one about one versus two semicolons.

You don't get to decide this unilaterally. I don't want to break the semicolon convention here without a good reason, and enough approval from Hylang's core to change the style guide to say that we're breaking with Lisp convention on this point. If you get that, then we can do one semicolon everywhere after the change. Not before.

It does have a use in Emacs. It will automatically indent it differently in a way that is useful. One semicolon is for [margin comments](https://stackoverflow.com/a/8825101). Over multiple lines it would take more work to align them if emacs didn't do it for you. Two semicolons automatically aligns with the other list elements. Three or more are never indented.

Kodiologist: If I make the editor indent automatically, then I have to wrestle with it when it's making a bad decision.

> Lisp must be indented correctly to be legible. That particular rule is absolute.

I think this code is legible (or at least, it was when I started). I know that's inherently subjective, but you and the other two Horsemen of Hy Maintenance have approved my PRs in the past, so I don't see what's changed. I'm not asking for new standards in reviewing PRs; I'm asking for what I'm used to.

Anyway, if things like the number of semicolons for a comment, or whether the indent lines up with the immediate parent form instead of an ancestor form without other hanging children, are going to be a blocker for you from now on, then I guess you're not going to approve my PRs. It's a shame, since you're whip-smart and notice genuinely important things that others miss, but enforcing style rules for their own sake is counterproductive.

I think [the review process here](https://github.com/hylang/hy/pull/1377) is a good model. The author indented something in a nonstandard way (for Python and our codebase), I pointed it out, the author said "I prefer it that way", I said "Eh, no biggie" and @kirbyfan64 said "it's not really a blocking issue" and merged it.

kirbyfan64:
Wait, why not compromise?? For minor style issues, the person who notices, when they can, just pushes a fix themselves? At least for now until we can actually reach some kind of consensus...

Kodiologist:
I think that pushing a style fix (or any other commit) onto somebody else's PR is fine. If the author reverts your change, you shouldn't make it again (the reviewers are gatekeepers for the repository, but authors are bosses of the content of their own PRs). But I don't see myself often reverting a change to indentation or semicolons or something. (Not on purpose, anyway. It might get blown away by a rebase to change the commit in some other way; in that case, you're free to make it again.)

Style-fix commits should generally be melded into other commits before the merge.

tests/native_tests/with_test.hy
gilch: > If I make the editor indent automatically, then I have to wrestle with it when it's making a bad decision.

I no longer trust you to judge when the editor is making a bad decision about formatting Lisp. But if you're using hy-mode and it's indenting things wrong, then it is Hy Society's responsibility to fix hy-mode. That's us.

> I think this code is legible (or at least, it was when I started).

The one-semicolon thing doesn't really affect legibility. I was not talking about the indentation from this PR.

> you and the other two Horsemen of Hy Maintenance have approved my PRs in the past, so I don't see what's changed.

If I've missed other indentation errors, that's a shame (and it ought to be fixed), but in the meantime, let's not make it *worse*.

> I think the review process here is a good model. The author indented something in a nonstandard way (for Python and our codebase), I pointed it out, the author said "I prefer it that way", I said "Eh, no biggie" and @kirbyfan64 said "it's not really a blocking issue" and merged it.

That was Python, this is Hy. If a Python block wasn't indented properly, it would fail the tests, because it's syntactically significant in Python. That was about leaving the parenthesis on its own line. It passed our flake8 test, didn't it? If flake8 doesn't complain, then I'm not really worried about the Python side. If they had to turn a test off or add a `noqa` tag, then it would require more scrutiny and a really good excuse. We need a similar automated check for Hy.

gilch:
> Wait, why not compromise?? For minor style issues, the person who notices, when they can, just pushes a fix themselves?

> But I don't see myself often reverting a change to indentation

I can work with that.

I shouldn't *have to*, but I *can*. This could slow down my approvals some, compared to the case where the code is legible in the first place. I have to fix it just so I can *read* it. I shouldn't have to count brackets just to tell which element belongs with which list. In well-formatted Lisp this would be clear from the indentation. I don't consider bad indentation "minor" in Lisp.

> Style-fix commits should generally be melded into other commits before the merge.

To be clear, I should add and push a *new* commit after all the others, with the style fixes, rather than a force-push with an amended commit? Then someone else will meld it before the merge? (For one so strict about commit structure, it's surprising you're so lax about formatting.) Also, if the formatting gets removed without comment, I'll assume it's accidental and try again.

> At least for now until we can actually reach some kind of consensus

I'd like to clarify the Style Guide and add an automated check to avoid indentation errors. And I'd like to settle the semicolon style in the Style Guide too. I expect most of the less-active, old-guard, Hy-Society members that actually use other Lisps much would agree with me about the indentation thing. "Consensus", in my mind, would have to include some of them. 

I'm *very concerned* that future Hy contributors that are used to properly formatted Lisp would be deterred from participating by our sloppy code if we don't enforce indents. I certainly felt deterred from participating by this dispute, to the point of contemplating a hostile fork.

Kodiologist:
> To be clear, I should add and push a *new* commit after all the others, with the style fixes, rather than a force-push with an amended commit? Then someone else will meld it before the merge?

Either seems fine to me.

> For one so strict about commit structure, it's surprising you're so lax about formatting.

Formatting can always be changed, but the commit history is forever. I guess that's why I feel paranoid about it.

> members that actually use other Lisps

I do use Emacs Lisp (and so do you, right?). To which you might reply that Emacs Lisp is a crummy Lisp anyway, which would be harsh but fair.

> I certainly felt deterred from participating by this dispute, to the point of contemplating a hostile fork.

That seems like cutting off your nose to spite your face, but it is free software, so it's always an option.

kirbyfan64:
> I certainly felt deterred from participating by this dispute, to the point of contemplating a hostile fork.

Please don't. That literally never ends well, we're short-staffed as it is, and IMO formatting, although important, isn't important *enough* for something that drastic.

Kodiologist:
> So what's the status here?

You mean @gilch's proposal 2? Nobody's been working on it that I know of, but not for any special reason. I like it, for what that's worth. Or at least, I'd appreciate an optional feature that requires you to declare variables before you use them, like Perl's `use strict`.

Kodiologist:
Thanks. Don't worry about it; I'll fix the conflict and merge.

PR #1344, opened 2017-07-26
Shadow `get`

### LINKED ISSUE #683: Shouldn't get be available as a function?
Here's a use case for this:

``` hy
=> (setv a {"foo" {"bar" {"baz" 3}}})
=> a
{u'foo': {u'bar': {u'baz': 3L}}}
=> (reduce get ["foo" "bar" "baz"] a)
NameError: name 'get' is not defined
=> (reduce (fn [x y] (get x y)) ["foo" "bar" "baz"] a)
3L
```

Kodiologist:
Closes #683.

### DIFF
--- a/hy/core/shadow.hy
+++ b/hy/core/shadow.hy
@@ -138,11 +138,18 @@
 (defn not-in [x y]
   (not-in x y))
 
+(defn get [coll key1 &rest keys]
+  (setv coll (get coll key1))
+  (for* [k keys]
+    (setv coll (get coll k)))
+  coll)
+
 (setv *exports* [
   '+ '- '* '** '/ '// '% '@
   '<< '>> '& '| '^ '~
   '< '> '<= '>= '= '!=
   'and 'or 'not
-  'is 'is-not 'in 'not-in])
+  'is 'is-not 'in 'not-in
+  'get])
 (if (not PY35)
   (.remove *exports* '@))
--- a/NEWS
+++ b/NEWS
@@ -12,6 +12,7 @@ Changes from 0.13.0
      as necessary, so you can write ``(eval `(+ 1 ~n))`` instead of
      ``(eval `(+ 1 ~(HyInteger n)))``
    * Literal `Inf`s and `NaN`s must now be capitalized like that
+   * `get` is available as a function
 
    [ Bug Fixes ]
    * Numeric literals are no longer parsed as symbols when followed by a dot
--- a/tests/native_tests/operators.hy
+++ b/tests/native_tests/operators.hy
@@ -288,3 +288,11 @@
   (assert (is (f 3 [1 2]) (!= f-name "in")))
   (assert (is (f 2 [1 2]) (= f-name "in")))
   (forbid (f 2 [1 2] [3 4])))
+
+
+(op-and-shadow-test [get]
+  (forbid (f))
+  (forbid (f "hello"))
+  (assert (= (f "hello" 1) "e"))
+  (assert (= (f [[1 2 3] [4 5 6] [7 8 9]] 1 2) 6))
+  (assert (= (f {"x" {"y" {"z" 12}}} "x" "y" "z") 12)))


Kodiologist:
Not to be confused with

Shine get

gilch:
It's still not the obvious implementation. Couldn't the proposed
```Hy
(defn get [thing &rest keys]
  (for [k keys]
    (setv thing (get thing k)))
  thing)
```
be simplified to
```
(defn get [thing &rest keys]
  (get thing #* keys))
```
?

I think it would still use the special form rather than recurse in that context. Did we test if `#*` works properly in special forms?

Kodiologist:
No, `#*` causes the parent expression to be interpreted as a function. It doesn't work with special forms. That was the issue we got hung up on in the PR comments.

hy/core/shadow.hy
```diff
@@ -138,11 +138,17 @@
 (defn not-in [x y]
   (not-in x y))
 
+(defn get [thing key1 &rest keys]
+  (for* [k (+ (, key1) keys)]
+    (setv thing (get thing k)))
+  thing)
```
gilch: Not the version I thought I approved.

Okay so why split off the first key? Is it for performance in the more common one-key case? In that case it should probably be
```Hy
(defn get [thing key1 &rest keys]
  (setv thing (get thing key1)
  (if keys
    (for* [k keys)]
      (setv thing (get thing k)))
    thing))
```
?
But complications due to performance optimizations like this need to be profiled to make sure they actually help.

Kodiologist: It's not for performance reasons but to ensure a minimum of two arguments. The test for this is `(forbid (f "hello"))`.

gilch: Ah, I see. We want to assert that get has at least one key. In the case of zero keys it might make sense to just return the thing. But we'd have to make the special form do that too. But the slightly optimized version I wrote above might perform better.

Kodiologist: > In the case of zero keys it might make sense to just return the thing. But we'd have to make the special form do that too.

Right. It wasn't clear to me that the 0-argument form would ever be useful, so I left that as-is.

hy/core/shadow.hy
```diff
@@ -138,11 +138,17 @@
 (defn not-in [x y]
   (not-in x y))
 
+(defn get [thing key1 &rest keys]
```
gilch: The name `thing` is kind of vague. I think we're using the term `coll` as short for "collection" like Clojure does elsewhere in core.

hy/core/shadow.hy
gilch: So, I profiled my version against yours. (It helps that we fixed contrib/profile.) Mine is about twice as fast with one key, and still faster with two. Adding tuples is too slow for something this fundamental.

Kodiologist: Using a shadow function in the first place is only going to worsen performance, but okay. What exactly is your version?

gilch:
```Hy
(defn get [coll key1 &rest keys]
  (setv coll (get coll key1))
  (if keys
    (for* [k keys]
      (setv coll (get coll k))))
  coll)
```

Kodiologist: Oh, I thought you meant some version using `reduce`. But why use the `if`?

    (defn get [coll key1 &rest keys]
      (setv coll (get coll key1))
      (for* [k keys]
        (setv coll (get coll k)))
      coll)

gilch: > version using reduce

I don't think `reduce` would be faster considering the overhead for Python's function calls.

> But why use the if?

So we immediately stop processing if the keys are empty, which should speed up the common 1-key case. I'm actually not certain if that helps. Python might do this check anyway in the for loop. I can profile and maybe disassemble the bytecode to see if it matters.

Kodiologist: I don't think this kind of micro-optimization is worth grubbying up the code. I'll use a leading `setv` instead of tuple addition, but I would only support adding an `if` for a performance rationale if I had good reason to believe that this function is going to be a CPU bottleneck in real code and the `if` makes a meaningful difference.

gilch: Looks like Hy is assigning an anon var because of the if. (even though it's not used). That might slow it down more than the equivalent Python would.

Let's remove the if then. It's not really helping, and it slows down the 2-key case noticeably when I profiled it.

Kodiologist:
Done and done.

Kodiologist:
Thanks, Ryan. Recall that the new policy is to create a merge commit when merging. You don't have to back this out, but remember that for next time.

ekaschalk:
Can `cut` be shadowed as well?

Kodiologist:
Probably, but you should make a new issue for it if you want it.

PR #1348, opened 2017-07-29
Fixes #1068

### LINKED ISSUE #1068: make assoc use a gensym

woodrush:
Closes #1068.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1386,11 +1386,11 @@ class HyASTCompiler(object):
                             step=step.expr),
             ctx=ast.Load())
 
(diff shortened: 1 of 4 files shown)


kirbyfan64:
FWIW usually it's best to let everyone know that you're starting work on something, that way wasted efforts are avoided. Sorry about that.

woodrush:
Thanks. No problem, I'll remember about it next time. :)

Issue #1217, opened 2017-02-04
Drop support for Python 2


Kodiologist:
This is something that's been brought up many times but has no dedicated issue. Even if we don't want to do this soon, we might want to put it on a milestone or something.

Dropping Python 2 would let us clean up the codebase a lot and make it easier to support Python 3's new features.

If there's anybody who uses Hy on Python 2 and really can't upgrade, at least until another library adds Python 3 support, you should say that here and mention what you're waiting for. We have no idea how many of you exist.

kirbyfan64:
Eh... I'd say Python 2 is still used widely enough to make this harder to justify.

We should probably make it clear which ones exactly we're supporting, though. Didn't we drop 2.6 at some point.

gilch:
https://pythonclock.org/

Python 2 support officially ends in 2020. It's open source, so others might continue to support a fork (Red Hat, in particular, has some long-term service obligations out to 2027, but only to paying customers on Linux), but I don't think Hy has any need or obligation to support Python 2 past this date. I think we should remove Hy's Python 2 support by that date.

The only reason I'd like to continue supporting Python 2 now is to get Hy on more platforms. Jython, in particular hasn't released a version 3 yet. Jython support would open up Hy to most of the Java ecosystem, except, perhaps, Android, which uses an incompatible Java VM.

Kodiologist:
If we release semi-annually, which I'd like to do, then we could drop Python 2 in mid-2019 or in early 2020. That sounds reasonable to me, although I hope we hit Hy 1.0 before then, so we might need to make that release Hy 2.0.

berkerpeksag:
+1 for dropping Python 2 support *now*. Note that some super popular Python projects like Django and Gunicorn will drop Python 2 in their next major release this year. I think we can be more aggressive since we have much smaller user base than these projects :)

davidkallie:
I am in favor of keeping support until 2020. I just ported hy to the FreeBSD ports tree, and python3 is not default. I think keeping in tune with support of Python 2 is reasonable and makes sense, as @glich mentions keeping compatibility ensures more exposure, which increases the chances of more users.

Rogof:
In my case, I'm working on a Standalone QGIS application. QGIS only supports Python 2 for now and while the next mayor version will support Python 3, it seems it won't be ready any time soon.

So I really would like for python 2 to be supported for a while.

Kodiologist:
It looks like this won't be happening soon, despite some support. So I'll close this.

[...]

PR #1349, opened 2017-07-29
Allow `unquote-splice` to accept any false value as empty

### LINKED ISSUE #1335: Allow ~@ splice on None
Related #896.

In most Lisps, `nil` is also the empty list. In Hy, `nil` used to be `None`. Python does not consider `None` to be iterable, but `unquote-splice` could have a special case for it. This would make macro writing easier. It's a common Lisp idiom to splice in an `if` expression during a macroexpansion which can optionally add some things or not. In Hy you have to put an empty iterable in the else part to make this work. If `None` were allowed, you could leave that off.

While we're at it, it would also be nice for macro writing if a `HyList` instance `+` `None` just returned the instance, but this is less consistent with Python lists.

woodrush:
The original `unquote_splice` received a form stored in the variable `f_contents`, and produced the code
```clojure
`(list ~f_contents)
```
This PR modifies this to the following code, providing an additional `var = self.get_anon_var()`:
```clojure
`(do
   (setv ~var ~f_contents)
   (if* (= ~var None)
        []
        (list ~var)))
```
I have ran and passed the `tox` tests a Fedora Linux with Python versions 2.7, 3.3.6, 3.4.6, 3.5.3, 3.6.2, pypy-5.7.1, installed through pyenv. I also ran
```clojure
(defmacro splicetest []
  `[1 2 ~@[3 4] ~@None 5 6 ~@(when False 0) 7 8 ~@(when True None) 9 10 ~@(when True [None]) 11 12 ~None])
(splicetest)
```
through `hy2py`, and got
```python
import hy
from hy import HyInteger, HyList

def _hy_anon_fn_1():
    _hy_anon_var_1 = [3, 4]
    _hy_anon_var_2 = None
    _hy_anon_var_3 = None
    _hy_anon_var_4 = None
    _hy_anon_var_5 = [None]
    return HyList((((((((((((((((([] + [HyInteger(1)]) + [HyInteger(2)]) + ([] if (_hy_anon_var_1 == None) else list(_hy_anon_var_1))) + ([] if (_hy_anon_var_2 == None) else list(_hy_anon_var_2))) + [HyInteger(5)]) + [HyInteger(6)]) + ([] if (_hy_anon_var_3 == None) else list(_hy_anon_var_3))) + [HyInteger(7)]) + [HyInteger(8)]) + ([] if (_hy_anon_var_4 == None) else list(_hy_anon_var_4))) + [HyInteger(9)]) + [HyInteger(10)]) + ([] if (_hy_anon_var_5 == None) else list(_hy_anon_var_5))) + [HyInteger(11)]) + [HyInteger(12)]) + [None]))
hy.macros.macro('splicetest')(_hy_anon_fn_1)
[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, None, 11, 12, None]
```
which is the expected result. (The last two `~`s are supposed to be evaluated to `None`.)

Edit: Closes #1335 .

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,7 @@
 Changes from 0.13.0
 
    [ Language Changes ]
+   * The unquote-splice or ~@ form now accepts any false value as empty.
    * `yield-from` is no longer supported under Python 2
    * `apply` has been replaced with Python-style unpacking operators `#*` and
      `#**` (e.g., `(f #* args #** kwargs)`)
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -1604,19 +1604,31 @@ unquote-splice
 --------------
 
 ``unquote-splice`` forces the evaluation of a symbol within a quasiquoted form,
-much like ``unquote``. ``unquote-splice`` can only be used when the symbol
+much like ``unquote``. ``unquote-splice`` can be used when the symbol
 being unquoted contains an iterable value, as it "splices" that iterable into
-the quasiquoted form. ``unquote-splice`` is aliased to the ``~@`` symbol.
+the quasiquoted form. ``unquote-splice`` can also be used when the value
+evaluates to a false value such as ``None``, ``False``, or ``0``, in which
+case the value is treated as an empty list and thus does not splice anything
+into the form. ``unquote-splice`` is aliased to the ``~@`` syntax.
 
 .. code-block:: clj
 
     (def nums [1 2 3 4])
     (quasiquote (+ (unquote-splice nums)))
-    ;=> (u'+' 1L 2L 3L 4L)
+    ;=> ('+' 1 2 3 4)
 
     `(+ ~@nums)
-    ;=> (u'+' 1L 2L 3L 4L)
+    ;=> ('+' 1 2 3 4)
 
+    `[1 2 ~@(if (< (nth nums 0) 0) nums)]
+    ;=> ('+' 1 2)
+
+Here, the last example evaluates to ``('+' 1 2)``, since the condition
+``(< (nth nums 0) 0)`` is ``False``, which makes this ``if`` expression
+evaluate to ``None``, because the ``if`` expression here does not have an
+else clause. ``unquote-splice`` then evaluates this as an empty value,
+leaving no effects on the list it is enclosed in, therefore resulting in
+``('+' 1 2)``.
 
 when
 ----
--- a/AUTHORS
+++ b/AUTHORS
@@ -78,3 +78,4 @@
 * John Patterson <john@johnppatterson.com>
 * Kai Lüke <kailueke@riseup.net>
 * Neil Lindquist <archer1mail@gmail.com
+* Hikaru Ikuta <woodrush924@gmail.com>
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -707,7 +707,9 @@ class HyASTCompiler(object):
                                                                          level)
                 imports.update(f_imports)
                 if splice:
-                    to_add = HyExpression([HySymbol("list"), f_contents])
+                    to_add = HyExpression([
+                        HySymbol("list"),
+                        HyExpression([HySymbol("or"), f_contents, HyList()])])
                 else:
                     to_add = HyList([f_contents])
 
--- a/tests/native_tests/quote.hy
+++ b/tests/native_tests/quote.hy
@@ -74,17 +74,17 @@
 (defn test-unquote-splice []
   "NATIVE: test splicing unquotes"
   (setv q (quote (c d e)))
-  (setv qq (quasiquote (a b (unquote-splice q) f (unquote-splice q))))
-  (assert (= (len qq) 9))
-  (assert (= qq (quote (a b c d e f c d e)))))
+  (setv qq `(a b ~@q f ~@q ~@0 ~@False ~@None g ~@(when False 1) h))
+  (assert (= (len qq) 11))
+  (assert (= qq (quote (a b c d e f c d e g h)))))
 
 
 (defn test-nested-quasiquote []
   "NATIVE: test nested quasiquotes"
-  (setv qq (quasiquote (1 (quasiquote (unquote (+ 1 (unquote (+ 2 3))))) 4)))
-  (setv q (quote (1 (quasiquote (unquote (+ 1 5))) 4)))
+  (setv qq `(1 `~(+ 1 ~(+ 2 3) ~@None) 4))
+  (setv q (quote (1 `~(+ 1 5) 4)))
   (assert (= (len q) 3))
-  (assert (= (get qq 1) (quote (quasiquote (unquote (+ 1 5))))))
+  (assert (= (get qq 1) (quote `~(+ 1 5))))
   (assert (= q qq)))
 
 


woodrush:
Note that this code assures that the form is evaluated only once using `setv`, and uses `self.get_anon_var()` to assure that unwanted captures of variables don't occur.

woodrush:
An alternative way would be to change https://github.com/woodrush/hy/blob/f90cb39a61359c54089b2251324a11cb353b3b96/hy/compiler.py#L703 so that it appends nothing when it receives `None`. This would mean that we change
```clojure
(+ [] contents)
```
to
```clojure
(do
  (setv ret [])
  (for [form contents]
    (setv var form)
    (if (None? var) (continue) (ret.append var)))
  ret)
```
Now, both of these solutions don't change the fact that the `None`-ness of the form is evaluated after the form is compiled to Hy AST. Therefore, in both solutions we must construct this code using `HySymbol` and `HyExpression` constructs, so the alternative way of changing L703 would be almost the same as this PR. Also, the `contents` that is received here is what is created at https://github.com/woodrush/hy/blob/f90cb39a61359c54089b2251324a11cb353b3b96/hy/compiler.py#L711 , where I've changed in this PR (it is added at https://github.com/woodrush/hy/blob/f90cb39a61359c54089b2251324a11cb353b3b96/hy/compiler.py#L721). Changing L703 would also be a slight overhead for expressions that are not `unquote-splice`ed since it would add redundant checks of `quotedform?`.

Kodiologist:
Cool, thanks. Don't forget tests. Also, change your commit message and PR title to be more descriptive. It's hard to remember what #1335 is.

woodrush:
Thanks! I ran tests with `tox` for all the versions in `tox.ini`. Oh, but I forgot to add tests that splice `None`, I'll work on that.

> Also, change your commit message and PR title to be more descriptive.

I see. Got that.

woodrush:
Thanks for the comments. I added some tests for splicing `None`, squashed and rewrote the commit messages. I also reformatted the code where it was more than 80 characters long.

I'm writing another test which tests for `(unquote None)`.

woodrush:
Hmm... I think I've hit an issue-ish thing but I'm not that sure.
I'm trying to write a test that compares `(quasiquote (unquote None))` with `(quote None)`,

```clojure
; Test 1
(setv qq (quasiquote (a b c (unquote (+ 1 3))
                      (unquote None)
                      (unquote (when False 1)))))
(assert (= (len qq) 6))
(assert (= qq (quote (a b c 4 None None))))
```

but this fails since `(quasiquote (unquote None))` becomes `<class 'NoneType'>` while `(quote None)` becomes `<class 'hy.models.HySymbol'>`. The same thing happens with the original master branch as well.

So I tried casting `None`,

```clojure
; Test 2
(setv qq (quasiquote (a b c (unquote (+ 1 3))
                      (hy.models.HySymbol (unquote None))
                      (hy.models.HySymbol (unquote (when False 1))))))
(assert (= (len qq) 6))
(assert (= qq (quote (a b c 4 None None))))
```

but this fails since `HySymbol` needs to be fed with a `str`. Same for the original master branch.

I wondered why the original test works with the integers, so I looked inside the definitions of `HyInteger`, and searched the codebase for
overrides of `__eq__`, but it seems that there's no overloading of `=` going on. I'm not sure why `(= (quote 1) 1)` is `True`
although `(type (quote 1))` is `<class 'hy.models.HyInteger'>` and `(type 1)` is `<class 'int'>`.

I'm not sure how, and also a little unsure if I should write this test for `unquote` somehow right now... any suggestions?

gilch:
I'm trying to think of ways to avoid that extra anon var. If we simply used `or`, it would work about as well. So you could use ``` `(list (or ~fcontents []))```. This is not quite the same, since any falsy value would be allowed to stand for an empty list in a splice. But I think this is fine. Empty sequences are falsy, so should be fine to substitute the empty list, since that's what they'd generate in a `list` call anyway. In Common Lisp, `NIL` is the one false value, and also the empty list. So there is about as good a reason to accept `False` as there is to accept `None` as the empty list. 

It's a little bit strange at first to think that `0` (or any custom falsy type via `__bool__`) is now also allowed in a splice as an empty list. But I think this is both harmless and perfectly consistent.

gilch:
> I'm trying to write a test that compares `(quasiquote (unquote None))` with `(quote None)`,

These aren't supposed to be the same though. ``` `~x``` is just `x`, not `'x`.

In a typical Lisp, `NIL` is the empty list and also a symbol whose named value is itself. But Hy is based on Python, so the symbol `'None` is different from it's value `None`. The former is of type `HySymbol`, while the latter is of the singleton type `NoneType`--Python's `None` being the only instance of `NoneType`.

gilch:
I came up with another option to avoid the anon var. We can use exactly the same rules as `cons` by actually using a `cons`.
```
`(list (rest (cons Ellipsis ~f_contents)))
```
Where `Ellipsis` could be anything.

kirbyfan64:
So...is there a reason we can't just do `(list ~(or f_contents []))`.

gilch:
@kirbyfan64, if I understand you right, you'd try
```
to_add = HyExpression([HySymbol("list"), f_contents or []])
```
Does this work?

woodrush:
Hmm, I tried that but that doesn't work. Although 
```clojure
`[1 2 ~@0] ;-> `[1 2],
`[1 2 ~@None] ;-> Becomes an error.
```
This worked:
```python
                    to_add = HyExpression([
                        HySymbol("list"),
                        HyExpression([HySymbol("or"), f_contents, HyList()])])
```

woodrush:
> These aren't supposed to be the same though. `~x is just x, not 'x.

I see. Well then I guess I'll just leave the test since it doesn't make that much sense as a practical expression.

> It's a little bit strange at first to think that 0 (or any custom falsy type via __bool__) is now also allowed in a splice as an empty list.

Since we have `lif`, I think we could create `lor` and put that in the core macros, where stuff like `defn` and `if` are. It would serve as part of a basis of what Hy considers as `nil`-ish.

woodrush:
Speaking of core macros, when I use the bootstrap macros that are in `bootstrap.hy` in where I wrote `to_add`, tests in pypy fail. I tried using `if` at first but failed, but `if*` didn't. `defn` also passed all versions except pypy but failed with it.

NEWS
```diff
@@ -1,6 +1,8 @@
 Changes from 0.13.0
 
    [ Language Changes ]
+   * Pythonic falsey values (in means of __bool__) can now be spliced
+     by `unquote-splice`
```
gilch: I can't tell what this means.

A better wording would be "The `unquote-splice` or `~@` form now accepts any false value as empty."

gilch:
What is the `tests/a_test_bin.py` file for? It seems like a duplicate of `tests/test_bin.py`.

NEWS
woodrush: I agree. Changed to that.

woodrush:
> What is the tests/a_test_bin.py file for? It seems like a duplicate of tests/test_bin.py.

Sorry, that was an artifact for me renaming it to avoid `tox` to run `test_bin`. I dropped it out.

Kodiologist:
Looking good!

For legibility, would you please edit your new tests to use `` ` `` and `~@` instead of `quasiquote` and `unquote-splice`? I realize this will be inconsistent with other tests in the same functions; you change those too if you like.

Would you please update the documentation, namely the section on `unquote-splice` in `api.rst`?

woodrush:
Thanks! The test now uses `` \` `` and `~@`. I also added the docs on `unquote-splice`, any feedback would be appreciated.

woodrush:
Awesome!

docs/language/api.rst
```diff
@@ -1604,9 +1604,12 @@ unquote-splice
 --------------
 
 ``unquote-splice`` forces the evaluation of a symbol within a quasiquoted form,
-much like ``unquote``. ``unquote-splice`` can only be used when the symbol
+much like ``unquote``. ``unquote-splice`` can be used when the symbol
 being unquoted contains an iterable value, as it "splices" that iterable into
-the quasiquoted form. ``unquote-splice`` is aliased to the ``~@`` symbol.
+the quasiquoted form. ``unquote-splice`` can also be used when the value
+evaluates to a false value such as ``None``, ``False``, and ``0``, in which
+the value is treated as an empty list, therefore splicing it to an empty value.
```
gilch: I'd reword this as 
> evaluates to a false value such as ``None``, ``False``, or ``0``, in which case the value is treated as an empty list and thus does not splice anything into the form.

woodrush: I see, that sounds better.

docs/language/api.rst
```diff
@@ -1617,6 +1620,15 @@ the quasiquoted form. ``unquote-splice`` is aliased to the ``~@`` symbol.
     `(+ ~@nums)
     ;=> (u'+' 1L 2L 3L 4L)
 
+    `[1 2 ~@(if (< (nth nums 0) 0) nums)]
+    ;=> (u'+' 1L 2L)
+
+Here, the last example evaluates to ``(u'+' 1L 2L)``, since the condition
+``(< (nth nums 0) 0)`` is ``False``, which makes this ``if`` expression
+evaluate to ``None``, since the ``if`` expression here does not have a
+second body. ``unquote-splice`` then evaluates this as an empty value,
```
gilch: I'd reword this to 
```
evaluate to ``None``, because the ``if`` expression here does not have an 
else clause.
```

docs/language/api.rst
```diff
@@ -1604,9 +1604,12 @@ unquote-splice
...
 
 ``unquote-splice`` forces the evaluation of a symbol within a quasiquoted form,
-much like ``unquote``. ``unquote-splice`` can only be used when the symbol
+much like ``unquote``. ``unquote-splice`` can be used when the symbol
 being unquoted contains an iterable value, as it "splices" that iterable into
-the quasiquoted form. ``unquote-splice`` is aliased to the ``~@`` symbol.
+the quasiquoted form. ``unquote-splice`` can also be used when the value
+evaluates to a false value such as ``None``, ``False``, or ``0``, in which
+case the value is treated as an empty list and thus does not splice anything
+into the form. ``unquote-splice`` is aliased to the ``~@`` symbol.
```
gilch: I'd change the last "symbol" to "syntax". I know that's how it was written before, but the equivalent of `(unquote-splice foo)` is `~@foo` rather than `(~@ foo)` as the current wording seems to imply.

docs/language/api.rst
```diff
@@ -1617,6 +1620,15 @@ the quasiquoted form. ``unquote-splice`` is aliased to the ``~@`` symbol.
     `(+ ~@nums)
     ;=> (u'+' 1L 2L 3L 4L)
 
+    `[1 2 ~@(if (< (nth nums 0) 0) nums)]
+    ;=> (u'+' 1L 2L)
```
gilch: Python 2 is considered legacy support now. We'll likely be removing support for Python 2 soon. #1217. All outputs in the documentation should be in the Python 3 form by then.

I'm not asking you to fix it everywhere in this file, just in the section you've added to to make it consistent. I'll even approve the pull without this particular fix, but we might as well do it now instead of making it worse.

woodrush:
@gilch, thanks for the review. All the changes are included now, including the Python 2 -> Python 3 notation for the section I edited.

gilch:
I think it's good to merge now. @Kodiologist, are you still OK with this version and commit structure?

Kodiologist:
Yes, all seems fine to me. I'll merge.

PR #1356, opened 2017-08-03
Added a check for docomplete truth on line 27 of completer.py. 

ChefAndy:
Without the check, the next line will reference readline even if the library was not successfully loaded.

### DIFF
--- a/hy/completer.py
+++ b/hy/completer.py
@@ -24,10 +24,11 @@ except ImportError:
     except ImportError:
         docomplete = False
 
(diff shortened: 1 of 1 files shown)


ChefAndy:
Sorry about not following protocol. In my haste, I inadvertently created the fork in my work development organization, so I deleted it when the changes were approved. Since it's a very minor change, I'm just going to make a new pull request.

ChefAndy:
Replaced by PR #1373

PR #1359, opened 2017-08-03
Fix TypeError when defn multi-arity method returns None

schaefed:
A defn mult-arity method returning a value evaluating to None currently raises a TypeError:
```
(require [hy.contrib.multi [defn]])
(defn fun                                                                                              
  ([] ""))
(fun)  ;; raises TypeError
```
With this PR the value is returned instead.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -79,3 +79,4 @@
 * Kai Lüke <kailueke@riseup.net>
 * Neil Lindquist <archer1mail@gmail.com
 * Hikaru Ikuta <woodrush924@gmail.com>
+* David Schaefer <david.schaefe@gmail.com>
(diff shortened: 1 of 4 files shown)


Kodiologist:
Yes, but now the `TypeError` is no longer raised when you really do call a function with no matching signature:

    => (require [hy.contrib.multi [defn]])
    => (defn fun ([a] "a"))
    => (print (fun 1 2))
    None

schaefed:
Ahh, you are right, that was the whole point of the exception... I updated the PR.

Kodiologist:
Great! Now please update NEWS and add yourself to AUTHORS. Also, rebase to remove the first commit (bf3f4da) since it's not doing anything now. Finally, clarify the commit message of 61fb432; per `CONTRIBUTING.rst`, ensure its first line (which will probably still be its only line) has 50 characters or less.

schaefed:
Alright, done!

Kodiologist:
I forgot to mention: in the future, create a new branch for each PR rather than using `master`.

PR #1365, opened 2017-08-04
Expose `read`, `read_str`, and `eval` as top-level functions in Python

### LINKED ISSUE #1256: Official support for evaluating strings of Hy code from Python
Is it possible to embed some hy code inside a python file? As opposed to having the whole file be full on hy?

jdanford:
Fixes #1256

### DIFF
--- a/hy/__init__.py
+++ b/hy/__init__.py
@@ -10,3 +10,7 @@ from hy.models import HyExpression, HyInteger, HyKeyword, HyComplex, HyString, H
 
 import hy.importer  # NOQA
 # we import for side-effects.
+
+
+from hy.core.language import read, read_str  # NOQA
+from hy.importer import hy_eval as eval  # NOQA


Kodiologist:
All right, how about some tests and documentation?

jdanford:
All three functions have docstrings, but tests would be good, so I'll add some as soon as I get a chance.

Kodiologist:
By "documentation", I mean the manual, not docstrings. Think of people like the guy who wrote #1256. How would they know without asking that using these functions is how you can use Hy in Python?

jdanford:
Of course, not sure why I misunderstood! Should I add a new section under [Using Hy from Python](http://docs.hylang.org/en/master/language/interop.html#using-hy-from-python)?

Kodiologist:
Sure, that makes sense.

jdanford:
Whoops, just noticed I did this on `master` instead of a branch! New PR is #1369

PR #1369, opened 2017-08-05
Expose `read`, `read_str`, and `eval` in Python

jdanford:
Fixes #1256 (and replaces #1365)

PR #1370, opened 2017-08-05
Add Jordan Danford to AUTHORS

jdanford:


Kodiologist:
This doesn't need to be a separate PR; just put this commit on top of the other one in #1369.

PR #1373, opened 2017-08-07
hy/completer.py:27 docomplete truth check

ChefAndy:
Replaces malformatted PR #1356 

Without the check, the next line will reference readline even if the library was not successfully loaded.


Kodiologist:
In the future, remember to make a feature branch rather than asking to pull from `master`.

ChefAndy:
Oy, of course. Such a small commit that I got ahead oh myself. Really batting .1000 with this change. 😁😬😳

Kodiologist:
We forgive you. Thanks for contributing. :)

PR #1375, opened 2017-08-07
Reimplement variadic if without recursion

### LINKED ISSUE #842: Conditionals waste assignments.

kirbyfan64:
Closes #842. Agaaain.

@gilch How does this look?

### DIFF
--- a/hy/core/bootstrap.hy
+++ b/hy/core/bootstrap.hy
@@ -13,9 +13,9 @@
         args (list args)
         result (if* (% n 2) (.pop args) `None))
 
-    (while args
-      (setv then-branch (.pop args)
-            cond (.pop args))
+  (while args
+    (setv then-branch (.pop args)
+          cond (.pop args))
     (setv result `(if* ~cond ~then-branch ~result)))
 
   result)
--- a/hy/core/bootstrap.hy
+++ b/hy/core/bootstrap.hy
@@ -8,13 +8,17 @@
 
 (defmacro if [&rest args]
   "if with elif"
-  (setv n (len args))
-  (if* n
-       (if* (= n 1)
-            (get args 0)
-            `(if* ~(get args 0)
-                  ~(get args 1)
-                  (if ~@(cut args 2))))))
+
+  (setv n (len args)
+        args (list args)
+        result (if* (% n 2) (.pop args) `None))
+
+    (while args
+      (setv then-branch (.pop args)
+            cond (.pop args))
+    (setv result `(if* ~cond ~then-branch ~result)))
+
+  result)
 
 (defmacro macro-error [location reason]
   "error out properly within a macro"
--- a/tests/native_tests/native_macros.hy
+++ b/tests/native_tests/native_macros.hy
@@ -306,6 +306,20 @@
   (assert (= (lif-not 0 "false" "true") "true")))
 
 
+(defn test-if-elif []
+  "test that sequential ifs get compiled to elifs"
+  (import ast)
+  (import [astor.codegen [to-source]])
+  (import [hy.importer [import-buffer-to-ast]])
+
+  (setv source "(if 1 11 2 (setv x 22) 3 33)")
+
+  (setv source (to-source (import-buffer-to-ast source "foo")))
+  (print source)
+  ; Make sure 1) there's a compiled elif present and 2) only one anon var is generated
+  (assert (and (in "elif" source) (not-in "_hy_anon_var_2" source))))
+
+
 (defn test-defmain []
   "NATIVE: make sure defmain is clean"
   (global --name--)


gilch:
I'd also like a test to make sure the wasted assignment isn't in the AST.

You forgot a pair of square brackets in that import #851, but we could avoid the `reduce` import and avoid re-implementing `partition` by using two pop/assignments in a loop, which conveniently produces our arguments in the desired reverse order.
```
(defmacro if [&rest args]
  "if with elif"
  (setv args (list args)
        result (if* (% (len args)
                       2)
                    (.pop args)
                    `None))
  (while args
    (setv result `(~(.pop args) ~result)
          result `(if* ~(.pop args) ~@result)))
  result)
```
This builds it inside-out just like your reduce. It's more imperative than functional, but we don't have all our functional tools yet it bootstrap.

We should also consider removing `if*` from the docs to discourage people from nesting it in their own macros.

gilch:
I also noticed the tutorial says there's no `elif` in Hy and you should use `cond`. We should clarify that `if` can have `elif` clauses.

Kodiologist:
You know, we could also get rid of the `if` macro, rename `if*` to `if`, and make `cond` a bootstrap macro. There's no real reason to have both `cond` and multi-branch `if` in the same language, right?

Kodiologist:
> We should also consider removing `if*` from the docs to discourage people from nesting it in their own macros.

Agreed; the starred special forms that are supposed to be internal to Hy shouldn't be documented for users.

Kodiologist:
> In Hy, `cond` has an implicit `do` after the condition, which is really only good for side effects in the imperative style.

And for `setv`, since there's no `let`.

gilch:
> And for `setv`, since there's no `let`.

Yup. That's a real pain point. I wish we could have a proper `let` in Hy. But `setv` *is* a side-effect--albeit a *localized* one when using...locals. It helps if you put the `if` inside the `setv` instead of the other way around.

With the threading macros, unpacking, `doto` or `comp`/`partial` (and sometimes or `juxt`), and small, focused, functions, I've found that I don't actually need `let` as much as I thought. When I do use a `setv`, it's usually just one at the top of the function, not in branches. But in the rare case you need it, you can use the `do` block. It's the exceptional case, not the common one.

kirbyfan64:
@gilch I changed the implementation like you said.

> I'd also like a test to make sure the wasted assignment isn't in the AST.

Is this even really possible right now?

gilch:
> Is this even really possible right now?

I think so. Even at the native Hy level, we can get the AST compilation of a test form using `disassemble`, which we can inspect programmatically. How much did the AST change between Python versions in a simple if branch? We could also use astor and see if the word `elif` appears in the string.

Kodiologist:
Travis is showing a failure because the Python 3.5 job was canceled before it ran. You can probably get it to run the tests again with something like `git commit --amend --reset-author && git push -f`. Or just test Python 3.5 locally.

Kodiologist:
> Is this even really possible right now?

My guess is yes. You can try making a test in `test_ast.py` that runs `can_compile` and recursively walks through the output counting the number of `ast.Assign` nodes.

kirbyfan64:
Let's see if Travis works now...

Kodiologist:
Yay, it does.

Kodiologist:
I'll take a crack at testing.

Kodiologist:
Actually, I don't think it works. `(if 1 1 (if 2 (setv x 2) (if 3 3)))` still uses two anon vars, not one.

kirbyfan64:
I should probably just check the number of elifs...

Kodiologist:
@kirbyfan64, are you still working on this?

kirbyfan64:
I mean, I'm not sure what exactly is wrong because of the `setv` interactions...

Kodiologist:
If you haven't already, I would try putting debugging print statements in `compile_if` to ensure that it's seeing the Hy model tree you're expecting it to see.

kirbyfan64:
*Whoops*, my test case was wrong. ¯\\_(ツ)_/¯

Kodiologist:
It still doesn't seem to work:

    $ echo '(cond [1 1] [2 (setv x 2)] [3 3])' | hy2py
    if 1:
        _hy_anon_var_2 = 1
    else:
        if 2:
            x = 2
            _hy_anon_var_1 = None
        else:
            _hy_anon_var_1 = (3 if 3 else None)

Kodiologist:
I'm closing this for inactivity. Please reopen it when you'd like eyeballs again.

PR #1376, opened 2017-08-09
Self-host Hy in Hy

woodrush:
I finally got [py2hy](https://github.com/woodrush/py2hy) (#876) working! There currently are no tests, but I created [HyHy](https://github.com/woodrush/hy/tree/hyhy), which is Hy where all *.py code is translated to *.hy, and all the pytests run correctly.

Using `hy2py`, we can maintain Hy in Hy, and run hy2py to create a Python version to simultaneously update the language as written in Hy. I know it rewrites too much files, but I think maintaining Hy in Hy makes things easier, and I wonder if there is a way we can work this out.

In the meantime I plan to integrate [py2hy](https://github.com/woodrush/py2hy) into Hy as well.


Kodiologist:
I'm super impressed that you managed to pull this off, and py2hy would be awesome to have, but I'd rather have Hy's lowest-level bits implemented in Python rather than Hy. Self-hosting comes with a lot of necessary complications. If the alternative is writing a compiler in assembly or C or something, it can be worth it. But Python is already a powerful language. I don't think we're going to much improve the readability or maintainability of the codebase, or even make it much shorter, by replacing all the Python with Hy.

gilch:
> I don't think we're going to much improve the readability or maintainability of the codebase, or even make it much shorter, by replacing all the Python with Hy.

You might be surprised what good macros can do. This might be worth it someday. But I think HyHy should be a separate project from Hy for the foreseeable future. I don't even want to *think* about self-hosting before we get to Hy version 1.0, get our tooling in order, and drop Python 2 support.

kirbyfan64:
Eh... I'm with @gilch on this. Especially since everything is still in a state of flux, it'd be a lot easier to just wait it out until at *least* 1.0.

gilch:
Three's a veto, so I'll close this, but discussion can continue. We might reconsider self-hosting later if HyHy is compelling as a separate project. Consider running it through Hydiomatic. The machine compiled code is kind of ugly compared to competent human-written Hy. And some of that probably can't be helped before we implement `return`. I know it's just a starting point.

woodrush:
Thanks for the feedback. I see. The code (the `py2hy`ed Hy, and the `py2hy | hy2py`ed Python) is not beautiful right now, and I could wait for self-hosting when both the Hy and Python code in this PR becomes readable enough. I also see that we could wait until when reach version 1.0, regarding the complications it would bring. Until then, I'll see if I could make the code more readable and more simple.

I tried the pip and HEAD version of Hydiomatic a little, but it seemed to need some workarounds to make it work with the current version of Hy. I'll look into it to see if I could use it to simplify the code.

Kodiologist:
Hydiomatic was last updated in March 2016, and Hy has changed a lot since, so it probably won't be useful without updates.

PR #1377, opened 2017-08-09
Fix copy behaviour of HyComplex

schaefed:
I encountered this, while trying to fix PR #1374:
```
(import [hy.models [HyComplex]])
(import [copy [copy deepcopy]])
(setv x (HyComplex 42j))
(copy x)      ;; returns 0j
(deepcopy x)  ;; returns 0j
```
using a literal or the python builtin `complex` everything works as expected:
```
(setv y 42j)
(copy y)       ;; returns 42j
(deepcopy y)   ;; returns 42j

(setv z (complex 42j))
(copy z)       ;; returns 42j
(deepcopy z)   ;; returns 42j
```
The proposed solution is to explicitly implement `__copy__` and `__deepcopy__` in HyComplex.

### DIFF
--- a/hy/models.py
+++ b/hy/models.py
@@ -171,18 +171,17 @@ class HyComplex(HyObject, complex):
     complex(foo) was called, given HyComplex(foo).
     """
 
-    def __new__(cls, num, *args, **kwargs):
-        value = super(HyComplex, cls).__new__(cls, strip_digit_separators(num))
-        if isinstance(num, string_types):
-            p1, _, p2 = num.lstrip("+-").replace("-", "+").partition("+")
+    def __new__(cls, real, imag=0, *args, **kwargs):
+        if isinstance(real, string_types):
+            value = super(HyComplex, cls).__new__(
+                cls, strip_digit_separators(real)
+            )
+            p1, _, p2 = real.lstrip("+-").replace("-", "+").partition("+")
+            check_inf_nan_cap(p1, value.imag if "j" in p1 else value.real)
             if p2:
-                check_inf_nan_cap(p1, value.real)
                 check_inf_nan_cap(p2, value.imag)
-            elif "j" in p1:
-                check_inf_nan_cap(p1, value.imag)
-            else:
-                check_inf_nan_cap(p1, value.real)
-        return value
+            return value
+        return super(HyComplex, cls).__new__(cls, real, imag)
 
 _wrappers[complex] = HyComplex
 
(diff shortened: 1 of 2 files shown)


Kodiologist:
I don't think this is correct, because HyModels may carry mutable attributes representing their source position.

    >>> import hy
    >>> x = hy.HyComplex("3+4j")
    >>> x
    (3+4j)
    >>> x.start_line = 5
    >>> import copy
    >>> y = copy.copy(x)
    >>> y
    (3+4j)
    >>> y.start_line = 6
    >>> x.start_line
    6

schaefed:
Okay, that was maybe to simplistic... 
After digging further, I guess I found the root of the problem. `HyComplex` only takes one positional argument, but when calling `copy` or `deepcopy` on an `HyComplex` object the real and imag parts a passed to the constructor as seperate arguments. The latter always ends up within `*args` which is not passed to the superclass. 
The proposed implementation handles that in a way, that there is no need to overwrite `__copy__` and `__deepcopy__` anymore.

tests/test_models.py
```diff
@@ -125,3 +126,17 @@ def test_cons_replacing():
         assert True is False
     except IndexError:
         pass
+
+
+def test_number_model_copy():
+    i = HyFloat(42)
```
Kodiologist: Did you mean to say `HyInteger` here? There's a `HyFloat` test just a few lines down.

hy/models.py
```diff
@@ -171,18 +171,17 @@ class HyComplex(HyObject, complex):
...
 
-    def __new__(cls, num, *args, **kwargs):
-        value = super(HyComplex, cls).__new__(cls, strip_digit_separators(num))
-        if isinstance(num, string_types):
-            p1, _, p2 = num.lstrip("+-").replace("-", "+").partition("+")
+    def __new__(cls, real, imag=0, *args, **kwargs):
+        if isinstance(real, string_types):
+            value = super(HyComplex, cls).__new__(
+                cls, strip_digit_separators(real)
+            )
```
Kodiologist: You don't need a separate line for the close parenthesis. It can go at the end of the preceding line.

schaefed: I know, but I prefer it that way... if you don't, I can change it...

Kodiologist: Eh, no biggie.

tests/test_models.py
schaefed: You are right, sorry! Changed.

Kodiologist:
@kirbyfan64, I should ask if you still approve, since the implementation has changed completely.

hy/models.py
kirbyfan64: FWIW most Python style guides prefer the closing paren on the same line.

PR #1378, opened 2017-08-10
Make `assoc` a macro instead of a special form

### LINKED ISSUE #841: Make assoc a function instead of a special form
Even though Hy implements `assoc`, somehow it's not a valid function in the local scope:

```
$ hy --spy
hy 0.11.0 using CPython(default) 2.7.10 on Darwin
=> assoc
assoc
NameError: name 'assoc' is not defined
```

Strange. Let's import `toolz` version:

```
=> (import [toolz.dicttoolz [assoc]])
from toolz.dicttoolz import assoc
=> assoc
assoc
<function assoc at 0x1022af140>
```

Good. Let's try:

```
=> (= (assoc {} "Say" "Hy") {"Say" "Hy"})
{}[u'Say'] = u'Hy'
(None == {u'Say': u'Hy', })
False
=> 
```

Double WTF.
1. Hy's `assoc` implementation is weird since it mutates instead of returning a new dict.
2. I can't shadow with a saner version of `assoc`.

:crying_cat_face:

### LINKED ISSUE #1068: make assoc use a gensym

Kodiologist:
- Closes #841.
- Fixes #1068.

See the discussion in #1346.

### DIFF
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -20,6 +20,22 @@
          ~@(interleave (repeat name) rest))
      ~name))
 
+
+(defmacro assoc [coll k1 v1 &rest other-kvs]
+  (if (odd? (len other-kvs))
+    (macro-error (last other-kvs)
+                 "`assoc` takes an odd number of arguments"))
+  (setv c (if other-kvs
+            (gensym "c")
+            coll))
+  `(setv ~@(+ (if other-kvs
+                [c coll]
+                [])
+              #* (genexpr [`(get ~c ~k) v]
+                          [[k v] (partition (+ (, k1 v1)
+                                               other-kvs))]))))
+
+
 (defmacro with [args &rest body]
   "shorthand for nested with* loops:
   (with [x foo y bar] baz) ->
--- a/NEWS
+++ b/NEWS
@@ -32,6 +32,7 @@ Changes from 0.13.0
    * try form now possible in defmacro/deftag
    * Fixed a crash when `with` suppresses an exception. `with` now returns
      `None` in this case.
+   * `assoc` now evaluates its arguments only once each
 
    [ Misc. Improvements ]
    * `read`, `read_str`, and `eval` are exposed and documented as top-level
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1406,30 +1406,6 @@ class HyASTCompiler(object):
                             step=step.expr),
             ctx=ast.Load())
 
-    @builds("assoc")
-    @checkargs(min=3, even=False)
-    def compile_assoc_expression(self, expr):
-        expr.pop(0)  # assoc
-        # (assoc foo bar baz)  => foo[bar] = baz
-        target = self.compile(expr.pop(0))
-        ret = target
-        i = iter(expr)
-        for (key, val) in ((self.compile(x), self.compile(y))
-                           for (x, y) in zip(i, i)):
-
-            ret += key + val + ast.Assign(
-                lineno=expr.start_line,
-                col_offset=expr.start_column,
-                targets=[
-                    ast.Subscript(
-                        lineno=expr.start_line,
-                        col_offset=expr.start_column,
-                        value=target.force_expr,
-                        slice=ast.Index(value=key.force_expr),
-                        ctx=ast.Store())],
-                value=val.force_expr)
-        return ret
-
     @builds("with_decorator")
     @checkargs(min=1)
     def compile_decorate_expression(self, expr):
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -328,9 +328,9 @@
     (do
       (defn merge-entry [m e]
         (setv k (get e 0) v (get e 1))
-        (if (in k m)
-          (assoc m k (f (get m k) v))
-          (assoc m k v))
+        (setv (get m k) (if (in k m)
+                          (f (get m k) v)
+                          v))
         m)
       (defn merge2 [m1 m2]
         (reduce merge-entry (.items m2) (or m1 {})))
--- a/tests/native_tests/language.hy
+++ b/tests/native_tests/language.hy
@@ -628,11 +628,26 @@
   (assoc vals "two" "three")
   (assert (= (get vals "two") "three")))
 
+
 (defn test-multiassoc []
   "NATIVE: test assoc multiple values"
   (setv vals {"one" "two"})
   (assoc vals "two" "three" "four" "five")
   (assert (and (= (get vals "two") "three") (= (get vals "four") "five") (= (get vals "one") "two"))))
+
+
+(defn test-assoc-eval-lvalue-once []
+  ;; https://github.com/hylang/hy/issues/1068
+  "`assoc` only evaluates its lvalue once"
+  (setv counter [])
+  (setv d {})
+  (defn f []
+    (.append counter 1)
+    d)
+  (assoc (f)  "a" 1  "b" 2  "c" 3)
+  (assert (= d {"a" 1  "b" 2  "c" 3}))
+  (assert (= counter [1])))
+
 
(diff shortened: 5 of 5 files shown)


hy/core/macros.hy
```diff
@@ -20,6 +20,20 @@
          ~@(interleave (repeat name) rest))
      ~name))
 
+
+(defmacro assoc [coll k1 v1 &rest other-kvs]
+  (if (odd? (len other-kvs))
+    (macro-error (last other-kvs) "`assoc` takes an odd number of arguments"))
+  (setv c (if other-kvs (gensym "c") coll))
+  `(setv
+    ~@(if other-kvs [c coll] [])
```
gilch: Since #1349, splice works on any False value. This could be `~@(if other-kvs [c coll])` (or `when`).

hy/core/macros.hy
```diff
@@ -20,6 +20,20 @@
...
+    ~@(sum
+      (list-comp [`(get ~c ~k) v] [[k v] (zip
+        (+ (, k1) (cut other-kvs 0 None 2))
+        (+ (, v1) (cut other-kvs 1 None 2)))])
+      [])))
```
gilch: `sum` and `zip` are not indented. And is there a reason you're avoiding `partition`? That would be simpler.
```Hy
~@(sum (list-comp [`(get ~c ~k) v] [[k v] (partition other-kvs))
       (and other-kvs `(get k1 v1)))
```

hy/core/macros.hy
Kodiologist: Oh, right.

gilch: `~@(and other-kvs [c coll])` would also work.

hy/core/macros.hy
gilch: `~@(sum (list-comp [`(get ~c ~k) v] [[k v] (cons [k1 v1] (partition other-kvs))])` should also work. I should be testing these at the repl.

gilch: Hmm, nope. A cons onto a generic iterable doesn't work like that for some reason. It would have to be `(chain [[k1 v1]] (partition other-kvs))`, or spliced like
```Hy
~@(sum (list-comp [`(get ~c ~k) v] [[k v] `(~[k1 v1] ~@(partition other-kvs))])
```

Kodiologist:
How's that?

gilch:
Not what I meant. Sorry I got it wrong the first try, you probably didn't see the edit. How about like this.
```Hy
(defmacro assoc% [coll k1 v1 &rest other-kvs]
  (if (odd? (len other-kvs))
    (macro-error (last other-kvs) "`assoc` takes an odd number of arguments"))
  (setv c (if other-kvs (gensym "g-assoc") coll))
  `(setv
    ~@(and other-kvs [c coll])
    ~@(sum (list-comp [`(get ~c ~k) v] [[k v] (partition other-kvs)])
           [`(get ~c ~k1) v1])))
```

Kodiologist:
I see the motivation for using `partition` instead of `zip`, but why take the `` [`(get ~c ~k1) v1])`` out of the `list-comp`?

gilch:
Hmm, yes. I don't like it there either. I think it's clearer to put all the assignments in the order they'd appear in the expansion.
```Hy
(defmacro assoc [coll k1 v1 &rest other-kvs]
  (if (odd? (len other-kvs))
    (macro-error (last other-kvs) "`assoc` takes an odd number of arguments"))
  (setv c (if other-kvs (gensym "g-assoc") coll))
  `(setv
    ~@(and other-kvs [c coll])
    (get ~c ~k1) ~v1
    ~@(+ #* (list-comp [`(get ~c ~k) v [[k v] (partition other-kvs)]]))))
```
This looks more like a template for the expansion. I like it there better.

We shouldn't ever be using `sum` on lists anyway. It's a mere implementation detail that this works at all. `sum`'s docstring says 
> This function is intended specifically for use with numeric values and may
    reject non-numeric types.

Note that `+` wouldn't have worked before #1349 because `(+)` is `0`, not `[]`. (We could have used `(+ [] #* (list-comp ...`)

We do have the shadow `+` here, right? If not, the `chain.from-iterable` would also avoid the last `[]` in the `sum` altogether, though it's an unfortunately long name for such a useful operation. Maybe we could give it a `*chain` alias or something, but that's another issue. I see this pattern a lot in macros, so I added a `chain-comp` macro in my destructure PR. It might be useful in core if we see this kind of pattern a lot there. But again, another issue.

Kodiologist:
That's true about `sum`, and we do indeed have shadow `+` here. Would you accept this?

hy/core/macros.hy
```diff
@@ -20,6 +20,16 @@
...
      ~name))
 
+
+(defmacro assoc [coll k1 v1 &rest other-kvs]
+  (if (odd? (len other-kvs))
+    (macro-error (last other-kvs) "`assoc` takes an odd number of arguments"))
+  (setv c (if other-kvs (gensym "c") coll))
+  `(setv ~@(+ (if other-kvs [c coll] []) #* (genexpr
+    [`(get ~c ~k) v]
+    [[k v] (partition (+ (, k1 v1) other-kvs))]))))
```
gilch: This is not indented properly, which makes it hard to read. Emacs can do this for you.

Kodiologist: It's indented with two spaces, same as everything else.

gilch: No, it is not indented 2 spaces. After bracket counting, I can see that arguments to `genexpr` are indented as if they were arguments to `setv`. It will take 42 more spaces to get it indented two spaces past the start of their list. To avoid excessively wide lines, instead of just inserting 42 spaces at the start of those two lines, add some newlines.

One must never put list elements on a column before (or equal to) the bracket opening the list, in any Lisp. Elements must always appear on a column following the opening bracket of the list they belong to. And when I say "list", this applies to all Hy models inheriting from HyList.

Kodiologist:
Okay.

hy/core/macros.hy
```diff
@@ -20,6 +20,15 @@
          ~@(interleave (repeat name) rest))
      ~name))
 
+
+(defmacro assoc [coll k1 v1 &rest other-kvs]
+  (if (odd? (len other-kvs))
+    (macro-error (last other-kvs) "`assoc` takes an odd number of arguments"))
+  (setv c (if other-kvs (gensym "c") coll))
+  `(setv ~@(+ (if other-kvs [c coll] []) #*
+    (genexpr [`(get ~c ~k) v] [[k v] (partition (+ (, k1 v1) other-kvs))]))))
```
gilch: `genexpr` is now indented at -7 from the opening bracket of its list. Line it up with the `(if`, or add a newline and line both the `(if` and `(genexpr` at +2.

gilch: It would also look better if `#*` were on the same line as the thing it's unpacking.

Kodiologist: [Please stop.](https://github.com/hylang/hy/pull/1338#discussion_r131307803)

gilch: This is not legible, Kodi.
I cannot approve obfuscated code.

You don't seem to understand how to format Lisp, or why it is important to. Please read up on it, or let Emacs do it for you.

gilch:
I added a formatting commit and I fixed the conflict with NEWS using the web editor, which added a merge commit. @Kodiologist is this commit structure OK?

Kodiologist:
I don't like to keep merge commits inside a PR (as opposed to the merge commit that brings the PR into master), so I rebased it away.

Kodiologist:
Thanks!

PR #1379, opened 2017-08-10
Add #[DELIM[ … ]DELIM] syntax for string literals

### LINKED ISSUE #1042: add heredocs to Hy
It would be nice if Hy had something like doctests, see #1019. It would also be nice to auto-generate reference docs from our docstrings, see #1040.

Doctests are only feasible in Python because of triple quotes. Hy doesn't have Python's triple quotes. It can do multi-line strings, but quotation marks must be escaped. (I had already complained about this a bit in #831.) Python can even have triple quotes inside triple quotes because `'''` and `"""` both work. Raw strings with quotes inside can (and do) potentially happen with regexes, so escaping the occasional quotation mark isn't really good enough. Python has raw triple quotes `r'''` and `r"""` for this. Triple quotes are a good solution.

But they might not be such a good fit for Hy. We're already using `'` for something else and you need both `'''` and `"""` to make triple quotes work well. Hy's string literal capabilities are weaker than Python's, but I think we can do even better than triple quotes.

I propose we add [heredocs](https://en.wikipedia.org/wiki/Here_document) to Hy's reader. This lets you specify an arbitrary terminator for your string. Racket Scheme and many other languages have this feature. Common Lisp has reader macros powerful enough to do it, but Hy doesn't. We'd have to build it in.

Heredocs could give us doctests, auto-generated reference docs, better regex support, and 80% of what Common Lisp reader macros can do, for e.g. DSL's (with the limitation that you have to wrap it in custom delimiters).

### LINKED ISSUE #1287: Alternative syntax for string literals

Kodiologist:
- Closes #1042.
- Closes #1287.
- Replaces #1327.

To do:
- [x] Documentation
- [x] NEWS

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -745,6 +745,12 @@ class HyASTCompiler(object):
             return imports, HyExpression([HySymbol(name),
                                           HyString(form)]).replace(form), False
 
+        elif isinstance(form, HyString):
+            x = [HySymbol(name), form]
+            if form.brackets is not None:
+                x.extend([HyKeyword(":brackets"), form.brackets])
+            return imports, HyExpression(x).replace(form), False
+
         return imports, HyExpression([HySymbol(name),
                                       form]).replace(form), False
 
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,8 @@ Changes from 0.13.0
      longer names
    * Periods are no longer allowed in keywords
    * `eval` is now a function instead of a special form
+   * Added a form of string literal called "bracket strings" delimited by
+     `#[FOO[` and `]FOO]`, where `FOO` is customizable
    * The compiler now automatically promotes values to Hy model objects
      as necessary, so you can write ``(eval `(+ 1 ~n))`` instead of
      ``(eval `(+ 1 ~(HyInteger n)))``
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -52,8 +52,8 @@ digits.
 
     (print 10,000,000,000 10_000_000_000)
 
-Unlike Python, Hy provides literal forms for NaN and infinity: `NaN`, `Inf`,
-and `-Inf`.
+Unlike Python, Hy provides literal forms for NaN and infinity: ``NaN``,
+``Inf``, and ``-Inf``.
 
 string literals
 ---------------
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -58,20 +58,41 @@ Unlike Python, Hy provides literal forms for NaN and infinity: ``NaN``,
 string literals
 ---------------
 
-Unlike Python, Hy allows only double-quoted strings (e.g., ``"hello"``). The
-single-quote character ``'`` is reserved for preventing the evaluation of a
-form (e.g., ``'(+ 1 1)``), as in most Lisps.
+Hy allows double-quoted strings (e.g., ``"hello"``), but not single-quoted
+strings like Python. The single-quote character ``'`` is reserved for
+preventing the evaluation of a form (e.g., ``'(+ 1 1)``), as in most Lisps.
 
 Python's so-called triple-quoted strings (e.g., ``'''hello'''`` and
 ``"""hello"""``) aren't supported. However, in Hy, unlike Python, any string
-literal can contain newlines.
+literal can contain newlines. Furthermore, Hy supports an alternative form of
+string literal called a "bracket string" similar to Lua's long brackets.
+Bracket strings have customizable delimiters, like the here-documents of other
+languages. A bracket string begins with ``#[FOO[`` and ends with ``]FOO]``,
+where ``FOO`` is any string not containing ``[`` or ``]``, including the empty
+string. For example::
 
-Whether running under Python 2 or Python 3, Hy treats string literals as
-sequences of Unicode characters by default, and allows you to prefix a literal
-with ``b`` to treat it as a sequence of bytes. So when running under Python 3,
-Hy translates ``"foo"`` and ``b"foo"`` to the identical Python code, but when
-running under Python 2, ``"foo"`` is translated to ``u"foo"`` and ``b"foo"`` is
-translated to ``"foo"``.
+   => (print #[["That's very kind of yuo [sic]" Tom wrote back.]])
+   "That's very kind of yuo [sic]" Tom wrote back.
+   => (print #[==[1 + 1 = 2]==])
+   1 + 1 = 2
+
+A bracket string can contain newlines, but if it begins with one, the newline
+is removed, so you can begin the content of a bracket string on the line
+following the opening delimiter with no effect on the content. Any leading
+newlines past the first are preserved.
+
+Plain string literals support :ref:`a variety of backslash escapes
+<py:strings>`. To create a "raw string" that interprets all backslashes
+literally, prefix the string with ``r``, as in ``r"slash\not"``. Bracket
+strings are always raw strings and don't allow the ``r`` prefix.
+
+Whether running under Python 2 or Python 3, Hy treats all string literals as
+sequences of Unicode characters by default, and allows you to prefix a plain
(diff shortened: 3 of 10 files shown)


Kodiologist:
@gilch @kirbyfan64 Is there any reason you guys aren't reviewing this? I thought we settled on the Lua style.

gilch:
> I thought we settled on the Lua style.

I though we did, or I did, at least. There are competing priorities, that's all. I can make this a priority since you think it's important.

This can come later, but we'll want to at least update hy-mode for the new syntax, and vim-hy too, but no-one seems to be maintaining that. Since hy-mode isn't settled even for what we've got so far, I've been de-prioritizing syntax changes that would break it compared to other things.

But even before then, both emacs and vim should mostly highlight properly if Hy files if we pick `#["[...]"]` as the delimiters. I'd also like to get a basic linter going for .hy files #1398. [Parlinter](https://github.com/shaunlebron/parlinter) looks promising, but it won't work with the new syntax generally, just with `#["[`, which it will recognize as a string inside `[]`. (The other proposed string syntax options probably can't even do that much.) We're not the only Lisp with this kind of problem, so we can probably get that fixed. I'd even translate it to Python and do it myself eventually, but priorities.

Kodiologist:
> There are competing priorities, that's all. I can make this a priority since you think it's important.

My own philosophy is that I should try to review PRs (that are ready for review, to my knowledge) before anything else (like writing issues, working on my own PRs, etc.), because PR reviewing seems to be the primary bottleneck for Hy development.

gilch:
```Hy
=> (. '#[foo[bar]foo] hashstring_delim)
from hy import HyString
HyString('bar').hashstring_delim
  File "c:\users\me\documents\github\hylang-hy\hy\importer.py", line 201, in hy_eval
AttributeError: 'HyString' object has no attribute 'hashstring_delim'
```
I can't seem to access the delimiter from the model. Is the compiler stripping it out somewhere?

gilch:
Looks like the macros can get to it though.
```Python
=> (defmacro foo [s] (. s hashstring_delim))
import hy
hy.macros.macro('foo')((lambda s: s.hashstring_delim))
<function <lambda> at 0x000001BDD82476A8>
=> (foo #[x[]x])
'x'
'x'
```

Kodiologist:
I think what you're seeing the first example is a limitation in how quoting works. The generated code is `HyString('bar')`, which doesn't set the attribute. It seems clear that this isn't right, but I'm not sure what needs to be fixed.

Kodiologist:
I think `HyASTCompiler._render_quoted_form` needs an upgrade to set the attribute. I'm looking into it now.

Kodiologist:
Done.

gilch:
Tested too, that looks good.

Seeing those `elif isinstance` statements in a super-long method makes me want to move that logic to methods in the individual Hy models themselves. It's another issue though. For now, it's more consistent this way. The refactor should be in #1399, or at least wait until after that's merged.

Now that the delimiter is a keyword argument, I think I should put it in the HyString repr in #1360, at least when it doesn't have the default value of `None`. (This will have to be merged first, so I can rebase.) But perhaps it should have a shorter name than `hashstring_delim` now? It's not kwonly, so I could just put it second positionally, but it might be clearer with the word.

We also need news and docs, which should also specify that some number of `=`'s are the preferred delimiters when `#[[` won't do, and there's not a specific reason to use something else.

Kodiologist:
> Seeing those `elif isinstance` statements in a super-long method makes me want to move that logic to methods in the individual Hy models themselves. It's another issue though. For now, it's more consistent this way.

Agreed on both counts.

> But perhaps it should have a shorter name than `hashstring_delim` now?

It doesn't make a difference to me, I guess. `hashstring_delim` just seemed like the most obvious choice. Is there any particular alternative you'd prefer?

> We also need news and docs

Will do.

> which should also specify that some number of `=`'s are the preferred delimiters when `#[[` won't do, and there's not a specific reason to use something else.

I disagree. I don't want to count `=`s. Let's let the reader decide, or at least defer this debate until a style-guide PR.

gilch:
> I don't want to count `=`s.

I don't want to count more than three, but how often does that come up?

> at least defer this debate until a style-guide PR

OK, I guess.

> hashstring_delim just seemed like the most obvious choice. Is there any particular alternative you'd prefer?

I didn't have a particular alternative in mind, but I guess `brackets` could work.

@kirbyfan64, anything else you'd like to add? Do you have a preference for the HyString reprs?

Kodiologist:
Done.

Kodiologist:
@gilch This good? I wasn't sure whether you wanted me to change `hashstring_delim` to `brackets` or you were just thinking out loud.

gilch:
Go ahead and change it to `brackets`.

Kodiologist:
@gilch Done.

hy/compiler.py
```diff
@@ -745,6 +745,12 @@ def _render_quoted_form(self, form, level):
             return imports, HyExpression([HySymbol(name),
                                           HyString(form)]).replace(form), False
 
+        elif isinstance(form, HyString):
+            l = [HySymbol(name), form]
+            if form.brackets is not None:
+                l.extend([HyKeyword(":brackets"), form.brackets])
+            return imports, HyExpression(l).replace(form), False
```
gilch: This is minor, but I'd prefer to avoid single-character names that can be confused with other characters in Python, like `lI1` or `0O`. These are distinguishable in a good programmer font, but you might have to squint. This is the kind of thing flake8 should be able to catch.

gilch:
The implementation seems good now.

I don't like the name "hashstring" it makes it sound like it has something to do with hash functions, but it doesn't. It's just from the `#` character. We could call them "bracket strings", unless you have a better idea?

Kodiologist:
"Bracket string" seems as good anything. I will likely get around to the mild inconvenience of editing this tomorrow.

Kodiologist:
Both changes made.

kirbyfan64:
Glad to see this finally getting in!

ekaschalk:
Bracket string literals are now fully supported in `hy-mode` master branch.

PR #1380, opened 2017-08-11
Add `return` as a special form

### LINKED ISSUE #739: Return expression

Kodiologist:
Closes #739.

To do:
- [ ] Documentation
- [ ] NEWS

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -322,21 +322,25 @@ def _raise_wrong_args_number(expression, error):
                                len(expression)))
 
 
+def _nargs(n):
+    return "%d argument%s" % (n, ("" if n == 1 else "s"))
+
+
 def checkargs(exact=None, min=None, max=None, even=None, multiple=None):
     def _dec(fn):
         def checker(self, expression):
             if exact is not None and (len(expression) - 1) != exact:
                 _raise_wrong_args_number(
-                    expression, "`%%s' needs %d arguments, got %%d" % exact)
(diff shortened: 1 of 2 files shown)


hy/compiler.py
```diff
@@ -2312,6 +2316,19 @@ def compile_function_def(self, expression):
 
         return ret
 
+    @builds("return")
+    @checkargs(max=1)
+    def compile_return(self, expr):
+        value = None
+        ret = Result()
+        if len(expr) > 1:
+            value = self.compile(expr[1])
```
kirbyfan64: Slightly bikeshedding, but here `value` is switching types which is kinda confusing. Usually in the compiler code, this would just be something like:

```python
compiled_expr = self.compile(expr[1])
ret += compiled_expr
value = compiled_expr.force_expr
```

or maybe another name still.

Kodiologist: I just realized that I can get rid of one of the assignments, because `ret.force_expr` works, so I did that.

Kodiologist:
@kirbyfan64 How's that?

Kodiologist:
Whoops, I hadn't written the NEWS item or documentation yet. I probably should've done that after the first approval. I'll do it in a new PR.

PR #1385, opened 2017-08-19
Add remarks in tutorial about different input functions in Python 2/3

freezas:
Hi,

I've gone through the tutorial for the first time and liked it!

I do think the tutorial can be improved by making things more clear and explaining a few more things. I still don't understand everything completely, but I think this is a first small change that can help other future (maybe more inexperienced in Python) readers.

If I should modify something or do something different to accept pull requests easier and faster, let me know.

Thanks!


Kodiologist:
Duplicate of #1272. It's probably best to make the whole tutorial Python 3 instead of switching versions mid-tutorial. (But in general, you're not wrong; the documentation isn't very good.)

PR #1386, opened 2017-08-19
Tutorial in Python3

freezas:
After discussions in #1385 and #1272 I decided to change all the examples to Python3.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -82,3 +82,4 @@
 * David Schaefer <david.schaefe@gmail.com>
 * Jordan Danford <jordandanford@gmail.com>
 * Andrew Silva <asilva@law.harvard.edu>
+* Zaheer Soebhan <z.soebhan@gmail.com>
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -26,6 +26,9 @@ This is pretty cool because it means Hy is several things:
    comfort of Python!
  - For everyone: a pleasant language that has a lot of neat ideas!
 
+Now this tutorial assumes you're running Hy on Python 3. So know things 
+are a bit different if you're still using Python 2.
+
 
 Basic intro to Lisp for Pythonistas
 ===================================
@@ -41,7 +44,7 @@ A "hello world" program in Hy is actually super simple. Let's try it:
 See?  Easy!  As you may have guessed, this is the same as the Python
 version of::
 
-  print "hello world"
+  print("hello world")
 
 To add up some super simple math, we could do:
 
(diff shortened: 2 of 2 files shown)


Kodiologist:
Looking good, thanks. There's still at least one `raw-input` in Hy code hanging around, so don't forget that. Also, you don't need a separate commit for each change when the changes are this small and to the same purpose. You can meld all your commits together with `git rebase -i master`.

freezas:
D'oh! I've replaced the `raw-input` in the Hy examples as well now.

I tried to rebase it, but it looks like the forked branch is still made of multiple commits... Or is this okay now?
I actually wanted to work on a seperate branch and merge it back to the fork's master, but I forgot to do that correctly. Would that be better next time?

Kodiologist:
Ah, yes, I missed the fact that you worked on `master` instead of using a feature branch. You should indeed use a feature branch in the future.

Kodiologist:
> `(1L 2L 3L)`

There shouldn't be trailing `L`s when you're running Hy on Python 3. Are you sure this is correct? It's worth running through the code and making sure it all works and produces the right output, if you haven't already.

Kodiologist:
It would also be wise to say something like "This tutorial assumes you're running Hy under Python 3; things are a bit different under Python 2" near the top.

freezas:
I've removed the trailing L's and added a remark about using Python 3.

I've ran most of the examples. Some of the examples don't work. The things I think need to be fixed are:

* an install of Django and import of 'models' in order to run the examples about class variables
* a 'require' in order to use the 'loop' expression in the example of Protips!

I wasn't sure if you want these in the same pull request. Should I try to fix those examples anyway?

There are also more issues I wanted to discuss first before changing more of the tutorial, but maybe it's better to discuss that on the mailing list. Or is there already a plan to work on improving the tutorial?

Kodiologist:
> I wasn't sure if you want these in the same pull request. Should I try to fix those examples anyway?

It's your choice. When you're ready, add yourself to AUTHORS in a new commit and I'll merge.

> There are also more issues I wanted to discuss first before changing more of the tutorial, but maybe it's better to discuss that on the mailing list.

Whichever you like, although I think issues and PRs are read more widely than the mailing list. I rarely look at the mailing list.

> Or is there already a plan to work on improving the tutorial?

Nope, no plans right now.

freezas:
I've add my name to the AUTHORS file.

I'll take another look at Hy and the tutorial another time again.

Kodiologist:
Sorry, when I said "a new commit", I meant a separate one. See CONTRIBUTING.rst.

Kodiologist:
I'll do it.

Kodiologist:
I couldn't push ("permission denied"; perhaps you unchecked "Allow edits from maintainers"), so I merged it manually.

freezas:
Thanks!

PR #1387, opened 2017-08-19
Update NEWS and docs for `return`

Kodiologist:


docs/language/api.rst
```diff
@@ -1469,6 +1469,38 @@ Given an empty collection, it returns an empty iterable.
     => (list (rest []))
     []
 
+return
+-------
+
+``return`` compiles to a :py:keyword:`return` statement. It exits the
+current function, returning its argument, or ``None`` if called without
```
kirbyfan64: Minor issue, but this would read a *lot* better as:

> returning either its argument or ``None`` if called without

e.g. adding *either* and a comma. Right now it's ambiguous which clause the *or None* is attached to.

Kodiologist: Well spotted. How do you like this wording?

PR #1399, opened 2017-08-26
Compiler refactor

Kodiologist:
More could be done, but this is progress. It saves a few hundred lines from the compiler by removing various kinds of duplication. There should be no changes to Hy semantics, although the exact generated AST is different in some cases.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -91,26 +91,18 @@ def ast_str(foobar):
     return "hy_%s" % (str(foobar).replace("-", "_"))
 
 
-def builds(_type):
-
-    unpythonic_chars = ["-"]
-    really_ok = ["-"]
-    if any(x in unpythonic_chars for x in str_type(_type)):
-        if _type not in really_ok:
-            raise TypeError("Dear Hypster: `build' needs to be *post* "
-                            "translated strings... `%s' sucks." % (_type))
+def builds(*types, **kwargs):
+    # A decorator that adds the decorated method to _compile_table for
+    # compiling `types`, but only if kwargs['iff'] (if provided) is
+    # true.
+    if not kwargs.get('iff', True):
+        return lambda fn: fn
 
     def _dec(fn):
-        _compile_table[_type] = fn
+        for t in types:
+            _compile_table[t] = fn
         return fn
     return _dec
-
-
-def builds_if(_type, condition):
-    if condition:
-        return builds(_type)
-    else:
-        return lambda fn: fn
 
 
 def spoof_positions(obj):
@@ -762,8 +754,7 @@ class HyASTCompiler(object):
         return imports, HyExpression([HySymbol(name),
                                       form]).replace(form), False
 
-    @builds("quote")
-    @builds("quasiquote")
+    @builds("quote", "quasiquote")
     @checkargs(exact=1)
     def compile_quote(self, entries):
         if entries[0] == "quote":
@@ -776,8 +767,7 @@ class HyASTCompiler(object):
         ret.add_imports("hy", imports)
         return ret
 
-    @builds("unquote")
-    @builds("unquote_splicing")
+    @builds("unquote", "unquote_splicing")
     def compile_unquote(self, expr):
         raise HyTypeError(expr,
                           "`%s' can't be used at the top-level" % expr[0])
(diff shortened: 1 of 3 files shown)


Kodiologist:
Huh, it looks like the redundant `arg` argument to some nodes is required by Python 3.3 but not later versions nor Python 2. [Dumb old 3.3 is going to officially go out of its support window in a month anyway.](https://docs.python.org/devguide/#status-of-python-branches) Do you guys think it's okay if we just drop 3.3 now? If I fix this, we'll just have to remove the fix again in October.

gilch:
Make a separate PR to drop 3.3, and we'll discuss it there. I'm okay with it, since it's only a month away. I doubt anyone will care, but they can install Hy from an older commit for a month if they have to. I don't think the date is final, but changes to that seem improbable at this point.

Kodiologist:
> Make a separate PR to drop 3.3, and we'll discuss it there.

Okay, it's #1401.

Kodiologist:
Rebased for the merge of that PR.

hy/compiler.py
```diff
@@ -2521,52 +2156,33 @@ def compile_symbol(self, symbol):
...
-    @builds(HyKeyword)
-    def compile_keyword(self, keyword):
-        return ast.Str(s=str_type(keyword),
-                       lineno=keyword.start_line,
-                       col_offset=keyword.start_column)
+        return asty.Name(symbol, id=ast_str(symbol), ctx=ast.Load())
+
+    @builds(HyString, HyKeyword, HyBytes)
+    def compile_string(self, string, building):
+        node = asty.Bytes if PY3 and building is HyBytes and PY3 else asty.Str
```
gilch: `PY3 and building is HyBytes and PY3` checking it twice?

Kodiologist: Oops, I must've been in a Christmasy mood ("He's makin' a list, he's checkin' it twice…"). Fixed.

tests/native_tests/language.hy
```diff
@@ -1592,3 +1592,14 @@
...
+#@(pytest.mark.xfail
+(defn test-assert-multistatements []
+  ; https://github.com/hylang/hy/issues/1390
+  (setv l (set))
+  (defn f [x]
+    (.add l x)
+    True)
+  (assert (do (f 1) (f 2)) (do (f 3) (f 4)))
+  (print l)
+  (assert (= l [1 2 3 4]))))
```
gilch: This seems unrelated to the rest of the PR.

hy/compiler.py
```diff
@@ -1125,117 +1080,60 @@ def compile_if(self, expression):
...
-                          lineno=expr.start_line,
-                          col_offset=expr.start_column)
-
     @builds("nonlocal")
     @checkargs(min=1)
-    def compile_nonlocal_expression(self, expr):
-        if not PY3:
+    def compile_global_or_nonlocal(self, expr):
+        form = expr.pop(0)
+        if form == "nonlocal" and not PY3:
```
gilch: shouldn't this be in the `@builds_if` decorator?

tests/native_tests/language.hy
Kodiologist: It is. I put that commit in because I noticed the bug while reading through the compiler. I figured that adding an xfail test wasn't worth its own PR.

hy/compiler.py
Kodiologist: Yeah, good idea.

Kodiologist: Okay, changed and rebased.

gilch:
@hylang/core can we get another ack? Any objections?

kirbyfan64:
I'd review, but this is a huge PR, and I have neither the time nor energy right now to go through it, and it'd be silly for me to approve for the sake of approving...

Kodiologist:
@kirbyfan64 The commit "General compiler refactoring pass" has most of the change in terms of sheer line count, but not much interesting is going on there. You could just review this PR on the basis of the other commits, if you like.

gilch:
@Kodiologist, there's a conflict due to #1418. #1419 might cause similar problems.

kirbyfan64:
hold on, manually testing and merging...

PR #1400, opened 2017-08-26
Forbid arguments to `break` and `continue`

Kodiologist:


gilch:
Good catch.

Kodiologist:
Yeah, it's not surprising that nobody seemed to notice it in real Hy code. The missing `@checkargs` in the implementation is more visible.

PR #1404, opened 2017-08-28
Simplify string parsing with `unicode_literals`

Kodiologist:


hy/lex/parser.py
```diff
@@ -265,38 +266,13 @@ def t_empty_list(p):
...
-    header = header.replace("u", "")
-    # remove bytes marker, since we'll need to exclude it for Python 2
-    is_bytestring = "b" in header
-    header = header.replace("b", "")
-    # build python string
-    s = header + '"""' + s + '"""'
-    return (hybytes if is_bytestring else uni_hystring)(s)
+    # Replace the single double quotes with triple double quotes to allow
+    # embedded newlines.
+    s = eval(p[0].value.replace('"', '"""', 1)[:-1] + '"""')
```
kirbyfan64: FWIW this probably would never happen, but I think it'd be safer (and more descriptive) to use `ast.literal_eval` here.

Kodiologist: Did you read the commit message?

kirbyfan64: *ahem*... in my defense I'm on a phone right now...

Kodiologist: :)

gilch:
> it'd be safer (and more descriptive) to use ast.literal_eval here.

@kirbyfan64 I thought about how to exploit `eval` here, but it's already got to be accepted by the parser as a string at this point, so I don't see how. F-strings might be a concern when we implement them later, but we could use a separate `@builds` function for that.

PR #1408, opened 2017-08-30
fix testing for windows

### LINKED ISSUE #1289: Tests are failing on Windows

gilch:
Also improve `test_bin_hy_builtins` and warn that it fails from IPython.

The shell on Windows needs double quotes. This is a partial fix for #1289. It's *still* not runnning the native tests, but I wanted to push this much to make sure switching the quotes doesn't break it for Travis.

[Edit: got the native tests working on Windows too. Fixes #1289.]

### DIFF
--- a/tests/test_bin.py
+++ b/tests/test_bin.py
@@ -5,11 +5,15 @@
 # license. See the LICENSE.
 
 import os
-import subprocess
+from pipes import quote
 import re
-from hy._compat import PY3, PY35
-from hy.importer import get_bytecode_path
+import shlex
+import subprocess
+
 import pytest
+
+from hy._compat import PY3, PY35, builtins
+from hy.importer import get_bytecode_path
 
 
 hy_dir = os.environ.get('HY_DIR', '')
@@ -24,12 +28,14 @@ def run_cmd(cmd, stdin_data=None, expect=0, dontwritebytecode=False):
     if dontwritebytecode:
         env = dict(os.environ)
         env["PYTHONDONTWRITEBYTECODE"] = "1"
-    p = subprocess.Popen(os.path.join(hy_dir, cmd),
+    cmd = shlex.split(cmd)
+    cmd[0] = os.path.join(hy_dir, cmd[0])
+    p = subprocess.Popen(cmd,
                          stdin=subprocess.PIPE,
                          stdout=subprocess.PIPE,
                          stderr=subprocess.PIPE,
                          universal_newlines=True,
-                         shell=True,
+                         shell=False,
                          env=env)
     if stdin_data is not None:
         p.stdin.write(stdin_data)
@@ -223,18 +229,22 @@ def test_hy2py():
                 if f == "py35_only_tests.hy" and not PY35:
                     continue
                 i += 1
-                output, err = run_cmd("hy2py -s -a " +
-                                      os.path.join(dirpath, f))
+                output, err = run_cmd("hy2py -s -a " + quote(os.path.join(dirpath, f)))
                 assert len(output) > 1, f
                 assert len(err) == 0, f
     assert i
 
 
 def test_bin_hy_builtins():
+    # hy.cmdline replaces builtins.exit and builtins.quit
+    # for use by hy's repl.
     import hy.cmdline  # NOQA
-
-    assert str(exit) == "Use (exit) or Ctrl-D (i.e. EOF) to exit"
-    assert str(quit) == "Use (quit) or Ctrl-D (i.e. EOF) to exit"
+    # this test will fail if run from IPython because IPython deletes
+    # builtins.exit and builtins.quit
+    assert str(builtins.exit) == "Use (exit) or Ctrl-D (i.e. EOF) to exit"
+    assert type(builtins.exit) is hy.cmdline.HyQuitter
+    assert str(builtins.quit) == "Use (quit) or Ctrl-D (i.e. EOF) to exit"
+    assert type(builtins.quit) is hy.cmdline.HyQuitter
 
 
 def test_bin_hy_main():
--- a/conftest.py
+++ b/conftest.py
@@ -1,10 +1,13 @@
 import _pytest
 import hy
+import os
 from hy._compat import PY3, PY35
+
+NATIVE_TESTS = os.path.join("", "tests", "native_tests", "")
 
 def pytest_collect_file(parent, path):
     if (path.ext == ".hy"
-            and "/tests/native_tests/" in path.dirname + "/"
+            and NATIVE_TESTS in path.dirname + os.sep
             and path.basename != "__init__.hy"
             and not ("py3_only" in path.basename and not PY3)
             and not ("py35_only" in path.basename and not PY35)):
--- a/hy/importer.py
+++ b/hy/importer.py
@@ -98,7 +98,7 @@ def import_file_to_module(module_name, fpath, loader=None):
         try:
             _ast = import_file_to_ast(fpath, module_name)
             module = imp.new_module(module_name)
-            module.__file__ = fpath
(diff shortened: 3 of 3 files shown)


Kodiologist:
I think we might be best off with the `shell` option of `Popen` off. Then we don't have to worry about this stuff.

gilch:
Like that?

gilch:
Hmm, nope. Travis did not like that at all.

Kodiologist:
With `shell` off, you need to pass `[Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen)` a list of the command and its arguments instead of a shell command as a single string.

gilch:
Strange, this version seems to work on Windows. It looks like the string interpretation is OS dependent, even with shell off.

I added a `shlex.split` to `run_cmd` so that any arguments in the cmd string in POSIX format will work as before. I also added an `args` kwarg, that takes a sequence to extend the list of args with, in case you have to specify them in non-POSIX format for some reason. I had to update one of the tests to do it this way, because `shlex.split` doesn't like Window's path format, but that's how `os.walk` provides it on Windows.

Kodiologist:
Confusingly, it looks like even when you turn `shell` off, a single string, rather than a list, will be interpreted in a platform-dependent manner. By contrast, if you pass a list, the library will [automatically](https://docs.python.org/3/library/subprocess.html#notes) convert it back into a string in the format Windows expects. It seems best to provide a list in any case, although we can still give `run_cmd` some syntactic sugar that the individual tests can use, of course.

I know this would be a bigger change than you were planning when you wrote this PR, so I'll do it if you want.

gilch:
> It seems best to provide a list in any case,

I'd already pushed an update to that last commit. I *am* passing a list to `Popen`.

> [...] although we can still give run_cmd some syntactic sugar that the individual tests can use, of course.

But I'm generating that list from a string passed to `run_cmd`, with `shlex.split`, which operates in a platform-independent way (i.e. POSIX sh). That's the sugar.

I could change line 231 from
```Python
output, err = run_cmd("hy2py -s -a", args=[os.path.join(dirpath, f)])
```
to
```Python
output, err = run_cmd("hy2py", args=["-s", "-a", os.path.join(dirpath, f)])
```
which puts *all* the args in a list, if you think that's clearer.

I'm not sure what else you want here. It seems to be doing everything you asked for.

Kodiologist:
So it does! I didn't read your latest changes closely. Sorry.

Is the change of quotation marks still necessary?

gilch:
> Is the change of quotation marks still necessary?

I doubt it actually. I could change it back and see.

gilch:
Not necessary. `tests\test_bin.py` passed on Windows when I changed it back.

I could squash the commits and omit the change to the quotes, if you want.

Kodiologist:
Sounds like a plan!

gilch:
I got pytest running the native tests on Windows.

Now this PR fixes #1289

gilch:
Can someone else look at this? It's making development on Windows difficult.

Kodiologist:
> development on Windows

There's your problem.

But seriously, Hy is mostly frozen whenever two Horsemen are inactive, and that seems to be the case right now.

tests/test_bin.py
```diff
@@ -223,18 +228,22 @@ def test_hy2py():
                 if f == "py35_only_tests.hy" and not PY35:
                     continue
                 i += 1
-                output, err = run_cmd("hy2py -s -a " +
-                                      os.path.join(dirpath, f))
+                output, err = run_cmd("hy2py -s -a", args=[os.path.join(dirpath, f)])
```
kirbyfan64: Sorry for the delays, but...adding an extra argument `args=` that's really...odd (you can pass arguments through both `cmd` *and* `args`) for just *one* test case feels really wrong. A simpler change would be:

```python
output, err = run_cmd("hy2py -s -a " + shlex.quote(os.path.join(dirpath, f)))
```

(Of course, using `pipes.quote` instead on Python 2.) Then you can drop the whole `args` mess.

gilch: I thought it might get used by future tests though.

gilch:
@kirbyfan64, how's that?

Issue #1405, opened 2017-08-28
Implement `else` clause for `while` loops


Kodiologist:
For some reason, we have `else` for `for`, but not `while`.

gilch:
Is there some reason the `else` in `for*` doesn't have an implicit `do`? It does in `try`.

rkday:
I've created #1419 to cover that (as a separate PR, in case there *is* some reason).

Kodiologist:
> Is there some reason the `else` in `for*` doesn't have an implicit `do`?

Nope, sounds like a bug to me.

PR #1418, opened 2017-09-13
Allow else after a while loop

rkday:
Closes https://github.com/hylang/hy/issues/1405

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2146,12 +2146,20 @@ class HyASTCompiler(object):
         expr.pop(0)  # "while"
         ret = self.compile(expr.pop(0))
 
+        orel = Result()
+        # (while cond body (else …))
+        if expr and expr[-1][0] == HySymbol("else"):
+            else_expr = expr.pop()
+            for else_body in else_expr[1:]:
+                orel += self.compile(else_body)
+                orel += orel.expr_as_stmt()
+
         body = self._compile_branch(expr)
         body += body.expr_as_stmt()
 
         ret += ast.While(test=ret.force_expr,
                          body=body.stmts,
-                         orelse=[],
+                         orelse=orel.stmts,
                          lineno=expr.start_line,
                          col_offset=expr.start_column)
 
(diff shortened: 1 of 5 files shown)


Kodiologist:
You can leave the matter of making the `else` of `for*` accept multiple statements to #1419, but for this PR, we might as well accept multiple statements from the beginning.

Otherwise, this looks great. Well done!

tests/native_tests/language.hy
```diff
@@ -266,6 +266,25 @@
   (assert (= count 0))
   (assert (= fact 120)))
 
+(defn test-while-loop-else []
+  "NATIVE: test while loops?"
```
Kodiologist: Why does this docstring end with a question mark? (To be fair, the docstrings of tests no longer have much effect since we switched from Nose to pytest. You can just not use one at all if you want.)

hy/compiler.py
```diff
@@ -2146,12 +2146,25 @@ def compile_while_expression(self, expr):
...
+                    else_expr,
+                    "`else' statement in `while' is too long")
+            elif len(else_expr) == 2:
+                orel += self.compile(else_expr[1])
+                orel += orel.expr_as_stmt()
+
         body = self._compile_branch(expr)
         body += body.expr_as_stmt()
 
+
```
Kodiologist: It looks like there's an accidental whitespace change here.

tests/native_tests/language.hy
rkday: Only because https://github.com/hylang/hy/blob/51fb807cc9eb5b005b79b4dbe819d8e11889de78/tests/native_tests/language.hy#L260 does - I'll remove.

rkday:
Thanks! I've made those changes.

Kodiologist:
Looks good. Just meld your changes back into the first commit. (I think this can be done directly with `rebase -i`.)

rkday:
Like so?

Kodiologist:
Yep. Thanks a bunch, and welcome aboard the Hy-pe train.

Issue #1409, opened 2017-08-31
Destructuring bind form in defn argslist breaks docstring


gilch:
```Python
=> (defn foo [a b]
... "doc?"
... "foo.")
def foo(a, b):
    'doc?'
    return 'foo.'
None
=> (help foo)
help(foo)
Help on function foo:

foo(a, b)
    doc?

=> (foo 1 2)
foo(1, 2)
'foo.'
=> (defn bar [[a b]]
... "doc?"
... "nope.")
def bar(_hy_anon_var_1):
    [a, b] = _hy_anon_var_1
    'doc?'
    return 'nope.'
None
=> (help bar)
help(bar)
Help on function bar:

bar(_hy_anon_var_1)

=> (bar [1 2])
bar([1, 2])
'nope.'
```

Kodiologist:
It makes sense that this happens: we need to make sure the generated destructuring code is placed after the docstring when there is one.

gilch:
We also need to make sure the first body element is not detected as the docstring if it is also the last. #1402.

PR #1421, opened 2017-09-16
Let argument destructuring work with docstrings

rkday:
Fixes https://github.com/hylang/hy/issues/1409.


Kodiologist:
Travis suggests it works on Python 3, but not 2.

rkday:
Yep - I had 'str' in a check but 2.7 wanted 'unicode'. Should be fixed now by using hy._compat.

rkday:
Done. (I didn't know about `none?`, and I agree it's nicer, so I've used that - thanks.)

PR #1427, opened 2017-09-17
Allow multiple expressions in a try

### LINKED ISSUE #1425: try form should have implicit do
It's strange that its subforms, `except/finally/else` have an implicit do, but the `try` form itself does not. After popping off any of the expected subforms at the end, the rest of the forms of the `try` body could be put in a `do` block, couldn't they?

rkday:
Fixes #1425.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -843,8 +843,14 @@ class HyASTCompiler(object):
         expr = copy.deepcopy(expr)
         expr.pop(0)  # try
 
-        # (try something…)
-        body = self.compile(expr.pop(0) if expr else [])
+        # (try something somethingelse…)
+        body = Result()
+        # Check against HyExpression and HySymbol to avoid incorrectly
+        # matching [except ...] or ("except" ...)
+        while expr and not (isinstance(expr[0], HyExpression)
+                            and isinstance(expr[0][0], HySymbol)
+                            and expr[0][0] in ("except", "else", "finally")):
+            body += self.compile(expr.pop(0))
 
         var = self.get_anon_var()
         name = asty.Name(expr, id=ast_str(var), ctx=ast.Store())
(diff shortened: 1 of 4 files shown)


hy/compiler.py
```diff
@@ -833,8 +833,11 @@ def compile_raise_expression(self, expr):
     def compile_try_expression(self, expr):
         expr.pop(0)  # try
 
-        # (try something…)
-        body = self.compile(expr.pop(0) if expr else [])
+        # (try something somethingelse…)
+        body = Result()
+        while expr and not (isinstance(expr[0], HyExpression) and expr[0][0] in ("except", "else", "finally")):
+            print(type(expr[0]))
```
kirbyfan64: Leftover debugging statement?

Kodiologist:
@rkday Did you see @kirbyfan64's comment above?

rkday:
Whoops, fixed.

Kodiologist:
Extending `try` to take multiple body forms means that we now need to be a bit more careful about identifying the `except`, `else`, and `finally` clauses. Consider

     => (try 1 ("except" [ValueError] 2)) 
     1

This should try to call the string `"except"` as a function, but instead the string literal is recognized as the start of an `except` form. You may wish to use some of my code from #1432 (or just use it as a reference).

Kodiologist:
By the way, a rebase doesn't trigger notifications to PR participants, so you might want to add a comment that's just "Done" or something in that case.

Kodiologist:
@kirbyfan64 Look good?

tuturto:
This looks good, but I would like to see docs to be updated with the same PR. It doesn't have to be anything fancy, just clarify how body of ``try`` is constructed.

Kodiologist:
@rkday Did you see @tuturto's comment above?

rkday:
Sorry for the delay in getting back round to this - I've now made that docs change and rebased. (I'm happy to squash the docs commit into the original one if that's better.)

Kodiologist:
Looks good to me. @tuturto?

PR #1433, opened 2017-09-21
Handle statements in the condition of `while`

### LINKED ISSUE #1362: A statement pulled out of the condition of a `while` loop is only evaluated once
Thinking about #1342 made me guess that the following would do the wrong thing, and indeed, it does:

    $ echo '(setv x 3) (while (do (print "hi") x) (print x) (-= x 1))' | hy2py
    x = 3
    print('hi')
    while x:
        print(x)
        x -= 1

The `print('hi')` only happens once instead of every time the `while` condition is checked, because it's pulled out to before the `while`.

The most straightforward fix is to make a `while` loop with a complex condition compile to something like:

    x = 3
    while True:
        # Condition
        print('hi')
        _hy_anon_var_1 = x
        if not _hy_anon_var_1:
            break
        # Body
        print(x)
        x -= 1

(In practice, `Result.rename` would remove the use of `_hy_anon_var_1` and just write `if not x:`.)

The problem with this method is that an `else` clause won't be executed when the loop terminates normally, because `break` skips the `else`. This doesn't actually matter right now because we haven't yet implemented `else` for `while`. But a method that's robust to this future added feature is:

    x = 3
    # Condition
    print('hi')
    _hy_anon_var_1 = x
    while _hy_anon_var_1:
        # Body
        print(x)
        x -= 1
        # Condition, repeated
        print('hi')
        _hy_anon_var_1 = x

Note that it will then be unwise to use `break` or `continue` inside the condition, because the condition executes once outside the `while` and subsequently inside it.

Kodiologist:
Fixes #1362.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1861,12 +1861,33 @@ class HyASTCompiler(object):
     @checkargs(min=2)
     def compile_while_expression(self, expr):
         expr.pop(0)  # "while"
-        ret = self.compile(expr.pop(0))
+        cond = expr.pop(0)
+        cond_compiled = self.compile(cond)
 
-        orel = Result()
-        # (while cond body (else …))
+        else_expr = None
         if ends_with_else(expr):
             else_expr = expr.pop()
+
+        if cond_compiled.stmts:
+            # We need to ensure the statements for the condition are
+            # executed on every iteration. Rewrite the loop to use a
+            # single anonymous variable as the condition.
+            def e(*x): return HyExpression(x)
+            s = HySymbol
+            cond_var = s(self.get_anon_var())
+            return self.compile(e(
+                s('do'),
+                e(s('setv'), cond_var, 1),
+                e(s('while'), cond_var,
+                  # Cast the condition to a bool in case it's mutable and
+                  # changes its truth value, but use (not (not ...)) instead of
+                  # `bool` in case `bool` has been redefined.
+                  e(s('setv'), cond_var, e(s('not'), e(s('not'), cond))),
+                  e(s('if*'), cond_var, e(s('do'), *expr)),
+                  *([else_expr] if else_expr is not None else []))).replace(expr))  # noqa
+
+        orel = Result()
+        if else_expr is not None:
             for else_body in else_expr[1:]:
                 orel += self.compile(else_body)
                 orel += orel.expr_as_stmt()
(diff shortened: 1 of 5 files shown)


hy/compiler.py
```diff
@@ -1850,24 +1850,41 @@ def compile_for_expression(self, expression):
...
+            # executed on every iteration. Rewrite the loop to use a
+            # single anonymous variable as the condition.
+            def e(*x): return HyExpression(x)
+            s = HySymbol
+            cond_var = s(self.get_anon_var())
+            return self.compile(e(
+                s('do'),
+                e(s('setv'), cond_var, s('True')),
+                e(s('while'), cond_var,
+                  e(s('setv'), cond_var, e(s('not'), e(s('not'), cond))),
```
kirbyfan64: IMO for simplicity purposes (also because `(not (not ...))` is pointless), this should be compiled to something like this instead:

```hy
(while True
  (setv anon-var expr-here)
  (if-not anon-var (break)))
```

Right now the condition is being tested twice, and again the double nots aren't necessary.

Kodiologist: @kirbyfan64 The problem with that design is that `break` skips its loop's `else` clause. As for the not-not, that's for the case in which `cond` produces a mutable value and its truth value changes. See the tests in this PR and the discussion in #1432.

kirbyfan64: @Kodiologist touche. Couldn't you still do `anon-var.__bool__()` though?

Kodiologist: @kirbyfan64 That doesn't seem to exist in Python 2.

kirbyfan64: It's `__nonzero__` on Python 2.

gilch:
@kirbyfan64, the `not not` seems more reliable than `bool` or `.__bool__` to me.

But as discussed in #1362, it can also be avoided in this case with an expansion like
```Python
x = 3
_hy_anon_var_1 = True
while _hy_anon_var_1:
    # Condition
    print('hi')
    _hy_anon_var_1 = x
    if _hy_anon_var_1:
        _hy_anon_var_1 = True
        # Body
        print(x)
        x -= 1
```
if you prefer.

hy/compiler.py
Kodiologist: @kirbyfan64 Not every class defines `__nonzero__`.

kirbyfan64:
I mean, I guess it's not *that* bad... the internal micro-optimizer in me is just yelling *extra bytecode, extra bytecode!*...

hy/compiler.py
kirbyfan64: FWIW I'd at least prefer a comment explaining the `not not `.

Kodiologist: @kirbyfan64 Done.

PR #1436, opened 2017-09-25
Delimit spy

### LINKED ISSUE #1435: Delimit python "--spy" output.
I've implemented highlighting python code outputted from `--spy` in `inferior-hy-mode`.

However I need to be able to distinguish where the python block ends.

I can't take the last line to be the limit as:

```
=> (defn x [])
def x():
    pass
;; nothing

=> (+ 1 2)
(1 + 2)
3
;; standard

=> x
x
  File "my-file-path", line 198, in hy_eval
NameError: name 'x' is not defined
;; many extra lines

=> (})
  File "<stdin>", line 1, column 2

  (})
   ^
LexException: Ran into a RCURLY where it wasn't expected.
;; no --spy output at all
```

A delimiter would do it unless you all know a better solution.

gilch:
Adds the `--control-codes` option for use with external tooling. For the moment, this means `hy-mode` in Emacs.

Fixes #1435 

For now, all this does is add a delimiter after `--spy` output (using string constant pre-generated from a UUID4). And it turns off `colorama`, so hopefully the ANSI color tags will show up in stdout/stderr even on Windows.

### DIFF
--- a/hy/__init__.py
+++ b/hy/__init__.py
@@ -1,3 +1,12 @@
+# colorama's deinit isn't working, probably due to
+# https://github.com/tartley/colorama/issues/145
+# This is located here to save the original streams before clint gets to them.
+# TODO: remove when colorama fixes it.
+from sys import stdout as _out, stderr as _err
+def colorama_deinit():
+    import sys
+    sys.stdout, sys.stderr = _out, _err
+
 __appname__ = 'hy'
 try:
     from hy.version import __version__
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -54,9 +54,11 @@ builtins.exit = HyQuitter('exit')
 
 class HyREPL(code.InteractiveConsole):
     def __init__(self, spy=False, output_fn=None, locals=None,
-                 filename="<input>"):
+                 filename="<input>", control_codes=False):
 
         self.spy = spy
+        self.control_codes = control_codes
+        self.control_codes and hy.colorama_deinit()
 
         if output_fn is None:
             self.output_fn = repr
@@ -95,6 +97,7 @@ class HyREPL(code.InteractiveConsole):
                     new_ast = ast.Module(main_ast.body +
                                          [ast.Expr(expr_ast.body)])
                     print(astor.to_source(new_ast))
+                    self.control_codes and print("#cbb4fcbe-b6ba-4812-afa3-4a5ac7b20501")
             value = hy_eval(do, self.locals, "__console__",
                             ast_callback)
         except HyTypeError as e:
@@ -275,6 +278,8 @@ def cmdline_handler(scriptname, argv):
         help="program passed in as a string, then stay in REPL")
     parser.add_argument("--spy", action="store_true",
                         help="print equivalent Python code before executing")
+    parser.add_argument("--control-codes", action="store_true",
+                        help="enable extra control codes in the repl, for external tooling")
     parser.add_argument("--repl-output-fn",
                         help="function for printing REPL output "
                              "(e.g., hy.contrib.hy-repr.hy-repr)")
@@ -321,7 +326,8 @@ def cmdline_handler(scriptname, argv):
     if options.icommand:
         # User did "hy -i ..."
         return run_icommand(options.icommand, spy=options.spy,
-                            output_fn=options.repl_output_fn)
+                            output_fn=options.repl_output_fn,
+                            control_codes=options.control_codes)
 
     if options.args:
         if options.args[0] == "-":
@@ -338,7 +344,8 @@ def cmdline_handler(scriptname, argv):
                 sys.exit(e.errno)
 
     # User did NOTHING!
-    return run_repl(spy=options.spy, output_fn=options.repl_output_fn)
+    return run_repl(spy=options.spy, output_fn=options.repl_output_fn,
+                    control_codes=options.control_codes)
 
 
 # entry point for cmd line script "hy"


gilch:
Wait, that should have been based on master.

gilch:
@ekaschalk try this branch. Does this work for #1436? Do the color codes come through on Windows? hylang/hy-mode#46

I'll add some tests and docs once we settle what this should be doing.

ekaschalk:
This works for delimiting `--spy`. I have it setup so you can replace the uuid with any string of your choosing, or replace with nothing. It doesn't solve color codes for erros on Windows however.

ekaschalk:
See https://github.com/hylang/hy-mode/commit/64397082d5ced6b1dc01129601ef84b4397c8b6c

Set `hy-shell-use-control-codes?` true to have font-locked python blocks.

ekaschalk:
If you want stderr vs stdout italicized, then implementing that as another UUID delimiter would make it trivial, otherwise I will table that addition for now.

gilch:
I re-implemented colorama's `deinit` and amended the commit. See tartley/colorama#145.

@ekaschalk, does this version show the color codes on Windows?

gilch:
That's weird. The tests do not fail on my machine. I can't imagine how those changes could cause those errors, as none of the old tests use the new features. I wonder if we've corrupted the cache somehow.

gilch:
I tried it on a Linux VM, and it failed due to `import colorama`. I guess that only installs on Windows?

gilch:
> If you want stderr vs stdout italicized, then implementing that as another UUID delimiter would make it trivial, otherwise I will table that addition for now.

That probably won't be possible. The user is free to print to stderr. It's not just for exceptions we could catch in the repl. Maybe we could monkeypatch `sys.stderr` with a custom class, but that would interfere with colorama (and who knows what else). And threaded code could even cause stderr to be interleaved with stdout, in ways not predictable in advance.

If emacs can't distinguish stderr, could it read from some other file instead? I think we could probably open a pipe in a cross-platform way on the Python side, even if emacs can't do it. If we can make that work, then we don't need the `--spy` delimiter either. It could just be another pipe.

We could also try a network loopback instead of file pipes. I think that's how nrepl does it.

ekaschalk:
It's not that it's impossible to use pipes here (I think), it's that I can't leverage current work. `comint-mode` provides some nice utilities, in particular hooks for modifying the process output. There are a number of other low hanging, high value fruit (like full auto completion and linting integrated) that I'd like to attack before going into any particular feature in too much depth.

kirbyfan64:
What about following the path of the Dart analysis server? It reads a one-line JSON response from stdin and writes an output JSON response to stdout. It's significantly easier to implement and work with than sockets, but it gets the job done.

ekaschalk:
I'm considering representing shell output similar to ipython, namely the prompt input has `In [x]:`, stdout is printed, and then the return is `Out [x]:`.

In this case, we could have an additional response, say `Spy [x]:`, with the python representation.

What if the REPL output was of this form? The UUID delimiter would then be unneeded.

gilch:
The UUID delimiter feels like a dirty hack. I'm not particularly attached to it, but I don't really have a better idea that doesn't use either pipes or network sockets. But I think `Spy[x]:` has the same issue. What if it appears in the output? Maybe you put it last so the `Spy[x]:` part follows the `Out[x]:` part and you only count the last one.

But say you put a repl session in a docstring for a function, including `Spy[x]:`. Not an unreasonable thing to do. Then you want to test it out, so you paste it into the repl. Now only the part following the last `Spy[x]:` in the docstring gets highlighting. And astor outputs the docstring with triple quotes, so the rest of your function looks like a string literal. It's not that easy.

ekaschalk:
I've been trying to determine how ipython handles it in Emacs.

Prompts are detected by `comint-prompt-regexp`, namely a regex, on which a match occurs, `comint` performs bookkeeping and start/end markers and etc... 

In `print("In [38]: print(\"hi\")"); 1+2`, somehow it knows that the printed line is not a prompt.

I'll keep investigating. I'd ideally like to use pipes/sockets, but that means reimplementing `comint-mode` (for example `cider-mode` doesn't use `comint`).

ekaschalk:
Actually outputting Spy [] right after the prompt will work. The first spy unambiguously identifies the beginning. Since it is python source, a newline Out [] block can only occur there if captured in a string, which I can check, so we can always identify the end of the spy block.

gilch:
```Python
=> (get Out "of town")
Out['of town']
```
If the `Out[x]:` has that trailing colon, and you check for strings, I think it works.

ekaschalk:
I meant for the `x` within `In/Spy/Out[x]: stuff` to be the prompt number like ipython. We could use the old prompt right arrow or the `In`, but the `Spy/Out` would be required.

ekaschalk:
I misunderstood your example, you're right the colon should do it for source blocks.

Kodiologist:
What's the status of this?

ekaschalk:
I haven't, and I assume no one else, has worked on it yet.

Doing the `In/Spy/Out[x]` in the shell would be nice on its own, but additionally allow for significant improvements emacs-side, including:

- Syntax highlighting Spy output.
- Concatenating spy (also in/out) output for things like sending files/regions/more than one form at a time.

I'm somewhat stuck, unable to merge the dev branch https://github.com/hylang/hy-mode/tree/new-shell-2 containing nice stuff like `send-defun-to-shell` until this is addressed, either by someone else or me.

I consider this tied to other (imo also important) shell usability improvements like formatting+coloring errors (I believe lexer errors only ones covered currently?), filepaths, and other traceback components.

They could be tied together in a single issue outlining shell improvements.

The code in this PR isn't relevant anymore and could be considered for discarding.

Kodiologist:
Okay, I'll close the PR, but open a new issue for those things if you want to and we don't already have one.

gilch:
Agreed, this PR is obsolete. Move discussion back to #1435 or new issues as applicable.

PR #1437, opened 2017-10-04
Implement `compl`, `compd`, `comps`, `compg`

### LINKED ISSUE #588: list-comp only iterates over the last expression when it is passed a do

```
print(n)
[print(n) for n in [1L, 2L]]
```

```
def _hy_anon_fn_1(a):
    print(a)
    return print(a)
1
1
2
2
[None, None]
```

### LINKED ISSUE #1371: Add :while :if and :let modifiers to genexpr, comprehensions, and for
Since we need to compile comprehensions containing statements to generator functions anyway #588, this gives us the opportunity to implement Clojure's optional `:let` and `:while` loop modifiers used in its [list comprehension](https://clojuredocs.org/clojure.core/for). 

Clojure also has a `:when` modifier for a filter predicate, which is equivalent to the `if` keyword in Python's comprehensions. So I propose we just call it `:if` in Hy. Hy currently can't use more than one filter predicate in comprehensions like Python can #848. This is a glaring deficiency for nested loops.

The modifiers follow the loop setup in the comprehension: 
```Hy
(genexpr <expression>
         [<binding1> <sequence1> <modifiers for loop 1>
          <binding2> <sequence2> <modifiers for loop 2>
           ...
          <bindingN> <sequenceN> <modifiers for loop N>])
```

```Python
def _hy_anon_fn_1():
    # pull out <sequence1> statements here
    _hy_anon_var_1 = <sequence1>
    for <binding1> in _hy_anon_var_1:
        <modifiers for loop 1>
        # pull out <sequence2> statements here
        _hy_anon_var_2 = <sequence2>
        for <binding2> in _hy_anon_var_2:
            <modifiers for loop 2>
             ...
            # pull out <sequenceN> statements here
            _hy_anon_var_N = <sequenceN>
            for <bindingN> in _hy_anon_var_N:
                <modifiers for loop N>
                # pull out <expression> statements here
                yield <expression>
_hy_anon_fn_1()
```
In that structure, an `:if <condition>` loop modifier compiles to
```Python
# pull out condition statements here
if not <condition>:
    continue
```
And a `:while <condition>` loop modifier compiles to
```Python
# pull out condition statements here
if not <condition>:
    break
```
And a `:let [<bindings>]` loop modifier compiles to `=` assignments for each binding.

If there aren't any statements, then the `:if` modifers will go directly in the comprehension, but the others modifiers will cause it to compile as a generator function.

We might also want to move the `<expr>` after the bindings list, like `(genexpr [<bindings>] <body>)`, which is the ordering Clojure uses. This way we can have an implicit `do` in the form.

Kodiologist:
- Closes #588.

This implements [my proposal here](https://github.com/hylang/hy/issues/1371#issuecomment-332986111). I wanted to get feedback before continuing.

To do:
- [ ] flake
- [ ] Test that variables are scoped properly
- [ ] Replace uses of the old forms with the new forms
- [ ] Get rid of the old forms
- [ ] Documentation
- [ ] NEWS

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -192,6 +192,22 @@ class Result(object):
         self.__used_expr = False
         self._expr = value
 
+    @property
+    def lineno(self):
+        if self._expr is not None:
+            return self._expr.lineno
+        if self.stmts:
+            return self.stmts[-1].lineno
+        return None
+
+    @property
+    def col_offset(self):
+        if self._expr is not None:
+            return self._expr.col_offset
+        if self.stmts:
+            return self.stmts[-1].col_offset
+        return None
+
     def add_imports(self, mod, imports):
         """Autoimport `imports` from `mod`"""
         self.imports[mod].update(imports)
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -123,13 +123,13 @@ def spoof_positions(obj):
 # ast.Foo(..., lineno=x.lineno, col_offset=x.col_offset)
 class Asty(object):
     def __getattr__(self, name):
-        setattr(Asty, name, lambda self, x, **kwargs: getattr(ast, name)(
+        setattr(Asty, name, staticmethod(lambda x, **kwargs: getattr(ast, name)(
             lineno=getattr(
                 x, 'start_line', getattr(x, 'lineno', None)),
             col_offset=getattr(
                 x, 'start_column', getattr(x, 'col_offset', None)),
-            **kwargs))
-        return getattr(self, name)
+            **kwargs)))
+        return getattr(Asty, name)
 asty = Asty()
 
 
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -385,19 +385,25 @@ def checkargs(exact=None, min=None, max=None, even=None, multiple=None):
     return _dec
 
 
+def eq_sym(x, sym):
+    return isinstance(x, HySymbol) and x == HySymbol(sym)
+
+
+def eq_kw(x, kw):
+    return isinstance(x, HyKeyword) and x == HyKeyword(kw)
+
+
 def is_unpack(kind, x):
     return (isinstance(x, HyExpression)
             and len(x) > 0
-            and isinstance(x[0], HySymbol)
-            and x[0] == "unpack_" + kind)
+            and eq_sym(x[0], "unpack_" + kind))
 
 
 def ends_with_else(expr):
     return (expr and
             isinstance(expr[-1], HyExpression) and
             expr[-1] and
-            isinstance(expr[-1][0], HySymbol) and
-            expr[-1][0] == HySymbol("else"))
+            eq_sym(expr[-1][0], "else"))
 
 
 class HyASTCompiler(object):
@@ -1427,6 +1433,114 @@ class HyASTCompiler(object):
             key=key.force_expr,
             value=value.force_expr,
             generators=gen)
+
+    @builds("compl", "compd", "comps", "compg")
+    @checkargs(min=2)
+    def compile_new_comp(self, expr):
+        node_class = dict(
+            compl=asty.ListComp,
+            compd=asty.DictComp,
+            comps=asty.SetComp,
+            compg=asty.GeneratorExp)[expr.pop(0)]
+
+        # Get the final value (and for dictionary compositions, the final key).
+        elt = self.compile(expr.pop())
+        key = None
+        if node_class is asty.DictComp:
+            key = self.compile(expr.pop())
+
+        # Parse the remaining arguments into a list of parts. Each part is a
+        # :for clause, an :if clause, or an expression.
+        parts = []
+        while expr:
+            part = expr.pop(0)
+            if eq_kw(part, ":for"):
+                if len(expr) < 2:
+                    raise HyTypeError(
+                        part, "Comprehension :for needs 2 arguments")
+                var = expr.pop(0)
+                parts.append(['for',
+                              self._storeize(var, self.compile(var)),
+                              self.compile(expr.pop(0))])
+            elif eq_kw(part, ":if"):
+                if not expr:
+                    raise HyTypeError(
+                        part, "Comprehension :if needs an argument")
+                parts.append(['if', self.compile(expr.pop(0))])
+            else:
+                parts.append(['expr', self.compile(part)])
+
+        # Produce a result.
+        if (elt.stmts or (key is not None and key.stmts) or
+            any(p[0] == 'expr' or (p[2].stmts if p[0] == "for" else p[1].stmts)
+                for p in parts)):
+            # The desired comprehension can't be expressed as a
+            # real Python comprehension. We'll write it as a nested
+            # loop in a function instead.
+            def f(parts):
+                # This function is called recursively to construct
+                # the nested loop.
+                if not parts:
+                    if node_class is asty.DictComp:
+                        ret = key + elt
+                        val = asty.Tuple(
+                            key, ctx=ast.Load(),
+                            elts=[key.force_expr, elt.force_expr])
+                    else:
+                        ret = elt
+                        val = elt.force_expr
+                    return ret + asty.Expr(
+                        elt, value=asty.Yield(elt, value=val))
+                p, parts = parts[0], parts[1:]
+                if p[0] == "for":
+                    return p[2] + asty.For(
+                        p[2], target=p[1], iter=p[2].force_expr, body=f(parts).stmts,
+                        orelse=[])
+                elif p[0] == "if":
+                    return p[1] + asty.If(
+                        p[1], test=p[1].force_expr, body=f(parts).stmts, orelse=[])
+                elif p[0] == "expr":
+                    return p[1] + p[1].expr_as_stmt() + f(parts)
+                else:
+                    raise ValueError("can't happen")
+            fname = self.get_anon_var()
+            ret = Result() + asty.FunctionDef(
+                expr,
+                name=fname,
+                args=ast.arguments(
+                    args=[], vararg=None, kwarg=None,
+                    kwonlyargs=[], kw_defaults=[], defaults=[]),
+                body=f(parts).stmts,
+                decorator_list=[])
+            generator_call = asty.Call(
+                expr,
+                func=asty.Name(expr, id=ast_str(fname), ctx=ast.Load()),
+                args=[], keywords=[], starargs=None, kwargs=None)
+            if node_class is asty.GeneratorExp:
+                return ret + generator_call
+            output_type = {
+                asty.ListComp: "list",
+                asty.DictComp: "dict",
+                asty.SetComp: "set"}[node_class]
+            return ret + asty.Call(
+                expr,
+                func=asty.Name(expr, id=ast_str(output_type), ctx=ast.Load()),
+                args=[generator_call],
+                keywords=[], starargs=None, kwargs=None)
+        else:
+            # Produce a comprehension.
+            generators = []
+            for p in parts:
+                if p[0] == "for":
+                    generators.append(ast.comprehension(
+                        target=p[1], iter=p[2].expr, ifs=[], is_async=0))
+                elif p[0] == "if":
+                    generators[-1].ifs.append(p[1].expr)
+                else:
+                    raise ValueError("can't happen")
+            if node_class is asty.DictComp:
+                return asty.DictComp(expr, key=key.expr, value=elt.expr, generators=generators)
+            return node_class(expr, elt=elt.expr, generators=generators)
 
     @builds("not", "~")
     @checkargs(1)
--- a/tests/native_tests/test_newcomp.hy
+++ b/tests/native_tests/test_newcomp.hy
@@ -1,0 +1,139 @@
+(import
+  types
+  pytest)
+
+(defn test-comprehension-types []
+
+  ; Simple comprehensions
+  (assert (is (type (compl :for x "abc" x)) list))
+  (assert (is (type (comps :for x "abc" x)) set))
+  (assert (is (type (compd :for x "abc" x x)) dict))
+  (assert (is (type (compg :for x "abc" x)) types.GeneratorType))
+
+  ; Statement comprehensions
+  (assert (is (type (compl :for x "abc" (setv y 1) x)) list))
+  (assert (is (type (comps :for x "abc" (setv y 1) x)) set))
+  (assert (is (type (compd :for x "abc" (setv y 1) x x)) dict))
+  (assert (is (type (compg :for x "abc" (setv y 1) x)) types.GeneratorType)))
+
+#@ ((pytest.mark.parametrize "form" ["compl" "comps" "compg"])
+(defn test-simple-comprehensions [form]
+  (defn f [a b]
+    (assert (= (sorted a) b)))
+  (setv form (HySymbol form))
+  (eval `(do
+    (f (~form :for x [] x)
+      [])
+    (f (~form :for j [1 2 3] j)
+      [1 2 3])
+    (f (~form :for x (range 3) (* x 2))
+      [0 2 4])
+    (f (~form :for x (range 4) :if (% x 2) (* x 2))
+      [2 6])
+    (f
+      (map list (~form
+        :for x (range 3)
+        :for y (range 3)
+        :if (> y x)
+        :for z [7 8 9]
+        :if (!= z 8)
+        (, x y z)))
+      [[0 1 7] [0 1 9] [0 2 7] [0 2 9] [1 2 7] [1 2 9]])
+    (f (~form :for (, x y) (.items {"1" 1 "2" 2}) (* y 2))
+      [2 4])
+    (f (~form :for x (range 2) :for y (range 2) (, x y))
+      [(, 0 0) (, 0 1) (, 1 0) (, 1 1)])))))
+
+(defn test-simple-dict-comprehensions []
+  (assert (= (compd :for x []  x x)
+    {}))
+  (assert (= (compd :for x (range 4) :if (% x 2)  (str x) (* 2 x))
+    {"1" 2  "3" 6}))
+  (assert (=
+    (compd
+      :for x (range 3)
+      :for y (range 3)
+      :if (> y x)
+      :for z [7 8 9]
+      :if (!= z 8)
+      (.format "{}{}{}" x y z) [x y z])
+    {"017" [0 1 7]  "019" [0 1 9]  "027" [0 2 7]  "029" [0 2 9]
+      "127" [1 2 7]  "129" [1 2 9]})))
+
+#@ ((pytest.mark.parametrize "form" ["compl" "comps" "compg"])
+(defn test-statement-comprehensions [form]
+  (defn f [a b]
+    (assert (= (sorted a) b)))
+  (setv form (HySymbol form))
+  (eval `(do
+
+    (setv l [])
+    (f (~form :for j [1 2 3] (.append l (inc j)) j)
+      [1 2 3])
+    (assert (= l [2 3 4]))
+
+    (f
+      (~form :for x (range 3) (*= x 10) x)
+      [0 10 20])
+
+    (setv l [])
+    (f
+      (map list (~form
+        :for x (range 3)
+        :for y (range 3)
+        :if (> y x)
+        :for z [7 8 9]
+        (.append l [x y z])
+        :if (!= z 8)
+        (, x y z)))
+      [[0 1 7] [0 1 9] [0 2 7] [0 2 9] [1 2 7] [1 2 9]])
+    (assert (= l
+      [[0 1 7] [0 1 8] [0 1 9] [0 2 7] [0 2 8] [0 2 9] [1 2 7] [1 2 8] [1 2 9]]))
+
+    (defclass E [Exception] [])
+    (setv l [])
+    (import pytest)
+    (with [(pytest.raises E)]
+      (list (~form
+        :for x (range 10)
+        (.append l x)
+        (when (= x 5)
+          (raise (E)))
+        x)))
+    (assert (= l [0 1 2 3 4 5]))))))
+
+(defn test-statement-dict-comprehensions []
+  (setv l [])
+  (assert (= (compd :for j [1 2 3] (.append l (inc j)) j j)
+    {1 1  2 2  3 3}))
+  (assert (= l [2 3 4]))
+
+  (assert (= (compd :for x (range 3) (*= x 10) (str x) x)
+    {"0" 0  "10" 10  "20" 20}))
+
+  (setv l [])
+  (assert (=
+    (compd
+      :for x (range 3)
+      :for y (range 3)
+      :if (> y x)
+      :for z [7 8 9]
+      (.append l [x y z])
+      :if (!= z 8)
+      (.format "{}{}{}" x y z) [x y z])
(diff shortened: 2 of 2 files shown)


gilch:
It looks like only `compg` has the correct expansion. `compl` and `comps` need to set a `[]` or `#{}` to a gensym, and then `.append` or `.add` to it (instead of `yield`) and finally return that; or it won't behave consistently with the expression version. And it needs to since it could expand to either. I also expect it will perform better, since this is how Python's authors chose to compile these expressions.

And `dcomp` returning the second-to-last expression as the key is kind of weird given the implicit `do`, but I'm not sure what to do about this. A `:` separator like Python? Hy's dicts aren't written like that. Return a tuple or cons pair? But then we can't expand it to an expression. Get rid of the implicit do? But then how could an explicit one possibly work?

We don't need two types of special forms for comprehensions. `list-comp` and friends should be removed. But then `compl` could be named `list-comp`. I don't might having a shorter name for something we expect to use often, but `s-comp` seems a lot less confusing than `comps`, which could easily be confused with `comp`.

And finally, the syntax doesn't seem to be compatible with `for`, which means only the comprehensions get it.

Kodiologist:
> `compl` and `comps` need to set a `[]` or `#{}` to a gensym, and then `.append` or `.add` to it (instead of `yield`) and finally return that; or it won't behave consistently with the expression version.

I think those two strategies should have the same effect. Can you provide an example of inconsistent behavior under the current implementation?

> `list-comp` and friends should be removed

Yep, it's on the to-do list.

> But then compl could be named list-comp. I don't might having a shorter name for something we expect to use often, but s-comp seems a lot less confusing than comps, which could easily be confused with comp.

So you'd prefer `l-comp`, `s-comp`, `d-comp`, and `g-comp`? (Honestly, I just picked the current names so they'd be alphabetized next to each other in the documentation etc.)

> And finally, the syntax doesn't seem to be compatible with `for`, which means only the comprehensions get it.

Right. I'm not suggesting any changes to `for` here.

gilch:
> I think those two strategies should have the same effect. Can you provide an example of inconsistent behavior under the current implementation?
```
=> (comps :for x [1 2 3] (yield x))
{yield x for x in [1, 2, 3]}
<generator object <setcomp> at 0x000001FBC90DFEB8>
=> (comps :for x [1 2 3] 42 (yield x))
def _hy_anon_var_1():
    for x in [1, 2, 3]:
        42
        yield yield x
set(_hy_anon_var_1())
{1, 2, 3, None}
```
The `42` is just to force the different expansion. They should produce the same result, but they have dramatically different answers.

Kodiologist:
[I think those kinds of constructs are actually a bug in CPython.](https://stackoverflow.com/a/32139977)

gilch:
> I think those kinds of constructs are actually a bug in CPython.

I think that's part of the spec. Comprehensions are _defined_ to work that way. It's not just CPython. PyPy does the same thing.
```
>>>> {(yield x) for x in [1, 2, 3]}
<generator object <setcomp> at 0x0280bdf0>
```
And besides, I think it performs better, which is why the Python authors chose to compile it that way.

> So you'd prefer l-comp, s-comp, d-comp, and g-comp? (Honestly, I just picked the current names so they'd be alphabetized next to each other in the documentation etc.)

I suppose. But I think the current names (`list-comp` etc.) are fine. If you want them to start the same way, maybe `for-list`, `for-set`, etc. could work. It's not much shorter, but the initial `:for` keyword is redundant. You're putting the generating expression at the end instead of the beginning like Python does. When would you ever *not* start these with `:for`? If you eliminate it, that name makes sense. Couldn't `(compl :for x [1 2 3] (* x x))` be written like `(for-list x [1 2 3] (* x x))`?

> Right. I'm not suggesting any changes to for here.

But I am. I've thought for some time now that Hy's `for` should act like `doseq`, including the modifiers, and Hy's comprehensions should have them too. They need a consistent syntax.

Kodiologist:
>> I think those kinds of constructs are actually a bug in CPython.
>
> I think that's part of the spec.

The language reference for Python 3 [says](https://docs.python.org/3/reference/expressions.html#yield-expressions) "The yield expression… can only be used in the body of a function definition." Besides, it's hard for me to see why this construct would've been designed to work seeing as it serves no apparent purpose.

But there's no reason I can't switch to `.append` / `.add` / `dictionary[key] = …` instead of `yield`, I think. It will just be a bit more verbose to implement, possibly.

> Couldn't `(compl :for x [1 2 3] (* x x))` be written like `(for-list x [1 2 3] (* x x))`?

Yes, but not, e.g., `(compl :for x [1 2 3] (f) :for y [1 2 3] (* x y))`.

>  I've thought for some time now that Hy's `for` should act like `doseq`, including the modifiers

I guess I've never seen the logic in that, beyond "Clojure did it".

gilch:
> Yes, but not, e.g., `(compl :for x [1 2 3] (f) :for y [1 2 3] (* x y))`.

I was suggesting `(for-list x [1 2 3] (f) :for y [1 2 3] (* x y))`. Only the first `:for` is redundant. But on second thought, special cases like that make macros that expand into comprehensions more difficult to implement. I think Clojure's `for` comprehensions do a better job here--
```Clojure
(for [x [1 2 3], y [1 2 3]] (* x y))
```
The `:for` keyword (as `compl` means it) is implied unless you use another modifier. Make the common case easy.

> I guess I've never seen the logic in that, beyond "Clojure did it".

Why do we have a nested `for` construct to begin with? Python doesn't have it. We got it from Clojure. We just didn't finish it. If you'd prefer to get rid of `for` altogether and just use `for*` (we'd rename it to `for`) that works for me too, I guess. You'd have to sell it to another core member. But when nesting loops, sometimes you _need_ to break/continue early. Hy's current approach is no better than `for*`. You have to nest the `for` expressions the same way.

Kodiologist:
> `(for [x [1 2 3], y [1 2 3]] (* x y))`

Where did the `(f)` go?

> If you'd prefer to get rid of for altogether and just use `for*` (we'd rename it to `for`) that works for me too, I guess.

That works for me too, I guess. It's out of scope of this PR, though.

gilch:
> It's out of scope of this PR, though.

Nice try, but it's certainly not out of the scope of #1371, which this PR purports to implement.

> Where did the (f) go?

I must have overlooked that when I copy/pasted. Clojure's not that in to side effects, but we can still do it--
```Clojure
user=> (defn f [] (println "called f!"))
#'user/f
user=> (vec (for [x "abc", y (do (f) "xyz")] (str x y)))
called f!
called f!
called f!
["ax" "ay" "az" "bx" "by" "bz" "cx" "cy" "cz"]
```
```Python
=> (defn f [] (print "called f!"))
def f():
    return print('called f!')
None
=> (compl :for x "abc" (f) :for y "xyz" (+ x y))
def _hy_anon_var_1():
    for x in 'abc':
        f()
        for y in 'xyz':
            yield (x + y)
list(_hy_anon_var_1())
called f!
called f!
called f!
['ax', 'ay', 'az', 'bx', 'by', 'bz', 'cx', 'cy', 'cz']
```
And really, so can Python's expressions.
```Python
>>> do = lambda *a: a[-1] if a else None
>>> f = lambda: print("called f!")
>>> [x+y for x in [1,2,3] for y in do(f(),[1,2,3])]
called f!
called f!
called f!
[2, 3, 4, 3, 4, 5, 4, 5, 6]
```

Kodiologist:
> it's certainly not out of the scope of `#1371`, which this PR purports to implement.

Okay, it's your issue. Removed above.

So you aren't on board with this PR because it doesn't implement the same keywords for `for`. Is that correct? I ask because I wouldn't want to make other changes you request only for you to reject it anyway.

gilch:
>  I wouldn't want to make other changes you request only for you to reject it anyway.
These aren't the changes I requested in #1371. It's your own made-up syntax.

As for the changes I requested in _this_ PR, #588 is still an issue. This is probably a good start on fixing it even if we stick to something closer to the old syntax for now, and my requested changes would still be needed for that. I'm not suggesting you throw it all away.

I want Hy to have a coherent, consistent design. I think about changes in the context of the whole language. I'm not on board with the new syntax because it doesn't fit well. Either it has to change, or Hy does.

Kodiologist:
> These aren't the changes I requested in #1371.

Correct. I meant changes to this PR.

To clarify, if I change to an `.append` / `.add` / `dictionary[key] = value` implementation, and change the names to `lcomp`, `scomp`, `dcomp`, and `gcomp` (the hyphens weren't necessary, come to think of it), would you approve this PR? You mentioned that you didn't like how `compd` uses its last two arguments as the final values, but you didn't seem to settle on an alternative you liked. Maybe the most obvious choice would be to require the last argument of `compd` to be two forms in a pair of square brackets; the brackets would be interpreted as special syntax like in `for` or `defn` or your `let` rather than a list display.

Kodiologist:
@gilch What do you think of the proposal?

gilch:
Hy currently has some bugs.

🐛The current reordering of statements is a bug. Evaluation must happen in the order written, and statements must be repeated properly in comprehensions. #588. The fix probably needs to be at the compiler level, since that's the part of Hy responsible for making statements act like expressions.

🐛 The current syntax is unable to express early filters in comprehensions. #848. Thus we _must_ change the syntax.

The proposal is a clear improvement over the broken status quo because it fixes these two bugs. I'd have to fix these myself (eventually) if no-one else does. We can't (eventually) release Hy 1.0 with these outstanding.

✔️ The `.append` / `.add` / `dictionary[key] = value` setup is necessary for consistent behavior. _Any_ final solution would have this.

🆗 The names `lcomp`, `scomp`, `dcomp`, and `gcomp` are OK. I don't really need the hyphens either. They're reasonably short and less likely to be confused with `comp`. I think we can do better. 
* `forlist`, `forset`, `fordict`, `forgen`. These have no chance of confusion with `comp` and are ordered together alphabetically like you originally wanted, but they're a bit longer to type at 6-7 characters instead of 5. I don't think that's a big deal for a special form name.
* `lfor`, `sfor`, `dfor`, `gfor`. I wouldn't shorten `forset` to `fors` for the same reason I wouldn't approve of `comps` but `sfor` works, so if you really care about length more that alphabetizing, these are even shorter at 4 characters instead of 5.

🆗 The square brackets for the dict comprehension pair seems OK to me. I don't particularly have a better idea either.

But the proposal is still noticeably sub-optimal and I'm concerned that if I approve this partial fix, we'll lack the will to clean it up _ever_. Then we'll be stuck with a proliferation of incompatible macros to fix it downstream or just have to live with it.

⚠️ The proposal makes up a new syntax instead of looking like Python, Clojure, or Common Lisp--the guidelines we've been using for everything else. I'd let this slide if the syntax is sufficiently elegant, but it's not.

⚠️ The first `:for` in the proposed syntax is redundant.

⚠️ Approving this would probably prevent me from doing #1371, which adds Clojure's `:let` and `:while` modifiers to comprehensions, and to our nested `for` macro.

⚠️ A deep change to syntax like this should probably have input from more of @hylang/core in an issue _before_ a PR. I'd be more confident in our final choice if I'd seen some alternatives.

Better late than never. Thoughts people? How do you guys _want_ this to work? Given that we _need_ early filters for full Python compatibility, and want `:let` and `:while` capabilities (or equivalent) in both comprehensions and the `for`/`doiter` macro (or equivalent), what syntax _should_ Hy have for comprehensions?

Kodiologist:
> Approving this would probably prevent me from doing #1371, which adds Clojure's `:let` and `:while` modifiers to comprehensions, and to our nested for macro.

I can't speak for `for`, for which I have no intention of adding additional features, but there is no reason that `:while` could not be added to the current implementation, and `:let` would be redundant with a regular `(setv ...)`, if I understand correctly.

> The first :for in the proposed syntax is redundant.

Coincidentally, this will no longer be the case if we add `:while`.

gilch:
> * When in doubt, defer to Python.
> * If you're still unsure, defer to Clojure.
> * If you're even more unsure, defer to Common Lisp.

I was only looking at Python and Clojure before, but have since discovered that Common Lisp can do comprehensions with its `loop` macro if you use the `collect` word. (This is very different from Clojure's `loop/recur`. It's more of a DSL for loops.) Studying it has been enlightening. For example,
```Lisp
* (loop for i from 1 to 5 when (oddp i) collect (* i i))

(1 9 25)
```
is like
```Python
>>> [i*i for i in range(1, 6) if i%2==1]
[1, 9, 25]
```
That wasn't too bad, but I don't think we should copy `loop` exactly. It's complex, verbose, and Python already does some of it better. But it did give me some ideas.

> I'd be more confident in our final choice if I'd seen some alternatives.

I've been experimenting with possible syntax. To match Python, we must have, at minimum,
* sequential binding from an iterable (`for`/`in`). This is like a `map()`.
* filtering (`if`). Like a `filter()`. 
* And the loops can nest and the filters must be allowed between them.  #848

Additionally, we want
* early abort (like Clojure's `:while`) #1371
* new bindings (like Clojure's `:let` and Haskell's `let`) #1371
* @Kodiologist wants arbitrary expressions at any point. Not sure why.
* And I want comprehensions to be consistent with `for`.

And finally, 
* We want the common case to be easiest!

Concrete example time. If we start with the verbose Common Lisp `loop` DSL style as applied to Hy, we get something like
```Clojure
(for
  :for i :in (range 6)
  :do (f)
  :if (even? i)
  :let k (* i i)
  :for j :in (range k)
  :while (< k 5)
  :list [i j k])
```
Here I'm using `for`, since name `loop` was already taken for `loop`/`recur`. It fits because that's what Clojure calls its comprehensions. I'm also using `:list` instead of Common Lisp's `collect`, since Hy needs to be able to make a `:set` too. Genexprs would use `:yield`. Dict comprehensions could use `:key`/`:val` or even just `:dict`. Normal side-effecting for loops would use a `:body`. So it doesn't conflict with Python's use of `for` either and we can get rid of our `for` macro.

In the example, `:let k (* i i)` is short for `:do (setv k (* i i))` and `:while (< k 5)` is short for `:do (if (< k 5) (break))`, and `:if (even? i)` compiles like `:do (if (even? i) (continue))`. `:if` would compile to a comprehension filter when possible, of course.

The above syntax is too verbose for my taste, but very readable. Can we simplify it?

`:for`/`:in` is redundant. One could imply the other. (Like how `:key`/`:val` could be just `:dict`.) @Kodiologist's current proposal just uses the (longer) `:for`. Python got comprehensions from Haskell in the first place. Haskell just uses the (shorter) `:in` (written `<-`). And Haskell doesn't need the `:if`, since it's implied when not using `<-` or `let`.

But let's make the common case easy.

@Kodiologist's current proposal gets this exactly backwards, making the `:do` implicit, even though "arbitrary expressions at any point" is the _least_ used of all our desired points. So it's noticeably suboptimal for comprehensions.

I use comprehension for mapping _far_ more than for filtering, so let's make the always-used `for`/`in` implicit instead. Besides, `(for :for` looks awkward.
```Clojure
(for
  i (range 6)
  :do (f)
  :if (even? i)
  :let k (* i i)
  j (range k)
  :while (< k 5)
  :list [i j k])
```
The next most common is the filter, but `:if` is quite short, and it matches Python's keyword here.

I think this is pretty good now, but can we do any better?

Kodiologist:
So in place of what is currently `(for [x xs] (foo x) (bar x))`, you'd write `(for x xs :body (do (foo x) (bar x)))`? That seems like a lot of boilerplate for a plain `for` loop.

gilch:
Ah, no. I removed a paragraph explaining that better. The `:body` would collect all remaining expressions in the `for`. So instead of `(for [x xs] (foo x) (bar x))`, you'd write `(for x xs :body (foo x) (bar x))`.

gilch:
On second thought, perhaps the final `:do` could be special-cased to act like `:body` if there isn't another collect word (`:list`, etc.) present. Then you could write `(for x xs :do (foo x) (bar x))`, which is pretty similar to what we're doing now.

Kodiologist:
To be honest, it makes me feel a little faint to think of rewriting every `for` loop in every one of my Hy programs. I know we're young enough that we don't have to maintain backward compatibility, but it seems like a lot of heat for a little light. So I would suggest we stick to comprehensions and leave `for` alone for now.

gilch:
Regex is your friend.

We can mostly keep compatibility with the current `for` syntax by using Clojure's approach.
```Clojure
(for :list
  [i (range 6)
  :do (f)
  :if (even? i)
  :let k (* i i)
  j (range k)
  :while (< k 5)]
  [i j k])
```
Here we put the collect word at the top, since you only ever need one of them. It might actually be easier to compile. But to distinguish the body from the implicit `for`/`in`, we add square brackets. Here's the side-affecting version.
```Clojure
(for :do
  [i (range 6)
  :do (f)
  :if (even? i)
  :let k (* i i)
  j (range k)
  :while (< k 5)]
  (print i)
  (print j k))
```
But now that the keyword is right next to the function name, we can just merge them, so
```Clojure
(forlist
  [i (range 6)
  :do (f)
  :if (even? i)
  :let k (* i i)
  j (range k)
  :while (< k 5)]
  [i j k])
```
Likewise `forset`, `fordo`, etc. or just plain `for` instead of `fordo`, really. Now most loops written in the old style will work. But you do still have to fix the `:if` filters. But you were proposing that anyway. It can't be helped! The current syntax is inadequate. #848. We _need_ early filters for reasonable performance.

Kodiologist:
Hmm, I like that better. Two suggestions:

- For the versions other than plain `for` (a.k.a. `fordo`), remove the square brackets. They're redundant, because there can be only one returning form.
- If you want `:let x 5` to work like `:do (setv x 5)`, which seems like the most obvious way to do it to me, let's call it `:setv` or something instead of `:let`, so as not to suggest a connection with your new `let`, which is quite a different animal.

Kodiologist:
Also, be aware of the fact that to maintain `continue` and `break` behavior in `for` loops, we'll need to compile `for` loops as something like

     for x, y in hy_anon_generator():
         [ body of the loop ]

which means that the body won't be in the same function as the earlier parts of the `(for ...)` construct. I guess we'd choose the variables on the left side of the `in` by taking each variable from a for clause and each variable from a `:let` clause.

 Are you okay with that?

gilch:
> call it `:setv`

That seems OK to me, if that's how we actually compile it. Python can already emulate `:let` to some degree by using `for` `in` a length-1 sequence. But it might actually be less overhead to use `setv` in all cases, even if that means creating an explicit generator function.

> maintain continue and break behavior in for loops,

We currently compile
```
(for [x xs y ys] [x y])
```
To
```
for x, y in ((x, y) for x in xs for y in ys):
    [x, y]
```
But up to about version 0.9 we used to compile it like
```
for x in xs:
    for y in ys:
        [x, y]
```
The main reason for the change was Python's lack of a labeled `break`/`continue`. Even with 3+ loops, we could only really pick the outer or inner, so which loop should that break? It made more sense to break the outer loop (the whole `for` form) than the inner one, especially since we didn't have `return` at the time, which is the usual way of breaking nested loops in Python.

But we have `return` now, and with `:if` and `:while` (and `:do`) available, we can break any of the loops we want. So I think it makes more sense for a `break` in the body to break only the inner loop now.

The old nested-loop style seems to play nicer with the new control words (like `:setv`?) anyway. If we want a `:do` available at any point in the loop, it should use the same scope as the body.

Kodiologist:
Okay. So `(for ...)` will generate nested loops, with no function declarations, whereas `forlist` (or `lfor`) and the other comprehensions will make an anonymous function, and hence, Python 3's distinction between leaky `for` loops and non-leaky comprehensions will be maintained. But in both cases, `break` and `continue` will refer to the innermost loop.

Shall I go ahead and implement this, then?

gilch:
A big motivation for updating Hy's `for` to be consistent with new comprehension features is that it makes it easy to build macros for comprehensions over custom data structures. For example, 
```Hy
(defmacro/g! fordoto [coll for-args method]
  `(do (setv ~g!coll ~coll)
       (for ~for-args
         (-> ~g!coll ~method))
       ~g!coll))
```
Now you can emulate a normal list comprehension like
```Hy
(fordoto [] [x (range 10) :if (odd? x)]  (.append x))
```
Note that our comprehension control words (like `:if`) "just work" "`for` free", because `for` has them too.

We could just as easily use some custom object instead of `[]` if it has an `.append` method. But we can use a different method too.
```Hy
(fordoto [] [x xs :if (odd? x)  y ys :if (even? y)] (.extend [x y]))
```
This version works like my `chain-comp`. It flattens one layer. (`chain-comp` is a very useful construct in it's own right, because monads. It appears in many languages with various names.)

`fordoto` is pretty general, but expects mutable objects. Immutable collections would be done differently. And if you use them a lot, you might a dedicated comprehension so you don't have to spell out the method names every time, e.g.  a `forvector` macro or something.

> For the versions other than plain for (a.k.a. fordo), remove the square brackets. They're redundant, because there can be only one returning form.

It seems awkward to remove square brackets from the above `fordoto`. Instead of `[coll for-args method]` you'd just have `[coll for-args]`. I guess it could be done though--manually pop off the last `for-args`. And custom loop macros might want to add custom keywords too. That might actually work better if comprehensions don't have the square brackets. One can always use a keyword instead. I'm not sure how I feel about this, but it seems like it could work.

Kodiologist:
That sounds like a "yes". Very good.

gilch:
> Okay. So (for ...) will generate nested loops, with no function declarations, whereas forlist (or lfor) and the other comprehensions will make an anonymous function, and hence, Python 3's distinction between leaky for loops and non-leaky comprehensions will be maintained. But in both cases, break and continue will refer to the innermost loop.

Sounds right.

> Shall I go ahead and implement this, then?

Yes, but details may come up after you start. Once you have the `for` implemented it seems like you could implement the generator functions using that. But `for` is a macro now. If you move it to the compiler and get rid of `for*`, I might have to adjust `let`. I'm not sure of your implementation approach.

Try it. I approve of the syntax, even dropping the square brackets, I think.

gilch:
> Yeah, I think for will have to be a special form.

It could probably still be done as a macro. I think the compiler can write code and then do a macroexpansion pass on it. I'm not sure if that's easier, but it wouldn't affect `let`. But I might have to adjust it for the new comprehension syntax anyway. Hmm.

Kodiologist:
I'm rethinking the issue of implementing `lfor` with `.append`, `sfor` with `.add`, etc. instead of using a generator for everything. The only behavior this seems to make a difference for is `yield` in a comprehension, which, if it isn't a bug (and I'm pretty sure it's a bug, even if implementations other than CPython do the same thing), still has no use I know of and defies the specification. So I don't see the benefit. And the cost is more complexity in the compiler logic for these constructs, which is already complex, as well as longer and more complex generated AST. It's a lot of overhead for nothing.

gilch:
If it's that difficult, I'm rethinking having comprehensions in the compiler _at all._ I kind of feel like we should be able to output comprehension AST in case a library uses it (like astor), but we'll always have some small differences from raw Python AST due to Hy anon vars required to make statements expressions and such, so I kind of doubt we'll have 100% compatibility with that kind of deep introspection. In the rare case it's really needed, one can always write that part in Python, I suppose.

At the bytecode level (at least in CPython), comprehensions are just sugar for generator functions. If we implement them the same way as the bytecode compiler in macros, there should be no performance impact.

I think macros are a lot easier to write than special forms. Given the enhanced `for` macro, I could already implement the base comprehensions in terms of the above`fordoto` macro.

Kodiologist:
As I'm writing it now, `for` uses pretty much the same compiler code as all the comprehension forms. I don't think you'd be able simplify the compiler much by farming out comprehensions to macros. But you can take a look for yourself once I'm done.

Kodiologist:
Closing in place of #1468.

PR #1438, opened 2017-10-15
Standardize hy.core docstrings

ekaschalk:
Now that we are getting some good documentation introspection utilities in `hy-mode`, Hy's exposed docstrings should be standardized.

Only docstrings were modified in this commit.

I changed formatting and wording in `hy.core.language`, `hy.core.shadow`, and `hy.core.macros` to:
1. First line is always one sentence with period.
2. Use the argument names in the first line and backtick them.
3. Further lines do not include spaces.

### DIFF
--- a/hy/core/bootstrap.hy
+++ b/hy/core/bootstrap.hy
@@ -7,7 +7,7 @@
 ;;; They are automatically required everywhere, even inside hy.core modules.
 
 (defmacro if [&rest args]
-  "if with elif"
+  "Conditionally evaluate alternating test and then expressions."
   (setv n (len args))
   (if* n
        (if* (= n 1)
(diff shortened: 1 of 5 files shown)


hy/core/bootstrap.hy
```diff
@@ -7,7 +7,7 @@
 ;;; They are automatically required everywhere, even inside hy.core modules.
 
 (defmacro if [&rest args]
-  "if with elif"
+  "Conditionally evaluate `args`, when given three conditions equiv. to `elif`."
```
Kodiologist: This is a comma splice, if I'm reading it correctly.

Kodiologist:
You also have the misspelling "peform" and comma splices in the docstrings of the shadow functions.

hy/core/bootstrap.hy
ekaschalk: I misread the documentation, I've never used `if` with more than one condition. I'll fix that up.

ekaschalk:
Fixed your comments.

hy/core/bootstrap.hy
```diff
@@ -7,7 +7,7 @@
 ;;; They are automatically required everywhere, even inside hy.core modules.
 
 (defmacro if [&rest args]
-  "if with elif"
+  "Conditionally evaluate `alternating test and then expressions `args`."
```
Kodiologist: I think you've got a stray backtick and `args` here.

ekaschalk: Fixed it.

PR #1443, opened 2017-10-23
Completer completes hy.core.language names

ekaschalk:
This is a trivial update to catch `hy.core.language` constructs in autcompletion.

I don't recall if Python2 has iteration over a dict equivalent to iterating over its keys, the only thing to watch.

I'm going to do a more involved rewrite to catch modules eg. `numpy.add`, that will come later.

### DIFF
--- a/hy/completer.py
+++ b/hy/completer.py
@@ -7,10 +7,9 @@ import os
 import re
 import sys
 
-import hy.macros
 import hy.compiler
+import hy.macros
 from hy._compat import builtins, string_types
-
 
 docomplete = True
 
@@ -38,6 +37,7 @@ class Completer(object):
             raise TypeError('namespace must be a dictionary')
         self.namespace = namespace
         self.path = [hy.compiler._compile_table,
+                     dir(hy.core.language),
                      builtins.__dict__,
                      hy.macros._hy_macros[None],
                      namespace]
@@ -75,11 +75,12 @@ class Completer(object):
     def global_matches(self, text):
         matches = []
         for p in self.path:
-            for k in p.keys():
+            for k in p:
                 if isinstance(k, string_types):
                     k = k.replace("_", "-")
                     if k.startswith(text):
                         matches.append(k)
+
         return matches
 
     def tag_matches(self, text):
--- a/hy/completer.py
+++ b/hy/completer.py
@@ -54,8 +54,8 @@ class Completer(object):
 
         if m:
             expr, attr = m.group(1, 3)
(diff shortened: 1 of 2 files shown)


gilch:
I had to cancel the test build. The `pypy` test was running for over an hour. That's not supposed to happen. All the others passed though, strange.

Kodiologist:
I think this doesn't account for mangling properly. I can't complete to `numeric?`, `is_numeric`, or `numeric!`, although I can complete to `is-numeric` and `defmacro-bang`. `hy.extra.reserved.names` exists for this sort of purpose, so try using that.

ekaschalk:
I'll look to fix the mangling - but isn't completing `is-numeric` but not `is_numeric` intended? Replacing the underscore with hyphens was already being done. As far as autocompletion goes - I only want to present the hyphenation as a candidate.

Kodiologist:
The thing about `is-numeric` is that it's an incompletely mangled form. The original is `numeric?` and the fully mangled form is `is_numeric`. Either would be a reasonable candidate for choice for a completable name, but not `is-numeric`.

ekaschalk:
Yes you're right I missed the question mark.

ekaschalk:
Is `hy.extra.reserved.names` even used anywhere? On my machine, it will throw an error that `hy.core` isn't found when I import it (implying the tests are only passing due to the sys.path extension). Trying to use it will throw `AttributeError: 'function' object has no attribute 'kwlist'`.

Anyway, it has overlap with the current completion but the current completion catches all symbols. Simply importing and using `hy_symbol_unmangle` is sufficient.

ekaschalk:
I've added support for completion of symbols with module member access like `pd.DataFrame.add` or `itertools.tee`.

I've also added tests for completion, there didn't seem to be any before.

It is strange that `_hy_macros` remains an empty dict python-side. This means I can't test that macros complete in python tests (though they do complete in Hy as it is nonempty).

ekaschalk:
Due to complexities with the `namespace`, I'm reimplementing completion entirely as emacs-side setup code and scrapping this PR.

This is how Eldoc works, where a simple `(.eval builtins obj (globals))` plus a macros/compiler-table check catches everything. Importing the `Completer` breaks the globals containing the imports.

It also lets me skip the readline specific functionality that's irrelevant and format the return results in a manner that allows skipping regex extraction of candidates from printed list.

More importantly I can also then extract the type of the object (eg. class/func/variable/keyword/macro) in some manner and have autocompletion display the type of the candidate, like python and all other autocompletion implementations.

Issue #1439, opened 2017-10-19
Copy toolz.curry to Hy?


ekaschalk:
Currying is useful regardless of threading macros. Consider

    (*map (fn [k v] (.format "[{} {}]" k v)) (.items {1 2 3 4}))

    (*map (tz.curry str.format "[{} {}]") (.items {1 2 3 4}))

    ;; I personally have it bound as a tag macro
    (*map #$(str.format "[{} {}]") (.items {1 2 3 4}))

    ;; Also possible but breaks code introspection and not ideal
    (*map (xi str.format "[{} {}]" x1 x2) (.items {1 2 3 4}))

I saw a previous issue discuss adding `functools.partial` but it was never merged and closed.

I specify toolz implementation because it will curry kwargs as well.

I'd also support `flip` getting added but saw it was ignored for lack of support. I'd say now that more operators are shadowed, its usecases are a bit more diverse, eg. flipping shadowed `get` in a thread or reversing the addition or division of items in a thread.

Kodiologist:
> breaks code introspection

What do you mean?

ekaschalk:
At the moment I'm referring to just eldoc. Because it's a lambda, unlike currying, I cannot make assumptions on its further sexps. In the tag macro case eldoc works already and the funcall case it can be made to work as a special case (by prepending "curried" to the eldoc output of the eldoc for the second sexp).

I dislike `xi` as a macro and think implementing it as a compiler construct like was done with the unpacking generalizations would be better. Much quicker/easier would be just porting curry.

Kodiologist:
I see.

> I dislike `xi` as a macro and think implementing it as a compiler construct like was done with the unpacking generalizations would be better.

Why's that? We always try to implement things as macros rather than special forms when possible, to make the language more flexible and the compiler simpler.

ekaschalk:
One reason is the above case.

Another is a bit more vague/personal - that `xi` breaks the rule that you can look at the first sexp of the form and know what's happening. The code being "conceptually evaluated" starts at the second sexp. This often results in a double-take whenever I glance at code I wrote using it. 

This is relative but to me currying is a fundamental operation on par with eg. unpacking. I feel a tag macro/compiler construct is the natural representation as the form is not polluted by the curry call. `with-decorator` as `#@` isn't all that different imo.

Kodiologist:
[...]

PR #1444, opened 2017-10-23
change xi macro to #% tag macro

gilch:
As discussed in #1439. This replaces `xi` with a more Clojure-like `#%` tag macro.

### DIFF
--- a/docs/extra/anaphoric.rst
+++ b/docs/extra/anaphoric.rst
@@ -234,20 +234,32 @@ Returns a function which applies several forms in series from left to right. The
 #%
 ==
 
-Usage ``#%(body ...)``
+Usage ``#% expr``
 
-Makes a function with an implicit parameter list from ``%`` parameters.
+Makes an expression into a function with an implicit ``%`` parameter list.
 
-A ``%i`` symbol designates the `i` th parameter (1-based, e.g. ``%1 %2 %3`` etc.).
+A ``%i`` symbol designates the (1-based) *i* th parameter (such as ``%3``).
+Only the maximum ``%i`` determines the number of ``%i`` parameters--the
+others need not appear in the expression.
 ``%*`` and ``%**`` name the ``&rest`` and ``&kwargs`` parameters, respectively.
-Nesting of ``#%()`` forms is not recommended.
-
-This is similar to Clojure's anonymous function literals (``#()``).
 
 .. code-block:: hy
 
-   => (#%(identity [%1 %5 [%2 %3] %* %4]) 1 2 3 4 5 6 7 8)
-   [1, 5, [2, 3,] (6, 7, 8), 4]
-   => (def add-10 #%(+ 10 %1))
-   => (add-10 6)
-   16
+    => (#%[%1 %6 42 [%2 %3] %* %4] 1 2 3 4 555 6 7 8)
+    [1, 6, 42, [2, 3], (7, 8), 4]
+    => (#% %** :foo 2)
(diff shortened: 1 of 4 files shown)


ekaschalk:
Do you have thoughts on whether to highlight the `%i`? I don't know how clojure does it.

gilch:
You mean, "does clojure-mode?" It does, I checked. Should Hy? I'm less certain. This is not builtin syntax. This is not even a core macro. 

But, if we conventionally used `%`-prefixed anaphors in other macros (as Clojure does), then highlighting such symbols would make sense. It does improve readability. But some might argue the old `xi` symbols were more readable, because that's what won out in #879, even though I presented `%` symbols as an option.

We could use `#xi` with the old `xi` symbols. Maybe it's more readable, but it's also more likely to interfere with user symbols. Names like `x1`, `x2` could totally come up.

ekaschalk:
I meant "does and what face", I can check into it.

A middle-ground is to implement is as an option, your call whether to enable it by default or not.

gilch:
I added a `%**` for the `&kwargs` parameter. This makes `#%` usable for partial application even if the function takes keyword arguments. I also changed the `&rest` parameter from `%&` to `%*` for consistency.

For example, `#%(foo #* %* #** %**)` would do the same thing as function `foo`, regardless of its signature.

Kodiologist:
This is cool. I like it. Is it ready for review? Still, I would finish up your outstanding PRs (and review those of others) before starting new ones like this.

gilch:
Hy can't be my top priority all the time. I'm busy with other things, but this seemed like a really easy fix, especially since I wrote the code in question in the first place.

It's a tag macro, so I've generalized `#%` to work on arbitrary expressions. 
For example, you can just use `#%[%1 %2]` instead of or
`(xi identity [x1 x2])` like before.

Clojure would have to use `#(identity [%1 %2])` or `#(vector %1 %2)` for that, but it saves the extra `%` in the more common case of a normal s-expression. We'd have to build function literals into the compiler to do it Clojure's way.

Kodiologist:
A lot of the diff is whitespace changes to old code. If you want to include them in this PR, could you put them in a separate commit?

Instead of "presence of a ``%`` parameter", I would write "presence of symbols named ``%*``, ``%**``, or ``%i`` for some number ``i``".

The tests seem to pass without the `(import hy)` in anaphoric.hy, despite the use of `HySymbol`. So I think you can remove it.

`(if (coll? expr) (flatten expr) (, expr))` can be written `(flatten [expr])`.

You have a commented-out `require` at the bottom of anaphoric.hy. I'm guessing it's old debugging code.

gilch:
> whitespace

Done.

> Instead of "presence

I reworded it, but not quite that way.

>  (import hy)

Yep.

> (flatten [expr])

More elegant, I like it.

> old debugging code

Removed.

@Kodiologist, how about now?

docs/extra/anaphoric.rst
```diff
@@ -229,21 +229,37 @@ Returns a function which applies several forms in series from left to right. The
...
+When used on an s-expression,
+``#%`` is similar to Clojure's anonymous function literals--``#()``.
+
+.. code-block:: hy
+
+    => (setv add-10 #%(+ 10 %1))
+    => (add-10 6)
+    16
+
+``#%`` determines the parameter list by the presence of a ``%*``, or ``%**``
```
Kodiologist: Stray comma.

PR #1447, opened 2017-10-31
update astor to 0.6

### LINKED ISSUE #1446: Astor update 0.6 10/31 change in API (astor.codegen to astor.code_gen) breaks hy.cmdline
Good timing, started a Hy project today and noticed the change.

Either update the import or specify version 0.5 as the requirement.

gilch:
Fixes #1446.

The new astor is much improved. You can actually read quasiquoted forms in Python now.

I've almost got this working, but Hy's `NaN` literal crashes astor and I'm not sure what to do about it.
Astor also prints off `Inf` as `1e1000` instead now.

One option is to remove the three literals, `Inf`, `-Inf`, and `NaN` from Hy. Python doesn't have them, but since 3.5 has a `math.nan` and `math.inf`. It would save on the extra code from #1294.

We're kind of exploiting an implementation detail of the ast to make these literals, so it's not that surprising that astor can't handle them. (We could kind of say the same of Hy's expanded character set for symbols, but that's used by more than Hy--e.g. by pytest's rewriting of `assert`s.) Our astor-based `hy2py` tool doesn't always produce valid Python because of this kind of thing. If we always mangled symbols to valid Python identifiers, we could improve on this though.

Another option is to request that astor allow the `NaN`s for us, or maintain a tweaked fork that does that. @berkerpeksag care to weigh in on this option? Can you remove the offending assertion?

Fixing Hy at astor 0.5 is not an option long-term. Hy has already fallen behind as Python adds new features. There are already things 0.5 can't render properly, like the extended * unpacking. We need to keep up with them.

### DIFF
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -11,7 +11,7 @@ import sys
 import os
 import importlib
 
-import astor.codegen
+import astor.code_gen
 
 import hy
 
@@ -413,17 +413,17 @@ def hy2py_main():
             else pretty_error(import_buffer_to_ast, stdin_text, module_name))
     if options.with_ast:
         if PY3 and platform.system() == "Windows":
-            _print_for_windows(astor.dump(_ast))
+            _print_for_windows(astor.dump_tree(_ast))
         else:
-            print(astor.dump(_ast))
+            print(astor.dump_tree(_ast))
         print()
         print()
 
     if not options.without_python:
         if PY3 and platform.system() == "Windows":
-            _print_for_windows(astor.codegen.to_source(_ast))
+            _print_for_windows(astor.code_gen.to_source(_ast))
         else:
-            print(astor.codegen.to_source(_ast))
+            print(astor.code_gen.to_source(_ast))
 
     parser.exit(0)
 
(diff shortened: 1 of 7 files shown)


Kodiologist:
I think astor should support NaN, but as a stopgap measure, we can compile NaNs to an expression that evaluates to NaN such as `1e1000 - 1e1000`.

gilch:
> compile NaNs to an expression that evaluates to NaN such as 1e1000 - 1e1000. 

It is not as easy as I expected to make the compiler do that. @Kodiologist I've allowed edits from maintainers if you want to use this as a starting point to try compiling NaN to expression.

Also @olasd tells me that we've had to patch astor before. I'm not sure how to go about that either.

Kodiologist:
> @Kodiologist I've allowed edits from maintainers if you want to use this as a starting point to try compiling NaN to expression.

Sure, I'll take a stab at it.

Kodiologist:
I think I did it. The tricky part was complex numbers.

gilch:
@Kodiologist, Travis is showing a failing test with your update.

gilch:
I think I got it. It wasn't due to NaN. Astor 0.6 added an extra terminating newline to our `disassemble` output for some reason. It's not exactly wrong, so I just updated the tests to expect that.

PR #1454, opened 2017-11-04
fix loop/recur nesting

gilch:
**mo\`** on IRC pointed out that `loop`/`recur` doesn't nest properly. 

Fixing this probably wasn't possible before the upgraded `macroexpand-all` from #1426. `loop` now uses a body pre-expansion strategy like our new `let`.

It was also surprising to **mo\`** that `loop` used extra brackets unlike Clojure's `loop` and our current `let`. This PR fixes both issues.

It also removes a useless extra assignment in the expansion.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -25,6 +25,7 @@ Changes from 0.13.0
    * `xi` from `hy.extra.anaphoric` is now the `#%` tag macro
    * `#%` works on any expression and has a new `&kwargs` parameter `%**`
    * new `doc` macro and `#doc` tag macro
+   * removed redundant brackets in loop/recur
 
    [ Bug Fixes ]
    * Numeric literals are no longer parsed as symbols when followed by a dot
@@ -51,6 +52,7 @@ Changes from 0.13.0
      a for loop
    * `else` clauses in `for` and `while` are recognized more reliably
    * Argument destructuring no longer interferes with function docstrings.
+   * In nested `loop`s, `recur` uses the current loop instead of the outermost.
 
    [ Misc. Improvements ]
    * `read`, `read_str`, and `eval` are exposed and documented as top-level
--- a/docs/contrib/loop.rst
+++ b/docs/contrib/loop.rst
@@ -29,15 +29,34 @@ tail-call optimization (TCO) in their Hy code.
 Macros
 ======
 
+.. _defnr:
+
+defnr
+-----
+Function definition with tail-call optimized ``recur`` anaphor.
+
+Tail recursion via the ``recur`` anaphor instead of the function name will not
+cause stack overflow.
+
+.. _fnr:
+
+fnr
+---
+Function with tail-call optimized ``recur`` anaphor.
+
+Tail recursion via the ``recur`` anaphor will not cause stack overflow.
+
 .. _loop:
 
 loop
 -----
+Tail-call optimized loop/recur macro.
 
-``loop`` establishes a recursion point. With ``loop``, ``recur``
-rebinds the variables set in the recursion point and sends code
-execution back to that recursion point. If ``recur`` is used in a
-non-tail position, an exception is raised.
+``loop`` declares a new function and immediately calls it with the given
+argument bindings--it establishes a recursion point. Within a ``loop``,
+``recur`` rebinds the variables set in the recursion point and continues
+execution from that recursion point, but without a new stack frame.
+If ``recur`` is used in a non-tail position, an exception is raised.
 
 Usage: ``(loop bindings &rest body)``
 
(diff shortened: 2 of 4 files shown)


Kodiologist:
The use of `macroexpand-all` means that `recur` in macro expansions will be replaced with the generated symbol:

    (require [hy.contrib.loop [loop]])

    (defmacro m []
      '(do
        (defn recur [x]
          (print "in m:" x))
        (recur 3)))

    (loop [i 1]
      (cond
        [(= i 1)
          (print "branch 1")
          (m)
          (recur 2)]
        [(= i 2)
          (print "branch 2")]
        [(= i 3)
          (print "branch 3")]))

This prints:

    branch 1
    in m: 3
    in m: 2

Bug or feature?

Kodiologist:
For comparison, if you change back to the old variable-setting syntax and run the example with master, you get:

    branch 1
    in m: 3
    branch 2

gilch:
Feature. That `m` macro is not [hygienic](https://en.wikipedia.org/wiki/Hygienic_macro). 

A hygienic macro would have used a gensym for that function name instead of a symbol like `recur` that could accidentally capture an identifier meant for something else. On the other hand, `loop` is _explicitly_ documented to be anaphoric--the capture of `recur` is part of its contract.

Kodiologist:
In that case, I think `recur` can be a regular function. For example, if you replace the whole of hy/loop.hy with:

    (defclass E [Exception] []
      (defn __init__ [self args]
        (setv self.args args)))

    (defn recur [&rest args]
      (raise (E args)))

    (defmacro/g! loop [bindings &rest body]
      (setv fn-args (cut bindings None None 2)
            init-args (cut bindings 1 None 2))
      `(do
        (import [hy.contrib.loop [E :as ~g!E]])
        (setv ~fn-args ~init-args)
        (while 1
          (try
            (setv ~g!output (do ~@body))
            (except [~g!e ~g!E]
              (setv ~fn-args (. ~g!e args)))
            (else
              (break))))
        ~g!output))

and add `(import [hy.contrib.loop [recur]])` to the tests, all the tests pass except for `test-recur-in-wrong-loc`, which is a feature that was already broken if the TODO comment in the original loop.hy is to be believed. I don't know; it's a thought.

Kodiologist:
Perhaps the best thing to do is, rather than forbidding a non-tail position, explain that putting `recur` there will cause any parent function calls or later statements to be ignored, since `recur` always jumps to the top of the `loop`.

gilch:
That changed the semantics. `recur` is supposed to be a function call. The old loop/recur can shadow variables. Yours clobbers them.

If we add this example to the tests,
```Hy
(loop [x 3]
  (loop [x x]
    (when (pos? x)
      (print x :sep " ")
      (recur (dec x))))
   (when (pos? x)
     (print)
     (recur (dec x))))
```
My version (and Clojure's) prints out
```
3 2 1
2 1
1
```
But yours prints
```
3 2 1
```
Using exceptions is an interesting approach. I had proposed something similar with `label`/`break`. #1055 

I'm not sure about the performance overhead of an exception every loop, but Python function calls have significant overhead too. 

Maybe we could add more looping options like these. The `loop` module seems like a good place for them. But it would be a new PR.

Kodiologist:
Okay, it makes sense that you'd want to preserve that aspect of variable scoping. Why don't you add a test for it? Otherwise, this PR looks good to me.

Kodiologist:
The test looks good, but I have questions about the other commit.

1. Why document and test `fnr` and `defnr`? Aren't they just part of the implementation of `loop`?
2. The documentation file says "If ``recur`` is used in a non-tail position, an exception is raised." The docstring of `loop` says "If recur is used in a non-tail-call position, None is returned, which causes chaos. Fixing this to detect if recur is in a tail-call position and erroring if not is a giant TODO." Which is correct?

gilch:
1. They turn out to be useful in their own right. In the fairly common case that a loop takes up the whole function body, it's easier to tack an `r` onto the `defn` (or `fn`) than to add a `loop` form. I had originally deleted `defnr` altogether. It was part of the old implementation, but turned out not to be required, and only added a useless assignment in the expansion.

2. I'm not sure if either is correct. The doc is the original intent of the author, but the implementation didn't quite meet the spec before being abandoned. I've also found deeper flaws, e.g. 
```
(loop [] (print 'recur))
```
prints a gensym.

I might have to rewrite the whole thing from scratch.

Kodiologist:
1. Okay.

2. I see. If you don't want to do the full rewrite for this PR, you can just remove the claim that this feature works from the documentation.

Kodiologist:
Although, if you care about tail-recursion optimization, your time might be better spent getting [the automatic version](https://github.com/hylang/hy/pull/1166) to work, rather than working on `hy.contrib.loop`. On the other hand, people who start work on that have historically not been able to finish it, so I guess it's harder than it sounds.

gilch:
I think this is not the right approach, so I'm closing it. But the issue still needs a fix #1465

Issue #1341, opened 2017-07-25
Add more recent repl history globals


gilch:
Python and Hy have `_` bound to the last result in the repl. And I think that's it, or at least it's not easy to find documentation. But a good repl usually has more than this.

IPython has `_` and also has `__` and `___` for the two results before that.
Clojure calls them `*1`, `*2`, and `*3`. It also has `*e` for the last exception.
(Common Lisp has them too, but calls them `*`, `**`, and `***`, which wouldn't be compatible with Python.)

Some repls also save the most recent *input*. IPython saves all inputs strings, but calls the last three `_i`, `_ii`, and `_iii`. (Common Lisp saves input *forms* and calls them `+`, `++`, `+++`, again, not compatible.)

Did we set up the `_` ourselves, or did that come with something else? If we did it, perhaps we could change it to be like Clojure with `*1`, `*2`, `*3` and `*e`.

Hy's `eval` works on *forms*, not strings, so that's what the input globals should store. I'm not sure what to name them though. Clojure's basic repl doesn't seem to have that feature, but if there's some other standard, we should consider that.

tuturto:
That's by us (https://github.com/hylang/hy/blob/7c53a07b93084a0970cdf192d4aa2d712d557154/hy/cmdline.py#L116 for relevant spot). I like the idea of ``*1``, ``*2`` and so on. Three should be enough I suppose?

Kodiologist:
`_` is now `*1` as of #1517.

waigx:
PSA: I will grab this one, please let me know if you have any concern.

PR #1458, opened 2017-11-16
Mangling makeover

### LINKED ISSUE #360: Find a way to keep the "original" symbol before the interop munging
``` clj
=> 'foo?
u'is_foo'
```

I think it'd be worthwhile to keep the original symbol, as it could be useful in pure hy stuff (typically printing quoted Hy expressions).

One way to do that would be to move all the munging to `HySymbol`'s `__init__`, and keep the original as an attribute.

### LINKED ISSUE #970: add option to hy2py that converts symbols to valid python identifiers
Hy would be more useful if it could compile to Python, and not just ast, see #923.

`hy2py` can translate ast back into Python source. It's very useful as a debugging tool, but the result is not always executable. The main issue seems to be that Hy identifiers are a superset of valid Python identifiers. The ast doesn't care, but when importing Hy from Python we have a system to resolve this problem using Punycode and a few other rules (`-` to `_` etc.)

I'm proposing we add an option to `hy2py` that converts all invalid symbols using the same system. This way the Python code will be still be executable. I'm not sure if there are other problems with the output, but it's a step in the right direction.

This might also be more useful as the default for `hy --spy`.

### LINKED ISSUE #1115: clean up rules for `?`

### LINKED ISSUE #1168: Should keywords be mangled?

### LINKED ISSUE #1173: Rethink mangling

### LINKED ISSUE #1336: add mangle function to public api
This only exists internally in the Python, but it's important for writing macros. Clojure, for example, [exposes `munge` in its core](https://clojuredocs.org/clojure.core/munge) namespace.

Kodiologist:
- Closes #360
- Closes #970 (no longer necessary)
- Closes #1115 (debatably)
- Closes #1168 (keywords are now indeed mangled when used for keyword arguments)
- Closes #1173 (or at least, changes the situation drastically)

This pull request overhauls mangling such that HySymbols remember their original names (it is no longer the case that `(= 'a-b 'a_b)`) and all Python variables have names that are legal Python identifiers. The latter accomplishes most of what we need to ensure that `hy2py` produces real Python code.

The major changes are:

- Mangling happens at compile-time, not read-time.
- Mangling rules are different.
  - `*foo*` → `FOO` and `foo!` → `foo_bang` are gone.
    - The first one seems misleading, and the second rarely useful.
  - The Punycode transformation (which was only enabled for Python 2 anyway) is gone.
  - `foo-bar` → `foo_bar` and `foo?` → `is_foo` remain.
  - Names that still aren't legal Python identifiers (either because they contain illegal characters or because they coincide with a Python reserved word) are prepended with `hyx_`, and each illegal character is replaced with `ΔfooΔ` (or for Python 2, `XfooX`), where `foo` is the Unicode character name in lowercase with spaces replaced by underscores and hyphens replaced by "H"es. If the character doesn't have a name, we use "U" followed by its code point in lowercase hexadecimal.
  - Any added `hyx_` or `is_` is added after any leading underscores (because leading underscores can have magical effects in Python).
- The names of keyword arguments are fully mangled.

The below things remain to be done, but I figured I'd get feeback before continuing.

- [ ] Watch out if the REPL variable `_` is confused with the shadow subtraction operator `-`
- [ ] In the unmangle function, undo `hyx_` when necessary; otherwise, the distinct variables "hyx_X61X" and "a" will both unmangle to "a"
- [ ] Perhaps make the mangle function stringify the input, so the caller doesn't have to
- [ ] Rename the mangling functions `mangle` and `unmangle` and add them to core (fixes #1336)
  - [ ] Testing
  - [ ] Documentation
- [ ] Document mangling in general
- [ ] Use `unmangle` instead of `hyify` in core
- [ ] NEWS


Kodiologist:
By the way, some of the underscore-to-hyphen edits in "Mangle symbols at compile-time instead of parse-time", particularly in `@build(...)` expressions, can be undone if desired, thanks to the later commit that mangles the names of special forms.

gilch:
I tried a few symbols, and I think legibility is a problem, especially for gensyms. I thought about how we could change how we construct the gensym names and realized that it's better to change the mangling system.

For example, the gensym `Hy, world` mangles as
`_hyx_X3bXHyX2cXX20XworldX7cX1240`.

But I think we could instead mangle it to something like
`_hyx_3B2C207C_XHyXXworldX1239`
which is much more readable. Here I've moved all the hexidecimal codes to the front, and then replaced the invalid characters with a single `X` each. The unmangle function could put them back together like this:
```Python
>>> from base64 import b16decode
>>> '_' + 'XHyXXworldX1239'.replace('X','{}').format(*b16decode('3B2C207C').decode('utf-8'))
'_;Hy, world|1239'
```

gilch:
Actually, it might be even better to use `_` as the elision marker instead of `X`. e.g.
`_hyx_3B2C207C__Hy__world_1239`.
An actual `_` would have a corresponding `5F` in the hex codes.

gilch:
I'm not sure if the `?` to `is_` mangling is worth keeping either, but I'd like to hear what the others think.

We could use `*1 *2 *3` as the repl history variables like Clojure instead of `_`. #1341 

But I am worried about the ambiguity introduced by treating `-` and `_` as the same sometimes, but not other times. It's not reversible. If we don't give `-` special treatment and use my suggested format above, `foo-bar` would render as `hyx_2d_foo_bar`, which is quite legible, but not something you'd want to import in Python. Perhaps when producing things for Python's use you simply write `foo_bar` instead.

Another possibility is to swap `-` with `_` before mangling, and swap them back when unmangling. But this means you must write magic dunder/private names with `-`, which is easily confused with negation. See #1062 for a discussion of why `_` is preferable as a prefix. Perhaps leading and trailing `_` could be special cased (like `?` is now) to avoid this issue but still be reversible.

Kodiologist:
> I thought about how we could change how we construct the gensym names and realized that it's better to change the mangling system.

I do see one way it could make sense to change the gensym format: instead of `_;foobar|1235`, use `_;foobar_1235`. I see why you want the leading underscore (to prevent automatic import) and the `;` (to create an illegal Hy identifier), but you don't need the `|`, right? Then this mangles to `_hyx_X3bXfoobar_1235`, which seems pretty reasonable.

We could also have the `;` do double duty as the digits separator, yielding `_foobar;1235`, which mangles to `_hyx_foobarX3bX1235`.

I think your suggestion of putting Base16 UTF-8 at the beginning works, too, even if we also make one of the aforementioned changes to the gensym format.

> Actually, it might be even better to use `_` as the elision marker instead of `X`.

But `_` is a common identifier character, whereas `X` is uncommon. 

> We could use `*1 *2 *3` as the repl history variables like Clojure instead of `_`.

That sounds reasonable.

gilch:
> But `_` is a common identifier character, whereas X is uncommon.

True, but why does that make X a better choice?

Underscores are more legible than X's.

I use the astor output from `hy --spy` all the time and I need to be able to read the identifiers. It's the one non-alphanumeric character allowed in Python2 identifiers. In the common case, you don't have to encode it anyway, and when you do, it still shows up as an underscore in the part you actually read!

gilch:
> it could make sense to change the gensym

I picked the current gensym format for legibility when debugging macro output with astor. I'd certainly consider improvements to it after we settle the encoding format. But what's best for gensyms depends on that, so let's figure it out later.

>  undo hyx_ when necessary

I think it makes sense to treat a symbol containing a `hyx__` prefix like an illegal identifier. That is, encode it again. It's just like double-escaping backslashes. So `hyx__foo` might be encoded as `hyx__hyx__foo`. Make it a 1:1 invertable function, so there would always be an unambiguous unmangling for any identifier.

Kodiologist:
> True, but why does that make X a better choice?

1. Because it means fewer characters have to be escaped. So, `some_l∞ng_identifier` becomes `hyx_E2889E_some_lXng_identifier` instead of `hyx_5FE2889E5F_some_l_ng_identifier`.

2. Because when you see an X in a `hyx_…` symbol, you can be pretty sure it's something else, whereas a `_` is as likely to be a `_` as something else.

Also, not related to the rarity of X, it's easier to count Xes than underscores. (This isn't important for my original proposal, where you never have more than two Xes (or rather, undercoress) in a row, but could become a pain point with your Base16 UTF-8 method.)

> I picked the current gensym format for legibility when debugging macro output with astor. I'd certainly consider improvements to it after we settle the encoding format. But what's best for gensyms depends on that, so let's figure it out later.

Okay, sure.

> I think it makes sense to treat a symbol containing a `hyx__` prefix like an illegal identifier. That is, encode it again.

That was my original plan, but I realized it wouldn't allow you to use the unmangled form in Hy. That is, you can write `foo?` as `is_foo` or `foo-bar` as `foo_bar`, but you wouldn't be able to write an identifier with illegal characters in its pure-ASCII `hyx_` form.

> Make it a 1:1 invertable function, so there would always be an unambiguous unmangling for any identifier.

That would be nice, but we can't have it so long as `foo-bar` and `foo_bar` in Hy both get mangled to `foo_bar`. Which I like and want to keep.

gilch:
> 1. Because it means fewer characters have to be escaped. So, `some_l∞ng_identifier` becomes `hyx_E2889E_some_lXng_identifier` instead of `hyx_5FE2889E5F_some_l_ng_identifier`.

Are you concerned about the greater length when using a common character? 

Because of runtime performance? Identifiers are interned in Python. Length has no effect. 

Because it causes lines to wrap? That's kind of unavoidable with astor, but we could use base64 instead. I preferred base16 because it's easier to lookup the Unicode by hand if necessary, but I don't feel too strongly about it, since we're providing an unmangle function.

Because it's harder to type from Python? Exports to Python should not use such names, but again base64 is an option.

> 2. Because when you see an X in a `hyx_…` symbol, you can be pretty sure it's something else, whereas a `_` is as likely to be a `_` as something else.

And you'd still have to check for a `58` to be sure it's not really an `X`. It's true that `X` is a rare letter in English, but it's a common letter in mathematics, so I'm not sure if it's as uncommon as you think it is in code.

But if that's your concern, wouldn't a rare Unicode character be a better choice? Python2 requries pure ASCII, but we're using different mangling rules for Python3 anyway. To be legible, it ought to be punctuation mark rather than a letter (or at least look that way), but valid in a Python3 identifier. Perhaps U+2054 `⁔`. These are also easier to count than underscores.

Maybe you want Python3 to use pure ASCII too for some reason? I can kind of understand not wanting force anyone to use Unicode in the command-line dubugger. But you can copy/paste identifiers or even the elision character if you're constructing them. You can even use strings and type out `\u2054` in Python.

I'm mainly concerned about losing my readable astor output for debugging.

We could instead avoid forcing Unicode on those who don't use it, but still use it as the elision character. At least for ASCII, we could replace invalid characters with a more mnemonic underscore-quoted substring via lookup table, e.g. `+` with `_plus_` or `_add_` or `_t_`; ` `(space)/`__` or even `_sp_; `?`/`_eh_` or `_qMark_`, etc. Clojure does something similar to make valid Java identifiers. There may be some tradeoff between obviousness and length, but this is much more readable than meaningless hex codes.

Suppose you wanted an increment function named `X+1` and an inverse function named `X-1`. Under the proposals they might look like
* `X+1`/`X-1` (status quo output, but invalid Python)
* `hyx_X58XX2bX1`/`hyx_X58XX2dX1` (X quotes)
* `hyx_582B_XX1`/`hyx_582D_XX1` (X elision)
* `hyx_2B_X_1`/`hyx_2D_X_2` (_ elision)
* `hyx_2B_X⁔1`/`hyx_2D_X⁔2` (⁔ elision)
* `hyx_X_add_1`/`hyx_X_sub_1` (mnemonic lookup table)
Only in the first and last case is the meaning clear, but elision seems easier to read than quotes, and `X` is the least readable of the elision characters.

Perhaps we'd use `⁔` only if there's a non-ASCII invalid character. But the higher Unicode would still have to use the hex codes, since a lookup table that big is probably not worth the trouble. But maybe we could write a program to convert the official Unicode names somehow, e.g. `foo⁔bar` becomes `hyx_foo_InvertedUndertie_bar`. Some of the names are pretty long though.

Kodiologist:
I agree that hexadecimal is a better choice than Base64 for this, whether we're using a quoting method or an elision method.

> It's true that `X` is a rare letter in English, but it's a common letter in mathematics

I don't think it's common as an uppercase letter in a Python identifier, because of how uppercase tends to be used. It also happens to fit with the use of `\x` to quote a character by its hex code.

Using a non-ASCII character is a pretty good idea, though. I don't think we should use `⁔`, because it seems to be some sort of combining character (in Emacs, it's being influenced by characters that I put around it in weird ways), but what about the Greek letter `Δ`?

Using Unicode character names is something that occurred to me but I didn't think about it much. How about, per your implicit suggestion, illegal characters get replaced with a CamelCase version of the name? So the identifier `#` becomes `hyx_ΔNumberSignΔ`, `+` becomes `hyx_ΔPlusSignΔ`, and `some-l∞ng-identifier` becomes `hyx_some_lΔInfinityΔng_identifier`? For a bit of brevity, we could use another `Δ` at the beginning instead of `hyx_`. Thus, `+` becomes `ΔΔPlusSignΔ`.

gilch:
I like where this is going. Using the Unicode names is much more legible in most cases.

Yes, combining characters could be trouble, but I'm not sure if `⁔` is one or not. I like the choice of delta better than X, but perhaps we could find a better one. It makes Greek variable names awkward anyway. A variable named `Δ` would mangle to `ΔΔGreekCapitalLetterDeltaΔ`, I suppose.

Python's `unicodedata` module has both a `name` function and its inverse `lookup`. So we wouldn't have to make our own Unicode table. But not all characters have names. We could make our own names table for the missing ASCII (the first 32 control characters), and fall back to the base16 representation for any nameless higher Unicode.

But Unicode names have spaces, which isn't allowed in identifiers. I figured that the simplest reversible transformation is `.replace(' ', '_')`, which gives e.g. `PLUS_SIGN`. The problem is names with hyphens, like  `LESS-THAN SIGN`. Camel case is also reversible, since Unicode names are all upper case. I had only suggested that so we could use `_` quotes, but we need it for the hyphen. So `<` would probably be `Less_thanSign`. I hope Unicode names don't have any other special characters. We should write a test enumerating all Unicode to make sure everything round-trips anyway though.

Kodiologist:
Oh, gross, I didn't realize that some Unicode characters don't have names. There are aliases, like "NULL" for U+0000, but private-use characters don't even have those. If we use hexadecimal for some characters, then we need a way to distinguish the hex codes from character names. I guess we can use a leading lowercase `x`, because names will always start with a capital letter.

> A variable named `Δ` would mangle to `ΔΔGreekCapitalLetterDeltaΔ`, I suppose.

In Python 3, it's a legal identifier, so it doesn't need mangling.

Speaking of which, Python 2 only allows ASCII in identifiers, so we can't use a non-ASCII escape character. I guess we'll fall back on "X" then, unless you still want to use "_", which would produce things like `hyx_LowLine_some_LowLine_l_Infinity_ng_LowLine_identifier`. 

I've checked that for every Unicode character with a name, the name matches `r"\A[-A-Z0-9 ]+\Z"`. So, are we good with Δ delimiters for Python 3, X delimiters for Python 2, names for characters with names, `x` followed by hex digits for characters without names, and hyphens in names replaced with underscores?

gilch:
Yeah let's go with that for now. I'm not sure of some of the details, but they should be more obvious when the actual code is available to test.

gilch:
Do the Unicode names match `r"\A[-A-Z ]+\Z"` too? That means we could replace `-` with a numeral. We could also count the empty string as `_` as a special case. 
* `hyx__LowLine_some_LowLine_l_Infinity_ng_LowLine_identifier` (_ quotes)
* `hyx__some__l_Infinity_ng__identifier` (empty is `_`)
* `hyx_some_lXInfinityXng_identifier` (X quotes)

`<foo_X>`
* `hyx__Less1ThanSign_foo__X_Greater1ThanSign_` (empty is `_` and `-` is `1`)
* `hyx__Less0ThanSign_foo__X_Greater0ThanSign_` (empty is `_` and `-` is `0`)
* `hyx_XLess_ThanSign_foo_XLatinCapitalLetterXX_Greater_ThanSign_` (X quotes)

The last one doesn't round trip! It worked when using hex codes, since X isn't in `[a-f0-9]`. We can't do X quotes for Python2 like that.

Kodiologist:
> Do the Unicode names match `r"\A[-A-Z ]+\Z"` too? 

Surprisingly, no. Even though digits don't appear in the names of the ASCII digits, they show up in the names of some higher characters.

> The last one doesn't round trip!

Ah, true. X quotes can't work for Python 2 if we're using character names. Nor can we use underscores, since we need those for hyphens in character names. That seems to leave us with my original plan of using hex for every character. I'd rather just use that for Python 3, for simplicity, but if you wouldn't accept that, and you would accept that for Python 2 and delta quoting for Python 3, I'll add delta quoting for Python 3. What do you say?

gilch:
I think we could special-case the empty string for `X`, so `<foo_X> becomes

`hyx_XLess_ThanSign_foo_XX_Greater_ThanSign_`

We could do that for Delta too.

Kodiologist:
But we'd still have the problem of `X` in other names, right? Like U+2423 OPEN BOX, ␣.

gilch:
> OPEN BOX

Ah, good point. Python2 is a pain.

We could try not using camel case. Convert to lower case, replace space with `_`, and `-` with `__`. This could only fail if a Unicode name contains two spaces in a row, which I don't think would ever happen, or if a Unicode name contains space-hyphen or hyphen-space sequence, which might.
`hyx_Xless__than_signXfoo_Xlatin_capital_letter_xXXgreater__than_signX`

If that's a concern, as above, but instead replace `-` with `H` (for "Hyphen").
`hyx_XlessHthan_signXfoo_Xlatin_capital_letter_xXXgreaterHthan_signX`
We converted the name to lower case, so `H` can't appear in the name, and neither can `X`. And `H` conveniently contains the horizontal line just like a hyphen, so it's pretty easy to read.

There are probably other solutions, but what do you think of this?

Kodiologist:
> Python2 is a pain. 

It always is. ;_;

I was 95% sure that space-hyphen and hyphen-space wouldn't occur, but it turns out that a handful of Tibetan letters have them (e.g., U+0F0A, "TIBETAN MARK BKA- SHOG YIG MGO"). So I think we have to go with `H` for hyphens. Then the only way that Python 2's and Python 3's mangling rules will need to be different, other than the fact that Python 2 needs more things escaped to begin with, is that Python 2 will use `X` and Python 3 will use `Δ`. Sound good?

gilch:
Pretty much, but I still want to test the resulting code before final approval.

Also what about the special case `XX` and `ΔΔ` to mean `X` and `Δ`? Otherwise we have to spell out `Xlatin_capital_letter_xX` and `Δgreek_capital_letter_deltaΔ` when there are any other illegal characters.

It seems like a bigger problem for Python2 or Greek speakers. `X` quotes does seem like the best of bad options for Python2. But I'm not sure about the choice of `Δ`. It's legible (at least when you're not writing Greek) and that seems easy to change if someone has a better idea.

Kodiologist:
> I still want to test the resulting code before final approval

Yes, that goes without saying.

> Also what about the special case `XX` and `ΔΔ` to mean `X` and `Δ`?

They're both pretty rare, and it's only an issue when other characters have to be escaped, so the additional mangling rule doesn't seem worth it to me.

Kodiologist:
Rebased. See the edited PR description for the new rules. I used `U` instead of `X` for hexadecimal code points to make parsing easier in the case of Python 2.

vodik:
Looks good to me.

And I don't think the `is_` mangling is worth keeping because Python isn't consistent with `is_foo` or `isfoo`

PR #1464, opened 2017-12-09
Update interop doc: print function is lower-case

MonsieurV:
(A typo made use of Print function, which is not defined)


Kodiologist:
Welcome aboard!

MonsieurV:
Can't be more reactive! Thanks

Just beginning to use Hy. The doc is clear, it is easy to progressively introduce it in a project: that's great!

PR #1468, opened 2017-12-17
Fancier `for` and comprehensions

### LINKED ISSUE #588: list-comp only iterates over the last expression when it is passed a do

Kodiologist:
- Closes #588.

This is a new version of #1437 following the discussion there. `:while` is not yet implemented. ~~There's one `macroexpand-all`-related test failure that I'm hoping @gilch can help with.~~ Never mind, it was much simpler than I expected.

To do:
- [ ] flake
- [ ] Test that variables are scoped properly
- [ ] Clean up test organization
- [ ] Replace uses of the old forms with the new forms
- [ ] Get rid of the old forms
- [ ] Documentation
- [ ] NEWS


Kodiologist:
@gilch What do you think?

gilch:
We're going to want async for comprehensions now too, and syntax to go with it. Perhaps an `:async` control word would work.

But then, to be consistent, wouldn't `(for/a [x (xs)] (foo))` be written like `(for [:async x (xs)] (foo))` instead? There should be only one obvious way to do it.

kirbyfan64:
Personally, I prefer `for/a`, because it's actually doing something entirely different...

Kodiologist:
`for/a` seems like the simplest option. I imagine it would make all the nested loops asynchronous as well, but to be honest, I don't use the asynchronous stuff nearly enough to guess whether that's the desirable behavior.

Kodiologist:
@gilch What do you think of that, and what I've done so far?

gilch:
I don't use asynchronous stuff much either. But my understanding is that nesting an async for inside a for or vice-versa is fairly common. Making them all async or not seems overly restrictive.

Kodiologist:
You can always nest constructs manually. I'm not sure what the alternative to `for/a` is unless you want to allow inserting `:async` in various positions, complicating the already somewhat complicated rules by which elements are grouped in these constructs, and creating an asymmetry with the other `/a` forms. Or, let's forget about it for now, since neither of us actually make much use of this feature and this PR is ambitious enough already.

vodik:
I don't mind filling it in one this lands. `await` inside comprehensions should work - I've used it with `list-comp` - a special form is only necessary for working with `__aiter__`.

vodik:
@gilch I think maintaining a separate syntax for now makes sense, as they are different underlying mechanisms with different-ish uses.

I personally interpret having multiple for expressions rolled together as effectively the same thing as an implicit `itertools.product`.

But that analogy kinda breaks down with async code. `async for` typically ends up being used more like a message or event bus (e.g. notify me as background tasks finish, as websocket messages come in), and I'd go so far to argue having multiple async for expressions rolled together is actually kinda weird conceptually and probably won't be of much practical use.

vodik:
I got async variants working on a dev branch, so this works for me.

hy/compiler.py
```diff
@@ -1421,6 +1443,168 @@ def compile_dict_comprehension(self, expr):
...
+                elif p[0] == "expr":
+                    return p[1] + p[1].expr_as_stmt() + f(parts)
+                else:
+                    raise ValueError("can't happen")
+            if is_for:
+                ret = f(parts)
+                ret.contains_yield = bool(contains_yield)
+                return ret
+            fname = self.get_anon_var()
+            ret = Result() + asty.FunctionDef(
```
vodik: Won't this possibly cause some NameErrors? I see you do this in examples:

```clojure
(lfor x "abc" :do (setv y 1) x))
```

Is `y` expected to be accessible to the outer scope? In Python it would be, but because of the function call added here, it won't be.

Kodiologist: None of the three types of comprehensions, nor generator expressions, leak their variables into the surrounding scope in Python 2 and Python 3, with one exception: list comprehensions in Python 2. This seems more like an artifact than a feature, so I didn't try to retain it.

vodik: Right, I see you point. The `for` and emitted ast threw me off. Thanks.

gilch:
I have to preemptively veto this one in light of the new merge policy. (Though to be clear, that policy only applies to newly opened issues, right?) I think this adds too much complexity to the compiler, which will force even more complexity on the already complicated `let` macro. We should be simplifying the compiler, but this is a step in the wrong direction.

#588 is still a bug so the status quo is unacceptable. I've been pondering alternatives that would let us move more of the logic out of the compiler and into macros. I think I've come up with something workable, but I'll need time to code it.

Kodiologist:
> Though to be clear, that policy only applies to newly opened issues, right?

Yes, it would only apply for these changes if I closed this PR and opened a new one for them.

Kodiologist:
I'm closing this in favor of updating #1437.

hy/compiler.py
```diff
@@ -370,19 +386,25 @@ def checker(self, expression):
     return _dec
 
 
+def eq_sym(x, sym):
+    return isinstance(x, HySymbol) and x == HySymbol(sym)
+
+
+def eq_kw(x, kw):
+    return isinstance(x, HyKeyword) and x == HyKeyword(kw)
```
vodik: You shouldn't need this, `HyKeyword.__eq__` should do the trick, unless I did something wrong.

vodik:
Well, you confused me for a bit as I didn't realize the last comment on both issues link to each other :wink:

asemic-horizon:
> We're going to want `async for` comprehensions now too, and syntax to go with it. Perhaps an `:async` control word would work.
> 
> But then, to be consistent, wouldn't `(for/a [x (xs)] (foo))` be written like `(for [:async x (xs)] (foo))` instead? There should be only one obvious way to do it.

Was a syntax for `async for` ever selected?

Kodiologist:
Yes. http://docs.hylang.org/en/stable/language/api.html#lfor

PR #1471, opened 2017-12-29
Initial commit of asyncfn/await support

vodik:
Adds basic support for defining coroutines and await expressions.

Putting this up for review as I'm sure there might be plenty of nitpicks how I did things. But once people are happy with it, I can proceed to add some test cases and anything else necessary.


vodik:
And I overlooked the lack of `AsyncFunctionDef` missing in older versions of python, I'll dig into this.

Kodiologist:
Why don't you write some tests first? That will make it plainer what you intend to implement.

vodik:
I can certainly get on that. I was more concerned if there would be any nitpicking over `asyncfn`.

Kodiologist:
There might, but it's not so important as the feature itself, and that's what mass search-and-replace is for, right?

kirbyfan64:
Agree with @Kodiologist, but anyway...

`asyncdefn` just feels wrong. What about an `a` abbreviation? (e.g. `adefn`).

Personally I always felt like something like this would be cool:

```hy
(fn async [a b] ...)
(defn async [] ...)
```

I mean, Hy's Lisp, so might as well take advantage of the loose syntax, right?

vodik:
Okay, tests pass, and added to basic tests to make sure it works, but I don't really know the depth at which I should be going for.

@kirbyfan64 I like the look of that, but I guess the issue is how well will that generalize. Technically there's also `async with`, `async for` and async generators to worry about too.

vodik:
Oh, didn't mean to push it, but I'll leave it up I guess then, I added a tag macro for await because I found myself writing a lot of `(->> ... await)` and found this a little cleaner.

vodik:
How about `fn/a` and `defn/a`?

Kodiologist:
This looks like a decent start. Remember to make name changes consistent (e.g., you renamed `asyncdefn` to `defn/a` but there are still appearances of "asyncdefn" in error messages and names).

hy/compiler.py
```diff
@@ -1139,6 +1141,14 @@ def compile_yield_expression(self, expr):
         node = asty.Yield if expr[0] == "yield" else asty.YieldFrom
         return ret + node(expr, value=ret.force_expr)
 
+    @builds("await", iff=PY35)
+    @checkargs(max=1)
```
Kodiologist: `(await)` should be illegal, since `await` needs an expression in Python.

hy/compiler.py
```diff
@@ -1139,6 +1141,14 @@ def compile_yield_expression(self, expr):
...
+    def compile_await_expression(self, expr):
+        ret = Result(contains_yield=(not PY3))
```
Kodiologist: Maintaining `contains_yield` is not necessary here, since the method can only trigger in Python 3.

tests/native_tests/py35_only_tests.hy
```diff
@@ -24,3 +26,23 @@
   (assert (= {1 "x" #**d1 #**d2 2 "y"} {"a" 1 "b" 2 "c" 3 "d" 4 1 "x" 2 "y"}))
   (defn fun [&optional a b c d e f] [a b c d e f])
   (assert (= (fun #**d1 :e "eee" #**d2) [1 2 3 4 "eee" None])))
+
+
+(defn run-coroutine [coro]
+  "Run a coroutine until its done in the default event loop."""
+  (-> (get-event-loop)
+      (.run-until-complete (coro))))
+
```
Kodiologist: `->` isn't necessary; this can be written `(.run-until-complete (get-event-loop) (coro))`.

tests/native_tests/py35_only_tests.hy
```diff
@@ -24,3 +26,23 @@
...
+
+(defn test-fn/a []
+  "NATIVE: testing fn/a"
```
Kodiologist: These sorts of comments are not necessary. I think they exist because they were somehow desirable when we were using our previous nose-based testing infrastructure.

Kodiologist:
I would vote for naming things by putting `async-` in the front, so for example Python's `async def` becomes `async-defn`.

hy/compiler.py
vodik: How do I do that? I based this block on `yield`/`yield-from`, so wouldn't they share that problem then?

kirbyfan64:
Personally I love the /a proposal. 🤷

vodik:
I'd second that, as, looking again into hy and where its going, the async stuff feels more like variants than its own set of things.

For example, looking at #1468, and assuming it lands, would we rather have `async-lfor` or `lfor/a` for PEP 530 coverage?

hy/compiler.py
Kodiologist: Write `@checkargs(1)` instead of `@checkargs(max=1)` and simplify the method body accordingly. In Python, nullary `yield` is equivalent to `yield None`, so `(yield)` should work similarly. On the other hand, the fact that `(yield-from)` is allowed is apparently a bug; `yield from None` will never work.

vodik: Maybe I should adjust this so that `yield` is its own block with `await` and `yield from` sharing their implementation instead?

Kodiologist: That would be fine, I think.

vodik:
Sorry, I keep tripping up that normalization happens before the compiling...

vodik:
I've been kicking around async with support, and got something working:

```hy
(import [asyncio [get-event-loop]]
        [aiohttp [ClientSession]])


(defn/a query [url]
  (with/a [session (ClientSession)
           resp (.get session url)]
      #a(.text resp)))


(defmain [&rest args]
  (->> (query "https://google.com")
       (.run-until-complete (get-event-loop))
       print))
```

Still need to write tests for it, so not pushing it just yet. I went back to using the `/a` suffix for this which I'm liking more and more as I keep adding support. And to reiterate myself, I think it'll also be cleaner when looking to add comprehension/for support.

Kodiologist:
Okay. I don't have a strong opinion about the name.

Don't forget to simplify `compile_yield_from_or_await_expression`; it is now guaranteed by `@checkargs` that `len(expr) > 1`. 

Include a test for #1472, since you're fixing it.

I don't think `#a` should be a core macro. It has a desirable name and it doesn't save many characters.

vodik:
@Kodiologist how would you have me writing a test, since I can't see any examples where we test for syntax errors.

vodik:
Fixed things up, added `with/a` and `for/a` syntax.

Kodiologist:
Try to avoid duplicated code, like the implementations of `for` and `for/a`. In that case, for example, you can use a function that each macro calls as a subroutine.

> I can't see any examples where we test for syntax errors.

Take a look at `tests/compilers/test_ast.py`.

vodik:
@Kodiologist got a naming convention for private macros? Or is there a better way to do what I just did?

Kodiologist:
Make the subroutine (`_for` or `_with`) a function rather than a macro, calling it at macro-expansion-time. Then it doesn't have to be added to the scope of the calling module.

vodik:
@Kodiologist not sure how to make that work, but I'm not super competent with macros just yet. I got into this to learn how this all works.

Kodiologist:
Do you want help, or to keep working at it for now?

vodik:
I would really appreciate some help

Kodiologist:
Sure, take a look at what I just pushed. It's the sort of thing that's hard to understand from a description but probably easier to get an intuition for once you've seen it done.

vodik:
Ahh, thanks. Looks like I was close then (but I don't think I would have solved it quickly) - I think I understand. The main problem I was hitting was forcing `_with` to resolve at expansion time.

Anything else needing to be done? I would like to look at comprehension support, but I don't think it makes sense until the other PRs are in.

Kodiologist:
And documentation. That's the annoying one. But since you're only providing Python features, you need not do much more than link to the appropriate parts of the Python documentation. The current Hy documentation often re-explains Python concepts, which I believe is a mistake, based on a mistaken idea we had earlier in Hy's development that you could master Hy without mastering Python semantics.

asemic-horizon:
I know I'm pretty dense, but what is the Hy syntax for "await", then?

Kodiologist:
Just `(await FORM)`. It looks like it was never documented. Whoops.

asemic-horizon:
Thanks!

I'm still getting used to Hy, but I hope to be able to contribute soon, even if at the level of improving the docs.

PR #1473, opened 2018-01-01
Update copyright years

Kodiologist:


rkday:
Do we need this? https://softwareengineering.stackexchange.com/questions/130478/what-copyright-date-for-an-update-to-an-open-source-project-from-last-year suggests year of first publication is all that's required.

gilch:
IANAL. Github is accessible worldwide and Hy has contributors from multiple jurisdictions. But in the US, at least, you are not required to declare copyright. The git repository is a sufficient record of authorship. This probably matters less than we think it does.

Kodiologist:
For the record, I'd still prefer not to have these notices at all (#1225), but everybody else seemed to want them. If you want to further shorten the notices to not even include a year, I would support that.

However, even if we remove dates from all the source files, there's still a date in the license file. I presume we wouldn't want to remove that because these licenses are written and approved by lawyers and the lawyers presumably know what they're doing.

Since the files frequently change, new versions are repeatedly published for the first time. I'd rather bump all the years in one commit than remember to do it every time I change an existing file.

Kodiologist:
@hylang/core C'mon, folks, let's merge this and then I'll put the next release together.

rkday:
👍

PR #1474, opened 2018-01-02
Safer returns

vodik:
Adds a sanity check to make certain forms no longer compile:

```
=> (fn [] return)

  (fn [] return)
  ^------------^
HyTypeError: Can't return a keyword: `return'
```

Without this patch, Hy will generate invalid Python (`lambda : return` in the case above).

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -64,9 +64,16 @@ def load_stdlib():
 # True, False and None included here since they
 # are assignable in Python 2.* but become
 # keywords in Python 3.*
+_keywords = set(keyword.kwlist)
+if PY3:
+    _keywords = _keywords.difference(('True', 'False', 'None'))
+
+_is_hy_reserved = frozenset(_keywords).__contains__
+
+
 def _is_hy_builtin(name, module_name):
     extras = ['True', 'False', 'None']
-    if name in extras or keyword.iskeyword(name):
+    if name in extras or _is_hy_reserved(name):
         return True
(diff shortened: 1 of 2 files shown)


kirbyfan64:
Actually, it generates invalid Python *syntax*, but the *AST* runs perfectly and treats `return` as an identifier.

vodik:
Yeah, but I still think its worth addressing since it isn't really meaningful, catches a possible error as compile time more clearly, and we already validate names on assignment.

vodik:
I guess I should probably remove "ast" from the test name, you're right, its valid AST.

Kodiologist:
I disagree with this change. I think `(fn [] return)` should remain legal, and in fact, `(setv return 1)` should be legal as well. My PR #1458 makes the latter change as well as mangling such variable names so as to produce valid Python code.

vodik:
@Kodiologist valid code as in it emits a return or returns a manged "return" variable? I'm taking it as you mean the later. In which case, yeah, that's a much better fix. Retracting this.

gilch:
Can we assign to our other special form names now? With #1458? Let's be consistent here.

Kodiologist:
> @Kodiologist valid code as in it emits a return or returns a manged "return" variable? I'm taking it as you mean the later.

Right.

> Can we assign to our other special form names now? With #1458?

Yep. Mangling occurs not when you use the name of a special form, but when you use the name of a Python reserved word (other than `True`, `False`, or `None`).

Issue #911, opened 2015-08-18
def vs setv


gilch:
A subissue from #240.

These do _exactly_ the same thing as far as I can tell. We don't need both. `def` is a Python keyword, so it gets priority in my mind. However, it's currently defined to do something very different than the Python `def`, so I'm less certain.

As pointed out in the discussion for #899, we don't actually _have_ to reserve Python keywords in Hy. It might be problematic for Hy/Python interop not to, but there are workarounds. So the opposite is also possible. We can just keep `setv` and treat `def` like any other symbol.

I've heard arguments that we should keep both, the gist being because `def` feels more like a global constant definition, and `setv` feels more like a variable change. I don't exactly buy this because,
- they're doing _exactly_ the same thing;
- we can have nested `defn`, `defclass`, and `defmacro`--meaning the def- prefix does not really mean global nor constant elsewhere in Hy;
- global "constants" already have earmuffs/caps to distinguish them from normal variables and we don't need another convention on top of that. 

But if this feeling is still that important, there are some alternative ways to deal with it.

gilch:
Alternative proposal 1:
**get rid of BOTH def and setv**

And replace with `:=` (a la ALGOL). Now they both feel like what they are--an assignment to a variable. 

Why `:=`? Well `=` was already taken to mean `==`. Confusing the two is a common source of errors in C. (Not using ALGOL's `:=` is widely considered one of C's historical design flaws.) 

Python fixed this by not counting `=` statements as expressions, so they can't accidentally be used as one. Hy has unfixed this since everything works as an expression, but mostly made up for this by having `setv`/`def` for assignment instead.

gilch:
Alternative proposal 2:
**Make def and setv different enough to be worth another keyword.**

Like I said before, filling keyword quotas is poor language design.

But there have been proposals to kind of enforce the existing convention for `def`, so perhaps we could try to enforce that better.

The best I've come up with is to make `def` a macro that asserts the target hasn't been assigned to yet. This is what you want for a constant. So you can only `def` any given symbol once. At least not without using an explicit `del` on it in between re`def`s. You can still `setv` a `def`ed variable though.

Many languages distinguish between delcaring a variable and assigning to it, and also variables vs constants--both points Python kind of glosses over. You could use the new asserting `def` for either purpose.

We could have our linter (hydiomatic) emit a warning whenever an earmuffed symbol was `setv`'d, but not when `def`ed.

[...]

PR #1475, opened 2018-01-02
[RFC] Variable annotation support

vodik:
Hey, so I really would like to see some type hinting support, and putting this up to see if there's interest in this. If there is, I'm willing to see this though, tests and all, but I don't want to spend a lot of time on this if there's fundamentally no interest.

I think its still worth adding support even if there isn't any direct static analysis support as the typing module has and gaining a number of useful runtime tools. And as these techniques become more popular, probably visible in 3rd party python libraries.

This PR, which I do not expect to get merged in its current state, adds support for emitting `ast.AnnAssign`. This allows us to support PEP 526 and define variables without an initial value. Then, to test this, I added `(annv ...)` expression for defining a variable without an initial value.

This enables us support things like `typing.NamedTuple` and the coming `dataclasses.dataclass` easily:

```clojure
(import [typing [NamedTuple]])

(defclass Person [NamedTuple]
   (annv name str
         age int)) 


(defmain [&rest args]
  (setv person (Person "Jeff" 30))
  (print person))
```

Now why not just manipulate `__annotations__` like I suggested in #640? It becomes cumbersome when dealing with things like `dataclasses` that expect both type annotations *and* a default value.

Feedback? Opinions? Should I carry on, is it there no interest in maintaining this?

### DIFF
--- a/docs/contrib/walk.rst
+++ b/docs/contrib/walk.rst
@@ -48,7 +48,7 @@ each sub-form, uses ``f`` 's return value in place of the original.
 .. code-block:: hy
 
     => (import [hy.contrib.walk [postwalk]])
-    => (def trail '([1 2 3] [4 [5 6 [7]]]))
+    => (setv trail '([1 2 3] [4 [5 6 [7]]]))
     => (defn walking [x]
     ...  (print "Walking:" x :sep "\n")
     ...  x)
@@ -128,7 +128,7 @@ each sub-form, uses ``f`` 's return value in place of the original.
 .. code-block:: hy
 
     => (import [hy.contrib.walk [prewalk]])
-    => (def trail '([1 2 3] [4 [5 6 [7]]]))
+    => (setv trail '([1 2 3] [4 [5 6 [7]]]))
     => (defn walking [x]
     ...  (print "Walking:" x :sep "\n")
     ...  x)
--- a/docs/extra/anaphoric.rst
+++ b/docs/extra/anaphoric.rst
@@ -225,7 +225,7 @@ Returns a function which applies several forms in series from left to right. The
 
 .. code-block:: hy
 
-   => (def op (ap-compose (+ it 1) (* it 3)))
+   => (setv op (ap-compose (+ it 1) (* it 3)))
    => (op 2)
    9
 
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -544,7 +544,7 @@ For example:
 
 .. code-block:: clj
 
-    => (def names ["Alice" "Bob" "Charlie"])
+    => (setv names ["Alice" "Bob" "Charlie"])
     => (print names)
     [u'Alice', u'Bob', u'Charlie']
 
@@ -590,7 +590,7 @@ below:
     ...
     ...  (defn speak [self] (print "Meow")))
 
-    => (def spot (Cat))
+    => (setv spot (Cat))
     => (setv spot.colour "Black")
     'Black'
     => (.speak spot)
@@ -1077,8 +1077,8 @@ immediately.
 
 .. code-block:: hy
 
-    => (def collection (range 10))
-    => (def filtered (genexpr x [x collection] (even? x)))
+    => (setv collection (range 10))
+    => (setv filtered (genexpr x [x collection] (even? x)))
     => (list filtered)
     [0, 2, 4, 6, 8]
 
@@ -1295,7 +1295,7 @@ passed to another function for filtering output.
 
 .. code-block:: clj
 
-    => (def people [{:name "Alice" :age 20}
+    => (setv people [{:name "Alice" :age 20}
     ...             {:name "Bob" :age 25}
     ...             {:name "Charlie" :age 50}
     ...             {:name "Dave" :age 5}])
@@ -1359,7 +1359,7 @@ conditional expression. Some examples:
 
 .. code-block:: clj
 
-    => (def collection (range 10))
+    => (setv collection (range 10))
     => (list-comp x [x collection])
     [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
 
@@ -1653,7 +1653,7 @@ counted starting from the end of the list. Some example usage:
 
 .. code-block:: clj
 
-    => (def collection (range 10))
+    => (setv collection (range 10))
 
     => (cut collection)
     [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -536,10 +536,10 @@ Gets help for macros or tag macros, respectively.
     Gets help for a tag macro function available in this module.
 
 
-def / setv
-----------
+setv
+----
 
-``def`` and ``setv`` are used to bind a value, object, or function to a symbol.
+``setv`` is used to bind a value, object, or function to a symbol.
 For example:
 
 .. code-block:: clj
@@ -563,6 +563,23 @@ They can be used to assign multiple variables at once:
     => b
     2L
     =>
+
+
+def
+---
+
+``def`` is used to declare a variable with a type annotation, while
+optionally providing a default value.
+For example:
+
+.. code-block:: clj
+
+    => (defclass Person []
+    ...   (def name str)
+    ...   (def age int))
+
+    => (print (tying.get-type-hints (Person)))
+    {'name': <class 'str'>, 'age': <class 'int'>}
 
 
 defclass
@@ -1296,9 +1313,9 @@ passed to another function for filtering output.
 .. code-block:: clj
 
     => (setv people [{:name "Alice" :age 20}
-    ...             {:name "Bob" :age 25}
-    ...             {:name "Charlie" :age 50}
-    ...             {:name "Dave" :age 5}])
+    ...              {:name "Bob" :age 25}
+    ...              {:name "Charlie" :age 50}
+    ...              {:name "Dave" :age 5}])
 
     => (defn display-people [people filter]
     ...  (for [person people] (if (filter person) (print (:name person)))))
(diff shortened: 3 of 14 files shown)


vodik:
An option for syntax could be to borrow from Racket for this:

```clojure
(defclass Person [NamedTuple]
   (: name str
     age int)) 
```

Which is already similar to Python's syntax. The problem with this though is that `:` is already a valid empty keyword, so it wouldn't work without even more work and maybe backwards compatibility concerns?

gilch:
I had proposed using `def` for this purpose. #911. Two arguments would declare a variable's type, and three arguments would also initialize it. Variables with unspecified type would continue to use `setv`.

vodik:
I like that idea. Lets see what I can do with that.

vodik:
@gilch quickly threw together something that implements Option A.

```clojure
(import [typing [NamedTuple]])

(defclass Person [NamedTuple]
   (def name str)
   (def age int))

(defmain [&rest args]
  (setv person (Person "Jeff" 30))
  (print person))
```

gilch:
Can it do the default arguments too?
```
(defclass Person [NamedTuple]
   (def name str "John Doe")
   (def age int))
```
The test are failing since you changed the meaning of `def` but didn't replace the existing usage with `setv`. This should be easy to fix with a project wide search and replace.

vodik:
It does. I need to clean it up and write tests, but at work. Will do it in the evening.

vodik:
Didn't have enough time to finish tests yet - need to go to sleep, but I replaced `def` with `setv` everywhere and did an initial draft of documentation changes.

Kodiologist:
Incidentally, if we're willing to change `def`, it should probably be an alias of (or replace) `defn`. It's unduly confusing that the name Python uses to define functions is used for something different in Hy. Then, for annotating variables, we can use `annv` or `var` or whatever.

gilch:
I think the current definitions in the PR are just fine.

Clojure does use `def` for vars and `defn` for functions. Hy's syntax was based primarily on Clojure. I don't find the "unduly confusing" argument very convincing in this case. This might get you once and then you read the docs. It's really not a problem. For someone coming from Clojure, `def` meaning "function" would be just as confusing--i.e. not that much.

Maybe we could remove `def` altogether and use `defv` to match `setv`--short for "define variable" and "set variable", respectively, which pretty well matches what they'd do. But I don't think it's necessary.

Kodiologist:
I'm not concerned with a speed bump for learners so much as the tight relationship between Hy and Python that often requires the programmer to mentally switch between syntaxes. The changing meaning of `def` is one more complication to this, and its only benefit is to remove a speed bump for people who already know Clojure.

gilch:
That's fair. Hy users might realistically switch between Hy and Python a lot. I had a similar complaint about symbols containing a trailing comma.

I've been trying to get rid of duplication (like `def`/`setv`) in core, so I wouldn't approve of making `def` an alias for `defn`.

I still think the current proposal is fine. But I'd like more feedback for a major syntax change.
@hylang/core, thoughts? Shall we
1. redefine `def` for annotations as currently proposed?
1. remove `def` altogether and use `defv` for variable annotations instead?
2. rename `defn` to `def` for functions and use `defv` for variable annotations? (weird from Clojure)
4. remove `def` altogether and just use `^foo` in `setv` forms for variable annotations?
3. rename `defn` to `def` and just use `^foo` in `setv` forms for variable annotations?

Other proposals?

Kodiologist:
I don't really want `def` and `defn` as synonyms, either, but it might be a good temporary measure to keep for one major release, with a deprecation warning for `defn`, because of the sheer amount of code that will break if we remove `defn`. We're still entitled to break backward compatiblity, but a change this big could be worth showing our poor early adopters a bit of mercy.

The `^` options don't seem so great because, presuming the proposal is to make `^` read as an operator rather than having `setv` treat symbols beginning with `^` specially, we're relinquishing the character `^` for a very limited purpose.

So I like proposal 3 best, although we might want to use `annv` or something instead of `defv` in case we want to reserve `defv` for some future feature of declaring variables a la Perl's `use strict`, as you discussed in #911. Of course, we may need to shake up the annotation syntax in that case regardless of what name we choose here, since we'd need to make sure annotations play nicely with declarations.

gilch:
We'd also use `^` for function annotations, like Clojure does, as proposed in #640. We *will* want those eventually. The single `^` could be special-cased as an operator, like we do for the `~` meaning both unquote and bitwise-not.

So, 
```Python
def foo(x: int) -> str:
    return str(x*x)
```
would be written
```Clojure
(defn ^str foo [^int x]
  (str (* x x)))
```
Alternatively, we could use a `#^`. Clojure actually used to do this, and it still works, last I checked.
```Clojure
(defn #^ str foo [#^ int x]
  (str (* x x)))
```
Symbols are already not allowed to start with `#`, so we're not losing anything. But a typechecked program would be using these a lot, so I'd prefer the shorter syntax. We'd be relinquishing `^` for something that gets used all the time, similar to `:` for keywords.

Alternatively, since Python annotations are only allowed in certain places, we could have `^` continue to be a symbol, but it would have special meaning in certain special forms, like `setv` and `defn`. But you'd always have to have a space after the `^`.
```Clojure
(defn ^ str foo [^ int x]
  (str (* x x)))
```
That would make naming the `^` shadow function a little awkward, since you probably couldn't just `setv` it, but it could be done with `(assoc (globals) "^" bit-xor)` or something.

Kodiologist:
Since only one other form needs that, namely `defn`, let's, again, try to change the lexer just to accommodate it.

gilch:
Not just one, `setv`, `defn`, `fn`, and many macros based on them, like `defmulti` or `let` might want to use these. And they do get used a lot--probably a lot more than the somewhat obscure bitwise-xor operator you seem to be protecting. Clojure uses metadata for a lot more than type hints. I think a lexer update might be worth it if we can get mypy to work with Hy.

But the `#^` and `^`-space alternatives wouldn't have to change the lexer. We could start with one of these and reconsider the question later.

We could have the `#^` be a tag macro shorthand for the constructor of a new `HyMeta` model type.

Kodiologist:
> Not just one, `setv`, `defn`, `fn`, and many macros based on them, like `defmulti` or `let` might want to use these.

Sure, but they're ultimately going to take advantage of it by calling the same code (which presumably would live in the compiler's definition of `fn` and `fn*`), right? So we should still only need to handle this syntax in one place.

> But the `#^` and `^-space` alternatives wouldn't have to change the lexer.

Right. Or, you could still obtain syntax like

    (defn ^str foo [^int x]
      (str (* x x)))

by having `defn` treat symbols beginning with `^`, as well as `^` itself, specially. This would be similar to how `&optional` is an ordinary symbol from the lexer and parser's perspective, but magical in a lambda list.

vodik:
> Not just one, `setv`, `defn`, `fn`, and many macros based on them

And don't forget `defclass` - it also takes a list of attributes.

> ... `defn` treat symbols beginning with `^`, as well as `^` itself, specially. This would be similar to how `&optional` is an ordinary symbol from the lexer's and parser's perspectives, but magical in a lambda list.

How would that interact with macros? Wouldn't it be desirable to allow type hints to be a macro argument for doing code generation? For example (contrived):

```clojure
(code-gen my-helper ^str)
```

I can't think how else to add annotations nicely, while keeping them optional and backwards compatible, without adding new syntax:

```clojure
(defn ^str foo [^int x]
  (str (* x x)))

(setv [^int x] 42)

```clojure
(defclass MyClass []
  [[^int x] 42])
```

vodik:
For what its worth, Python also reuses operators as syntax too: `@` is used for matmul and decorators.

Kodiologist:
> > ... `defn` treat symbols beginning with `^`, as well as `^` itself, specially. This would be similar to how `&optional` is an ordinary symbol from the lexer's and parser's perspectives, but magical in a lambda list.
>
> How would that interact with macros? Wouldn't it be desirable to allow type hints to be a macro argument for doing code generation? For example (contrived):
>
> ```clojure
> (code-gen my-helper ^str)
> ```

That should work fine, with no special effort: `code-gen` treats `^str` as an ordinary symbol and passes it down to `defn`, which treats it specially.

vodik:
Should I split the `def` removal into a separate PR?

Kodiologist:
If you end up using this PR to add variable annotations in a way that doesn't affect `def`, then yeah, you don't need to worry about `def` in this PR.

Kodiologist:
So what's the status of this now, @vodik?

vodik:
I can continue on with this, but I figured I'd wait to see what we decide for #1482. Or if we'll accept a new type annotation form. The more I've been reflecting on this, the less confident it can even be done without something new going into the language.

For example, I've seen this proposal thrown around for function argument annotation:

```clojure
(defn foo [test ^int] ...)
; or
(defn foo [[test ^int]] ...)
```

But without the `^` in there, you can't distinguish between two arguments, or tuple unpacking support.

But if that's how its going to have to be, I'd like to see *that* added first rather than adding `annv` to later remove it/adjust it.

Kodiologist:
I guess that depends on what counts as "something new going into the language". Like I was saying, you get can `^` to work like that by only changing how `defn` treats symbols beginning with `^`.

vodik:
Yes, but I just want some sort of agreement. If we agree it should be ^int, then I want to use that form in as well `annv`.

And I just want to know ahead of time if people are okay enough with the idea that its worth spending time on and would be at least considered and not rejected immediately.

Kodiologist:
> If we agree it should be `^int`, then I want to use that form in as well `annv`.

That seems fine.

vodik:
I'm significantly more comfortable with Hy and how the internals work now - and I see where I was tripping up in my various musing and proposals on how to implement this (thanks for the patience!).

I think implementing #1482 would be great and I think I understand how to tackle it now. @gilch if you don't mind, I think I can handle it, unless you wanted to do it.

PR #1483, opened 2018-01-05
Just remove `def`

vodik:
Simply removed `def` keyword and replaces it with `setv` everywhere. See #1475 and #911.


Kodiologist:
I think we should decide what we're going to do with it before removing it; for example, replacing `defn` with it (as I'd like to do).

vodik:
Yup, that's fair, but that should be easy to add on top of this.

Kodiologist:
That's true, and this alone is more likely to get consensus. So let's go with it. Just update the `def / setv` section of `api.rst` and update NEWS and I think it's good to go.

vodik:
Done. I'm pretty sure `def` has been removed *everywhere* now.

Issue #1481, opened 2018-01-04
Add implicit tuples to . DSL


gilch:
Something I mentioned in #911, Python code often passes tuples to the subscript operator `[]`, in numpy code especially (for multidimensional slices), and now in the typing module's generics. Currently you can write things like,
```Hy
;; foo[:, i]
(. foo [(, (slice None)
           i)])

;; Generator[User, None, None]
(. Generator [(, User None None)])
```
But since we're already declaring a list, the tuple seems redundant. So we could simplify it to.
```Hy
;; foo[:, i]
(. foo[(slice None)
       i])

;; Generator[User, None, None]
(. Generator[User None None])
```
So a single item would be passed to the subscript as-is, but multiple items would implicitly have a tuple wrapped around them. In the rare case you actually want a single-item tuple, you can still do that explicitly, since it's a single item passed as-is.
```Hy
;; foo[i,]
(. foo[(, i)])
```
In fact, the old syntax would still work as-is, even if the tuple had multiple items, since it would be a single item, passed as-is.
```
;; foo[i, j, k]
(. foo [(, i j k)])
;; same as
(. foo [i j k])
```

With more compact notation for slice objects, #541 we could improve the slice sytanx further.

I'm reluctant to propose anything that complicates the compiler. But there's really no reason the `.` DSL couldn't be a macro, provided a more basic `.*` special form for attribute access only. We could also add calls to the chain #1108.

Kodiologist:
Looks pretty cool. It is a bit weird that square brackets, which are usually used to define a list, end up defining a tuple here. But the same criticism applies to Python, so that's not our fault.

gilch:
Technically, it's the comma that makes a tuple in Python, not the parentheses (with the single exception of the empty tuple). It's just that syntactically, you often need them. Inside a subscript operator is one of the cases where you don't.

ekaschalk:
Somewhat related - how do you all feel about implementing `...` as an alias for `Ellipsis`?

gilch:
I just demonstrated that in #541. In the context of the tag macro, it's an alias. But dots do weird things to Hy expressions, since they interfere with our syntax for dotted lists. You can't just `(setv ... Ellipsis)`. We'd have to special case it in the grammar the way Python does. I agree that spelling out `Ellipsis` makes slicing numpy arrays too annoying. You could `(setv ,,, Ellipsis)` though, which is my current workaround.

vodik:
I managed to get hy to support Ellipsis:

[...]

PR #1491, opened 2018-01-21
Native support for Ellipsis

vodik:
Looks like its a really trivial patch to add a special case to interpret `...` as `Ellipsis`.

Related to #541, #1481

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2163,7 +2163,12 @@ class HyASTCompiler(object):
 
     @builds(HySymbol)
     def compile_symbol(self, symbol):
-        if "." in symbol:
+        if symbol == "...":
+            if not PY3:
+                raise HyTypeError(symbol, 'ellipsis syntax is only supported '
+                                          'for Python 3')
+            return asty.Ellipsis(symbol)
+        elif "." in symbol:
             glob, local = symbol.rsplit(".", 1)
 
             if not glob:
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -272,6 +272,10 @@ class Result(object):
         if isinstance(other, ast.excepthandler):
             return self + Result(stmts=[other])
 
+        if isinstance(other, ast.slice):
+            raise TypeError("Can't add %r with slice specific syntax %r" % (
+                self, other))
+
         if not isinstance(other, Result):
             raise TypeError("Can't add %r with non-compiler result %r" % (
                 self, other))
--- a/NEWS
+++ b/NEWS
@@ -26,6 +26,7 @@ Changes from 0.13.0
    * `#%` works on any expression and has a new `&kwargs` parameter `%**`
    * new `doc` macro and `#doc` tag macro
    * support for PEP 492 with `fn/a`, `defn/a`, `with/a` and `for/a`
+   * add support to understand `...` as `Ellipsis`
 
    [ Bug Fixes ]
    * Numeric literals are no longer parsed as symbols when followed by a dot
--- a/tests/compilers/test_ast.py
+++ b/tests/compilers/test_ast.py
@@ -664,3 +664,9 @@ def test_ast_good_yield_from():
 def test_ast_bad_yield_from():
     "Make sure AST can't compile invalid yield-from"
     cant_compile("(yield-from)")
+
+
+@pytest.mark.skipif(not PY3, reason="Python 3 required")
+def test_ast_ellipsis():
+    "Make sure AST can compile Ellipsis"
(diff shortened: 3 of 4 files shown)


vodik:
Oh right, this is only going to work for Python 3... Well that still works for me.

hy/compiler.py
```diff
@@ -2163,7 +2163,12 @@ def nan(): return asty.BinOp(
 
     @builds(HySymbol)
     def compile_symbol(self, symbol):
-        if "." in symbol:
+        if symbol == "...":
+            if not PY3:
+                raise HyTypeError(symbol, 'ellipsis syntax is only supported '
```
rkday: Although the `...` syntax is only supported on Python3, the `Ellipsis` object exists in 2.x (https://docs.python.org/2.4/lib/bltin-ellipsis-object.html). Couldn't we do:

```
if PY3:
    return asty.Ellipsis(symbol)
else:
    symbol = "Ellipsis"
```

Kodiologist: The output should probably be `__builtins__.Ellipsis`, since `Ellipsis` isn't a reserved word.

vodik: Good call

vodik: Unfortunately, it doesn't work. When you try to use Ellipsis inside a slice in python 2 that's been generated this way, python throws `ValueError: invalid integer value: None` when trying to load the ast.

The reason, I think, is because that patch generates code to load `Ellipsis`: 

```
slice=Index(value=Tuple(elts=[Num(n=1L), Name(id='Ellipsis', ctx=Load()), Num(n=6L)], ctx=Load()))
```

However, `Ellipsis` **is** a thing an Python 2's ast. This is what I get from ast.parse -> ast.dump on Python 2:

```
slice=ExtSlice(dims=[Index(value=Num(n=1)), Ellipsis(), Index(value=Num(n=6))])
```

Kodiologist: So why not use `ast.Ellipsis`, just like in Python 2?

vodik: Now, the reason I gated this to Python2 is that there's some deeper issue with constructing Ellipsis on Python2:

```
⤷ TypeError: Can't add <hy.compiler.Result object at 0x7fc4c1834e30> with non-compiler result <_ast.Ellipsis object at 0x7fc4c17caa50>
```

I'll dig into what's going on.

Kodiologist: Possibly Python 2 will only allow `ast.Ellipsis` inside a slice. In that case, you'll probably need to edit `compile_cut_expression` to peek at its arguments and make an `ast.Ellipsis` itself, whereas `compile_symbol`  will need to create `__builtins__.Ellipsis` or raise an error.

vodik: Yeah, so the problem is in python 2, `ast.Ellipsis` is a subclass of ast.slice, not ast.expr, and `Result.__add__` then doesn't know what do with it.

But the problem with that is it means its only valid as a dim inside `ExtSlice`, which we don't generate.

I think this is only going to work on Python 3. Which personally I'm fine with. If it where up to me, I'd be seriously looking at dropping Python 2 support 😉

Kodiologist: So we'll need to produce an ExtSlice under Python 2, right?

> If it where up to me, I'd be seriously looking at dropping Python 2 support

[Me too, but it isn't.](https://github.com/hylang/hy/issues/1217)

vodik: @Kodiologist yeah, I think so. With @rkday proposed patch, the following snippet only partially works:

```clojure
(defn test-slice []
  (defclass EllipsisTest []
    (defn --getitem-- [self [start ellipsis end]]
      (assert (= ellipsis Ellipsis))
      (list (range start end))))

  (assert (= (get (EllipsisTest) (, 1 ... 6))
             [1 2 3 4 5])))
```

When manually copy and pasted into hy's REPL, it works as expected. When imported, Python decides it doesn't like the generated AST.

Which really gaslighted me while trying to dig into it, heh.

> Me too, but it isn't.

Maybe I should pipe up on there, but I think its worth reconsidering now. Do one more release for both Python 2 and Python 3, then drop Python 2 for the next one.

Python 3 only has got some serious momentum now (http://www.python3statement.org/) and maybe it makes sense to follow the scientific computing's timeline. I mean, IPython is Python 3 only now, and IPython Notebook is going to be as well by June this year.

But alas, that's neither here nor there.

vodik:
Okay, I'm working on cleaning up slice generation, I'll add this as part of that work to make it both py2/py3 compatible.

Issue #1492, opened 2018-01-21
Properly Generate ExtSlice


vodik:
Multi-dimensional lookup should emit `ast.ExtSlice` instead of a normal lookup with `ast.Tuple`.

Basically any time we'd generate this:

```
slice=Index(value=Tuple(elts=[Num(n=3), Ellipsis(), Num(n=5)], ctx=Load())
```

But if any of the elements of that tuple is an instance of `ast.slice` (`ast.Slice` and, on Python 2, `ast.Ellipsis`), it should be constructed as `slice=ExtSlice(dims[...])` instead.

Related to #1491 if we want to add Python 2 support for Ellipsis. Otherwise its not too pressing since we don't have anything that emits `ast.Slice` directly, its always done as a function call. But if we where to add something, this issue would have to be resolved first.

Kodiologist:
Does this distinction matter at all in Python 3, or is it effectively a Python 2 issue?

gilch:
Possibly related to #1481. If the distinction matters, Hy will have to have special forms for either case, which means `.` can't just be a macro.

vodik:
@Kodiologist as far as I can tell, yes, but we get away with it because we never emit multidimensional slices anywhere with `ast.Slice`, always with a call to the `slice` constructor.

As far as I can tell, only `cut` emits `ast.Slice`, but its single dimension.

vodik:
And for what its worth, its probably worth implementing - even if its an optimization when we notice `slice` construction inside indexing - because there's an order of magnitude in performance between the two forms:

```python
In [1]: class Demo:
   ...:     def __getitem__(self, key):
   ...:         return key
   ...: 

In [2]: demo = Demo()

In [3]: %timeit demo[1:2]
183 ns ± 0.666 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: %timeit demo[slice(1, 2)]
266 ns ± 2.23 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

gilch:
On second thought, we could have a `cuts` special form for multiple dimensions and `.*` and implement `.` as a macro in terms of those. Or we could rethink `cut`. But how should the syntax look in Hy if we're not using `slice`?

vodik:
I don't know how you could properly disambiguate it without using `slice`:

[...]

Issue #1363, opened 2017-08-04
Hy models have wrong equality


gilch:
Hy models are equal to other types.

For atoms, maybe this is okay?
```Python
=> (= 1 '1)
from hy import HyInteger
(1 == HyInteger(1))
True
```
It might even be convenient for symbols to be equal to strings, because Python uses strings pervasively when referring to its identifiers at runtime (e.g. `getattr`), though symbol mangling might confuse things in some cases.

But for compound types, it gets weird.
```Python
=> (= [1 2 3] '[1 2 3] '#{1 2 3} '{1 2 3} '(1 2 3)) ; ORLY?
from hy import HyDict, HyExpression, HyInteger, HyList, HySet
([1, 2, 3] == HyList(((([] + [HyInteger(1)]) + [HyInteger(2)]) + [HyInteger(3)])) == HySet(((([] + [HyInteger(1)]) + [HyInteger(2)]) + [HyInteger(3)])) == HyDict(((([] + [HyInteger(1)]) + [HyInteger(2)]) + [HyInteger(3)])) == HyExpression(((([] + [HyInteger(1)]) + [HyInteger(2)]) + [HyInteger(3)])))
True
```

Kodiologist:
The fact that `(= 'foo "foo")` has created a lot of strange bugs.

vodik:
I'm honestly not a fan of how `(= 'foo "foo")`. Sure `getattr` is a neat trick but its also technically represents a slight deviation away from the Python object model.

- It's got all the weirdness of implicit type conversion:
```
=> (+ 'foo "bar")
'foobar'
```
- It leads to more verbose and error prone macros `(if (and (symbol? x) (= x "foo")) ...)` vs just `(if (= x 'foo) ...)`
- I think its somewhat surprising and non-intuitive

I've been playing with a branch that makes `(= 'foo "foo")` false and similar strictness on HyKeywords. I'll put it up once I'm done fixing tests.

allison-casey:
[...]

PR #1493, opened 2018-01-30
Better slicing with ellipsis support

vodik:
Proof that its possible to optimize slicing. This code:

```clojure
(setv foo [1 2 3])
(get foo (slice 1 2))
(slice 1 2)
```

Is now equivalent to:

```
foo = [1, 2, 3]
foo[1:2]
slice(1, 2)
```

I've also demoted `cut` to a macro as its now redundant, but imho still a useful form (still nicer to use when updating a list in place: `(setv (cut foo) [1 2 3])`

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -27,6 +27,7 @@ Changes from 0.13.0
    * new `doc` macro and `#doc` tag macro
    * support for PEP 492 with `fn/a`, `defn/a`, `with/a` and `for/a`
    * remove `def`
+   * optimize `slice` inside subscriptions to generate slice AST
 
    [ Bug Fixes ]
    * Numeric literals are no longer parsed as symbols when followed by a dot
--- a/NEWS
+++ b/NEWS
@@ -28,6 +28,7 @@ Changes from 0.13.0
    * support for PEP 492 with `fn/a`, `defn/a`, `with/a` and `for/a`
    * remove `def`
    * optimize `slice` inside subscriptions to generate slice AST
+   * add support to understand `...` as `Ellipsis`
 
    [ Bug Fixes ]
    * Numeric literals are no longer parsed as symbols when followed by a dot
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -311,6 +311,7 @@ class Result(object):
         )
 
 
+
 def _branch(results):
     """Make a branch out of a list of Result objects
 
@@ -384,6 +385,13 @@ def is_unpack(kind, x):
             and len(x) > 0
             and isinstance(x[0], HySymbol)
             and x[0] == "unpack_" + kind)
+
+
+def is_builtin(kind, x):
+    return (isinstance(x, HyExpression)
+            and len(x) > 0
+            and isinstance(x[0], HySymbol)
+            and x[0] in (kind, "__builtin__" + kind))
 
 
 def ends_with_else(expr):
@@ -544,6 +552,24 @@ class HyASTCompiler(object):
             return compiled_exprs, ret, keywords, oldpy_starargs, oldpy_kwargs
         else:
             return compiled_exprs, ret, keywords
+
+    def _compile_subscript(self, expr, value, sub_expr):
+        if is_builtin("slice", sub_expr):
+            ret = Result()
+            nodes = [None] * 3
+            for i, e in enumerate(sub_expr[1:]):
+                ret += self.compile(e)
+                nodes[i] = ret.force_expr
+
+            index = ast.Slice(lower=nodes[0], upper=nodes[1], step=nodes[2])
+        else:
+            index = ast.Index(self.compile(sub_expr).force_expr)
+
+        return asty.Subscript(
+            expr,
+            value=value,
+            slice=index,
+            ctx=ast.Load())
 
     def _compile_branch(self, exprs):
         return _branch(self.compile(expr) for expr in exprs)
@@ -1222,18 +1248,12 @@ class HyASTCompiler(object):
 
     @builds("get")
     @checkargs(min=2)
-    def compile_index_expression(self, expr):
-        expr.pop(0)  # index
+    def compile_index_expression(self, exprs):
+        exprs.pop(0)  # index
 
-        indices, ret, _ = self._compile_collect(expr[1:])
-        ret += self.compile(expr[0])
-
-        for ix in indices:
-            ret += asty.Subscript(
-                expr,
-                value=ret.force_expr,
-                slice=ast.Index(value=ix),
-                ctx=ast.Load())
+        ret = self.compile(exprs[0])
+        for expr in exprs[1:]:
+            ret += self._compile_subscript(exprs, ret.force_expr, expr)
 
         return ret
 
@@ -1287,21 +1307,6 @@ class HyASTCompiler(object):
                                               ast.Del))
 
         return ret + asty.Delete(expr, targets=del_targets)
-
-    @builds("cut")
-    @checkargs(min=1, max=4)
-    def compile_cut_expression(self, expr):
-        ret = Result()
-        nodes = [None] * 4
-        for i, e in enumerate(expr[1:]):
-            ret += self.compile(e)
-            nodes[i] = ret.force_expr
-
-        return ret + asty.Subscript(
-            expr,
-            value=nodes[0],
-            slice=ast.Slice(lower=nodes[1], upper=nodes[2], step=nodes[3]),
-            ctx=ast.Load())
 
     @builds("with_decorator")
     @checkargs(min=1)
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -311,7 +311,6 @@ class Result(object):
         )
 
 
-
 def _branch(results):
     """Make a branch out of a list of Result objects
 
@@ -387,11 +386,8 @@ def is_unpack(kind, x):
             and x[0] == "unpack_" + kind)
 
 
-def is_builtin(kind, x):
-    return (isinstance(x, HyExpression)
-            and len(x) > 0
-            and isinstance(x[0], HySymbol)
-            and x[0] in (kind, "__builtin__" + kind))
+def expr_startswith(kind, x):
+    return (isinstance(x, HyExpression) and len(x) > 0 and x[0] == kind)
 
 
 def ends_with_else(expr):
@@ -552,26 +548,6 @@ class HyASTCompiler(object):
             return compiled_exprs, ret, keywords, oldpy_starargs, oldpy_kwargs
         else:
             return compiled_exprs, ret, keywords
-
-    def _compile_subscript(self, expr, value, sub_expr):
-        if is_builtin("slice", sub_expr):
-            ret = Result()
-            nodes = [None] * 3
-            for i, e in enumerate(sub_expr[1:]):
-                ret += self.compile(e)
-                nodes[i] = ret.force_expr
-
-            index = ast.Slice(lower=nodes[0], upper=nodes[1], step=nodes[2])
-        elif not PY3 and isinstance(sub_expr, HySymbol) and sub_expr == "...":
-            index = ast.Ellipsis()
-        else:
-            index = ast.Index(self.compile(sub_expr).force_expr)
-
-        return asty.Subscript(
-            expr,
-            value=value,
-            slice=index,
-            ctx=ast.Load())
 
     def _compile_branch(self, exprs):
         return _branch(self.compile(expr) for expr in exprs)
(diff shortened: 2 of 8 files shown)


vodik:
See #1492, but still not there yet as this is only concerned with a single dimension. Maybe the nicest would be to let `get` take a variable number of arguments...

gilch:
`get` **does** take a variable number of arguments. But that means a chain of lookups, not one slice over multiple dimensions.

vodik:
Oh, so like `itemgetter`? Interesting, didn't know that.

Beyond that, though, any complains with this approach?

hy/compiler.py
```diff
@@ -1221,17 +1221,31 @@ def _compile_import(expr, module, names=None, importer=asty.Import):
     def compile_index_expression(self, expr):
         expr.pop(0)  # index
 
-        indices, ret, _ = self._compile_collect(expr[1:])
-        ret += self.compile(expr[0])
+        if isinstance(expr[1], HyExpression) and expr[1][0] == "slice":
```
Kodiologist: This will trigger on the string `"slice"` instead of just the symbol.  You may also need to check that `expr[1]` is nonempty.

Kodiologist:
Don't forget about those test failures.

Kodiologist:
One issue with this approach is that `slice` isn't a reserved word, so there's no way to tell at compile-time that the symbol `slice` corresponds to `__builtins__.slice`.

vodik:
> One issue with this approach is that `slice` isn't a reserved word, so there's no way to tell at compile-time that the symbol `slice` corresponds to `__builtins__.slice`.

Oh right, is there any suggestions on how to deal with this? Should I promote `slice` to a builtin? Or is there another trick that I might not be aware of?

Kodiologist:
`slice` is already a builtin, because Hy inherits Python's builtins. Making `slice` a special form just so we have the right of always defining what `(slice ...)` means, when in fact we intend it to expand to a regular function call whenever the compiler actually gets to a `@builds("slice")` method, is probably overkill. Perhaps the best approach is to check for `__builtins__.slice` instead. It's probably best to make `cut` emit this instead of a bare `slice`, anyway. While it's technically possible for the user to set `__builtins__` to something weird, the double underscores in the name mean that this is officially a bad idea.

vodik:
> Perhaps the best approach is to check for `__builtins__.slice` instead.

So if I understand this correctly, I should compile the expression and then peek inside the results of that?

Kodiologist:
I think it will suffice to check for the symbol `__builtins__.slice`. This won't work for something like `(get foo ((. __builtins__ slice) ...))`, but that's probably okay. It's just an optimization, after all.

Kodiologist:
It should be possible to compile it first and check the AST instead of checking the symbol, but the code will probably be a bit wordier that way.

vodik:
Right, so I'm not seeing an easy way to do this. Either I check the `HyExpression`/`HySymbol` for "slice", or I compile it first and look into the resulting `ast.Name`. But my understanding, and maybe this is where I'm missing something, but `ast.Name` is still going to result in runtime lookup, so there is no sensible to know for sure if builtins has been messed with.

Maybe I should add `slice*` to force a slice object to be emitted?

Or maybe this idea isn't workable then and we'll just have to live with the slightly less optimal code generation.

> ... to check for `__builtins__.slice` instead. It's probably best to make `cut` emit this instead of a bare `slice`, anyway.

`cut` still compiles to `ast.Slice` as it used to, so it doesn't matter what anyone has done to builtins. That's the only reason I moved it to a macro.

Kodiologist:
It's true that both the name `__builtins__` and the attribute `slice` of the referent have to be looked up at runtime. But what I'm saying is, the double underscores in the name imply that the user isn't supposed to reassign the name. So it's fair of us to assume it hasn't been reassigned. Does that make sense?

vodik:
Oh, sorry, I understand what you're saying now. **Rewrite** to builtins.__slice__...

vodik:
Ah, sorry, I've misunderstood you entirely.

Okay, so let me see if I'm following: you're suggesting that the `cut` macro emits `__builtins__.slice` instead of `slice`. Then I'll look for `__builtins__.slice` and emit ast.Slice in that case? Is this on top of rewriting `slice` too, but just to make it more bullet proof?

vodik:
Okay, so if this is acceptable, I'll continue working on it. Currently `get` will either accept its current list of keys, or a single slice. I need to do some deeper refactoring to make the slice optimization work as expected, and then additionally dealing with #1492 and #1491 should be easy.

vodik:
Okay, merged #1491 into this PR, need to just rework tests, but I think this should do the trick.

The following code compiled as expected on both 2 and 3:

```clojure
(get foo (slice 1 2 3) ... bar)
```

will compile to

```python
foo[1:2:3][...][bar]
```

The optimization currently falls apart when we use a tuple to have a multidimensional slice, but I think we can merge this and leave that as a separate PR.

Just need to add some tests...

Kodiologist:
> Okay, so let me see if I'm following: you're suggesting that the `cut` macro emits `__builtins__.slice` instead of `slice`. Then I'll look for `__builtins__.slice` and emit ast.Slice in that case?

Bingo.

> Is this on top of rewriting `slice` too, but just to make it more bullet proof?

No, leave `slice` as a Python builtin.

> Just need to fix the tests and add new ones...

Sounds good. Let us know when you want us to take a look again.

vodik:
I think it can stand for review now. The code now properly emits the exact same AST that python does, multi or single dimension.

I took another step further and I deconstruct tuples when used as arguments to `get` to turn them into multi dimentional slices. This, as I understand it, is safe as Python represents them internally as tuples so there's no room for ambiguity. That means we now have the following code working:

```clojure
(get foo 1 2 (slice 3 4) ... (slice 5 6))
(get foo (, 1 2 (slice 3 4) ... (slice 5 6)))
```

generating the following equivalent Python:

```python
foo[1][2][3:4][...][5:6]
foo[(1), (2), 3:4, (...), 5:6]
```

The second example generates the following AST:

```python
Module(
    body=[
        Expr(
            value=Subscript(value=Name(id='foo'),
                slice=ExtSlice(
                    dims=[Index(value=Num(n=1)),
                        Index(value=Num(n=2)),
                        Slice(lower=Num(n=3), upper=Num(n=4), step=None),
                        Index(value=Ellipsis),
                        Slice(lower=Num(n=5), upper=Num(n=6), step=None)])))])
```

hy/compiler.py
```diff
@@ -382,6 +387,13 @@ def is_unpack(kind, x):
             and x[0] == "unpack_" + kind)
 
 
+def is_builtin(kind, x):
+    return (isinstance(x, HyExpression)
+            and len(x) > 0
+            and isinstance(x[0], HySymbol)
+            and x[0] in (kind, "__builtin__" + kind))
```
Kodiologist: Shouldn't this be `x[0] == "__builtins__." + kind`? You don't want to allow `kind` alone because then you'll trigger on `slice` alone, which might be rebound. Also, you need the period.

hy/compiler.py
```diff
@@ -1216,20 +1228,72 @@ def _compile_import(expr, module, names=None, importer=asty.Import):
...
+                dims.append(ret.force_expr)
+
+        def box_as_slice(expr):
+            if isinstance(expr, ast.slice):
+                return expr
+            return ast.Index(value=expr)
+
+        # How the subscript is constructed will depend on if any of
+        # the components subclass ast,slice. If they do, then we have
+        # to construct an ExtSlice instead of a Tuple.
```
Kodiologist: Is there any reason not to use ExtSlice instead of a tuple? That would simplify things a bit.

hy/core/bootstrap.hy
```diff
@@ -30,6 +30,10 @@
          (fn ~(+ `[&name] lambda-list)
            ~@body))))))
 
+(defmacro cut [node &rest body]
+  "Take a subset of a list and create a new list from it."
+  `(get ~node (__builtin__slice ~@body)))
```
Kodiologist: `__builtins__.slice`

hy/compiler.py
```diff
@@ -2163,7 +2212,12 @@ def nan(): return asty.BinOp(
 
     @builds(HySymbol)
     def compile_symbol(self, symbol):
-        if "." in symbol:
+        if symbol == "...":
+            if not PY3:
+                raise HyTypeError(symbol, 'ellipsis is only supported in '
+                                          'slices in Python 2')
```
Kodiologist: The error message might be better worded as "in Python 2, ellipsis is only supported in slices".

hy/compiler.py
vodik: Well `__builtins__.slice` doesn't work because __builtins__ is a dictionary, not an object. So we'd have to either change that too or write `(get __builtins__ "slice")`. And I'm not particularly eager to use throw that large expression down every time I want a slice to be slightly faster (and to be honest, same for `__builtins__.slice`).

If we don't feel comfortable with rewriting `slice`, I can just drop that entire part of the PR.

hy/compiler.py
vodik: Not sure, but I suspect its fine. I figured it made the most send to match exactly what Python does.

hy/compiler.py
Kodiologist: > Well `__builtins__.slice` doesn't work because `__builtins__` is a dictionary, not an object.

Are you sure?

    $ python2
    Python 2.7.14 (default, Sep 23 2017, 22:06:14) 
    [GCC 7.2.0] on linux2
    >>> __builtins__.slice
    <type 'slice'>

    $ python3
    Python 3.6.3 (default, Oct  3 2017, 21:45:48) 
    [GCC 7.2.0] on linux
    >>> __builtins__.slice
    <class 'slice'>

Kodiologist:
> the following equivalent Python:
>
> ```python
> foo[1][2][3:4][...][5:6]
> foo[(1), (2), 3:4, (...), 5:6]
> ```

These don't seem to be equivalent:

    import numpy as np

    foo = np.ones((7, 7, 7,  2, 2, 7))

    print(foo[(1), (2), 3:4, (...), 5:6])
    # prints: [[[[ 1.]
                [ 1.]]

               [[ 1.]
                [ 1.]]]]
    print(foo[1][2][3:4][...][5:6])
    # prints: []

hy/compiler.py
vodik: hy complains though:

```
=> (cut [1 2 3] 1 2 3)
  File "/home/simon/src/hy/hy/importer.py", line 201, in hy_eval
AttributeError: 'dict' object has no attribute 'slice'
```

Kodiologist: Huh, that's a bizarre bug. We don't set `__builtins__` anywhere in our code and yet Hy still sees a different object than Python does.

Kodiologist: [Oh shoot, that's a feature.](https://docs.python.org/3/reference/executionmodel.html) Since we hope to be setting `__main__` appropriately someday, we can't safely use `get` nor attribute lookup. I'm afraid I'm not sure what design would make sense, then.

vodik: I'm honestly quite comfortable letting slice mean slice here and just documenting the behaviour. We already prevent `slice` from being assigned to. Overriding `__builtins__.slice` in Python also won't change the behaviour of the `:` operator.

I could also add a check to make sure no `slice` macro has been defined and/or add `slice*` to force the construction of a slice object and that probable covers 99% of weird use cases.

Kodiologist: > We already prevent `slice` from being assigned to.

Are you sure?

     hy 0.13.0+211.g5c34296 using CPython(default) 3.6.3 on Linux
     => (setv slice "foo")
     => slice
     'foo'

Anyway, if we want to make `get` recognize a magic symbol, why not use `:` rather than `slice`? So you'd write e.g. `(get foo (: 3 4))` to mean `foo[3:4]`. This is concise, and literal keywords are already magical as function arguments.

vodik: I do like this better. My concern was that `:` is `HyKeyword(":")`, but there's less room for things to blow up.

> Are you sure?

Yes your right.

I think I was playing `@builds("slice")` at the time I tested it and forgot that I had put that in there.

hy/compiler.py
vodik: Oh, to expand on this, its probably worth at least merging the identical behaviour. If a simplification works, that's awesome, but that can be added next and at least there's always something more technically correct to revert to.

Kodiologist: Okay.

vodik:
I think you misunderstood - I'm not saying those two lines of Python are equivalent. Let me rephrase, and remove some noise from that complicated expression:

```
(get foo 1 2) ;; foo[1][2]
(get foo (, 1 2)) ;; foo[1, 2]
(get foo (, 1 2) (, 3 4)) ;; foo[1, 2][3, 4]
```

There is now support for multidimensional slices while also preserving chaining. I do it by destructing the tuple - which should be safe __getitem__ interface will present it as a tuple in the first place.

vodik:
Is this more to your taste then?

Kodiologist:
Yes, it is. I'll take a closer look tomorrow.

NEWS
```diff
@@ -27,6 +27,8 @@ Changes from 0.13.0
    * new `doc` macro and `#doc` tag macro
    * support for PEP 492 with `fn/a`, `defn/a`, `with/a` and `for/a`
    * remove `def`
+   * optimize `slice` inside subscriptions to generate slice AST
```
Kodiologist: Update this line for the change to `:`.

tests/native_tests/language.hy
```diff
@@ -860,6 +860,14 @@
...
 
+(defn test-slice []
+  (defclass CutTest []
+    (defn --getitem-- [self index]
+      (assert (= index (slice 1 6)))))
+
+  (get (CutTest) (: 1 6)))
+
+
 (defn test-cut []
```
Kodiologist: Shouldn't this be removed from `native_tests/langauge.hy` now that it's in `native_tests/native_macros.hy`?

hy/compiler.py
```diff
@@ -1216,20 +1224,69 @@ def _compile_import(expr, module, names=None, importer=asty.Import):
...
+            root,
+            value=value,
+            slice=box_as_slice(index),
+            ctx=ast.Load())
+
     @builds("get")
     @checkargs(min=2)
-    def compile_index_expression(self, expr):
-        expr.pop(0)  # index
+    def compile_index_expression(self, exprs):
```
Kodiologist: The reason we call this sort of parameter `expr` rather than `exprs` is because it's set to a single `HyExpression`.

hy/compiler.py
```diff
@@ -272,6 +272,10 @@ def __add__(self, other):
         if isinstance(other, ast.excepthandler):
             return self + Result(stmts=[other])
 
+        if isinstance(other, ast.slice):
+            raise TypeError("Can't add %r with slice specific syntax %r" % (
+                self, other))
+
```
Kodiologist: I get how the nature of `ast.slice` really bit you during the development of this PR, but will this check really be useful for later programmers?

hy/compiler.py
```diff
@@ -1216,20 +1224,69 @@ def _compile_import(expr, module, names=None, importer=asty.Import):
 
         return rimports
 
+    def _compile_subscript(self, root, value, exprs):
+        dims = []
+        ret = Result()
+
+        for expr in exprs:
+            if expr_startswith(HyKeyword(":"), expr):
```
Kodiologist: Unfortunately, this check isn't strict enough (it allows `(get [1 2 3 4 5] ("\ufdd0:" 1 2))`) because equality of Hy model objects is very loose about type. You probably need to check the type of `expr` explicitly.

vodik:
I won't be able to get this done till tomorrow, but I'm thinking of adding some examples to the documentation as well citing numpy.

hy/compiler.py
vodik: Would it maybe make sense to have a crack at #1363 and fix that first?

Kodiologist: You can try, but it's a deep rabbit hole to go down because the compiler currently depends on that "feature" in certain ways. Changes to that may also conflict with #1493. I'd advise waiting for #1493 before you do that, and finishing this first in any case.

vodik: I can wait, honestly. I'd rather see it some in on a more solid foundation.

gilch:
>  By default, when in the __main__ module, __builtins__ is the built-in module builtins; when in any other module, __builtins__ is an alias for the dictionary of the builtins module itself.

> Oh shoot, that's a feature. Since we hope to be setting __main__ appropriately someday, we can't safely use get nor attribute lookup. I'm afraid I'm not sure what design would make sense, then.

At least in Python3, `__import__("builtins").slice` should always work, whether or not you're in `__main__`. Unfortunately, Python2's equivalent is called `__builtin__`. But we could explicitly dump the builtins we need into the `hy` module or something, then it would be `__import__("hy").slice` in both 2 and 3. We're already normalizing some of the differences in Hy core, e.g. importing izip and renaming it zip. In this case, we'd conditionally import `slice` from `builtins` into `hy` when on Python 3 and from `__builtin__` when on Python 2.

Kodiologist:
@vodik, how are we doing here?

Kodiologist:
Closed for lack of response.

PR #1494, opened 2018-01-30
Fix runtime compliation of quoted expressions.

vodik:
So this is kinda insane, but also kinda cool and probably should be completely valid. Its probably better than `lineno` and `col_offset` defaults to 0 rather than None anyways, since that is what Python does if it can't give those fields a meaningful value.

The problem is any quoted expressions (or pure Python generated expressions) will generate AST with those attributes set to None.

I came across this playing with this snippet:

```clojure
(defn hy-compile [expr]
  (import [hy.compiler [hy-compile]]
          [hy.importer [ast-compile]])

  (setv namespace {})
  (-> (hy-compile expr "<STDIN>")
      (ast-compile "<STDIN>" "exec")
      (exec namespace))
  namespace)

(as-> `(defn test [] "Hello World") it
      (hy-compile it)
      (get it "test")
      (print (it)))
```

And as a pure Python alternative:

```python
from hy import *

expr = HyExpression([] + [HySymbol('defn')] + [HySymbol('test')] + [HyList([])] + [HyString('Hello')])
ast = hy.compiler.hy_compile(expr, "<STDIN>")
code = hy.importer.ast_compile(ast, filename="<STDIN>", mode="exec")
namespace = {}
exec(code, namespace)
namespace['test']()
```

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -129,9 +129,9 @@ class Asty(object):
     def __getattr__(self, name):
         setattr(Asty, name, lambda self, x, **kwargs: getattr(ast, name)(
             lineno=getattr(
-                x, 'start_line', getattr(x, 'lineno', None)),
+                x, 'start_line', getattr(x, 'lineno', 0)),
             col_offset=getattr(
-                x, 'start_column', getattr(x, 'col_offset', None)),
+                x, 'start_column', getattr(x, 'col_offset', 0)),
             **kwargs))
         return getattr(self, name)
 asty = Asty()


Kodiologist:
I would be against this change because it means that failing to set the attributes properly will cause Hy to point to the wrong line and column instead of raising an error. Hence, it will make bugs harder to find. Note that `hy.compiler.hy-compile` and `hy.importer.ast-compile` are internal functions.

vodik:
Python exposes these things as builtins, and considering our relationship with the language, I don't see why we should shy away from supporting something similar. Especially since we're in the position to do it better. Its effectively a runtime "`unquote`".

And I get that these are internals, but when `ast-compile` is just a thin wrapper around the `compile` builtin - the above snippet works just as well with `hy-compile`, `compile` and `exec` - I don't know if that's a particularly compelling argument against it.

vodik:
Maybe the right thing to do is make sure `lineno` and `col_offset` are correct with `quote`, but it does leave us with a question:

Why shouldn't a fully run-time specified `HyExpression` not be compilable?

I think its also worth noting that, as far as I know, this patch brings Hy's behaviour inline with Python's.

Kodiologist:
> Python exposes these things as builtins

I'm not totally sure what "these things" are. In Hy, as in other Lisps, if you want to run code objects (HyModels), you `eval` them. There's no need to compile them to Python AST first. `eval` does that on its own. Of course, it's better to `eval` an appropriately parametrized function than to `eval` inside a loop.

To evaluate Hy code from a Python program, you can use `hy-eval`, which is the same function, provided by `hy`.

>  the above snippet works just as well with `hy-compile`, `compile` and `exec`

I'm not totally sure what you're trying to do, or how you'd do it with only public functions.

vodik:
> I'm not sure what "these things" are.

Just meant commenting that `exec` and `compile` are Python builtins.

But looking back, I think I misunderstood your point and I'm talking past you. I never noticed `hy-eval` (or `read-str` / `disassemble`), which is functionality that I was looking for and was trying to implement by hand.

Thanks for the help and sorry.

Kodiologist:
Oh, very good, then. No problem.

PR #1497, opened 2018-02-05
Relative imports

### LINKED ISSUE #1496: Support relative imports (PEP 328)
Module names with a leading dot need to have that dot stripped:

```clojure
(disassemble '(import [.foo [bar]]))
```

Incorrectly generates `ImportFrom(module='.foo', names=[alias(name='bar', asname=None)], level=0)`, but `module` should be `"foo"` (no dot).

And 

```clojure
(disassemble '(import [. [bar]]))
```

Generates `ImportFrom(module='.', names=[alias(name='bar', asname=None)], level=0)` when `module` should be `None`.

vodik:
Closes #1496 

Just need to figure out how to test this - might make sense to just inspect the emitted ast.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -27,6 +27,7 @@ Changes from 0.13.0
    * new `doc` macro and `#doc` tag macro
    * support for PEP 492 with `fn/a`, `defn/a`, `with/a` and `for/a`
    * remove `def`
+   * support for relative imports (PEP 328)
 
    [ Bug Fixes ]
    * Numeric literals are no longer parsed as symbols when followed by a dot
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1156,10 +1156,17 @@ class HyASTCompiler(object):
         def _compile_import(expr, module, names=None, importer=asty.Import):
             if not names:
                 names = [ast.alias(name=ast_str(module), asname=None)]
+
+            ast_module = ast_str(module)
+            module = ast_module.lstrip(".")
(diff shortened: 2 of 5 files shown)


Kodiologist:
You could have a file in `native_tests` import from `..resources.tlib`.

vodik:
Yeah, that's a good idea, but it looks like that doesn't work unless the package contains `__init__.py` - `__init__.hy` isn't good enough. I wonder if that's something we can fix.

vodik:
Ah, no, this is another bug - but looks like the bug is *specifically* inside the pytest glue code.

vodik:
Yeah, its this function in particular that causes a problem from `_py.path` which pytest uses to figure out the import. So unless we want to maintain much, much more pytest code (subclass `pytest.Module` to replace the import logic), not much that can be done about this constraint.

```python
    def pypkgpath(self):
        """ return the Python package path by looking for the last
        directory upwards which still contains an __init__.py.
        Return None if a pkgpath can not be determined.
        """
        pkgpath = None
        for parent in self.parts(reverse=True):
            if parent.isdir():
                if not parent.join('__init__.py').exists():
                    break
                if not isimportable(parent.basename):
                    break
                pkgpath = parent
        return pkgpath
```

Kodiologist:
Add a bit more to the test to make sure we're importing the right file; for example, you could add `x = "hello"` to tlib.py and then check the value of `x` after the import. Otherwise, looks good to me.

PR #1498, opened 2018-02-07
Release 0.14.0

Kodiologist:
I have a lot of old PRs that I had hoped to get into this release, but reviews don't seem to be forthcoming, so there's no need to keep waiting.

Once this is merged, I'll take care of uploading the release files.


Kodiologist:
@hylang/core

PR #1499, opened 2018-02-08
Simplify conftest for loading hy modules

vodik:
Just a small cleanup - this works now since python 3.1.0, and already pin travis to use >= 3.2.1


Kodiologist:
> this works now since python 3.1.0

I presume you mean pytest.

vodik:
Yes, opps

PR #1500, opened 2018-02-08
Deal with some deprecation and resource warnings / fix (return) semantics

vodik:
Deal with almost everything except the deprecation of `inspect.getargspec`. I have code that deals with that as well, but I haven't made it fully py2/py3 compatible just yet. Since that's a larger change, I figured I'll put it up separately and put in smaller changes first.

There is one necessary AST generation change that needs to land to let us properly handle PEP 479, and something that I think that should be considered a bug to begin with: `(return)` transforms into `return None` when it should transform to just `return`.

This is because a naked `return` is valid in Python 2 as a way to exit a generator. `return` with a value isn't supported till Python 3.2 or 3.3, can't remember.

While its a deprecation warning now, its default behaviour come python 3.7 (which is in beta) so probably needs landing.

### DIFF
--- a/hy/inspect.py
+++ b/hy/inspect.py
@@ -1,0 +1,37 @@
+# Copyright 2018 the authors.
+# This file is part of Hy, which is free software licensed under the Expat
+# license. See the LICENSE.
+
+from __future__ import absolute_import
+
+import inspect
+
+try:
+    # Check if we have the newer inspect.signature available.
+    # Otherwise fallback to the legacy getargspec.
+    inspect.signature  # noqa
+except AttributeError:
+    def get_arity(fn):
+        return len(inspect.getargspec(fn)[0])
+
+    def has_kwargs(fn):
+        argspec = inspect.getargspec(fn)
+        return argspec.keywords is not None
+
+    def format_args(fn):
+        argspec = inspect.getargspec(fn)
+        return inspect.formatargspec(*argspec)
+
+else:
+    def get_arity(fn):
+        parameters = inspect.signature(fn).parameters
+        return sum(1 for param in parameters.values()
+                   if param.kind == param.POSITIONAL_OR_KEYWORD)
+
+    def has_kwargs(fn):
+        parameters = inspect.signature(fn).parameters
+        return any(param.kind == param.VAR_KEYWORD
+                   for param in parameters.values())
+
+    def format_args(fn):
+        return str(inspect.signature(fn))
--- a/tests/compilers/test_ast.py
+++ b/tests/compilers/test_ast.py
@@ -8,7 +8,7 @@ from __future__ import unicode_literals
 from hy import HyString
 from hy.models import HyObject
 from hy.compiler import hy_compile
-from hy.importer import import_buffer_to_hst
+from hy.importer import hy_eval, import_buffer_to_hst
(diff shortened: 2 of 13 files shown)


vodik:
All these warning came from pytest, so we could also turn that on in our CI build too, for what its worth.

tests/compilers/test_ast.py
```diff
@@ -664,3 +668,8 @@ def test_ast_good_yield_from():
 def test_ast_bad_yield_from():
     "Make sure AST can't compile invalid yield-from"
     cant_compile("(yield-from)")
+
+
+def test_eval_generator_with_return():
+    """Ensure generators with a return statement works."""
+    can_eval("(fn [] (yield 1) (yield 2) (return))")
```
vodik: Has to be evaluated as the outputted AST is fine. Python rejects it only on import.

Kodiologist:
You're on fire lately!

I had been under the impression that `return` and `return None` were interchangeable; mea culpa.

> All these warning came from pytest

I don't see any warnings when I run pytest. Do you have to turn them on somehow?

vodik:
> I don't see any warnings when I run pytest. Do you have to turn them on somehow?

`pytest -q` since about 3.1.0

Kodiologist:
`-q` didn't do it for me, but adding this passage to `setup.cfg` (mentioned in the linked page) did:

    filterwarnings =
        once::DeprecationWarning
        once::PendingDeprecationWarning

That change should probably be committed once we've tended to all the warnings.

vodik:
@Kodiologist looking at that pypy3 failure... did we maybe find a pypy bug?

vodik:
I've just dealt with everything and turned on the warning reporter.

hy/inspect.py
```diff
@@ -0,0 +1,37 @@
...
+    def get_arity(fn):
+        return len(inspect.getargspec(fn)[0])
+
+    def has_kwargs(fn):
+        argspec = inspect.getargspec(fn)
+        return argspec.keywords is not None
+
+    def format_args(fn):
+        argspec = inspect.getargspec(fn)
+        return inspect.formatargspec(*argspec)
```
Kodiologist: I think the `try` includes too much code. It would catch an `AttributeError` raised anywhere in those three functions.

Kodiologist: Or maybe not; maybe it's only in effect when the functions are defined, not called. Still, the code looks confusing in that way, so it seems ideal to just include that first statement in the `try` block.

Kodiologist:
Lookin' really good. Just see that one comment.

hy/inspect.py
Kodiologist: Easy solution: put everything after the first statment in an `else` block.

vodik: Done

Kodiologist: I don't see a change. Did you forget to push?

vodik: Yeah... ssh key wasn't loaded and didn't notice.

vodik:
Updated NEWS -> NEWS.rst, not sure what you want to put as a header for the next version.

Kodiologist:
You could use "Unreleased" as the new header.

vodik:
Done

PR #1502, opened 2018-02-09
Implement take-nth with islice

vodik:
I was wondering why `take-nth` was implemented by hand, but it turns out its because of a pypy bug. Its been fixed upstream (works on 5.10), but pypy3 on Travis is still on 5.8.

Not sure if this one will be taken, but I've added a workaround to detect if we're on pypy and keep using the kludge and use islice everywhere else.

### DIFF
--- a/hy/core/bootstrap.hy
+++ b/hy/core/bootstrap.hy
@@ -85,3 +85,10 @@
   (if (< (get sys.version_info 0) 3)
     python2-form
     python3-form))
+
+(defmacro if-pypy [pypy-form default-form]
+  "If running on pypy, execute pypy-form, else, execute the other form"
+  (import platform)
+  (if (= (.python-implementation platform) "PyPy")
+    pypy-form
+    default-form))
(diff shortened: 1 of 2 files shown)


Kodiologist:
Since the workaround also works with CPython, I don't think there's any need for this. You could just add a comment to the existing code pointing out that it's a workaround, or you could make a reminder for yourself to clean this up once Travis has upgraded its PyPy.

vodik:
Fair enough, I thought it be a stretch. But it did lead me to another idea. There's [PEP 508](https://www.python.org/dev/peps/pep-0508/#environment-markers)... what if we exposed that instead if `if-python2` and family? It would capture all that information and more.

Kodiologist:
`if-python2` is only for internal use. I don't think there would be much of a benefit to expanding it to handle PEP 508 in general, but I guess it's hard to forsee.

vodik:
Fair enough, I'll see how easy it is to implement as maybe there's already something exposed that handles most of the workload when I got some down time.

In the meanwhile, I'll withdraw this.

Issue #1501, opened 2018-02-08
Python 3.7 support


vodik:
Heh, instead of finishing the type annotation work, I've gone down the rabbit hole of seeing how it works on Python 3.7:

Known issues:

- [ ] importer is broken as the pyc header data structure has changed, adding a flags field to enable hash based checking on top of timestamp: PEP 552. Emitting another long of 0s preserved the legacy behaviour.
- [ ] docstrings are no longer the first statement of a function, they're stored in the AST: bpo-29463
- [ ] something broke in macro expansion (test_bin_hy_byte_compile breaks, not sure why)
- [ ] astor breaks because it uses `async` as a keyword argument https://github.com/berkerpeksag/astor/issues/86

Beyond that, all other unit tests passed.

Kodiologist:
> astor breaks because it uses `async` as a keyword argument

We should probably give the Astor folks an issue for that, if they don't already have one.

vodik:
Yeah, was planning on doing it later tonight. Just got some work stuff I need to wrap up first

Kodiologist:
We should consider Python 2.7 support, and testing of it on Travis, a blocker for 0.15.0. It'll be one of the first things I work on next.

PR #1503, opened 2018-02-09
Add python 3.7-dev to travis, allow it to fail

vodik:


vodik:
Starts off #1501, this gets the ball rolling that we can at least monitor progress towards full support.

vodik:
Well it frees us up from having to merge all changes at once. For example, I suspect the docstring AST change is going to be big enough to be worth considering separately.

Issue #1352, opened 2017-07-31
Perplexing keyword behavior in splice


gilch:
```Hy
(defn print-args [&rest args &kwargs kwargs]
  (print args kwargs))
```
```Python
=> (print-args :key 42)  ; normal keyword magic
print_args(key=42)
() {'key': 42}
=> (print-args ':key 42)  ; quote to pass keyword as data
from hy import HyKeyword
print_args(HyKeyword('\ufdd0:key'), 42)
('\ufdd0:key', 42) {}
```
That was strange for a Lisp, but easy enough to understand.

```Python
=> (eval `(print-args ~@[:key 42]))  ; raw keyword
from hy.core.language import eval
from hy import HyExpression, HySymbol
eval(HyExpression((([] + [HySymbol('print_args')]) + list(['\ufdd0:key', 42]))))
('\ufdd0:key', 42) {}
=> (eval `(print-args ~@[':key 42]))  ; quoted keyword
from hy.core.language import eval
from hy import HyExpression, HyKeyword, HySymbol
eval(HyExpression((([] + [HySymbol('print_args')]) + list([HyKeyword('\ufdd0:key'), 42]))))
() {'key': 42}
```
Somehow it became the opposite? Is this how it's supposed to work?

Kodiologist:
I think so. The trick is that in Hy, unlike most Lisps, keywords don't evaluate to themselves. They evaluate to strings, not `HyKeyword`s. So the expression `` `(print-args ~@[:key 42])`` evaluates to `(print-args "\\ufdd0:key" 42)`; if you evaluate this, the compiler will see the string `"\\ufdd0:key"` instead of a literal keyword. By contrast, `` `(print-args ~@[':key 42])`` evaluates to `(print-args :key 42)`. If you want to pass a real `HyKeyword` to `print-args` here, you need **two** single quotes (i.e., `` `(print-args ~@['':key 42])``) because you're evaluating `:key` twice, once by `~@` and once by the `eval`.

PR #1504, opened 2018-02-10
No implicit conversions between HySymbol and HyKeyword and str

vodik:
Related to #1363 

This is a very large diff, and totally something I'm partially expecting to be a no, but this branch makes the following comparisons now resolve to false:

```clojure
(= 'foo "foo")
(= :foo '\ufdd0:foo')
```

Maybe this is my hubris, but I haven't seen good use case to support it. Clojure, nor any other lisp dialect I've played with does this. And if anything, it makes things the Python compiler easier understand and reason about (granted its gotten a bit wordier, but there's room to refactor this nicely).

It also helps open the door to adding type annotation for extra safety and not having to worry about having a large union type everywhere.

And one last detail, keywords in the repl now look like this:

```
=> :foo
HyKeyword(':foo')
```

Would it maybe be desirable to use hy-repl here?

### DIFF
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -21,8 +21,8 @@ from hy.compiler import HyTypeError
 from hy.importer import (hy_eval, import_buffer_to_module,
                          import_file_to_ast, import_file_to_hst,
                          import_buffer_to_ast, import_buffer_to_hst)
-from hy.completer import completion
-from hy.completer import Completer
+from hy.completer import Completer, completion
+from hy.contrib.hy_repr import hy_repr
 
 from hy.errors import HyIOError
 
@@ -59,7 +59,7 @@ class HyREPL(code.InteractiveConsole):
         self.spy = spy
 
         if output_fn is None:
-            self.output_fn = repr
+            self.output_fn = hy_repr
         elif callable(output_fn):
             self.output_fn = output_fn
         else:
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -113,6 +113,10 @@ def builds(*types, **kwargs):
     return _dec
 
 
+def builds_symbol(*names, **kwargs):
+    return builds(*(HySymbol(name) for name in names), **kwargs)
+
+
 def spoof_positions(obj):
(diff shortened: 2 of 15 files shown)


Kodiologist:
> Maybe this is my hubris, but I haven't seen good use case to support it.

The loose equality was never deliberately chosen, I think. At any rate, we'll all be happy to see it gone, all other things being equal.

gilch:
> And one last detail, keywords

Related to #846, #1352?

vodik:
Keywords as arguments could fit in nicely anywhere positional only arguments are taken, which is thus far only in the standard library, but I can't think of a nice way of implementing it without adding function introspection overhead...

vodik:
Still working on some refactoring as I think I can do something nice to counteract the verbostiy of needing to specify HySymbol everywhere now explicitly.

@Kodiologist this now works in the repr as expected:
```clojure
=> `(print-args ~@[':key 42])
'(print_args :key 42)
```

Kodiologist:
I may not be very available for Hy stuff till Tuesday, thanks to the NIH grant deadline. Sorry about that.

vodik:
No worries, gives me plenty of time to chew on this slowly.

vodik:
I'm happy with this. I got some more work for dealing with all the tables, but I'll put that aside for now. I think we should get #1458 in first and rebase this on top of that.

vodik:
Couple of notes:

- I'll pull out the repl work and put it up separately.
- I think it makes sense to start treating HySymbol like how I handle HyKeyword.
- I've also got mypy coverage for most of the compiler and its working really well.

Kodiologist:
What's the status on this now? It looks like you finished it and then I completely neglected it, but now we have some conflicts, presumably from #1517, that might take work to unsnarl.

vodik:
Not neglecting, just don't have the free time. Hopefully will go through everything pending on the weekend.

Kodiologist:
In case you missed it, I was accusing **myself** of neglecting it, not you. ;)

PR #1505, opened 2018-02-11
Add metaclass support, support PEP 3115 and PEP 487

### LINKED ISSUE #855: Metaclasses under Python 3

vodik:
Fixes #855 (though I didn't think to search for the issue until after I implemented it...)

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -14,6 +14,7 @@ Other Breaking Changes
 * Non-shadow unary `=`, `is`, `<`, etc. now evaluate their argument
   instead of ignoring it. This change increases consistency a bit
   and makes accidental unary uses easier to notice.
+* `hy-repr` uses registered functions instead of methods
 
 New Features
 ------------------------------
@@ -22,10 +23,6 @@ New Features
 Bug Fixes
 ------------------------------
 * Fix `(return)` so it works correctly to exit a Python 2 generator
-
-Other Breaking Changes
------------------------------
-* `hy-repr` uses registered functions instead of methods
 
 Misc. Improvements
 ----------------------------
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -19,6 +19,8 @@ Other Breaking Changes
 New Features
 ------------------------------
 * Added `mangle` and `unmangle` as core functions
+* `defclass` in Python 3 now supports specifying metaclasses and other
+  keyword arguments
 
 Bug Fixes
 ------------------------------
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2057,11 +2057,12 @@ class HyASTCompiler(object):
 
         bases_expr = []
         bases = Result()
+        keywords = []
         if expressions:
             base_list = expressions.pop(0)
             if not isinstance(base_list, HyList):
                 raise HyTypeError(base_list, "Base classes must be a list.")
-            bases_expr, bases, _ = self._compile_collect(base_list)
+            bases_expr, bases, keywords = self._compile_collect(base_list, with_kwargs=PY3)
 
         body = Result()
 
@@ -2092,7 +2093,7 @@ class HyASTCompiler(object):
             expressions,
             decorator_list=[],
             name=ast_str(class_name),
-            keywords=[],
+            keywords=keywords,
             starargs=None,
             kwargs=None,
             bases=bases_expr,
--- a/tests/compilers/test_ast.py
+++ b/tests/compilers/test_ast.py
@@ -214,6 +214,13 @@ def test_ast_good_defclass():
     can_compile("(defclass a [])")
 
 
+@pytest.mark.skipif(not PY3, reason="Python 3 supports class keywords")
+def test_ast_good_defclass_with_metaclass():
+    "Make sure AST can compile valid defclass with keywords"
+    can_compile("(defclass a [:metaclass b])")
+    can_compile("(defclass a [:b c])")
+
+
 def test_ast_bad_defclass():
     "Make sure AST can't compile invalid defclass"
     cant_compile("(defclass)")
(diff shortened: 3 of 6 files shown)


hy/compiler.py
```diff
@@ -2071,14 +2071,35 @@ def rewire_init(expr):
...
+            if not isinstance(keywords[0].value, ast.Name):
+                raise HyTypeError(base_list, "specified metaclass must be "
+                                             "a symbol")
+
+            e = HyExpression([
+                    HySymbol("setv"),
+                    HySymbol("__metaclass__"),
+                    HySymbol(keywords[0].value.id)
+                ]).replace(base_list)
+            spoof_positions(e)
```
Kodiologist: Do you need both `.replace` and `spoof_positions` here? Shouldn't one be enough?

tests/compilers/test_ast.py
```diff
@@ -217,6 +218,12 @@ def test_ast_bad_defclass():
     cant_compile("(defclass a None None)")
 
 
+@pytest.mark.skipif(PY3, reason="Python 3 supports class keywords")
+def test_ast_bad_defclass_with_kwargs():
+    "Make sure AST can't compile invalid yield-from"
```
Kodiologist: The comment doesn't match the test.

tests/native_tests/test_metaclass.hy
```diff
@@ -0,0 +1,15 @@
...
+
+(defn test-metaclass []
+  (defclass QuestBase [type]
+     [--new-- (fn [cls name bases classdict]
+               (if (in "__swallow__" classdict)
+                   (assoc classdict "swallow" (get classdict "__swallow__")))
+               (-> (super QuestBase cls) (.--new-- cls name bases classdict)))])
+
+  (defclass Quest [:metaclass QuestBase]
+    [--swallow-- "african"])
```
Kodiologist: Why the use of double underscores? `__swallow__` isn't a special Python name, which is what double underscores on both sides of a name are typically reserved for, and it doesn't trigger the automatic Python mangling that a name like `__swallow` would, which we probably don't need for this test, anyway.

Kodiologist: If all you want to do is distinguish `swallow` from a similar but distinct variable, how about `swallow_`?

hy/compiler.py
vodik: Yeah, I thought it was weird when I looked at it, `replace` should be enough and idea. But I copy and pasted the approach and elsewhere both are used:

https://github.com/hylang/hy/blob/10cd36d13d71010c38e8b411ebf13390f80b9542/hy/compiler.py#L414-L424

tests/native_tests/test_metaclass.hy
vodik: No good reason. Honestly, I'm not married to this particular test case at all. The other two came from their respective PEP's, this one I derived.

hy/compiler.py
Kodiologist: Such cases might also be redundant, couldn't they? (Not that I'm expecting you to fix them in this PR.)

vodik: Yeah, agreed. Just pointing it out so we know. I'm certainly going to investigate it.

hy/compiler.py
```diff
@@ -2071,14 +2071,35 @@ def rewire_init(expr):
...
             base_list = expressions.pop(0)
             if not isinstance(base_list, HyList):
                 raise HyTypeError(base_list, "Base classes must be a list.")
-            bases_expr, bases, _ = self._compile_collect(base_list)
+            bases_expr, bases, keywords = self._compile_collect(base_list, with_kwargs=True)
 
         body = Result()
 
+        if not PY3 and keywords:
+            # Dirty shim for Python 2 to make :metaclass keyword work.
```
gilch: I'm not sure if this is worth the added complexity to the compiler. We can already use metaclasses in Python2 using `(setv __metaclass__ foo)`. Our Python2 support is ending in 2020 at the latest. It's fine to fix bugs, but we shouldn't be adding new features for it.

vodik: I'd be fine with this if everyone else is.

Kodiologist: I can confirm that I would be fine with that.

Kodiologist:
Looks good. Why not put the new tests in `py3_only_tests.hy` and `py36_only_tests.hy` instead of their own files?

vodik:
Just taste. pytest doesn't require that we have to put them all together and I like having the file name reflect what's inside of it.

Can do though, just a sec.

Kodiologist:
It's only two new tests, so it seems a little silly to make two new files. When (or if) the version-specific files get large, then we should start breaking them up, by all means. We already have some test files that are probably too big, like language.hy and core.hy.

vodik:
No worries, I get it.

Kodiologist:
@gilch, do you approve?

NEWS.rst
```diff
@@ -1,5 +1,13 @@
 .. default-role:: code
 
+Unreleased
+==============================
+
+New Features
+------------------------------
+* `defclass` can now take keyword arguments for metaclass for
+  Python 3 metaclass support.
```
gilch: "for metaclass for Python 3 metaclass support"? I think this could be worded better.

vodik:
Fixed up NEWS... finally

Kodiologist:
@gilch, do you approve?

gilch:
Approved. We still have a NEWS conflict though.

Issue #1506, opened 2018-02-12
`with-decorator` should work with lambdas


vodik:
If the preferred way of writing is with the attribute list, then we probably want to make `with-decorator` work with lambda by forcing a function to be defined instead (rewrite `fn` -> `fn*`).  Otherwise we'll trip up around `classmethod` (and `property`, technically, even though its more natural to construct it directly).

For example:

```clojure
(defclass Foo []
  [from-string #@(classmethod (fn [cls str] ...))])
```

Fails with `HyTypeError: Decorated a non-function`

See #855 where I had to workaround this.

vodik:
I have a patch for this, should we agree its desirable, but I'll wait for #1504 so its significantly cleaner.

vodik:
Actually, another option is to drop lambdas altogether and only ever emit functions. They have the same performance characteristics anyways...

Kodiologist:
Decorator syntax is not necessary in this case. You can write simply

    (defclass Foo []
      [from-string (classmethod (fn [cls str] ...))])

vodik:
Should I close this then?

Kodiologist:
Works for me.

PR #1507, opened 2018-02-12
Never generate lambas, only function definitions

vodik:
There is no downside to this change:

- Lambdas and functions has identical performance characteristics on python 3 (and functions are slightly faster on python 2).
- It simplifies the compiler slightly.
- And there should be less surprises with the generated code (decorating lambas, for example, see #1506 ).

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1922,15 +1922,11 @@ class HyASTCompiler(object):
 
         return ret
 
-    @builds("fn", "fn*")
+    @builds("fn")
     @builds("fn/a", iff=PY35)
-    # The starred version is for internal use (particularly, in the
-    # definition of `defn`). It ensures that a FunctionDef is
-    # produced rather than a Lambda.
     @checkargs(min=1)
     def compile_function_def(self, expression):
         root = expression.pop(0)
-        force_functiondef = root in ("fn*", "fn/a")
         asyncdef = root == "fn/a"
 
         arglist = expression.pop(0)
@@ -1998,10 +1994,6 @@ class HyASTCompiler(object):
             defaults=defaults)
 
         body = self._compile_branch(expression)
-        if not force_functiondef and not body.stmts:
-            ret += asty.Lambda(expression, args=args, body=body.force_expr)
-            return ret
-
         if body.expr:
             if body.contains_yield and not PY3:
                 # Prior to PEP 380 (introduced in Python 3.3)
--- a/hy/contrib/multi.hy
+++ b/hy/contrib/multi.hy
@@ -93,4 +93,4 @@
      ret)
     (do
      (setv [lambda-list body] (head-tail bodies))
-     `(setv ~name (fn* ~lambda-list ~@body)))))
+     `(setv ~name (fn ~lambda-list ~@body)))))
--- a/hy/contrib/walk.hy
+++ b/hy/contrib/walk.hy
@@ -246,7 +246,7 @@ Arguments without a header are under None.
   ;; and local bindings should shadow those made by let.
   (defn handle-call [self]
     (setv head (first self.form))
-    (if (in head '[fn fn*]) (self.handle-fn)
+    (if (in head '[fn fn]) (self.handle-fn)
         (in head '[import
                    require
                    quote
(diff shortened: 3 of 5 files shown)


Kodiologist:
If I'm not mistaken, this complicates the generated code: what could be a lambda will now take an extra assignment. Also, I'd like Hy to be able to generate every Python construct, even if it's not strictly necessary to do so (Python has a lot of constructs that are semantically redundant). Given, we don't have that yet (we're missing, e.g., variable annotations), but if we eliminate lambdas, we're taking a step back.

tuturto:
But on the other hand, it simplifies code generation. In this case, that's a trade off I would be willing to make. If lambda is supported, we need to work around some special cases like decorating it (#1506).

Kodiologist:
But it doesn't make sense to decorate a `fn`. You can just use a regular function call for that. You can decorate a `defn`, but that's already guaranteed not to produce a lambda.

vodik:
But I think its weird that `fn` then *may* generate a function, if it detects it contains statements. Its certainly deterministic, but its an internal that leaks outwards.

As far as I'm concerned, conceptually, `fn` generates a function but may choose to pick the more "optimal" representation of a lambda. But when you can't use a lambda everywhere you'd use a function, small weird things arise.

Anyways, there is another approach to the other issue which I have a proof of concept: have a mechanism to promote `fn` to `fn*`.

Kodiologist:
> As far as I'm concerned, conceptually, `fn` generates a function but may choose to pick the more "optimal" representation of a lambda.

It may be more useful to think of it the other way round. `fn` produces an anonymous function and returns it. That's the reason we have `fn`. If you want to assign your function a name from the beginning, instead of passing it as an argument or putting it in a list or something, you should probably be using `defn`. It is an implementation detail that sometimes `fn` needs to produce a Python `def` statement because of Python's limitations on what can go in a `lambda`.

vodik:
Okay, so this is going to be a little on the stream of consciousness side, I apologize in advance.

Right, let me be clear, I understand that is how its intended to be interpreted, and its how the docs present it as well.

> But it doesn't make sense to decorate a `fn`...

But there's also a lot of effort put into rewriting `setv` + `fn`. This example:

```clojure
(setv foo (with-decorator bar (fn [x] (import x) x)))
```

emits equivalent code to:

```python
@bar
def foo(x):
    import x
    return x
```

Treating Hy like a black box, when I seeing this much effort in emitting optimal AST, I can't help but start to think of it the other way: that lambda is the optimization.

If we're going to say `fn` shouldn't be decorated, maybe we shouldn't let it be decorated at all. But I honestly don't see anything wrong with allowing `fn` to be decorated:

- We can force a function expansion
- We can make `with-decorator` work with function calls instead of decorator syntax when its a lambda (maybe this is the best solution `#@(classmethod my-decorator foobar) -> classmethod(my_decorator(lambda ...))`

> Python has a lot of constructs that are semantically redundant

My understanding is that that they're literally the same thing since Python 3 - internally at least.

> Also, I'd like Hy to be able to generate every Python construct, even if it's not strictly necessary to do so

Well, what about if we promoted `#%` and supported it first party to generate lambda's instead like in Clojure?

Anyways, its no hill I want to die on, what do you guys think about my ideas on augmenting `with-decorator` then?

gilch:
I'm thinking of putting decorators inside the `defn` form itself, instead of wrapping them in another special form. #952. But `fn` wouldn't have this.

Kodiologist:
> But there's also a lot of effort put into rewriting `setv` + `fn`

This might have been a mistake. I mean, it would be more intuitive if `defn` were a special form, replacing `fn*`, rather than `fn*` and `defn` both existing just to guarantee the creation of a Python `def`. I'm not sure how we got here. Full disclosure: I created `fn*`, but that was as part of a change from an earlier even less sensible design, which included `lambda` as a separate special form (#1228; see also the follow-up #1259).

> Well, what about if we promoted `#%` and supported it first party to generate lambda's instead like in Clojure?

I think that would just pass the buck, in a sense: like `fn` does now, `#%` would be able to produce a lambda in some cases and forced to produce in a `def` in others.

> Anyways, its no hill I want to die on, what do you guys think about my ideas on augmenting `with-decorator` then?

You can see how in #1228 I generalized `with-decorator` only to revert it in #1259. Since it's only necessary for `def` blocks, it makes sense to only support `def` blocks.

In conclusion, it looks like what got us down this path was the fact that `(#@(foo (fn …)))` works inconsistently. The easiest solution to that would be to make `#@` or `with-decorator` raise an error saying that you can just write `(foo (fn …))`.

Issue #1044, opened 2016-04-13
set up Sphinx autodoc to put our docstrings in our reference docs.


gilch:
There is considerable duplication of documentation between the reStructuredText docs and Hy's internal docstrings. I propose we move as much as reasonably possible into the docstrings themselves, to make the references more accessible at the repl, and for language tools like emacs. @jkitchin has started this effort with #1040.

I think Sphinx can fetch docstrings [programmatically](http://www.sphinx-doc.org/en/stable/ext/autodoc.html) from Python objects rather than parsing them from text. This means that it could theoretically work for Hy docstrings as well, though I've never seen this done before.

Our documentation effort is lagging behind Hy's actual capabilities. Part of this is due to the extra effort required to document things in two places, and then keep them in sync afterwards.

The new setup will require considerable reorganization of the documentation. For each function (and potentially class, module etc.) we'll have to take the better of the two or combine them somehow.

Our testing is similarly lagging behind. While Hy's doctests shouldn't completely replace unit tests (the priority must be understandable examples, rather than completeness), it is better than the nothing we sometimes get. More importantly, making our reference examples into doctests in our docstrings will ensure that they actually work as intended. Currently, many of our example docs are broken. They should all be automatically testable, even those from the tutorial.

Doctests show a Python repl session and don't currently work with a Hy repl, but we need our Hy examples to be in Hy, not Python. See #1019.

Hy lacks Python's (raw) triple quotes. Hy's strings can certainly represent any possible text given enough backslashes, but this will make it difficult to write reStructuredText (which also needs backslash escapes) and Hy doctests (which will certainly use quoted strings themselves) in our docstrings. I've proposed adding heredocs as string literals to the reader (#1042) to resolve this.

jkitchin:
I am not sure you can get the autodoc feature of Sphinx to work with hy the way it does with Python. There is a lot of redirection between hy functions/macros and where the doc strings are. For example, here https://github.com/jkitchin/hyve/blob/master/hylp.hy#L103 I expanded some code you posted earlier to get the Python object for a given hy symbol, so that I can use the python tools for accessing docstrings, source files and linenumbers. I am pretty sure Sphinx would see things from the Python side in using the autodoc features.

On the other hand it is not that difficult to write code that would generate html docs from these docstrings. pydoc already has a lot of this framework. So there probably could be some code in the Sphinx docs that generates documentation for the hy functions and macros.

[...]

Issue #957, opened 2015-10-13
Can you make it possible to contribute with money?


ghost:
I'm sure I am not the only one that would like to help out.

Foxboron:
Interesting request. CC: @paultag @olasd 

Side note: Hei fra Bergen!

gilch:
The question is how to distribute it fairly. I'm sure we're not the only project with this "problem", so maybe somebody has solved it. I stumbled across https://whispersystems.org/blog/bithub/ recently, which looks like an option. I have no idea how well it works in practice because it seems like it might create some perverse incentives.

Kodiologist:
I guess the simplest approach would be to give money to @paultag and leave it up to him how to spend it. Right now, I don't think we have any hosting expenses (since we're using gratis sites like GitHub), and we can't expect to get enough donations to support a developer working full-time (that would be something like 15k USD a year at minimum). Maybe we could use money to send developers to conferences.

asemic-horizon:
I previously offered to help draft a business model in a couple months' time when I can give it my full attention. But to use procrastination time productively, here's some informal notes that do not amount to expert advice (the standard disclaimer - I'm not currently consulting for Hy)

Because nonprofits are not naturally interfacing with market forces, they kind of have to "give shape to their own supply and demand". I'm using informal language so this doesn't come of as formal advice. But loosely:

* Supply forces -- here we have costs, developer time, etc. At first blush I don't think a full-time developer would do; rather, we'd have to review who are the stakeholders and who are the significant Hy hackers to figure out with them a model that would allow them to spend more time with Hy. In a formal project this would be preceded by an industry benchmark where we'd look at Clojure and Julia etc. to figure out what they're doing.

* Demand forces -- here we have raising awareness and establishing the importance of the project. For charities etc this involves a lot of fuzzy stuff but here we have two core groups to woo, (1) potential Hy users and (2) researchers/hackers who can get into Hy language development right away. Kodiologist very wisely said "send developers to conferences" -- if you can get them to give talks this would be GREAT. It also gives us focal points for fundraisers that obviate the need to structure an org for the time being.

Those are some random droppings from what might be something organized with spreadsheets and slide decks and organizational structure proposals after lots of data gathering yadda yadda yadda -- I could, in a professional (even if donated) capacity look into what open source funders such as NumFOCUS are doing and prepare everything.  But in the short term, Kodiologist has already identified a critical point.

[...]

PR #1508, opened 2018-02-14
Policy: allow PRs to be merged more easily

Kodiologist:
@hylang/core

It's not a secret that in the year and a half I've been working on Hy, I've been frustrated by how slowly PRs get reviewed and merged. I think the current policy of requiring two core-team reviewers other than the author would be fine if the core team were more active. But the core team isn't very active. It's nobody's fault; it's just that we're often busy with our jobs and other responsibilities.

I'm proposing a policy change that allows for an alternative way for a PR to be cleared to merge: if it's at least **two weeks** old and **no** member of the Hy core team has expressed disapproval of the PR in its current state. This means that the author can merge the PR without help if he's a core member, or with only one core member's help if he isn't. At the same time, any core member who has concerns with the PR can effectively veto this method (and later rescind the veto if the PR is changed to his satisfaction).

I realize this is a less conservative approach than we've taken in the past and hence may result in a less stable Hy. But I think an approach leaning more towards the move-fast-and-break-things end of the spectrum could be what Hy needs to speed up progress on the long road to 1.0 after more than 5 years of development. The potential for things to go wrong will be checked by the veto part of the rule, as well as the fact that PRs that simply undo previous PRs can be merged through the same policy.

If this policy is approved, let it not be applied retroactively to preexisting PRs. But, authors are free to close such PRs and make new ones with substantially the same commits to take advantage of the policy.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -120,15 +120,20 @@ handy. Conversely, avoid this exact language if you want to mention
 an issue without closing it (because e.g. you've partly but not
 entirely fixed a bug).
 
-Before any PR is merged, it must be approved by members of Hy's core team
-other than the PR's author. Changes to the documentation, or trivial changes
-to code, need to be approved by **one** member; all other PRs need to be
-approved by **two** members.
+There are two situations in which a PR is allowed to be merged:
 
-Anybody on the team may perform the merge. Merging should create a merge commit
-(don't squash unnecessarily, because that would remove separation between
-logically separate commits, and don't fast-forward, because that would throw
-away the history of the commits as a separate branch), which should
+1. When it is approved by **two** members of Hy's core team other than the PR's
+   author. Changes to the documentation, or trivial changes to code, need only
+   **one** approving member.
+2. When the PR is at least **two weeks** old and **no** member of the Hy core
+   team has expressed disapproval of the PR in its current state. (Exception: a
+   PR to create a new release is not eligible to be merged under this criterion,
+   only the first one.)
+
+Anybody on the Hy core team may perform the merge. Merging should create a
+merge commit (don't squash unnecessarily, because that would remove separation
+between logically separate commits, and don't fast-forward, because that would
+throw away the history of the commits as a separate branch), which should
 include the PR number in the commit message.
 
 Contributor Code of Conduct


berkerpeksag:
+1 from me, but I'd make the threshold a month rather than two weeks (or require one approval after two weeks and no approval needed after a month)

tuturto:
+1 and two weeks is fine.

gilch:
I'd prefer 1 month.

Kodiologist:
A month seems really slow, but this policy with a one-month delay is still much better than the status quo.

gilch:
There are probably more things we can do to speed up Hy development.

We could add more core members. @vodik seems like the next candidate, considering recent activity. 

We could add an experimental branch that takes only one approval. Merging that in to master would be more difficult, but less frequent. @hylang/core would a bigger, more difficult, but less frequent review process fit your schedules better?

We could try to drum up more interest in Hy via social media, etc. to get more developers. More cool open-source projects using Hy would help, and they don't require approval. We might even consider a little bit of paid advertising. Some have expressed interest in donations #957 and that's a possible use for it, even if it's too small to pay anyone to work on Hy full time. There are various ways to set this up, e.g. Patreon.

We could drop support for Python 2 early, or at least threaten to. #1217 This would simplify the compiler a lot, but if developers care enough to protest, we could perhaps recruit them. @ddkn @Rogof any interest in working on Hy issues? Know anyone else who cares?

We could prioritize tooling to help with PRs, like a Hy linter and formatter and a Sphinx autodoc plugin #1044. Updating Hydiomatic and adding it to Travis or something might help.

@hylang/core what is holding you back? Is there any process improvement that could address this?

tuturto:
Having to review bigger changes would probably mean that I would be reviewing them even less frequently. At least I'm having hard time finding enough time to really look into changes introduced in any given PR and understand what implications they might have.

Having more members in the core would be a good idea.

ddkn:
@gilch I would be interested in popping through some bugs. However, at the moment I have been busy with the end of a MSc. 

Currently was going to port over to OpenBSD in the next week or two, as this is my daily driver now -- and they conveniently have flavours built in to ports, which mean packages for both v2.7+, and v3+. Might get some more interest that way.

I cannot promise to commit anything any time soon until I finish what I am currently doing, but I will try. My end date is August, and should see more activity after that point.

Kodiologist:
It looks like there's acceptance of the policy, but we're 2-to-2 on whether the delay should be two weeks or one month. @kirbyfan64 @rkday would you care to weigh in and break the tie?

kirbyfan64:
What about:

- 1 month if no one has expressed disapproval.
- 2 weeks if someone has approved, but no one else has yet.

rkday:
I'm keen on two weeks - it's nice to be faster-moving (and having fewer open pull requests gives new contributors a better impression of the project - I'm much less inclined to contribute to projects where it looks like PRs just become stale).

"We could add an experimental branch that takes only one approval. Merging that in to master would be more difficult, but less frequent."

Maybe some variation on this idea? For example, PRs can be merged after two weeks with only one core member's approval, but every PR needs to get two approvals before a Hy release (which may be a more efficient "if this was going to break anything, it would have by now". That keeps the master branch faster-moving, but doesn't reduce the quality of actual releases.

Kodiologist:
We now have 3 votes for a two-week delay versus 3 votes for a one-month delay (2 of which have the caveat that a two-week delay could be okay if there's one approval). 

Regarding the compromise Rob offers, I'm afraid that reviewers would have little interest in going back and indicating approval of old PRs that have been in master for a while, and releases would never happen. But I appreciate the point that we should be more conservative about releases than about master. So how about we keep the 2-week interval, but say that release PRs themselves are not eligible for the "absence of disapproval" method, so they need two approvals other than the author, per the traditional policy?

berkerpeksag:
+1

Kodiologist:
Okay, great. That will be our new policy.

Issue #1134, opened 2016-10-12
Circular import fails


Kodiologist:
With Python, this works:

```
$ echo 'import b; print("in a")' > a.py
$ echo 'import a; print("in b")' > b.py
$ echo 'import a, b; print("done")' > main.py
$ python main.py 
in b
in a
done
```

The equivalent with Hy doesn't work:

```
$ echo '(import b) (print "in a")' > a.hy            
$ echo '(import a) (print "in b")' > b.hy
$ echo '(import a b) (print "done")' > main.hy
$ hy main.hy
Traceback (most recent call last):
  ...
  File "/tmp/b.hy", line 1, in <module>
    (import a) (print "in b")
ImportError: import of 'a' halted; None in sys.modules
```

One consequence of this is that an `__init__.hy` file can't import other files in its own package, as `__init__.py` files usually do.

Kodiologist:
Self-import doesn't work, either. Compare

    a = 1

    def f():
       import foo
       return foo.a

    print(f())

to

    (setv a 1)

    (defn f []
      (import foo)
      foo.a)

    (print (f))

Issue #1512, opened 2018-02-16
`load_module` is deprecated (since 3.4)


vodik:
https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.load_module

> Deprecated since version 3.4: The recommended API for loading a module is exec_module() (and create_module()). Loaders should implement it instead of load_module(). The import machinery takes care of all the other responsibilities of load_module() when exec_module() is implemented.

vodik:
And as a heads up, the `Finder` API (what we've incorrectly named `MetaImporter`) as also deprecated. Looking into this.

Kodiologist:
> And as a heads up, the `Finder` API (what we've incorrectly named `MetaImporter`) as also deprecated.

#1085 and #1188 might be related.

vodik:
I'm knee deep in this now, and I think there's some really interesting potential with a reworked importer... I got some ideas I'll put up in a proposal, if this work pans out.

vodik:
Yeah, digging into this and #1513 got me looking under the hood and I think I've been effectively nerd snipped. 😉 

I've got something that's working on both python 2 and python 3, using only 3.4 and newer `importlib` API on python 3, and much simpler than the reverted polyloader.

I've also took the opportunity to deal with some of other bugs like #1466. The following application:

```clojure
"Tool to dump __spec__, __loader__, and other metadata"
(print "name:" __name__)
(print "doc:" __doc__)
(print "package:" __package__)
(print "loader:" __loader__)
(print "spec:" __spec__)
(print "file:" __file__)
(print "cached:" __cached__)
```
prints
```
$ hy .
name: <run_path>
doc: Test tool to dump __spec__, __loader__, and other metadata
package: 
loader: <hy.importlib.bootstrap.HyLoader object at 0x7f7c2ff00c88>
spec: ModuleSpec(name='__main__', loader=<hy.importlib.bootstrap.HyLoader object at 0x7f7c2ff00c88>, origin='/home/simon/src/hy/dumpers/__main__.hy')
file: /home/simon/src/hy/dumpers/__main__.hy
cached: /home/simon/src/hy/dumpers/__pycache__/__main__.cpython-36.hyc
```

(I'm also playing with changing the bytecode cache extension to `.hyc` at the moment. I won't try to get that merged in - well, unless you guys like that idea - but I've got some ideas on some extra information we can store in the cache that would justify changing the extension).

vodik:
Going to close this in favour of a bigger issue.

Issue #1482, opened 2018-01-05
Hy metadata proposal


gilch:
Clojure has a metadata system that applies an arbitrary map to symbols and its built in collection types (i.e. the kind code is made of). This is used for, among other things, type hints.

The `^` reader macro (formerly `#^`), applied to a mapping, followed by a compatible object (symbol or coll) attaches that mapping to the object as metadata. E.g. `^{:spam "eggs"} foo` attaches the metadata mapping `{:spam "eggs"}` to the symbol object `foo`.

There are shortcuts. Applying `^` to a keyword adds it as a flag, e.g. `^:spam foo` is the same as `^{:spam true} foo`. Applying `^` to a string adds it under the `:tag` key, e.g. `^"spam" foo` is the same as `^{:tag "spam"} foo`.

You can also attach metadata more than once. E.g. `^:x ^:y foo` is the same as `^{:x true, :y true} foo` i.e. the mappings are merged. Clojure mappings are immutable, so it returns a new symbol with different metadata.

Clojure metadata does not impact equality. The symbol `foo` is equal to the symbol `foo` with metadata attached. But they are different objects.

In Hy, I propose adding metadata to all Hy model objects. These would usually be unavailable at runtime, since the compiler discards the model objects after translating them to Python AST. But macros could use them, and so could the compiler. The line/column number we use for the debugger and stack trace would then be metadata.

And we'd also use them for function and variable annotations. Anything under `:tag` would end up compiled in the AST and thereby `__annotations__`. We could generalize the shortcut so that not just strings, but anything besides a mapping or a keyword gets put under `:tag`.

Kodiologist:
At this point, I'm reasonably sure such a feature would be overkill. Ordinary attributes for model objects work fine. The applications for arbitrary user-defined metadata on models are few, and supporting it would require both the compiler and macros to be very careful about the identity of the models they generate; it wouldn't be enough to return a model that's equal to what's desired.

Issue #1466, opened 2017-12-16
__main__.hy doesn't work like __main__.py


Zirak:
(All snippets below tested with hy master and python 3.6.3)

When running a directory with a `__main__` file, hy fails while python succeeds:

    % ls
    __main__.hy __main__.py
    % python .
    hey
    % hy .
    hy: Can't open file '.': [Errno 21] Is a directory

Same applies to egg files, but that may be its own independent problem.

Passing the `-m` flag does execute the `__main__`, *unlike python*:

    % hy -m .
    hy
    % python -m .
    python: Relative module names not supported

Fixing the first block is relatively simple: In [`cmdline`, `run_module`](https://github.com/hylang/hy/blob/e0e664c030266ead171de2a7fe887ceff4832d32/hy/cmdline.py#L199) already does the correct thing when running a directory/file according to [the respective function in cpython](https://github.com/python/cpython/blob/9454060e84a669dde63824d9e2fcaf295e34f687/Modules/main.c#L1439).

However, making the 2nd block fail is a breaking change, as `run_module` does not do the same thing as [the respective function in cpython](https://github.com/python/cpython/blob/9454060e84a669dde63824d9e2fcaf295e34f687/Modules/main.c#L228) (wraps [`runpy._run_module_as_main`](https://github.com/python/cpython/blob/9454060e84a669dde63824d9e2fcaf295e34f687/Lib/runpy.py#L167))

In the end, the question may be is this worth fixing on a lower level? Does hy want to copy the behaviour of `run_file` and `run_module` as implemented in cpython, or is it good enough to be able to run a directory/zip/whatever with a `__main__`?

Kodiologist:
Huh, I didn't even know this Python feature existed. #459 might be related. Thanks for reporting.

PR #1513, opened 2018-02-16
Fix self imports

vodik:
Make sure the proper `sys.modules` entry it created as soon as the module is loaded.

This was just a subtle bug - self import only worked so long as there was a module cache miss.

I need this fixed for #1482, as I'm attaching metadata to the module when appropriate and I'd like to be able to read it (like how `__annotations__` works).

I'll have a look at #1512 later.

### DIFF
--- a/hy/importer.py
+++ b/hy/importer.py
@@ -98,7 +98,7 @@ def import_file_to_module(module_name, fpath, loader=None):
         try:
             _ast = import_file_to_ast(fpath, module_name)
             module = imp.new_module(module_name)
-            module.__file__ = os.path.normpath(fpath)
+            sys.modules[module_name] = module
             code = ast_compile(_ast, fpath, "exec")
             if not os.environ.get('PYTHONDONTWRITEBYTECODE'):
                 try:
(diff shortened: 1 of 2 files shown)


Kodiologist:
Does this change anything with respect to #459 or #1134?

vodik:
Fixes #1134. Not sure about #459 - while I don't touch sys.executable, I haven't seen `__name__` to be incorrect yet anywhere.

Kodiologist:
> Not sure

The other issues referencing that one have some example code you could try.

vodik:
Yeah, the code in #1134 works now. I assume you're referring to:
```
$ echo '(import b) (print "in a")' > a.hy            
$ echo '(import a) (print "in b")' > b.hy
$ echo '(import a b) (print "done")' > main.hy
$ hy main.hy
  ...
  File "/tmp/b.hy", line 1, in <module>
    (import a) (print "in b")
ImportError: import of 'a' halted; None in sys.modules
```

vodik:
I've got a large refactoring of the importer to properly support python2, python3, and properly implement the necessary APIs that allow `runpy` to work on both platforms (which in turn solves #1466)

PR #1516, opened 2018-02-24
hy-repr improvements

Kodiologist:
Reopened for the new merge policy.

### DIFF
--- a/hy/contrib/hy_repr.hy
+++ b/hy/contrib/hy_repr.hy
@@ -6,6 +6,7 @@
   [math [isnan]]
   re
   datetime
+  collections
   [hy._compat [PY3 PY36 str-type bytes-type long-type]]
   [hy.models [HyObject HyExpression HySymbol HyKeyword HyInteger HyFloat HyComplex HyList HyDict HySet HyString HyBytes]])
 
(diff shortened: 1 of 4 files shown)


hy/contrib/hy_repr.hy
```diff
@@ -4,95 +4,161 @@
...
+  (hy-repr-register types :placeholder fmt (mkrepr fmt)))
+
+(defn -cat [obj]
+  (.join " " (map hy-repr obj)))
 
-(defn base-repr [x]
+(defn -base-repr [x]
   (unless (instance? HyObject x)
     (return (repr x)))
   ; Call (.repr x) using the first class of x that doesn't inherit from
```
vodik: Why not check for `__hy_repr__` as well? I don't think defining that as a protocol was a bad idea, tbh.

Kodiologist: It seems an unnecessary complication. It would bring up issues such as what code gets run when both a `__hy_repr__` and a registered function apply.

vodik: What if you just threw an exception registration in there's a `__hy_repr__`?

Not a strong opinion either way, but `__hy_repr__` feels like it fits into the Python way a bit better, and its easier for first class support, leaving registration for everything else.

Kodiologist: I doubt it would be that simple—for example, there's nothing stopping somebody from monkey-patching a `__hy_repr__` onto a class after a function for it has been registered.

A priori, `__hy_repr__` does make more sense than registered functions, but Hy is a second-class citizen in the Python world, and so a Hy programmer is likely to want a Hy representation of a type from a package written by somebody who doesn't care about Hy.

Kodiologist: If you're crazy, by the way, you can always register a function for `object` that looks for `__hy_repr__`. ;)

vodik: Good points. I'm comfortable with registration, so don't let me stop this...

Kodiologist:
That was the 777th closed PR, for those of you keeping score at home.

Issue #1324, opened 2017-07-15
.pyc files don't get recreated when they use a macro from a file that was changed


paultag:
It seems like Hy is using the .pyc cache even if the file we're loading has a newer filesystem mtime than the .pyc file -- Python will default to overwriting the pyc.

This winds up biting people who update a script and wind up very confused about why something isn't being invoked :)

(Just bit me on the version on pypi - anyone know if this is being tracked with another issue?)

Kodiologist:
I can't reproduce this with Hy 0.13.0:

    > echo '(print "bye")' > foo.hy
    > hy foo.hy 
    hello
    > hy foo.hy
    hello
    > echo '(print "bye")' > foo.hy 
    > hy foo.hy                     
    byte

gilch:
I'd start looking here https://github.com/hylang/hy/blob/master/hy/importer.py#L57

@paultag, a reliable reproduction guide would be helpful. There was a recent atomation of hyc #1094 #1269.

paultag:
Weird. The above works for me too, on `hy 0.13.0`. The exact issue is pretty tied up to some code, but there's a few macros and imports happening all over the place. I'm not sure what's going on, exactly yet.

I can't get a minimal reproduction at the moment - I'm seeing weird stuff happening, but I did see this happen in the code, and removing the exact .pyc from the cache caused it to recompile the code and start working. I lost a few minutes on that.

Let me try to figure out why it's doing this

paultag:
Oh crap I know why.

I have a macro in a file and I was only changing the macro. I bet that when you have a macro in one file, require it in another, and generate the file off the macro, it will cache the pyc, and modifying the original file won't cause anything using that macro to update.

paultag:
Writing a reproduction now

gilch:
Is this related to #1268?

paultag:
Sorta, this would still be an issue if that was fixed

paultag:
```
(projector) [paultag@nyx:~/x][05:36 PM] ♥  cat api.hy 
(defmacro yeah []
  (import mmm)
  (mmm.generate-function))
(yeah)
(projector) [paultag@nyx:~/x][05:36 PM] ♥  cat mmm.hy 
(defn generate-function [] `(defn yeah [] (print "hi")))
(projector) [paultag@nyx:~/x][05:36 PM] ♥  cat main.py 
import hy
import api
api.yeah()
[...]

PR #1517, opened 2018-02-26
Mangling makeover

### LINKED ISSUE #360: Find a way to keep the "original" symbol before the interop munging

### LINKED ISSUE #970: add option to hy2py that converts symbols to valid python identifiers

### LINKED ISSUE #1115: clean up rules for `?`

### LINKED ISSUE #1168: Should keywords be mangled?

### LINKED ISSUE #1173: Rethink mangling

### LINKED ISSUE #1336: add mangle function to public api

Kodiologist:
Reopened for the new merge policy.

- Closes #360
- Closes #970 (no longer necessary)
- Closes #1115 (debatably)
- Closes #1168 (keywords are now indeed mangled when used for keyword arguments)
- Closes #1173 (or at least, changes the situation drastically)

This pull request overhauls mangling such that HySymbols remember their original names (it is no longer the case that `(= 'a-b 'a_b)`) and all Python variables have names that are legal Python identifiers. The latter accomplishes most of what we need to ensure that `hy2py` produces real Python code.

The major changes are:

- Mangling happens at compile-time, not read-time.
- Mangling rules are different.
  - `*foo*` → `FOO` and `foo!` → `foo_bang` are gone.
    - The first one seems misleading, and the second rarely useful.
  - The Punycode transformation (which was only enabled for Python 2 anyway) is gone.
  - `foo-bar` → `foo_bar` and `foo?` → `is_foo` remain.
  - Names that still aren't legal Python identifiers (either because they contain illegal characters or because they coincide with a Python reserved word) are prepended with `hyx_`, and each illegal character is replaced with `ΔfooΔ` (or for Python 2, `XfooX`), where `foo` is the Unicode character name in lowercase with spaces replaced by underscores and hyphens replaced by "H"es. If the character doesn't have a name, we use "U" followed by its code point in lowercase hexadecimal.
  - Any added `hyx_` or `is_` is added after any leading underscores (because leading underscores can have magical effects in Python).
- The names of keyword arguments are fully mangled.

The below things remain to be done, but I figured I'd get feeback before continuing.

- [x] Watch out if the REPL variable `_` is confused with the shadow subtraction operator `-`
- [ ] ~~In the unmangle function, undo `hyx_` when necessary; otherwise, the distinct variables "hyx_X61X" and "a" will both unmangle to "a"~~ Not worth it.
- [x] Perhaps make the mangle function stringify the input, so the caller doesn't have to
- [x] Rename the mangling functions `mangle` and `unmangle` and add them to core (fixes #1336)
  - [x] Testing
  - [x] Documentation
- [x] Document mangling in general
- [x] Use `unmangle` instead of `hyify` in core
- [x] NEWS


hy/_compat.py
```diff
@@ -35,3 +35,24 @@ def wr_long(f, x):
...
+        # it a non-keyword.
+        return True
+    if keyword.iskeyword(x):
+        return False
+    if PY3:
+        return x.isidentifier()
+    if x.rstrip() != x:
+        return False
+    import tokenize as T
+    from StringIO import StringIO
```
vodik: You should prefer `io.StringIO`/`io.BytesIO` over the `StringIO` package

Kodiologist: Done.

gilch:
This really should have had a link to #1458 where much of the original discussion happened.

One-approval merges should say so here in the conversation, for the record.
As a courtesy I'd like to ask that you give a one day warning before such merges. I'd have requested that in the original rule change, but never got the chance.

Now that we've got the mangling makeover format settled, it's time to rethink the gensym format as discussed in #1458. I'd have mentioned this in time for the merge had I gotten warning.

Kodiologist:
> This really should have had a link to #1458 where much of the original discussion happened.

Fair enough.

> One-approval merges should say so here in the conversation, for the record.

It isn't hard to figure out if something was validly merged (and, if so, under what rules it was merged), is it?

> As a courtesy I'd like to ask that you give a one day warning before such merges.

But the creation of a PR is itself a 14-day warning. In theory, if you wanted a 1-day warning, you could set an alarm for yourself for 13 days hence (we could even write a bot that does this automatically; in Hy, no less), but why wait to do whatever it is you were going to do?

> Now that we've got the mangling makeover format settled, it's time to rethink the gensym format as discussed in #1458. I'd have mentioned this in time for the merge had I gotten warning.

No worries, that can be done in a new issue or PR. This PR is already big, anyway.

Kodiologist:
> we could even write a bot that does this automatically

Come to think of it, this could be useful as a reminder to PR authors to check that the PR is in a mergeworthy state.

vodik:
Heads up, we'll have to be careful with this at release time. I found I had to flush all my compiled pyc files.

Kodiologist:
"Be careful" in the sense of warning users to remove stale bytecode? NEWS could be a good place to put such a warning.

vodik:
Yes.

gilch:
Shouldn't the installer replace all the stale .pyc's automatically? It's fine to warn users about the issue #1324 since it will come up in their own development. But we shouldn't ask the users to do an extra installation step that we could have automated.

vodik:
Well, its not only about Hy, right?

Lets say I package Hy for Arch Linux (which there exists a package in AUR), and some other useful Hy library/tool (honestly don't know of any, but for examples sake, suppose there is).

Because the best practice for packaging Python like this is to also package the .pyc files, the eventual update to Hy to 0.15 would also mean **all** other Hy packages have to be rebuilt.

Not the worst thing in the world to require, but its a kind of change that really needs to be announced.

I'll put something up for adding it to the NEWS file tonight.

tests/native_tests/mangling.hy
```diff
@@ -0,0 +1,189 @@
...
+(defn test-late-mangling []
+  ; Mangling should only happen during compilation.
+  (assert (!= 'foo? 'is_foo))
+  (setv sym 'foo?)
+  (assert (= sym "foo?"))
+  (assert (!= sym "is_foo"))
+  (setv out (eval `(do
+    (setv ~sym 10)
+    [foo? is_foo])))
+  (assert (= out [10 10])))
```
vodik: @Kodiologist what's the rational for late mangling?

Just asking because its easier to make #1504 nice if I mangle symbols before comparing:

```closure
(defmacro foo [symbol]
  (if (= symbol 'foobar?)
      ...
	  ...))
```

Means this macro work on either `'foobar?` and `is_foobar`. Not sure if there's something I'm missing, so asking.

Kodiologist: So that you have access to the literal symbol that was entered, in case you want it. If you want to compare things post-mangling, you can always write code like `(= (mangle symbol) (mangle 'foobar?))`. By contrast, before this PR, when symbols were mangled at read-time, there was no way to tell whether the user entered `foobar?` or `is_foobar`, if you wanted to do that. After all, not every symbol ends up as a Python name, so not every symbol eventually needs to be mangled.

Kodiologist: The specific feature request was #360.

vodik: Thanks.

ekaschalk:
The new `mangle` has different behavior on the empty string, namely an assertion error whereas before it returned back the empty string.

Is there a reasoning for this? Note that `unmangle` on an empty string doesn't fail, returning back an empty string.

If not, the blank string assertion error would be helpful to have dropped.

Kodiologist:
I'm not sure what `unmangle` should do with an empty string, but `mangle` errors on an empty string because it's supposed to always return a valid Python identifier, and we have no mangling rule to produce a Python identifier for the empty symbol.

PR #1524, opened 2018-03-06
Cleanup sys.modules after a Hy specific import failure.

### LINKED ISSUE #1523: Import failures prevent from re-trying
During development, it is common to import faulty modules from the *repl*.  If a Hy module fails to load due to a `HyTypeError` or `LexException`, it cannot be loaded again until the *repl* is restarted.

There is a different exception treatment [in the importer][l].  One of the exception blocks is **not** cleaning up `sys.modules` before returning and that leaves a trace of the failure.

[l]: https://github.com/hylang/hy/blob/c1e5c3e48ca68ad2372e1fd6dca5a1dda9419e9b/hy/importer.py#L114-L122

etanol:
When there is a HyTypeError or a LexException, the None placeholder inserted in
sys.modules is not removed.  This prevents future import attempts because the
Python import machinery finds the None reference during the "fast path" of the
default importer.

This fixes #1523.

### DIFF
--- a/tests/importer/test_importer.py
+++ b/tests/importer/test_importer.py
@@ -68,6 +68,37 @@ def test_import_autocompiles():
     os.remove(get_bytecode_path(f.name))
 
 
+def test_import_failure_retryable():
+    "Test that a failed import does not prevents from re trying"
+    module_name = "again"
+
+    f = open("again.hy", "wb")
+
+    def _import_test():
+        try:
+            import again
+        except HyTypeError:
+            return "Failed"
+        else:
+            return "Succeeded"
+
+    f.write(b'(import "sys")')
+    f.flush()
+
+    assert _import_test() == "Failed"
+
+    f.seek(0)
+    f.truncate()
+    f.write(b'(import sys)')
+    f.flush()
+
+    assert _import_test() == "Succeeded"
+
+    f.close()
+    os.remove(f.name)
+    os.remove(get_bytecode_path(f.name))
+
+
 def test_eval():
     def eval_str(s):
         return hy.eval(hy.read_str(s))
--- a/hy/importer.py
+++ b/hy/importer.py
@@ -116,6 +116,7 @@ def import_file_to_module(module_name, fpath, loader=None):
                 with open(fpath, 'rt') as fp:
                     e.source = fp.read()
                 e.filename = fpath
+            sys.modules.pop(module_name, None)
             raise
         except Exception:
             sys.modules.pop(module_name, None)


Kodiologist:
This is cool, but seems likely to conflict with #1518.

Does this help with #712?

Is there any reason the test uses `f.close` instead of a `with` block?

etanol:
> This is cool, but seems likely to conflict with #1518.

How far from completion is that?

> Does this help with #712?

Not really.

> Is there any reason the test uses f.close instead of a with block?

No reason, I just forgot about it.  I'll change that.

Kodiologist:
> > This is cool, but seems likely to conflict with #1518.
>
> How far from completion is that?

@vodik?

vodik:
Close. The hard parts are done, the branch works. I just need to revert some of my experiments (patching `run_path` to support Hy might not have been the best idea), and write some documentation on it.

But I got a busy week ahead of me with renovations, so I'll probably not have too much programming time. I'd have no problem merging this in the meanwhile. No point making this painful when the fix is just a few lines.

And on that note, if you guys decide to merge this in the interim, have a look at #1513 as well, @etanol, and add that to this PR. Might as well work around two issues at once.

tests/importer/test_importer.py
```diff
@@ -68,6 +68,37 @@ def test_import_autocompiles():
     os.remove(get_bytecode_path(f.name))
 
 
+def test_import_failure_retryable():
+    "Test that a failed import does not prevents from re trying"
+    module_name = "again"
+
+    f = open("again.hy", "wb")
```
vodik: @Kodiologist what would you think about starting to use the `pytester` plugin for this kind of stuff? Comes stock with pytest for this kind of stuff, just needs to be turned on:

```python
def test_import_failure_retryable(testdir):
    testdir.makefile("hy", again="""(import sys)""")
    testdir.syspathinsert()  # magically cleans up post test too

    import again
```

Keeps current working directory clean and makes sure everything is fresh each pytest run. I suspect stale .pyc files could cause a false pass here.

Kodiologist: I'm fine with using pytest plugins, sure; just make sure everything works.

tests/importer/test_importer.py
```diff
@@ -68,6 +68,37 @@ def test_import_autocompiles():
...
+            import again
+        except HyTypeError:
+            return "Failed"
+        else:
+            return "Succeeded"
+
+    f.write(b'(import "sys")')
+    f.flush()
+
+    assert _import_test() == "Failed"
```
vodik: Probably better to do:

```python
with pytest.raises(HyTypeError):
    import again

# ...

import again
```

tests/importer/test_importer.py
vodik: Okay, in which case `pytest_plugins = "pytester"` needs to go in our `conftest.py` and its loaded.

etanol:
Dropping in favor of #1518

PR #1528, opened 2018-03-11
Python 3.7 support

### LINKED ISSUE #1501: Python 3.7 support

Kodiologist:
These changes should suffice to make us Python 3.7-ready (#1501), except that berkerpeksag/astor#86 is still a blocker.

Update: the astor bug is fixed, so this closes #1501.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -12,7 +12,7 @@ from hy.lex.parser import mangle
 
 import hy.macros
 from hy._compat import (
-    str_type, string_types, bytes_type, long_type, PY3, PY35,
+    str_type, string_types, bytes_type, long_type, PY3, PY35, PY37,
     raise_empty)
 from hy.macros import require, macroexpand, tag_macroexpand
 import hy.importer
@@ -1949,10 +1949,11 @@ class HyASTCompiler(object):
                 ) + expression
                 expression = expression.replace(arg[0])
 
-        # Docstrings must come at the start, so ensure that happens even if we
-        # generate anonymous variables.
-        if docstring is not None:
+        # Before Python 3.7, docstrings must come at the start, so ensure that
+        # happens even if we generate anonymous variables.
+        if docstring is not None and not PY37:
             expression.insert(0, docstring)
+            docstring = None
 
         if PY3:
             # Python 3.4+ requires that args are an ast.arg object, rather
@@ -1989,7 +1990,7 @@ class HyASTCompiler(object):
             defaults=defaults)
 
         body = self._compile_branch(expression)
-        if not force_functiondef and not body.stmts:
+        if not force_functiondef and not body.stmts and docstring is None:
             ret += asty.Lambda(expression, args=args, body=body.force_expr)
             return ret
 
@@ -2012,7 +2013,9 @@ class HyASTCompiler(object):
                     name=name,
                     args=args,
                     body=body.stmts,
-                    decorator_list=[])
+                    decorator_list=[],
+                    docstring=(None if docstring is None else
+                        str_type(docstring)))
 
         ast_name = asty.Name(expression, id=name, ctx=ast.Load())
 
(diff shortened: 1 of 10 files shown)


hy/compiler.py
```diff
@@ -2029,7 +2030,9 @@ def compile_function_def(self, expression):
                     name=name,
                     args=args,
                     body=body.stmts,
-                    decorator_list=[])
+                    decorator_list=[],
+                    **({} if docstring is None else
+                        {'docstring': str_type(docstring)}))
```
vodik: Just doing `docstring=str_type(docstring)` should work. The ast module's API is basically all `self, /, *args, **kwargs`, and drops unknown `kwargs`. Might make this stuff a little simpler to maintain.

vodik: Actually, I tested it on your branch. All tests passed on py27 and py36. py37 fails for a few tests (running a patched astor, its not that) - but the macro tests start passing. Haven't dug into it yet.

It also seems like Python 3 always sets docstring to None as a default, so its probably best to match that behaviour.

```python
>>> print(astor.dump_tree(ast.parse('''
... def test():
...     pass
... ''')))
Module(
    body=[
        FunctionDef(name='test',
            args=arguments(args=[], vararg=None, kwonlyargs=[], kw_defaults=[], kwarg=None, defaults=[]),
            body=[Pass],
            decorator_list=[],
            returns=None,
            docstring=None)],
    docstring=None)
```

vodik:
Look really good, just think we can simplify the AST generation a touch.

hy/compiler.py
Kodiologist: I'm getting no test failures locally with a patched astor.

vodik: Just to be clear, I mean testing it with a modification on top of your patch to always set `docstring`, not sure if that was clear.

Kodiologist: Try it without your modification, just with my PR and a fixed astor, and see if you still get test failures.

vodik: I have `test_bin_hy_byte_compile` failing, with or without my changes, on py37 only. Everything else is working. Weird...

Okay, so something that needs to get investigated but doesn't need to block this PR.

I can't reproduce the failures from Travis either.

vodik: Aaaah, I fixed the bytecode serializer for PEP 552, but I forgot to fix the loader.

I patched it: https://github.com/vodik/hy/commit/34929ec7ba3252fdc13fa5c59b7dbd01fd0af1de.patch

Feel free to borrow it if you want, or, write it as you see fix.

Kodiologist: I used `docstring=(None if docstring is None else str_type(docstring))` so that empty docstrings are still recognized.

I thought all the tests were passing because I forgot to run `test_bin`. I added your patch and everything should be good now (except for astor).

vodik: Nice. If you notice from the above snippet ^ python also sets `docstring=None` on the Module, so you can get rid of that spat trick everywhere.

Kodiologist: Right, oops. Done.

vodik:
Technically class docstrings also go in the AST:

```
In [3]: print(astor.dump_tree(ast.parse('''
   ...: class foo():
   ...:     "test"
   ...: ''')))
Module(
    body=[ClassDef(name='foo', bases=[], keywords=[], body=[], decorator_list=[], docstring='test')],
    docstring=None)
```

I approved it since we don't currently do class docstrings correctly. I'd say merge this as it is, and I'll address the class changes in #1536.

Kodiologist:
That seems fair.

PR #1529, opened 2018-03-12
Add pytest plugin

### LINKED ISSUE #1526: Expose pytest integration as a plugin
So, was going to put this off until I got some of my other ideas completed, but someone just opened an issue on an old project I started the last time I was playing with Hy: https://github.com/vodik/hytest/issues/1

That project was me fiddling with writing a pytest plugin that integrated Hy into pytest.

But we already do this. So, my proposal is to do the following:

- Move the Hy specific logic out of `conftest.py` and move it somewhere inside the `hy` package (say `hy.contrib.pytest_plugin`)
- Update `conftest.py` to use `pytest_plugins = ['hy.contrib.pytest_plugin']`

Now we have two options for exposing this to the general pytest community:

1. Hy itself could register a pytest plugin directly as an entry_point.
2. Publish another package under hylang organization called something like `pytest-hy` that registers a pytest plugin which contains a single line along the lines of:

```python
from hy.contrib.pytest_plugin import *
```

I'm more in favour of the later, as its more explicit. It is also how I've seen other libraries do it, for example [pytest-aiohttp](https://github.com/aio-libs/pytest-aiohttp/blob/master/pytest_aiohttp/__init__.py)

We could also promote a few useful functions like `can_compile`, `can_eval`, etc as pytest fixtures for other projects to reuse.

vodik:
Adds a Hy pytest plugin.

With this patch, pytest - though a special Hy plugin - will automatically pick up tests written in Hy. No need to have every project using Hy come up with their own testing solution.

Closes #1526

Now back to the importer...

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -1,0 +1,45 @@
+=======
+Testing
+=======
+
+The Hy package provides a very basic pytest plugin out of the box to
+help testing Hy code.
+
+A simple test function with just four lines of code::
+
+    # content of test_sample.hy
+    (defn func [x]
+      (+ x 1))
+
+    (def test-answer []
+      (assert (func 3) 5))
+
+You can now execute the test function::
+
+    $ pytest
+    =========================== test session starts ============================
+    platform linux -- Python 3.x.y, pytest-3.x.y, py-1.x.y, pluggy-0.x.y
+    rootdir: $REGENDOC_TMPDIR, inifile:
+    plugins: hy-x.y
+    collected 1 item
+
+    test_sample.hy F                                                     [100%]
+
+    ================================= FAILURES =================================
+    _______________________________ test_answer ________________________________
+
+	(defn test-answer []
+    >     (assert (= (func 3) 5)))
+    E     AssertionError
+
+    test_sample.hy:5: AssertionError
+    ========================= 1 failed in 0.12 seconds =========================
+
+This test returns a failure report because ``func(3)`` does not return ``5``.
+
+``pytest`` will run all files of the form test_*.hy or \*_test.hy in
+the current directory and its subdirectories. It will follow the
+`standard test discovery rules
+<https://docs.pytest.org/en/latest/goodpractices.html#test-discovery>`_,
+but that its customized with ``hy_files`` rather than with
+``python_files``.
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -18,6 +18,8 @@ Other Breaking Changes
 
 New Features
 ------------------------------
+* Added a first pytest plugin that teaches pytest how to collect Hy
+  modules
 * Added `mangle` and `unmangle` as core functions
 * `defclass` in Python 3 now supports specifying metaclasses and other
   keyword arguments
--- a/conftest.py
+++ b/conftest.py
@@ -1,15 +1,7 @@
-import pytest
-import hy
-import os
 from hy._compat import PY3, PY35, PY36
 
-NATIVE_TESTS = os.path.join("", "tests", "native_tests", "")
 
-def pytest_collect_file(parent, path):
-    if (path.ext == ".hy"
-            and NATIVE_TESTS in path.dirname + os.sep
-            and path.basename != "__init__.hy"
-            and not ("py3_only" in path.basename and not PY3)
-            and not ("py35_only" in path.basename and not PY35)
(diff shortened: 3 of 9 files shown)


Kodiologist:
The methods of the hytest fixture are functions we use for testing the compiler itself, so why would they be used by other projects?

Shouldn't the Hy plugin have to somehow enabled, as by adding it to `pytest_plugins`?

vodik:
> The methods of the hytest fixture are functions we use for testing the compiler itself, so why would they be used by other projects?

Yeah, now I think about it, it might not make sense. I was rushing though it. I was thinking it be useful for testing macros and these kinds of things downstream projects might write.

> Shouldn't the Hy plugin have to be somehow enabled, as by adding it to pytest_plugins?

See setup.py. So long as you use tox (which installs hy) or you do a development install, the plugin is registered with pytest.

Kodiologist:
Yeah, the fixture is probably not necessary.

> See setup.py. So long as you use tox (which installs hy) or you do a development install, the plugin is registered with pytest.

So, supposing some user is writing a Hy-based package (other than Hy itself) and wants to run some Hy tests with pytest, it suffices just to have Hy installed?

When I do `git clean -dfx && pip install -e . && pytest`, I get 586 tests, compared to 545 on master. Perhaps you're running some non-test functions as if they were tests.

vodik:
> So, supposing some user is writing a Hy-based package (other than Hy itself) and wants to run some Hy tests with pytest, it suffices just to have Hy installed?

Correct. See https://docs.pytest.org/en/latest/writing_plugins.html#setuptools-entry-points for more details

> ... so perhaps you're running some non-test functions as if they were tests.

Probably running under slightly different collection rules. Thanks, I'll look into that.

vodik:
Okay, so this is now the bare minimum required to make Hy code collect under pytest. Now its just a matter of having Hy installed in your environment to get things working.

I added support for setting a `hy_files` ini variable to override the default collection pattern for Hy files, since "tests_*" is a sensible default, but we don't follow that structure (nothing wrong with that, just figured out of scope).

I started playing a little bit with trying to add some deeper pytest integration as I'd like to see a bit more control over how Hy tests themselves are collected, but doesn't look like the pytest Collector is customize enough to support it (yet...)

A good friend of mine is a pytest maintainer, I'll see if I can get something useful out of him with over a beer.

Kodiologist:
Are you aiming to merge this roughly as-is, or do you want to add that more control first?

vodik:
I'm aiming to leave it as is, as its "good enough" - as in, works for our use cases and lets me drop comparable code with the one major project I've started using Hy with.

Now for what I think is possible, having working on pytest itself in the past, is much deeper integration. All this is hypothetical, but should be doable:

Most practically, fix up the corner cases around test collection. For example, its weird that there's `hy_files` but not `hy_tests` ini field. With our own collection logic, we could add it in, and make it work so it considers pre-mangled names - making it less confusing to use.

We could also, for example, support something closer to 'clojure.test' package with our own way of specifying fixtures, tests, and have it all integrate nicely with pytest. It lets us also hook into suppressing parts of the stack from the traceback and hook into the assert rewriter.

I've actually done work like this in the past (pytest-sipp, for example, creates a new kind of test item for pytest to collect), but wasn't satisfied that I wasn't relying on too many pytest internals that weren't part of its public API. So I'd like to see if my buddy has any ideas on the topic.

Kodiologist:
Okay, `pytest` with no arguments seems to be collecting the right number of tests now, but `pytest --ignore=tests/test_bin.py` runs `test_bin.py` despite the argument.

docs/testing.rst
```diff
@@ -0,0 +1,45 @@
+=======
+Testing
+=======
+
+The Hy package provides a very basic pytest plugin out of the box to
+help testing Hy code.
+
+A simple test function with just four lines of code::
+
+    # content of test_sample.hy
```
furiel: I suggest use semicolon for this line. Not a big deal, but I ran into this as I accidentally copied this line too.

docs/testing.rst
```diff
@@ -0,0 +1,45 @@
...
+    (defn func [x]
+      (+ x 1))
+
+    (def test-answer []
```
furiel: Small typo: def -> defn.

Kodiologist:

PR #1534, opened 2018-03-15
Fix `defclass` validation

### LINKED ISSUE #1533: Triple-quoted docstrings trigger internal compiler bug
If I accidentally use triple quotes in my docstrings (say, when playing with the Hy tutorial and assuming too much similarity to Python class docstrings), I trigger an internal compiler bug (`IndexError: string index out of range`), rather than a more readable Hy syntax error.

I first saw this in v0.14 and am able to replicate it in the bleeding-edge commit b023ebd0b5340a666df83e3fc0b72d7734636e04.

(Maybe the discussion in #1527 is relevant.)

vodik:
Mostly fixes the problem. This one however sneaks through:

```clj
(defclass Foo [] 'test)
```

Which would be a tricky case to fix. The problem is `defclass` sees it as `(quote test)` so its a valid expression.

Mostly fixes #1533

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -20,10 +20,8 @@ import hy.inspect
 
 import traceback
 import importlib
-import codecs
 import ast
 import sys
-import keyword
 import copy
 
 from collections import defaultdict
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2033,7 +2033,12 @@ class HyASTCompiler(object):
     def compile_class_expression(self, expressions):
         def rewire_init(expr):
             new_args = []
-            if expr[0] == HySymbol("setv"):
+            if not isinstance(expr, HyExpression):
+                raise HyTypeError(expr,
+                                  "`defclass` expects expressions, received `{}`".format(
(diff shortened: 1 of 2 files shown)


hy/compiler.py
```diff
@@ -2035,7 +2033,12 @@ def compile_return(self, expr):
     def compile_class_expression(self, expressions):
         def rewire_init(expr):
             new_args = []
-            if expr[0] == HySymbol("setv"):
+            if not isinstance(expr, HyExpression):
+                raise HyTypeError(expr,
+                                  "`defclass` expects expressions, received `{}`".format(
```
gilch: Technically, things like `42` are also expressions, so this message is confusing. It should say "HyExpressions" if that's what you actually want. But I'm not sure why we'd want to reject such things. Python doesn't.

vodik: No, that's a very good point. But that, I think, makes this PR simpler.

vodik:
Closing and making a new PR, as I've modified it further and this one is already improved.

PR #1536, opened 2018-03-15
Fix `defclass`

### LINKED ISSUE #1533: Triple-quoted docstrings trigger internal compiler bug

vodik:
Okay, so back to square one, and this one I think is a much better fix. Address a few issues all at once.

- Fixes #1533. The problem there was in the \_\_init__ return suppression.
- Tweaks class construction so we emit a list of statements.
- Remove `__doc__` attribute setting. Python looks at the first ast.Expr statement in its body, we don't need to really handle it specially.

Now the only problem I have, and could use some help, is with `force_stmt`. I duplicated the function because `expr_as_stmt` drops statments. For example, consider I get back:

    Result(imports=[], stmts=[FunctionDef(name='foo', args=arguments(args=[], vararg=None, kwonlyargs=[], kw_defaults=[], kwarg=None, defaults=[]), body=[Pass()], decorator_list=[])], expr=None, contains_yield=False)
	
Its already a statement but `expr_as_stmt` returns `Result()`.

So I tried putting that second condition I put in the `force_stmt` variant inside `expr_as_stmt`, and every statement suddenly was duplicated. For example, that new test I added looked like this:

	def test_class_sideeffects():
		"""NATIVE: test that defclass runs all expressions"""

		def set_sentinal():
			set_sentinal.set = True

		def set_sentinal():
			set_sentinal.set = True
		set_sentinal.set = False
		set_sentinal.set = False

		class A:
			set_sentinal()

		class A:
			set_sentinal()
		assert set_sentinal.set

So I suspect that this is an issue you guys solved differently, but I can't seem to dig into it. And I'm kinda short on time anyways (I shouldn't be hacking on Hy right now either....)

**EDIT** rewrote the test, updated it here too

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -20,10 +20,8 @@ import hy.inspect
 
 import traceback
 import importlib
-import codecs
 import ast
 import sys
-import keyword
 import copy
 
 from collections import defaultdict
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2036,7 +2036,10 @@ class HyASTCompiler(object):
     def compile_class_expression(self, expressions):
         def rewire_init(expr):
             new_args = []
-            if expr[0] == HySymbol("setv"):
+            if (isinstance(expr, HyExpression)
+                and len(expr) > 1
+                and isinstance(expr[0], HySymbol)
+                and expr[0] == HySymbol("setv")):
                 pairs = expr[1:]
                 while len(pairs) > 0:
                     k, v = (pairs.pop(0), pairs.pop(0))
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2072,12 +2072,7 @@ class HyASTCompiler(object):
 
         # grab the doc string, if there is one
         if expressions and isinstance(expressions[0], HyString):
-            docstring = expressions.pop(0)
-            symb = HySymbol("__doc__")
-            symb.start_line = docstring.start_line
-            symb.start_column = docstring.start_column
-            body += self._compile_assign(symb, docstring)
-            body += body.expr_as_stmt()
+            body += self.compile(expressions.pop(0)).expr_as_stmt()
 
         if expressions and isinstance(expressions[0], HyList) \
            and not isinstance(expressions[0], HyExpression):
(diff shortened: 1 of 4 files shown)


hy/compiler.py
```diff
@@ -2035,7 +2049,9 @@ def compile_return(self, expr):
     def compile_class_expression(self, expressions):
         def rewire_init(expr):
             new_args = []
-            if expr[0] == HySymbol("setv"):
+            if (isinstance(expr, HyExpression)
+                and len(expr) > 1
+                and expr[0] == HySymbol("setv")):
```
Kodiologist: This will trigger on the string literal `"setv"` without #1504.

tests/native_tests/defclass.hy
```diff
@@ -127,3 +127,14 @@
   (setv b (B))
   (assert (= a.x 1))
   (assert (= b.x 2)))
+
+(defn test-class-sideeffects []
+  "NATIVE: test that defclass runs all expressions"
+  (defn set-sentinal []
+    (setv (. set-sentinal set) True))
+  (setv (. set-sentinal set) False)
```
Kodiologist: `(. set-sentinel set)` is better written `set-sentinel.set`.

Kodiologist: And yes, it's spelled "sentinel". :P

hy/compiler.py
vodik: Put an `isinstance` check. Don't have the time at the moment to dig into Hy stuff (very soon though), but might as well try and get these bug fixes in as things work currently.

vodik:
Thanks!

Kodiologist:
Could you rebase this and check that Python 3.7 now passes?

vodik:
Yeah, lets get this fixed, so I'll get this one out the way.

vodik:
Aaand I broke something. I'll look into it in the evening.

vodik:
@Kodiologist working as expected.

PR #1538, opened 2018-03-19
In ->>, don't modify the arguments

### LINKED ISSUE #1537: ->> can duplicate an argument when called in a macro
I came across a bug (feature) that I can't really understand. In macros, threading with ``->>`` into a ``tuple`` gives error "TypeError('tuple() takes at most 1 argument (2 given)',)" if I have ``tuple`` as expression. Instead, if I have that as just a symbol (ie. without enclosing parentheses), the error isn't raised.

Looking into code, I thought that the both would result the same code. Both ``(->> body (tuple))`` and ``(->> body tuple)`` should turn into ``(tuple body)``. This has been around at least since 0.12.1, so I don't expect it to be anything new.

Minimal example I could come up with:
```
(hy-master)tuukka@eevertti:~$ hy
hy 0.14.0+48.gcb72a8c using CPython(default) 3.4.3 on Linux
=> (defmacro not-ok [&rest body]
...   (setv uniques (->> body
...                      (tuple)))
...   `(print "ok"))
<function _hy_anon_var_1 at 0xb6a53a04>
=>
=> (not-ok (foo bar (baz)))

  (not-ok (foo bar (baz)))
  ^----------------------^
HyMacroExpansionError: expanding `not-ok': TypeError('tuple() takes at most 1 argument (2 given)',)


=> (defmacro ok [&rest body]
...   (setv uniques (->> body
...                      tuple))
...   `(print "ok"))
<function _hy_anon_var_1 at 0xb6a47b24>
=>
=> (ok (foo bar (baz)))
ok
```

Kodiologist:
Fixes #1537.

### DIFF
--- a/tests/resources/macros.hy
+++ b/tests/resources/macros.hy
@@ -1,0 +1,9 @@
+(defmacro thread-set-ab []
+  (defn f [&rest args] (.join "" (+ (, "a") args)))
+  (setv variable (HySymbol (-> "b" (f))))
+  `(setv ~variable 2))
+
+(defmacro threadtail-set-cd []
+  (defn f [&rest args] (.join "" (+ (, "c") args)))
+  (setv variable (HySymbol (->> "d" (f))))
+  `(setv ~variable 5))
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -25,6 +25,7 @@ New Features
 Bug Fixes
 ------------------------------
 * Fix `(return)` so it works correctly to exit a Python 2 generator
+* Fixed a case where `->` and `->>` duplicated an argument
 
 Misc. Improvements
 ----------------------------
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -150,8 +150,8 @@ the second form, the second result is inserted into the third form, and so on."
   (setv ret head)
   (for* [node args]
     (setv ret (if (isinstance node HyExpression)
-      `(~(first node) ~ret ~@(rest node))
-      `(~node ~ret))))
+                  `(~(first node) ~ret ~@(rest node))
+                  `(~node ~ret))))
   ret)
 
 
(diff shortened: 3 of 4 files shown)


tests/resources/macros.hy
```diff
@@ -0,0 +1,4 @@
+(defmacro threadtail-set-ab []
+ (defn f [&rest args] (.join "" (+ (, "a") args)))
+  (setv variable (HySymbol (->> "b" (f))))
+  `(setv ~variable 2))
```
vodik: Is this idiomatic style with the `join` on the same line as `defn`?

Kodiologist: Eh, I dunno. I forgot a space at the beginning of this line, though, so I've fixed that.

gilch:
I'm not sure I understand why the error only shows up inside `defmacro`. Doesn't that indicate the problem is a mutation in the compiler itself, rather than a problem with the definition of `->>`?

Kodiologist:
Yes, I think that either a mutation, or an expectation that something won't mutate that actually can mutate, somewhere in the compiler, is also at fault. Well, lots of such issues, actually. But finding and fixing them is another matter. Probably the way to make progress on that is to make Hy models immutable and see what breaks.

Kodiologist:
And before we do that, we should probably take a stab at writing the compiler more in terms of pattern matching of some kind, in place of all the manual inspection and modification of trees.

Kodiologist:
I might as well do it here.

Kodiologist:
Done.

tuturto:
Looking good to me

Kodiologist:
There's no rule that says core membership doesn't apply retroactively (just like there's no rule that says a dog can't play basketball), so I'm going to merge this.

vodik:
Something else worth noting about making HyExpression immutable: debugging could be easier.

I've hit this enough times: while trying to dig into something in the core with pdb, go up the stack trace and try to see what was passed in, only to realize that I can't because we've popped entries off it.

PR #1543, opened 2018-03-25
Make HyKeyword a first class objects

vodik:
HyKeywords are no longer string instances with a particular prefix, but a completely separate object.

It also means HyKeywords now evaluate to themselves as well.

I'm splitting up #1504 into two parts, especially since I basically need to rewrite the `HySymbol` work now that the mangling changes went in.

I want to break the current loose typing between `HyKeyword` / `HySymbol` / strings. I think it mostly  makes the compiler's core messier than it should be.

I think I've changed my mind and that it makes a lot of sense to continue to keep `HySymbol` a subclass of `str`, but I think we might be able to make some improvements that improve the type safety of the compiler's core without needing `isinstance(foo, HySymbol)` checks everywhere.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -747,7 +747,7 @@ class HyASTCompiler(object):
         elif isinstance(form, HyString):
             x = [HySymbol(name), form]
             if form.brackets is not None:
-                x.extend([HyKeyword(":brackets"), form.brackets])
+                x.extend([HyKeyword("brackets"), form.brackets])
             return imports, HyExpression(x).replace(form), False
 
         return imports, HyExpression([HySymbol(name),
@@ -811,7 +811,7 @@ class HyASTCompiler(object):
             ret += self.compile(expr.pop(0))
 
         cause = None
-        if len(expr) == 2 and expr[0] == HyKeyword(":from"):
+        if len(expr) == 2 and expr[0] == HyKeyword("from"):
             if not PY3:
                 raise HyCompileError(
                     "raise from only supported in python 3")
@@ -1170,7 +1170,7 @@ class HyASTCompiler(object):
             if isinstance(iexpr, HyList) and iexpr:
                 module = iexpr.pop(0)
                 entry = iexpr[0]
-                if entry == HyKeyword(":as"):
+                if entry == HyKeyword("as"):
                     if not len(iexpr) == 2:
                         raise HyTypeError(iexpr,
                                           "garbage after aliased import")
@@ -1464,7 +1464,7 @@ class HyASTCompiler(object):
                 else:
                     assignments = {}
                     while names:
-                        if len(names) > 1 and names[1] == HyKeyword(":as"):
+                        if len(names) > 1 and names[1] == HyKeyword("as"):
                             k, _, v = names[:3]
                             del names[:3]
                             assignments[k] = v
@@ -1473,7 +1473,7 @@ class HyASTCompiler(object):
                             assignments[symbol] = symbol
                     require(module, self.module_name, assignments=assignments)
             elif (isinstance(entry, HyList) and len(entry) == 3
-                    and entry[1] == HyKeyword(":as")):
+                    and entry[1] == HyKeyword("as")):
                 # e.g., (require [foo :as bar])
                 module, _, prefix = entry
                 __import__(module)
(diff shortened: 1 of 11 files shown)


gilch:
Fixes #1352. I think.

vodik:
It does, I tested it. Also related to #846 but I didn't implement `__call__`, so didn't know if its worth marking as fixed.

tests/native_tests/language.hy
```diff
@@ -552,9 +552,7 @@
   (assert (= (.a.b.meth x) "meth"))
   (assert (= (.a.b.meth x "foo" "bar") "meth foo bar"))
   (assert (= (.a.b.meth :b "1" :a "2" x "foo" "bar") "meth foo bar 2 1"))
-  (assert (= (.a.b.meth x #* ["foo" "bar"]) "meth foo bar"))
-
-  (assert (is (.isdigit :foo) False)))
```
Kodiologist: If you don't replace this line, then we're no longer testing whether `(.method :keyword)` is interpreted as intended.

tests/native_tests/language.hy
```diff
@@ -1221,9 +1219,11 @@
   "NATIVE: test if keywords are recognised"
 
   (assert (= :foo :foo))
+  (assert (= :foo ':foo))
```
Kodiologist: You might also test `(is (type :foo) (type ':foo))`.

hy/models.py
```diff
@@ -121,22 +121,39 @@ def __init__(self, string):
...
 
-    def __new__(cls, value):
-        if not value.startswith(cls.PREFIX):
-            value = cls.PREFIX + value
-
-        obj = str_type.__new__(cls, value)
-        return obj
+    def __init__(self, value):
+        if value[0] != ':':
+            value = ':' + value
```
Kodiologist: Shouldn't the empty keyword be built with `value` set to the empty string instead of `":"`?

hy/contrib/hy_repr.hy
```diff
@@ -82,10 +82,10 @@
...
     (+ "(" (-cat x) ")"))))
 
-(hy-repr-register HySymbol str)
-(hy-repr-register [str-type bytes-type HyKeyword] (fn [x]
-  (if (and (instance? str-type x) (.startswith x HyKeyword.PREFIX))
-    (return (cut x 1)))
+(hy-repr-register [HySymbol HyKeyword] str)
+(hy-repr-register [str-type bytes-type] (fn [x]
+  (if (and (instance? str-type x) (.startswith x ":"))
+      (return x))
```
Kodiologist: I don't think this is right, since `:foo` doesn't evaluate to the string ":foo"; it now produces a HyKeyword. This ought to be causing a test failure, so I guess I screwed up the test for this.

hy/compiler.py
```diff
@@ -2182,7 +2184,18 @@ def compile_symbol(self, symbol):
 
         return asty.Name(symbol, id=ast_str(symbol), ctx=ast.Load())
 
-    @builds(HyString, HyKeyword, HyBytes)
+    @builds(HyKeyword)
+    def compile_keyword(self, string, building):
```
Kodiologist: You can remove the `building` parameter, since you're not using it.

hy/models.py
```diff
@@ -121,22 +121,39 @@ def __init__(self, string):
...
+
+    def __str__(self):
+        return self._value
+
+    def __hash__(self):
+        return hash(self._value)
+
+    def __eq__(self, other):
+        if not isinstance(other, HyKeyword):
+            return NotImplemented
```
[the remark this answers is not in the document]
vodik: That's actually got different semantics, as I understand it.

Because, abstractly, when running `a == b`, python tried first `a.__eq__(b)`, and if `NotImplemented` is returned, it'll try `b.__eq__(a)`.  Returning `False` here means we'll break code that might want to test equality with a `HyKeyword` (I don't know why you would, but I don't think that justifies breaking that intentionally).

hy/contrib/hy_repr.hy
vodik: Good catch

hy/models.py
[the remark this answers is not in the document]
Kodiologist: Oh, you're right, I think. With your code, `(= :hello "hello")` returns false instead of throwing an error.

hy/contrib/hy_repr.hy
vodik: Yeah, I left that in because of this assert:

```clj
(assert (= (hy-repr (str ':mykeyword)) ":mykeyword"))
```

But I guess that doesn't make sense, we're breaking that.

hy/models.py
vodik: I did this as its the easier repr function. I think your right that it makes sense, so I'll work on it, but what would we prefer to see: `HyKeyword(':foo')` or `HyKeyword('foo')`?

Currently they're actually equivalent, but maybe I should put some extra effort into making sure there's only one way.

tests/native_tests/language.hy
vodik: Good catch, also deleted too much...

tests/native_tests/language.hy
vodik: Done

hy/models.py
Kodiologist: The colon is mandatory, so I think `HyKeyword('foo')` should produce `:foo`, whereas `HyKeyword(':foo')` would produce `::foo`.

vodik: :+1: was thinking the same but I think thats a deeper change it seems. Might as well fix it though.

Kodiologist:
This is great work, my dude. Fix or remove the parts of the documentation that are wrong now, update NEWS, and we're good to go.

Issue #1542, opened 2018-03-25
Consider making Hy models immutable.


gilch:
We've discussed this a bit recently in other issues. Particularly #1519. (also #1538). But it's important enough for a dedicated issue.

The mutable models are causing serious but subtle bugs (the worst kind) in the compiler which are most obvious in macros. User macros may run into the same kind of issues, even if we add workarounds to ours. Not good.

Fixing the compiler without switching to immutable Hy models may be very difficult. Our tests are not adequate--so we can't be very sure we've fixed all the problems, and it would be easy to reintroduce similar problems by accident. It not clear we could add enough tests to catch everything either. The best practice without changing the interface may be to copy on receipt and copy-on-write everywhere in the compiler, which would be slow. Switching to immutable models may actually be easier and perform faster.

The obvious approach would be to base our models on `tuple` instead of `list`, but that's not the only option. We could use Pyrsistent, or roll our own basic linked list, or add an immutable seq abstraction for iterators in general #1242.

Regardless of the chosen approach, this is a very deep change that will break the world--to fix it. Fixing Hy's core will be a pain. Most basic user code would be unchanged, but anything that touches the Hy models directly (particularly macros) will break and have to be rewritten to the new interface.

vodik:
Also related to this, I think this could improve hy2py as well with macro output. It be nicer to see this:

```python
HyExpression([HySymbol('setv'), HySymbol('x'), HyInteger(1)])
```

instead of 

```python
HyExpression([] + [HySymbol('setv')] + [HySymbol('x')] + [HyInteger(1)])
```

gilch:
It's a consequence of the implementation of quasiquoting. We'd basically need the list monad to get the `~@` unquote splice, i.e. if it's already an iterable, just convert it to a list instead of wrapping it in one. E.g.
```
`(0 ~@[1 2 3] 4)
```
currently compiles to
```
HyExpression([] + [HyInteger(0)] + list([1, 2, 3] or []) + [HyInteger(4)])
```
but could be
```
HyExpression([HyInteger(0), *list([1, 2, 3] or []), HyInteger(4)])
```
but only since PEP 448 in v3.5. We're still supporting 3.4 though.

I'm not sure how immutable models help us here.

Kodiologist:
#1575 does at least make the copying step that we currently depend on use more obvious.

Kodiologist:
[...]

Issue #1407, opened 2017-08-29
Get rid of autoimports via new namespaces


gilch:
> Namespaces are one honking great idea -- let's do more of those!
> -- The Zen of Python

Autoimports are causing us headaches. #1367, #791.

Macros are also brittle due to lack of namespacing and hygene. #277. Even some compiler builtins have this problem--
```Python
=> ((fn [list] (eval `(print ~@list))) ["Hy" "world!"])
from hy.core.language import eval
from hy import HyExpression, HySymbol
(lambda list: eval(HyExpression((([] + [HySymbol('print')]) + list((list or []))))))(['Hy', 'world!'])
Traceback (most recent call last):
  File "c:\users\me\documents\github\hy\hy\importer.py", line 201, in hy_eval
    return eval(ast_compile(expr, "<eval>", "eval"), namespace)
  File "<eval>", line 1, in <module>
  File "<eval>", line 1, in <lambda>
TypeError: 'list' object is not callable
```
It seems like that should work. I didn't even use any macros. Can you spot the error in the Python expansion?

One might be tempted to scoff and say, "You should *never* use a builtin name as a local!". That might work for Python, which has a fairly small number of builtins, but for Clojure, which has ~600 symbols in core, that's an *unreasonable cognitive burden* on the programmer. Hy isn't quite there yet, but it's still got a *much* bigger core than Python's builtins. The `name` builtin is already a problem. #525 Maybe restricting the names of special forms is okay, but we really need to be able to shadow the other core names with a local.

Clojure's namespacing may hold the answer. If the compiler had used `__import__("builtins").list` instead of just `list` in its quasiquote expansion, the above would have worked properly. Clojure automatically namespaces symbols in its syntax-quote forms, so they work more like that.

An inline import is not a big deal in Python, since modules are cached. It's basically just an extra dict lookup--a fast action that you do all the time in Python code.

Using `__import__` directly is frowned upon in Python--`__import__` is kind of an implementation detail (but we could say the same of the AST itself). The recommendation is to use the import statement, like we're doing now. But that's harder to use than an expression, and requires an extra gensym, and doesn't play nice with `__future__`.  So if you need a more portable import expression, you're supposed to use the one from importlib. But how do we get to *that* without an autoimport? Is `__import__("importlib").import_module("foo").bar` any better than `__import__("foo").bar`? Chicken/egg.

The solution to a chicken/egg problem is bootstrapping.
### step 1 new builtin
[...]

PR #1545, opened 2018-03-25
unquote/unpacking improvements

### LINKED ISSUE #1519: Looks like `replace` does unnecessary work
I think I'm seeing some unnecessary work going on: `HyList.replace` calls `replace_hy_obj`, which in turn can `wrap_value`. But the result of `replace_hy_obj` is discarded and thus, depending on how a particular expression is constructed, its possible `wrap_value` is evaluated 3 or 4 times on the same object.

Fixing `HyList.replace` so it instead self mutates:

```python
    def replace(self, other):
        self[:] = [replace_hy_obj(x, other) for x in self]
        HyObject.replace(self, other)
        return self
```

Wrapping only happens once now. But this seems to have some consequences with using iterables in unquote. The following expression now does this:

```
=> (defn numbers [] (yield 1) (yield 2) (yield 3))
=> (macroexpand `(do ~(numbers)))
HyExpression([
  HySymbol('do'),
  HyList([
    HyInteger(1),
    HyInteger(2),
    HyInteger(3)])])
```

instead of

```
=> (defn numbers [] (yield 1) (yield 2) (yield 3))
=> (macroexpand `(do ~(numbers)))
HyExpression([
  HySymbol('do'),
  <generator object numbers at 0x7f2d2da38db0>])
```

(which can't be compiled: TypeError: Don't know how to wrap a <class 'generator'> object to a HyObject)

It also means `unquote-splice`'s compiler logic can be simplified a bit as well.

Are there any other concerns or thoughts on the topic before I dig further into this? I couldn't find any other open issues around this, but maybe I just don't know what to look for?

vodik:
A pile of changes dealing with some ugly corner cases in the language

Fixes #1519

The problem here is calling `replace` can trigger a call to `replace_hy_obj`, which will try to also wrap values inside a `HyObject`.

However, this means when we try to recursively replace inside a nested object, like a `HyList`, we should try and capture these new wrapped objects.

Hy still works if we don't, but it results in the object getting rewrapped multiple times during a compile.

But with that change in place, `wrap_value` should now only be called once per expression. This opens the door to safely wrap iterables. So instead of crashing when trying to wrap an iterable, first expand it and then try to wrap the resulting list.
    
This allows `unquote` to work with iterators and generators, which previously didn't work unless you used `unquote-splice` or explicitly expanded a list first.

And lastly, use PEP 448 for better unquoted expressions (inspired by @gilch's comment in #1542)
    
With Python 3.5, we can have an arbitrary number of unpackings. We can use this to emit cleaner expressions.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -699,28 +699,31 @@ class HyASTCompiler(object):
         imports = set([name])
 
         if isinstance(form, (HyList, HyDict, HySet)):
-            if not form:
-                contents = HyList()
-            else:
-                # If there are arguments, they can be spliced
-                # so we build a sum...
-                contents = HyExpression([HySymbol("+"), HyList()])
-
+            contents = HyList()
             for x in form:
                 f_imports, f_contents, splice = self._render_quoted_form(x,
                                                                          level)
                 imports.update(f_imports)
                 if splice:
-                    to_add = HyExpression([
-                        HySymbol("list"),
-                        HyExpression([HySymbol("or"), f_contents, HyList()])])
+                    # On Python 3.5 and newer, leverage PEP 448 to generate
+                    # expression using tuple unpacking instead of addition.
+                    #
+                    # Should be marginally faster too.
+                    contents.append(HyExpression([
+                        HySymbol("unpack-iterable" if PY35 else "list"),
+                        HyExpression([HySymbol("or"), f_contents, HyList()])
+                    ]))
                 else:
-                    to_add = HyList([f_contents])
+                    contents.append(f_contents if PY35
+                                    else HyList([f_contents]))
 
-                contents.append(to_add)
+            if not PY35:
+                # Without Python 3.5's unpacking improvements, we need
+                # to concatenate the expression manually
+                contents = HyExpression([HySymbol("sum"), contents, HyList()])
 
-            return imports, HyExpression([HySymbol(name),
-                                          contents]).replace(form), False
+            expr = HyExpression([HySymbol(name), contents])
+            return imports, expr.replace(form), False
 
         elif isinstance(form, HyCons):
             ret = HyExpression([HySymbol(name)])
--- a/hy/models.py
+++ b/hy/models.py
@@ -226,11 +226,8 @@ class HyList(HyObject, list):
     """
 
     def replace(self, other):
-        for x in self:
-            replace_hy_obj(x, other)
-
-        HyObject.replace(self, other)
-        return self
+        self[:] = [replace_hy_obj(x, other) for x in self]
+        return HyObject.replace(self, other)
 
     def __add__(self, other):
         return self.__class__(super(HyList, self).__add__(other))
@@ -387,11 +384,11 @@ class HyCons(HyObject):
 
     def replace(self, other):
         if self.car is not None:
-            replace_hy_obj(self.car, other)
+            self.car = replace_hy_obj(self.car, other)
         if self.cdr is not None:
-            replace_hy_obj(self.cdr, other)
+            self.cdr = replace_hy_obj(self.cdr, other)
 
-        HyObject.replace(self, other)
+        return HyObject.replace(self, other)
 
     def __repr__(self):
         if PRETTY:
--- a/hy/models.py
+++ b/hy/models.py
@@ -3,6 +3,7 @@
 # license. See the LICENSE.
 
 from __future__ import unicode_literals
+import collections
 from contextlib import contextmanager
 from math import isnan, isinf
 from hy._compat import PY3, str_type, bytes_type, long_type, string_types
@@ -62,9 +63,11 @@ def wrap_value(x):
 
     wrapper = _wrappers.get(type(x))
     if wrapper is None:
+        if (not isinstance(x, HyObject)
+            and isinstance(x, collections.Iterable)):
+            return wrap_value(list(x))
         return x
-    else:
-        return wrapper(x)
+    return wrapper(x)
 
 
 def replace_hy_obj(obj, other):
--- a/tests/native_tests/quote.hy
+++ b/tests/native_tests/quote.hy
@@ -98,3 +98,12 @@
       [1 2 3]
       [4 5 6])
     [4 5 6])))
+
+(defn test-unquote-with-generators []
+  (defn values []
+    (yield '1)
+    (yield '2)
+    (yield '3))
+
+  (assert (= (eval `[~@(values)]) [1 2 3]))
+  (assert (= (eval `[~(values)]) [[1 2 3]])))


vodik:
So the big use case for this patch is letting `map` and `unquote` work well together.

I started hitting this when looking at some Clojure macro tutorials to brush up. The following snippet doesn't work currently on Python 3 because map is a generator:

```clojure
(eval-and-compile
  (defn criticize-code
    [criticism code]
    `(print ~criticism (quote ~code))))

(defmacro code-critic
  [bad good]
  `(do ~(map (fn [x] (criticize-code #* x))
        [["Cursed bacteria of Liberia, this is bad code:" bad]
         ["Sweet sacred boa of Western and Eastern Samoa, this is good code:" good]])))

(code-critic (1 + 1) (+ 1 1))
```

But it feels natural to work. I understand tossing an explicit `list` in there will make it work, but it also makes it noisier.

I think we should embrace the strengths of Python and try to leverage generators like this.

vodik:
And I've been trying to think what would be the best wording for the NEWS entry because of the generator handling:

```clojure
=> (eval-and-compile
... (defn numbers [] (yield-from [1 2 3])))
=> (quote (sum (unquote (numbers))))
'(sum ~(numbers))
=> (quasiquote (sum (unquote (numbers))))
'(sum <generator object numbers at 0x7ff6050f5bf8>)
=> (eval *1)
6
```

vodik:
Yeah, something interesting is happening here for sure: it works, but:

```clojure
=> (setv sym `(sum ~(genexpr x [x (range 10)] (odd? x))))
=> sym
'(sum <generator object <genexpr> at 0x7fa63dc589e8>)
=> (eval sym)
25
=> sym
'(sum [1 3 5 7 9])
```

vodik:
Okay, so addressing the generator rewriting, I have a fix for it, but its not trivial and I can only make it work by completely breaking line numbering...

The problem is that we should probably be calling `wrap_obj` in `HyList` constructor. I think we probably want to do that anyways as other Hy objects (like `HyCons`) do wraps on construction.

But that opens a whole different can of worms.

It seems we throw away line numbering information on a lot of intermediate forms inside the compiler (for example, slicing a `HyExpression` results in a new `HyExpression` without line numbers) which is fix up later. Wrapping forms in `HyList.__init__` starts exposing with this.

So probably there are some subtle bugs here related to #1542.

Don't know how to proceed. I'd be comfortable merging it with the mutation and then trying to fix these issues as I really would like to see the functionality work.

Any ideas?

gilch:
We could make a core tag macro to convert to tuple:
```Hy
(deftag , [iterable]
  `(tuple ~iterable))
```
Python3 uses iterators a lot more than Python2 did, so realizing one into a tuple is something we do often. This would make using generators in macros pretty easy.
```
=> #,(range 10)
tuple(range(10))

(0, 1, 2, 3, 4, 5, 6, 7, 8, 9)
```

vodik:
I actually really like that idea, should we do both? Or are you suggesting this as an alternative?

I mean, `~#,(map ...)` seems a little noisy to me, and I think its inconsistent that `~@(map ...)` works just.

I do think we should embrace generators in the core like this considering all the extensive places that Python 3 uses them and everywhere they're accepted.

Kodiologist:
This PR looks pretty good to me as is (but don't forget NEWS, of course).

I don't think the new promotion semantics for generators are strictly necessary, because you can always write `(, ~@foo)` or `[~@foo]` and then it's explicit what you're getting, but I'm okay with them.

[Your example](https://github.com/hylang/hy/pull/1545#issuecomment-376000499) is kind of inevitable considering that iterators in Python mutate whenever you take items from them.

> The problem is that we should probably be calling `wrap_obj` in `HyList` constructor. I think we probably want to do that anyways as other Hy objects (like `HyCons`) do wraps on construction.

I don't think we should promote things to Hy models in the model constructors because [promotion should happen in only one place, which is just before compilation](https://github.com/hylang/hy/pull/1314#issuecomment-311220657). HyCons is the only exception, and even that may ultimately be wrong, because HyCons has always been pathological (#568).

gilch:
Generators are mutable, may have side effects, and can only be read once. Laziness is great, but iterators are a pain to use compared to Clojure's seq abstraction. #1242.

I'm confused by your motivating examples. The output isn't using the Hy model reprs, and the code critic isn't indented right. After counting brackets, I get,
```
(defmacro code-critic
  [bad good]
  `(do ~(map (fn [x] (criticize-code #* x))
             [["Cursed bacteria of Liberia, this is bad code:" bad]
              ["Sweet sacred boa of Western and Eastern Samoa, this is good code:" good]])))
```
So is the intended expansion is like,
```
(do
  (print "Cursed bacteria of Liberia, this is bad code:" '(1 + 1))
  (print "Sweet sacred boa of Western and Eastern Samoa, this is good code:" '(+ 1 1)))
```
?
But in that case you need to splice with `~@`! I don't see how making `~` act like `~@` when it happens to be unquoting a generator is a good idea at all. Let's be consistent here.

Here's another example with the default pretty repr and --spy.
```
=> (setv sym `(sum ~(genexpr x [x (range 10)] (odd? x))))
from hy.core.language import is_odd
sym = HyExpression([] + [HySymbol('sum')] + [(x for x in range(10) if
    is_odd(x))])
None

=> sym
sym

HyExpression([
  HySymbol('sum'),
  <generator object <genexpr> at 0x000002B637B10B48>])
```
Of course it's a generator here, what did you expect? What bothers me is that `eval` can compile this at all, since it's not a proper hytree. And in fact, it doesn't.
```
=> (eval sym)
from hy.core.language import eval
eval(sym)

  ...
TypeError: Don't know how to wrap a <class 'generator'> object to a HyObject
```
Now suppose we had converted this to a tuple,
```
=> sym
sym

HyExpression([
  HySymbol('sum'),
  (1, 3, 5, 7, 9)])
=> (eval sym)
from hy.core.language import eval
eval(sym)

25
```
It works, but it probably shouldn't. Using `cons` shows why.
```
=> (cons (, 1 2 3) 1)
cons((1, 2, 3), 1)

<HyCons (
  HyList([
    HyInteger(1),
    HyInteger(2),
    HyInteger(3)])
. HyInteger(1))>
```
The tuple isn't inserted as a tuple. It's getting autoconverted into a HyList. Hy is strange for a Lisp in that its model types are distinct from the data they represent. In Clojure there's no distinction between, say, `HyInteger` and `int`. Hy can autoconvert certain datatypes (it didn't always #1314), but other types are not representable as Hy models. #919. There doesn't seem to be a way to insert arbitrary datatypes into the ast. Only Python code is allowed. I find the autoconversion of tuples to `HyList` a questionable choice to begin with.

Kodiologist:
You're basically objecting to #1314, then.

Kodiologist:
I mean, making the user promote everything by hand to the appropriate Hy model type would be pretty annoying. Right?

gilch:
Not all of it maybe. The promotion of the atomic types is certainly convenient. But I'm less comfortable with the compound types. We can splice them in like `[~@foo]` when constructing macros, without referring to `HyList` and friends explicitly. This would be more consistent with the generators.

gilch:
I mentioned in #1542 that `'(0 ~@[1 2 3] 4` could compile to 
`HyExpression([HyInteger(0), *list([1, 2, 3] or []), HyInteger(4)])`,
but it's probably more efficient to use tuple instead of list,
`HyExpression([HyInteger(0), *tuple([1, 2, 3] or []), HyInteger(4)])`.

There's also a hygene issue that I mentioned in #1407. Shadowing `list` with a local can break quasiqutes. It's probably safer to use something like `__import__("builtins").tuple` or `__import__("builtins").list` here. We have not been careful enough in the compiler. This does add noise to the output, which we were trying to clean up here. We need a better system for macro hygiene.
`HyExpression([HyInteger(0), *__import__("builtins").tuple([1, 2, 3] or []), HyInteger(4)])`
Even if we don't do the `__import__`, `tuple` is less likely to cause an accidental capture than `list`.

Kodiologist:

PR #1546, opened 2018-03-25
Use io.StringIO everywhere

vodik:
Small change that drops some compatibility code. Python 2.6 got a backport of the `io` module, and it contains a better `StringIO` implementation


vodik:
Never mind, right, I forgot its unicode only though because its a backport.

PR #1547, opened 2018-03-25
Improve testing of hy2py

Kodiologist:


tests/test_hy2py.py
```diff
@@ -0,0 +1,120 @@
...
+
+    assert m.closed == ["v2", "v1"]
+
+
+def import_from_path(name, path):
+    if PY36:
+        import importlib.util
+        spec = importlib.util.spec_from_file_location(name, path)
+        m = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(m)
```
vodik: This works on PY35 as well, for what its worth. `spec_from_file_location` came with 3.4 and `module_from_spec` with 3.5

Kodiologist: I tried enabling this for Python 3.5 and I got a weird crash inside pytest, so I've changed it back.

vodik: Yeah, probably won't work until #1518 is in. I tested it there. We don't implement the importlib APIs correctly.

PR #1550, opened 2018-03-26
Move hy-repr to core and rename `eval` to `hy-eval`

Kodiologist:
`hy-repr` is about a year old and it seems to work pretty well. With this PR, `hy-repr` is used to print REPL output by default, and Hy gets a pair of functions `hy-repr` and `hy-eval` analogous to Python's `repr` and `eval`, the latter of which is no longer shadowed by Hy.

The change of default `repl-output-fun` will require a lot of documentation changes, of course, which I haven't yet done, but I will before this PR is merged.


gilch:
Veto. I prefer the Python-compatible reprs as default, so you can see what's actually there instead of having to guess. I think changing the default repl output function to hy-reprs would only cause confusion.

I don't mind changing `eval` back to Python's, so we don't have to use `builtins.eval`. But I'd prefer we keep the old name. So we'd invoke it as `hy.eval` instead of `hy-eval`, or `hy.hy_eval` on the Python side.

vodik:
Shame, because I was going to suggest doing it too.

> I don't mind changing `eval` back to Python's, so we don't have to use `builtins.eval`. But I'd prefer we keep the old name.

I need to jig it around a bit more, and maybe as my time frees up this week I'll pull it out and do it a separate PR, but the importer rework has this change included. But its not just `hy.eval` - it tries to expose Hy equivalents to Python's `parse`, `compile` and `exec` as well.

Kodiologist:
You can still review and approve this PR if you like it, and the PR can get merged with two approvals. What the veto means is just that the PR can't get merged under the two-week rule. For what it's worth, I fully expected Matthew to veto this because he disapproved the same thing a year ago, when I first wrote `hy-repr` (#1246).

tuturto:
I like idea of renaming ``eval`` to ``hy-eval``. I most of the time use ``hy-repr`` when using REPL, but I recognize that not everyone is doing that. Maybe we'll just vote on this and let the majority decide if ``hy-repr`` is enabled by default?

Kodiologist:
That sounds reasonable. Tuuka and I are for and Matthew is against. @vodik, are you for? Does anybody else (developers or users) have an opinion one way or the other?

gilch:
I'll reproduce examples from my previous objection below. This is never going to work right except for a narrow subset of data types. This is a niche use that shouldn't be the default. Going our own way makes us _less compatible_ with the Python ecosystem. We cannot afford to do that. Hy's _only_ chance to thrive is to embrace what Python offers. We can find other ways to fill these needs without breaking compatibility. 
***
Suppose you have some Python objects (`foo`s) from some Python library. They have reprs like 
`foo(1, duck())`, but no hy-repr. Now you make some Hy objects (`bar`s) that take some `foo`s as arguments. They have reprs like `bar(foo(1, 2), foo(3, bar(foo(4, 5))))`. So far, so good, the `bar` repr defers to the repr of its `foo` arguments, which defer to the `int` and `bar` reprs and the net result is still valid Python, and it will generate an equivalent `bar` object.

But wait, now we have to write a `__hy-repr__`. Now it prints `(bar foo(1, 2) foo(3, bar(foo(4, 5))))`, and that's one of those weird hybrids that's neither Hy nor Python. It's not clear at all to the future user how to parse this. Note that `bar` appears both in the form `(bar ...)` and `bar(...)`. Why? Because `foo` is written in Python and knows nothing about `__hy-repr__`, so of course it has to use `__repr__` from `bar` instead.

So what if we use generic placeholders instead? Now you've got 
`(bar <foo instance at 0xXXX1> <foo instance at 0xXXX2>)`. The user can totally parse this. But, you've hidden important information. Not good. It's not really valid Hy, but at least you know exactly where the holes are.

Okay, maybe we can auto-wrap the repr with the placeholders. Now you get 
`(bar <Py foo(1, 2)> <Py foo(3, bar(foo(4, 5)))>)`. It's still not valid Hy, but now we know both where the holes are and what goes in those holes. But look, the Hy repr version only applies to the first layer! We're still getting the form `bar(...)`. What's the point of a Hy repr if we're printing it all off in Python anyway?

Even though this is a contrived example, it is very representative of the common case. It's very typical in Python for the repr to be a constructor call that defers to the reprs of its arguments. There are more Python libraries written in Python than in Hy, so almost all Python objects will have no hy-repr.
***
@vodik consider my objections before deciding.

@kirbyfan64 @olasd @algernon, you seem to have objected before to similar proposals.

@hylang/core anyone else who wants to stop this, speak up.

Do we know any other users? @ekaschalk? If we don't have a strong consensus, maybe we should leave this open for a while to hear from them?

@paultag care to weigh in? Your opinion counts for a lot around here. We might also need an arbiter.

paultag:
This is a tough one. Let me think about it a bit. I can see both sides of this; on the one hand, having `eval` `eval` hy code when you're in Hy seems sensible (after all, `eval`'ing Python from Hy seems.... weird. `eval`'ing Hy forms seems more likely), but I also understand that overloading internals causes confusion, and one would expect accessing a Python internal from Hy would give you that internal.

I half-heartedly reach for the bashism `/` out of sheer instinct, but I'm also sure that's not the right solution either (e.g.:

```
$ alias ls=echo
$ ls /tmp
/tmp
$ \ls /tmp
pulse-PKdhtXMmr18n
```
)

Let me work this over a bit

Kodiologist:
To be clear, the point of most controversy here is using `hy-repr`, rather than Python's `repr`, as the default method for printing results in the Hy REPL.

gilch:
Right, I'm OK with having `eval` pointing to Python's version. The proposal here was `hy-eval`. I made a counter proposal of keeping it in the `hy` namespace and accessing it as `hy.eval` by convention. We just have to call the two versions different names so we don't lose access to either. We could keep Hy's `eval` and rename Python's to `py-eval`. We could even call Hy's `evaluate` and Python's `eval`.

Anyway, the part I'm objecting to most is the Hy repl not printing valid Python reprs by default. If it's valid Python we could at least paste them back in the repl and `eval` them (using Python's eval) most of the time. With Hy reprs, it's much more fragile, since they can contain Python objects that don't have a Hy repr. Thus, it only works on a very restricted set of data types, or we force the user to write a hy repr for any Python type they might ever use. I don't think this can be automated reliably, not that we've even tried.

Kodiologist:
@tuturto, it doesn't look like anybody else is weighing in, so the majority vote as it stands is aye.

Kodiologist:
To be clear, `hy-repr` uses registered functions, rather than methods, as of #1516.

> I would suggest that (if we want Hy reprs at all) we automatically convert Python reprs to Hy syntax.

I don't think this is a good idea. It seems error-prone. But if I did it, would you approve this pull request?

Kodiologist:
@brandonwillard I'm thinking of picking this back up again, but I recall that you wrote in a blog post about a major overhaul of `hy-repr` you were working on. Do you think you'll submit that soon, and that it will make the changes here irrelevant?

Kodiologist:
This should be made obsolete by the bigger core changes I'd like to make in the future.

Kodiologist:
The PyPy bug fix in the first commit has since gotten in by other means.

PR #1551, opened 2018-03-27
Add `hy --spy` in quickstart

waigx:
Since many people came to `hy` from Python community, knowing this would be helpful for learning as well as debugging.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -26,23 +26,29 @@ Quickstart
        etc
 
 6. Hit CTRL-D when you're done.
+7. If you're familiar with Python, try ``hy --spy`` to check what happens inside::
+
+       => (+ "Hyllo " "World" "!")
+       'Hyllo ' + 'World' + '!'
+       
+       'Hyllo World!'
 
 *OMG! That's amazing! I want to write a Hy program.*
 
-7. Open up an elite programming editor and type::
+8. Open up an elite programming editor and type::
 
        #! /usr/bin/env hy
        (print "I was going to code in Python syntax, but then I got Hy.")
 
-8. Save as ``awesome.hy``.
-9. Make it executable::
+9. Save as ``awesome.hy``.
+10. Make it executable::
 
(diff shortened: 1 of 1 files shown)


kirbyfan64:
Thanks!

docs/quickstart.rst
```diff
@@ -26,23 +26,29 @@ Quickstart
        etc
 
 6. Hit CTRL-D when you're done.
+7. If you're familiar with Python, try ``hy --spy`` to check what happens inside::
```
gilch: @kirbyfan64 Doc fixes only require one approval. Is it clear to a new user that you start the repl with `--spy`, rather than entering it in the repl itself?

waigx: Just updated a little bit: [quickstart.rst](https://github.com/waigx/hy/commit/e0d37d63f703a0c461a1be2465df041a538a2973)
Not sure if it worths another pull request.

kirbyfan64: @gilch Uh, yeah...did I miss something? Using a mobile app, so I'm not sure if I screwed something up...

IMO considering the previous step was closing the REPL, and this one looks like a shell command, it seemed obvious to me.

gilch: > not sure if I screwed something up

You didn't. I just noticed you hadn't merged yet, so I suggested an improvement. (about the same time you merged).

> it seemed obvious to me.

Could be. It's hard to tell what's obvious to new people when you're already familiar with it. Don't worry about merging doc fixes that make sense. We have the one-approval rule for docs to reduce friction.

> Not sure if it worths another pull request.

I personally don't mind small doc merges if its a clear improvement. Wiki edits are often tiny, but they seem to accumulate good quality. Others may feel differently about adding noise to Hy's git history. If that's an issue, github does do wikis too. We could also put the docs in a separate branch or separate repository and only combine them for the tagged releases. idk

waigx: Thanks for the suggestions - opened a small PR for above concerns

PR #1552, opened 2018-03-28
Fix `mangle` for Pythons compiled with UCS-2

waigx:
When `hy` python compiled with UCS-2, the unit tests failed because in the [mangle function](https://github.com/hylang/hy/blob/master/hy/lex/parser.py#L54), `😂` unicode represents as two 16bit int `0xD83D` `0xDE02` rather than one 32bit int `0x1F602`.

This problem only affects before Python version 3.3 ([PEP 393](https://www.python.org/dev/peps/pep-0393/) uses UCS-4 on all systems).

ref:
[1] https://docs.python.org/2/howto/unicode.html#the-unicode-type
[2] https://docs.python.org/3/library/sys.html#sys.maxunicode

### DIFF
--- a/hy/lex/parser.py
+++ b/hy/lex/parser.py
@@ -10,7 +10,7 @@ import string, re, unicodedata
 
 from rply import ParserGenerator
 
-from hy._compat import PY3, str_type, isidentifier
+from hy._compat import PY3, str_type, isidentifier, UCS4
 from hy.models import (HyBytes, HyComplex, HyCons, HyDict, HyExpression,
                        HyFloat, HyInteger, HyKeyword, HyList, HySet, HyString,
                        HySymbol)
@@ -25,14 +25,28 @@ pg = ParserGenerator(
 
 mangle_delim = 'Δ' if PY3 else 'X'
 
+def unicode_to_ucs4iter(ustr):
+    # Covert a unicode string to an iterable object,
+    # elements in the object are single USC-4 unicode characters
+    if UCS4:
+        return ustr
+    ucs4_list = list(ustr)
+    for i, u in enumerate(ucs4_list):
+        if 0xD7FF < ord(u) < 0xDC00:
+            ucs4_list[i] += ucs4_list[i + 1]
+            del ucs4_list[i + 1]
+    return ucs4_list
+
 def mangle(s):
     """Stringify the argument and convert it to a valid Python identifier
     according to Hy's mangling rules."""
+    def unicode_char_to_hex(uchr):
+        # Covert a unicode char to hex string, without prefix
+        return uchr.encode('unicode-escape').decode('utf-8').lstrip('\\U').lstrip('\\u').lstrip('0')
 
     assert s
 
     s = str_type(s)
-
     s = s.replace("-", "_")
     s2 = s.lstrip('_')
     leading_underscores = '_' * (len(s) - len(s2))
@@ -50,8 +64,8 @@ def mangle(s):
                  # allowed at the start of an identifier.
                else '{0}{1}{0}'.format(mangle_delim,
                    unicodedata.name(c, '').lower().replace('-', 'H').replace(' ', '_')
-                   or 'U{:x}'.format(ord(c)))
-            for c in s)
+                   or 'U{}'.format(unicode_char_to_hex(c)))
+            for c in unicode_to_ucs4iter(s))
 
     s = leading_underscores + s
     assert isidentifier(s)
--- a/hy/_compat.py
+++ b/hy/_compat.py
@@ -25,6 +25,10 @@ PY35 = sys.version_info >= (3, 5)
 PY36 = sys.version_info >= (3, 6)
 PY37 = sys.version_info >= (3, 7)
 
+# The value of UCS4 indicates whether Unicode strings are stored as UCS-4.
+# It is always true on Pythons >= 3.3, which use USC-4 on all systems.
+UCS4 = sys.maxunicode == 0x10FFFF
+
 str_type     = str   if PY3 else unicode      # NOQA
 bytes_type   = bytes if PY3 else str          # NOQA
 long_type    = int   if PY3 else long         # NOQA


Kodiologist:
I don't think it's the test that's wrong, but the `mangle` function. `😂` is Unicode code point U+1F602, so on Python 2, presuming the character name isn't available, it should be mangled to `hyx_XU1f602X`, not `hyx_XUd83dXXUde02X`.

waigx:
@Kodiologist - Yeah, I tried to fix `mangle` function yesterday but didn't figured out a working solution. Unicode on python 2 seems really tricky.

I think it is fine as long as `unmangle` converts the emoji back. The internal represents shouldn't really matter.

Kodiologist:
A mangled name isn't just an internal representation, though. It's what you have to use to access the variable in Python. We do unfortunately have to tolerate some inconsistency here in terms of Python 2 vs. Python 3, because of the changes in Python's own identifier rules, but let's not tolerate an inconsistency caused by an esoteric detail of one's Python build.

waigx:
That sounds reasonable. But since `mangle` a frequently called function, I just little bit concern about the performance trade off on fixing that.

Kodiologist:
I doubt it will make a perceptible speed difference, but you can always benchmark and find out.

vodik:
Just a quick reminder we don't actually support 3.3. Its 3.4 and newer.

Expanding support to 3.3 has implications that complicates the importer work I'm doing.
-- 

Kodiologist:
Okay, this should also affect Python 2, then.

waigx:
Ah, right. This only affects Python 2. The performance difference depends on the solution we come up with.

If 'hyx_XU1f602X' is correct on all systems, essentially we want a "literal" unicode or coverting UTF-16s to UTF-32. Not sure if some internal methods could do that.

vodik:
Yeah, good point.

Would it maybe make more sense to mangle to the utf8 representation?

On March 27, 2018 8:36:58 PM EDT, Kodi Arfer <notifications@github.com> wrote:
>Okay, this should also affect Python 2, then.
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/hylang/hy/pull/1552#issuecomment-376720597

-- 

Kodiologist:
>  Would it maybe make more sense to mangle to the utf8 representation?

I don't think so. It would be longer and harder for the casual user to figure out what character is being represented. Code points are the most widely used numeric identifiers for Unicode characters.

waigx:
OK, talked with @gilch in the IRC and sounds like there won't be a performance problem.

Just updated the `mangle` function for `UCS-2` case. Please take another look when you have a chance. Thanks.

hy/lex/parser.py
```diff
@@ -50,8 +60,8 @@ def mangle(s):
                  # allowed at the start of an identifier.
                else '{0}{1}{0}'.format(mangle_delim,
                    unicodedata.name(c, '').lower().replace('-', 'H').replace(' ', '_')
-                   or 'U{:x}'.format(ord(c)))
-            for c in s)
+                   or 'U{}'.format(c.encode('unicode-escape').decode('utf-8').lstrip('\\U').lstrip('\\u').lstrip('0')))
```
vodik: So long as `c` is unicode, `ord` should work:

```python
or 'U{:x}'.format(ord(c))
```

hy/lex/parser.py
```diff
@@ -29,6 +29,16 @@ def mangle(s):
...
     according to Hy's mangling rules."""
 
+    def convert_to_ucs4_list(ustr):
+        if UCS4:
+            return ustr
+        ucs4_list = list(ustr)
+        for i, u in enumerate(ucs4_list):
+            if 0xDC00 > ord(u) > 0xD7FF:
+                ucs4_list[i] += ucs4_list[i + 1]
+                del ucs4_list[i + 1]
```
vodik: Mutating a list while iterating over it fields weird to me. Using a generator here might be cleaner, but I suspect this would be faster.

hy/lex/parser.py
waigx: If c is a two 16-bit int unicode, `ord(c)` throws in a UCS2 python system:

```
>>> ord(u'\U0001f5b1')
TypeError: ord() expected a character, but string of length 2 found
```

But I have to say my solution is really ugly - just a patch I can think of without refactor the function.

hy/lex/parser.py
waigx: @vodik - Actually I guess `del` operation is quite expensive - I wrote like that for pythonic and simplicity.

But I can change this part if we'd prefer something like this (from my understanding):
``` Python
i, ucs4lst = 0, []
while i < len(ustr):
  usc4lst.append(usrt[i])
  if LOW < ustr[i] <HIGH:
    usc4lst[-1] += ustr[i]
    i += 1
  i += 1
```

hy/lex/parser.py
vodik: Oh of course it be the same problem.

hy/lex/parser.py
vodik: You know what? I don't think its worth putting a lot of effort in. Whatever is cleanest.

We don't support 3.3, and if its just for Python 2 support - well, that's also rapidly approaching EOL

waigx: Agreed, we probably want to polish related code a little bit after ditching Python 2

Kodiologist:
Great work.

- `convert_to_ucs4_list`, despite being `mangle`'s subroutine, should probably be defined outside it, since we don't need to redefine it every time we call `mangle`.
- `convert_to_ucs4_list` would be better named `convert_to_ucs4` or something, since it might return a `unicode` or `str` instead of a list.
- I think you should write `0xD7FF < ord(u) < 0xDC00` rather than `0xDC00 > ord(u) > 0xD7FF`.
- I would reword the comment for `UCS4` to "The value of UCS4 ndicates whether Unicode strings are stored as UCS-4. It is always true on Pythons >= 3.3, which use USC-4 on all systems."
- Add a NEWS item.
- Rebase your work into one commit.
- As a separate commit, add yourself to AUTHORS.

waigx:
@Kodiologist - Thanks for suggestion, the code has been updated accordingly:
  - pulled it out and renamed `convert_to_ucs4_list` --> `unicode_to_ucs4iter`;
  - styling, comment updated;
  - checked the NEWS, I feel "Mangling rules have been overhauled, such that mangled names are always legal Python identifiers" is already good. Since the `Mangling` change hasn't been release yet, this commit will be part of it.
  - PR is squashed and rebased
  - thanks for the reminder, I will do :D

Kodiologist:
> checked the NEWS, I feel "Mangling rules have been overhauled, such that mangled names are always legal Python identifiers" is already good. Since the Mangling change hasn't been release yet, this commit will be part of it.

Oh, okay, that makes sense.

Kodiologist:
@vodik What do you think?

kirbyfan64:
I'm a bit confused by this...in Python 2, strings are basically byte strings, which means that *only* the `unicode` type would use UCS-2?

Kodiologist:
Yes, since `str` on Python 2 is always a sequence of bytes.

hy/lex/parser.py
vodik: And I just noticed the old code did ord. Okay. So I just have one more suggestion to make and I'm happy with this:

I have a feeling someone is going to come along, look at this, and say someone "Wait... isn't this ord?" and break this all over again. Not to mention we have to remember to pull it out once we finally drop Python 2 support.

So may I suggest adding a `multibyte_ord` helper?

```
def multibyte_ord(ustr):
    if UCS4:
        return ord(ustr)
    return ustr.encode('unicode-escape').decode('utf-8').lstrip('\\U').lstrip('\\u').lstrip('0')
```

Makes it clearer that this klugde is in there intentionally?

Kodiologist: That makes sense to me.

PR #1554, opened 2018-03-28
Report objects when raising type error

tuturto:
Improves error reporting just a tiny bit by including offending objects and thus making debugging easier


Kodiologist:
It might be better to get the `repr`s (`r!`) than the stringifications. Also, you don't need the numbers in the format string.

tuturto:
good catch, should be better now

kirbyfan64:
I just squashed + merged.

vodik:
@tuturto heads up in case you didn't know, `"{!r}".format(foo)` does repr (and `!s` forces str, which is the default).

PR #1555, opened 2018-03-29
Add UCS-2 Python 2 to Travis testing environments

waigx:
This commit adds USC-2 Python 2 to the testing environments.

USC-2 Python 2 is the default python environment comes with `macOS`. Also, USC-2 is the choose value if compiling python 2 from source code. Considering the user base for languages like `Hy`. It is reasonable to add it to the Travis environment.

To do so, we could add platform `osx` in the matrix configuration, but `osx` platform usually has a long waiting time before task starts (~1h).

Another way is using pyenv to compile a new python instance, which what this PR does.


waigx:
Testing failure on the first environment is expected:
https://travis-ci.org/hylang/hy/jobs/359655542

Kodiologist:
This is interesting, but it might be overkill, seeing how long it took us to create a bug related to UC2-2.

Kodiologist:
@waigx By the way, you might want to avoid slashes in branch names because of how Git uses slashes for remotes. `git checkout feat/travis` could mean checking out a branch named `feat/travis` or it could mean checking out a branch named `travis` from a remote named `feat`.

waigx:
Thank all for reviewing and suggestions.

@Kodiologist - Yes, that is a valid concern.
This is all about how git manages these pointers. Internally git have all items inside `.git/refs` folder, inside of it, has sub-folders. git provides a extensive tools to manipulate these folders/refs. In my case, my `.git/refs` file system structure will looks like:
```
- refs/
  - remotes/
     ...
  - heads/
     - fix/
        - ucs2/
        - problem1/
     - feat/
        - travis/
        - interesting/
```
Back to your example, if git got confused, say `feat/travis`, it will warning you and always checkout the local branch. git also could be also confused if you have same tag and branch name.

I just used to organize my code with `/` durning past 2 years. typically I have `fix`, `feat`, `chore`, `doc` folders on my working git. And have my remote names as `upstream`, `origin`, `upstream-gitlab` etc. I feel the `/` notation is pretty "nature" for me to organize my code.

That been said, I am totally open to suggestions of organizing my git. Also I am willing to follow the existing rules or conventions on projects I am working on.

Kodiologist:
I didn't know that Git prints a warning message in that case. That's good.

Kodiologist:
I thought about it and ultimately, I disagree with this change. We already have a lot of Travis jobs, meaning that Travis builds already take a while, and all this one is doing is checking for an extremely unusual kind of bug.

waigx:
Travis jobs run in parallel, typically `pypy3` is the last one to pass even with this change in.

But I do agree that all this is meant to test a Python2 only pit fall, and makes `.travis.yml` unclear. Hope other developers could give some advices.

Kodiologist:
I'm not too familiar with the innards of Travis, but my impression is that rather than every job running in parallel, only 3 or 4 jobs at a time actually run while the rest wait.

waigx:
@Kodiologist - I honestly not really sure either... but seems their document say so.
`Stages group jobs that run in parallel, while their stages run sequentially.`[ref](https://docs.travis-ci.com/user/build-stages/#How-do-Build-Stages-work%3F)

But either way I agree this increase the demand on QA boxes resources. (wait I think that's the point of QA: reduce potential bug with pre-resources 😺)

Issue #1488, opened 2018-01-10
Support more of CPython's command-line options


Kodiologist:
[See the documentation here.](https://docs.python.org/3/using/cmdline.html) A few that we might want to support are `-b`, `-B`, `-E`, `-O`, and `-S`.

waigx:
This sounds a good idea. I am going to grab `-E` supporting.

Kodiologist:
But then I'd have to run icky proprietary software.

waigx:
I am honestly not familiar with `PYTHONCASEOK`... How does this effect the project?

kirbyfan64:
Because it only has an effect on Windows and macOS, which are "icky" proprietary systems...

waigx:
@kirbyfan64 haha gotcha

PR #1557, opened 2018-03-30
Add -E support for Hy REPL

waigx:
This commit adds `-E` support for Hy. Similar to Python, hy will ignore
all PYTHON* environment variables, e.g. PYTHONPATH and PYTHONHOME,
that might be set.

This is an enhancement mentioned in https://github.com/hylang/hy/issues/1488

### DIFF
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -271,6 +271,8 @@ def cmdline_handler(scriptname, argv):
                         help="program passed in as a string")
     parser.add_argument("-m", dest="mod",
                         help="module to run, passed in as a string")
+    parser.add_argument("-E", action='store_true',
+                        help="ignore PYTHON* environment variables")
     parser.add_argument("-i", dest="icommand",
                         help="program passed in as a string, then stay in REPL")
     parser.add_argument("--spy", action="store_true",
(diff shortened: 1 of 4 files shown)


hy/cmdline.py
```diff
@@ -436,3 +442,10 @@ def _print_for_windows(src):
             print(line)
         except:
             print(line.encode('utf-8'))
+
+# remove PYTHON* environment variables,
+# such as "PYTHONPATH"
+def _remove_python_envs():
+    for key in os.environ.copy().keys():
```
vodik:
```python
for key in list(os.environ.keys()):
    # ...
```
You only need to copy the key view, not the whole dictionary

waigx: Thanks for suggestion, code has been updated

(I actually used `list` before but for simplicity changed it to something like this, now changing it back :D)

Kodiologist:
Also, @waigx, I didn't realize you went to Stony Brook. I got my PhD in psychology there. Small world.

waigx:
@Kodiologist - Ahhh, I totally forgot to update NEWS — my bad. Thanks for adding that.

Really? I got my masters in SBU two years ago (and had an amazing time there). Small world indeed! Go Seawolves! 🐾

PR #1558, opened 2018-03-31
Fix unit tests when run with PYTHONDONTWRITEBYTECODE

vodik:
When set, it will conflict with any tests that generate bytecode because they don't expect it to be set.

Fixable by sanitize the environment before forking, but we can't do anything about import tests.

This is a pipenv default, and possibly a sane flag to set while doing Hy development, so lets not let it be a hazard for developers to trip over.

### DIFF
--- a/tests/importer/test_importer.py
+++ b/tests/importer/test_importer.py
@@ -10,6 +10,7 @@ import os
 import ast
 import tempfile
 from fractions import Fraction
+import pytest
 
 
 def test_basics():
@@ -50,6 +51,8 @@ def test_import_error_reporting():
     assert _import_error_test() is not None
 
 
+@pytest.mark.skipif(os.environ.get('PYTHONDONTWRITEBYTECODE'),
+                    reason="Bytecode generation is suppressed")
 def test_import_autocompiles():
     "Test that (import) byte-compiles the module."
 
--- a/tests/test_bin.py
+++ b/tests/test_bin.py
@@ -24,10 +24,12 @@ def hr(s=""):
 
 
 def run_cmd(cmd, stdin_data=None, expect=0, dontwritebytecode=False):
-    env = None
+    env = dict(os.environ)
     if dontwritebytecode:
-        env = dict(os.environ)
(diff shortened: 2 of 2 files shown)


vodik:
Oh, and for what its worth, any opinions on including `Pipfile` and `Pipfile.lock` files?  Its personally won me over and I use it in all my new projects.

Kodiologist:
What do those files do and what would we accomplish with them?

vodik:
Pipfile is a toml based replacement for requirements.txt that does a nicer job of describing the environment

The lock file pins the very specific package versions for reproducible builds

https://github.com/pypa/pipfile/blob/master/README.rst
-- 

tests/importer/test_importer.py
```diff
@@ -10,6 +10,10 @@
 import ast
 import tempfile
 from fractions import Fraction
+import pytest
+
+
+PYTHONDONTWRITEBYTECODE = os.environ.get('PYTHONDONTWRITEBYTECODE')
```
Kodiologist: We only use this once, so we don't need the assignment.

Kodiologist:
The documentation for Pipfiles says they're a work in progress, whereas requirements files are stable. And we don't seem to need anything we can't get with requirements files. So let's stick with requirements files for now.

tests/importer/test_importer.py
vodik: Fair enough

vodik:
@Kodiologist technically true, but its as a point where its very stable. I doubt they'll remove that warning until Pipfile support ends up in pip itself.

pipenv has become the officially recommended Python packaging tool now, and it creates and manages a Pipfile for you, by reading requirements.txt and other sources. So we don't need to check it in technically, but it does give us some nicer control.

If you're not familiar with the tool, I'd really suggest having a look whenever you got the moment.

Kodiologist:
Okay, I doubt it'll make too much of a different either way.

vodik:
Yeah, I just mainly want this in for my sanity :grin:

PR #1559, opened 2018-03-31
Python 3.8 development has started 😉

vodik:
https://docs.python.org/3.8/whatsnew/3.8.html

Lets keep an eye on its compatibility.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,12 @@ python:
   - "3.5"
   - "3.6"
   - "3.7-dev"
+  - "3.8-dev"
   - pypy
   - pypy3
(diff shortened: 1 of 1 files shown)


vodik:
Ah, it was pulled because of OpenSSL issues: https://github.com/travis-ci/travis-ci/issues/9069

Kodiologist:
Seeing as it doesn't work, we can close this, yes?

vodik:
Yeah, let's move it to an issue.
-- 

Kodiologist:
Even an issue seems premature. Python 2.7 isn't even out yet.

PR #1560, opened 2018-03-31
Add recent REPL results globals *i and most recent error *e

waigx:
Closes https://github.com/hylang/hy/issues/1341

### DIFF
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -73,8 +73,20 @@ class HyREPL(code.InteractiveConsole):
         code.InteractiveConsole.__init__(self, locals=locals,
                                          filename=filename)
 
+        # Pre-mangle symbols for repl recent results: *1, *2, *3
+        self._repl_results_symbols = [mangle("*{}".format(i + 1)) for i in range(3)]
+        self.locals.update({sym: None for sym in self._repl_results_symbols})
+
     def runsource(self, source, filename='<input>', symbol='single'):
         global SIMPLE_TRACEBACKS
+
+        def error_handler(e, use_simple_traceback=False):
+            self.locals[mangle("*e")] = e
+            if use_simple_traceback:
+                print(e, file=sys.stderr)
+            else:
+                self.showtraceback()
+
         try:
             try:
                 do = import_buffer_to_hst(source)
@@ -84,7 +96,7 @@ class HyREPL(code.InteractiveConsole):
             if e.source is None:
                 e.source = source
                 e.filename = filename
-            print(e, file=sys.stderr)
+            error_handler(e, use_simple_traceback=True)
             return False
 
         try:
@@ -101,24 +113,23 @@ class HyREPL(code.InteractiveConsole):
             if e.source is None:
                 e.source = source
                 e.filename = filename
-            if SIMPLE_TRACEBACKS:
-                print(e, file=sys.stderr)
-            else:
-                self.showtraceback()
+            error_handler(e, use_simple_traceback=SIMPLE_TRACEBACKS)
             return False
-        except Exception:
-            self.showtraceback()
+        except Exception as e:
+            error_handler(e)
             return False
 
         if value is not None:
-            # Make the last non-None value available to
-            # the user as `*1`.
-            self.locals[mangle("*1")] = value
+            # Shift exisitng REPL results
+            next_result = value
+            for sym in self._repl_results_symbols:
+                self.locals[sym], next_result = next_result, self.locals[sym]
+
             # Print the value.
             try:
                 output = self.output_fn(value)
-            except Exception:
-                self.showtraceback()
+            except Exception as e:
+                error_handler(e)
                 return False
             print(output)
         return False
--- a/tests/test_bin.py
+++ b/tests/test_bin.py
@@ -75,8 +75,15 @@ def test_bin_hy_stdin_multiline():
 
 
 def test_bin_hy_history():
-    output, _ = run_cmd("hy", '(+ "a" "b")\n(+ *1 "y" "z")')
-    assert "'abyz'" in output
+    output, _ = run_cmd("hy", '''(+ "a" "b")
+                                 (+ "c" "d")
+                                 (+ "e" "f")
+                                 (.format "*1: {}, *2: {}, *3: {}," *1 *2 *3)''')
+    assert "'*1: ef, *2: cd, *3: ab,'" in output
+
+    output, _ = run_cmd("hy", '''(raise (Exception "TEST ERROR"))
+                                 (+ "err: " (str *e))''')
+    assert "'err: TEST ERROR'" in output
 
 
 def test_bin_hy_stdin_comments():
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -20,6 +20,8 @@ New Features
 ------------------------------
 * Python 3.7 is now supported
 * Added `mangle` and `unmangle` as core functions
+* More REPL history result variables: `*2`, `*3`. Added last REPL error
+  variable: `*e`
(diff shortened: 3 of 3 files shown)


Kodiologist:
Looks like we've got test failures.

waigx:
D’oh, seems it only works on python 2. Will update it once I got back to home. Also the NEWs ...

gilch:
In #1341, I also asked for the last exception (the one caught by the repl not quitting) to be stored in `*e`. This lets us examine exception objects in the repl without having to write a `try` form every time to do it. I don't want to close that issue without `*e`.

It would also be nice to store the last three input strings (or forms), like IPython and Common Lisp do. This is less important than `*e`, since we can copy/paste it.

A nicer repl, like IPython, stores all of them and updates the prompt to show the current entry number. @ekaschalk has discussed some about how we would integrate that into emacs. I've also mentioned that IPython is quite extensible, and could be adapted to Hy. It might be easier to do that as a separate project that has both IPython and Hy as a dependency, rather than trying to bring Hy's built-in repl up to that level. So the last three would suffice.

vodik:
> A nicer repl, like IPython, stores all of them and updates the prompt to show the current entry number. @ekaschalk has discussed some about how we would integrate that into emacs. I've also mentioned that IPython is quite extensible, and could be adapted to Hy.

Another option, and one I've been meaning to try as a weekend project: build something on top of the [python prompt toolkit](https://github.com/jonathanslenders/python-prompt-toolkit)

hy/cmdline.py
```diff
@@ -111,9 +119,17 @@ def ast_callback(main_ast, expr_ast):
             return False
 
         if value is not None:
+            # Shift exisitng REPL results queue
+            for i in xrange(1, self._repl_results_queue_len):
```
vodik: Stick to Python 3 isms (in this case, prefer `range` over `xrange`) and leave Python 2 a second class citizen.

Its harsh, and I wish we could drop Python 2 support, but I also don't think any of the maintainers are running Hy on Python 2 either.

hy/cmdline.py
```diff
@@ -58,6 +58,9 @@ def __init__(self, spy=False, output_fn=None, locals=None,
 
         self.spy = spy
 
+        # Queue size of stored REPL results
+        self._repl_results_queue_len = 3
```
vodik: Why not, instead of going through a range of numbers and redoing the mangling every time, do it once ahead of time?

```python
self._repl_results_vars = [mangle("*{}".format(i) for i in range(3))]
```

Then its just a mater of going through the list:

```python
value = ...
for var in self._repl_results_vars:
    self.locals[var], value = value, self.locals[var]
```

vodik:
Looks good, just needs some work. And yeah, getting `*e` in there would be awesome!

hy/cmdline.py
waigx: I was thinking about this, after checking the behavior of python. It seems that `_` will be undefined if no result from previous lines: Accessing `_` throws error.

I didn't check iPython but assume it might be the same. I guess we would like keep such behavior for `*x`, what do you think?

Kodiologist: I don't think that's necessary, or even particularly desirable.

waigx: @Kodiologist - What the default value then? It might be a little weird If we use `None` as the default value, as we don't allow None appears in `*x`.

The implementation could be much simpler if we don't want this behavior, just initialize and shift (I don't even think the queue size need to be configurable):

vodik: @waigx you sure?

```
ipython
Python 3.6.4 (default, Jan  5 2018, 02:35:40) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.2.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: _
Out[1]: ''
```

waigx:
```
I didn't check iPython but assume it might be the same.
```

hahaha 😄

waigx: alright, you guys win 😤

waigx: What the initial values then - apparently iPython is using `''`

vodik: `None` or `''` works for me.

waigx: no problem at all - I just used `None` as default since that makes more sense to me. The code has been updated (literally copied you suggestion above). Please take another look.

waigx:
Hi @gilch , the `*e` has been added (I saw that that but didn't add it - not meant to close that issue). Please take a looks see if that looks good to you. Thanks.

Kodiologist:
Looks good to me. How about adding some simple tests to `test_bin.py`?

waigx:
@Kodiologist - Thanks for the reminder, I just added the unit tests.

hy/cmdline.py
```diff
@@ -73,8 +73,20 @@ def __init__(self, spy=False, output_fn=None, locals=None,
...
                                          filename=filename)
 
+        # Pre-mangle symbols for repl recent results: *1, *2, *3
+        self._repl_results_symbols = [mangle("*{}".format(i + 1)) for i in range(3)]
+        self.locals.update({sym: None for sym in self._repl_results_symbols})
+
     def runsource(self, source, filename='<input>', symbol='single'):
         global SIMPLE_TRACEBACKS
+
+        def error_handler(e, use_simple_traceback=SIMPLE_TRACEBACKS):
```
vodik: Be careful of how positional arguments work here. The default value is created once and shared between all function calls:

```python
>>> TEST = "Hello!"
>>> def foobar():
...     global TEST
...     
...     def inner(test=TEST):
...         print(test)
...         
...     inner()
...     TEST = "ANOTHER MESSAGE"
...     inner()
... 
>>> foobar()
Hello!
Hello!
```

Technically this should be fine because we only read from it, but its something linters will complain about.

I think it be best to just make it an explicit argument always, since you rely on the implicit default twice and explicitly set it twice.

tests/test_bin.py
```diff
@@ -75,8 +75,15 @@ def test_bin_hy_stdin_multiline():
 
 
 def test_bin_hy_history():
-    output, _ = run_cmd("hy", '(+ "a" "b")\n(+ *1 "y" "z")')
-    assert "'abyz'" in output
+    output, _ = run_cmd("hy", '(+ "a" "b")\n\
+                               (+ "c" "d")\n\
+                               (+ "e" "f")\n\
+                               (.format "*1: {}, *2: {}, *3: {}," *1 *2 *3)')
```
vodik: Its a matter of style, but I prefer using brackets over `\`:

```python
output, _ = run_cmd("hy", (
    '(+ "a" "b")'
    '(+ "c" "d")'
    '(+ "e" "f")'
    '(.format "*1: {}, *2: {}, *3: {}," *1 *2 *3)'
))

Technically Hy doesn't care about the newlines either, so no need to have the visual noise.

hy/cmdline.py
waigx: Thanks, I didn't know that could be a concern....

Just updated the code and make the default value to `False`. I am thinking that to use the key value when calling `error_handler`, so the code is easier to reason. made it default to false to void verbose.

tests/test_bin.py
waigx: Thanks! Again, I honest didn't know the `run_cmd` could be used that way. This looks nicer ... but unfortunately guessing hy treat the tuple as 4 individual code pieces. So the `*1` ... is `None`.

I updated to `'''` -- hope it looks better.

Kodiologist:
Don't forget another commit to update AUTHORS, by the way.

waigx:
@Kodiologist - Thanks for the reminder!

vodik:
Well, with #1570 merged, lets merge this.

PR #1561, opened 2018-04-01
Update .gitignore to exclude pytest cache folder

waigx:
The `.pytest_cache/` folder appears after running `make test` locally.
This commit also updates vim temporary files exclusion rules.


vodik:
Nice, been meaning to deal with that.
-- 

PR #1566, opened 2018-04-02
Port the completer to Hy

vodik:
Deals with the mangling problems in the REPL by properly mangling and unmangling before completing.

Also ported it to Hy, since, well, might as well be as self-hosting as possible.


hy/lex/parser.py
```diff
@@ -44,7 +44,8 @@ def unicode_char_to_hex(uchr):
         # Covert a unicode char to hex string, without prefix
         return uchr.encode('unicode-escape').decode('utf-8').lstrip('\\U').lstrip('\\u').lstrip('0')
 
-    assert s
+    if not s:
+        return ""
```
Kodiologist: Is this line (and the corresponding one in `unmangle`) returning the same type as the rest of the function?

hy/completer.hy
```diff
@@ -0,0 +1,111 @@
...
+      (import readline)
+      (except [ImportError]
+        (setv docomplete False)))))
+
+(defclass Completer []
+  (defn --init-- [self &optional [namespace {}]]
+    (when (not (instance? dict namespace))
+      (raise (TypeError "namespace must be a dictionary")))
+
+    (setv self.namespace namespace
```
Kodiologist: Shouldn't you copy `namespace` (or set the default argument of `namespace` to `None` and then assign `{}` in the body)?

Kodiologist:
I think you should allow completion of the mangled form, too. Thus, if you imported a symbol `is_foo` from a Python module, `is_f` will be recognized by the completer, as a user would probably expect.

hy/completer.hy
```diff
@@ -0,0 +1,111 @@
...
+                               (instance? string-types key))
+                     (viewkeys builtins.--dict--)
+                     (viewkeys (get hy.macros.-hy-macros None))
+                     (viewkeys namespace)]
+          self.tag-path [(viewkeys (get hy.macros.-hy-tag None))])
+
+    (setv module-name (.get namespace "__name__"))
+    (when module-name
+      (->> (get hy.macros.-hy-macros module-name) (viewkeys) (.append self.path))
+      (->> (get hy.macros.-hy-tag module-name) (viewkeys) (.append self.tag-path))))
```
Kodiologist: Why use `->>` and write these lines backwards when you could just write them forwards?

hy/completer.hy
```diff
@@ -0,0 +1,111 @@
...
+
+  (defn attr-matches [self text]
+    (setv match (re.match r"(\S+(\.[\w-]+)*)\.([\w-]*)$" text))
```
Kodiologist: This doesn't seem to allow for attributes with weird characters (that would need to be mangled) in their name.

hy/lex/parser.py
vodik: Yeah, so I don't mind putting an error back in (probably should raise a `ValueError` though) and handling the edge case inside the completer.

Kodiologist: That sounds reasonable to me.

hy/completer.hy
vodik: Yeah, good catch. I ported over the old code without thinking hard enough.

ekaschalk:
Have you all seen: https://github.com/ekaschalk/jedhy? Example usages here: https://github.com/ekaschalk/jedhy/blob/master/tests/test_models.hy

It performs completion on all constructs (macros, compile table, module members, etc...), among other actions.

I use it in the develop branch of `hy-mode`, pending merge while I'm still learning how to manage its auto-installation.

Kodiologist:
That's cool. Perhaps the completion parts should be moved into Hy, or perhaps its own completion should be replaced with Hy's.

vodik:
@ekaschalk I'd be in favour of it, and we always leverage optional dependencies for it. The Python repl doesn't  have any autocomplete and we can always keep a minimal install by default. Do something like `pip install hy[autocomplete]` or something.

Kodiologist:

PR #1568, opened 2018-04-04
Add AppVeyor for Windows CI

vodik:
In light of a few recent Windows issues, maybe it makes sense to leverage AppVeyor and get a Windows CI build going.

Its already catching some issues.

### DIFF
--- a/hy/contrib/hy_repr.hy
+++ b/hy/contrib/hy_repr.hy
@@ -10,6 +10,13 @@
   [hy._compat [PY3 PY36 str-type bytes-type long-type]]
   [hy.models [HyObject HyExpression HySymbol HyKeyword HyInteger HyFloat HyComplex HyList HyDict HySet HyString HyBytes]])
 
+(defmacro strftime [time formatstr]
+  "Version of strftime that handles some Windows/POSIX portability"
+  (import platform)
+  (when (= (platform.system) "Windows")
+    (setv formatstr (.replace (str formatstr) "%-" "%#")))
+  `(.strftime ~time ~formatstr))
+
 (try
   (import [_collections_abc [dict-keys dict-values dict-items]])
   (except [ImportError]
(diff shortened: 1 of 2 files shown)


vodik:
This should be stable: https://ci.appveyor.com/project/vodik/hy/build/1.0.8

Once merged I can turn it on for hylang/hy

Kodiologist:
Can you turn off the automatic AppVeyor check for every new PR?

vodik:
I set it to skip branches without `appveyor.yml` present, so its on but shouldn't spam every PR anymore until this gets merged.

vodik:
@Kodiologist style mostly. IMHO format strings should be treated as constants as much as possible - though I guess it would be safe in this case.

hy/contrib/hy_repr.hy
```diff
@@ -10,6 +10,13 @@
   [hy._compat [PY3 PY36 str-type bytes-type long-type]]
   [hy.models [HyObject HyExpression HySymbol HyKeyword HyInteger HyFloat HyComplex HyList HyDict HySet HyString HyBytes]])
 
+(defmacro strftime [time formatstr]
+  "Version of strftime that handles some Windows/POSIX portability"
+  (import platform)
+  (when (= (platform.system) "Windows")
+    (setv formatstr (.replace (str formatstr) "%-" "%#")))
```
vodik: Slightly annoying that `HyObject.replace` collides with `str.replace`.

Kodiologist:
> @Kodiologist style mostly. IMHO format strings should be treated as constants as much as possible - though I guess it would be safe in this case.

Wouldn't it suffice to write something like

    (.replace (.strftime x "%Y %m %d %H %M %S") "  " " ")

? Then the format string is a constant, and you don't need to check the platform.

vodik:
@Kodiologist Is the `strftime` macro acceptable? Similar idea but we shunt it to compile time.

Kodiologist:
I don't see how that would work. The line I wrote manipulates the return value of `.strftime`, which doesn't exist till runtime.

Kodiologist:

Issue #1571, opened 2018-04-04
Malformed escape sequence can crash the repl.


gilch:
Minimal example.
```
=> "\x8"
```
The repl should _never_ crash due to malformed input. If it accepts the line as complete, it should report the syntax error, not quit to the shell.

PR #1572, opened 2018-04-05
Raise LexException when codec can't decode some bytes

waigx:
Fixes https://github.com/hylang/hy/issues/1571

### DIFF
--- a/tests/test_lex.py
+++ b/tests/test_lex.py
@@ -69,6 +69,13 @@ bc"
     assert objs == [HyString("abc")]
 
 
+def test_lex_strings_exception():
+    """ Make sure tokenize throws when codec can't decode some bytes"""
+    with lexe() as execinfo:
+        tokenize('\"\\x8\"')
+    assert "Can't convert \"\\x8\" to a HyString" in str(execinfo.value)
+
+
 def test_lex_bracket_strings():
 
     objs = tokenize("#[my delim[hello world]my delim]")
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -31,6 +31,7 @@ Bug Fixes
 * Fix `(return)` so it works correctly to exit a Python 2 generator
 * Fixed a case where `->` and `->>` duplicated an argument
 * Fixed bugs that caused `defclass` to drop statements or crash
+* Fixed a REPL crash caused by illegle unicode escape string inputs
 
 Misc. Improvements
 ----------------------------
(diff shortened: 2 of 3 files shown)


tests/test_lex.py
```diff
@@ -69,6 +69,12 @@ def test_lex_strings():
     assert objs == [HyString("abc")]
 
 
+def test_lex_strings_exception():
+    """ Make sure tokenize throws when codec can't decode some bytes"""
+    with lexe() as e: tokenize('\"\\x8\"')
```
vodik: Do this across two lines *cough* *cough* PEP8 :wink: 

That said, I'd suggest renaming `e` to `execinfo`. Its consistent with other unit tests, and I think also better names what's provided as it isn't the exception itself.

waigx: Ah, sorry I was just copying and pasting other part of this file...

Really should have a good read of PEP8, and maybe find some PEP8 plugin for vim...

tests/test_lex.py
```diff
@@ -69,6 +69,13 @@ def test_lex_strings():
     assert objs == [HyString("abc")]
 
 
+def test_lex_strings_exception():
+    """ Make sure tokenize throws when codec can't decode some bytes"""
+    with lexe() as execinfo:
+        tokenize('\"\\x8\"')
+    assert "Can't convert \"\\x8\" to a HyString" in str(e.value)
```
vodik: Forget to update `e.value`? :wink:

waigx: D'OH

Kodiologist:
N.B. GitHub won't detect "Fix issue #1571" properly; [you need the issue number right after the word "fix"](https://help.github.com/articles/closing-issues-using-keywords/).

waigx:
@Kodiologist - Thanks, good to know that. I just just updated the initial comment.

tests/test_lex.py
Kodiologist: @waigx You can use `make flake` to run flake8 on the Hy codebase.

Kodiologist:
Looks good.

1. Don't you mean "**Throw** when codec can't decode some bytes"?
2. Update NEWS, unless you know this to be a regression since the previous release.

waigx:
@Kodiologist - Thanks, just added the NEWS (shamefully forgot that again... ) and re-worded the title. Please take another look.

PR #1575, opened 2018-04-06
Fix bug in macro-expanding NaN

### LINKED ISSUE #1574: (macroexpand 'NaN) enters infinite macro-expansion loop

Kodiologist:
Fixes #1574.

### DIFF
--- a/hy/macros.py
+++ b/hy/macros.py
@@ -181,8 +181,6 @@ def macroexpand_1(tree, compiler):
         fn = tree[0]
         if fn in ("quote", "quasiquote"):
             return tree
-        ntree = HyExpression(tree[:])
-        ntree.replace(tree)
 
         opts = {}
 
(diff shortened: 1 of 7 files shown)


hy/macros.py
```diff
@@ -174,48 +174,41 @@ def macroexpand(tree, compiler):
...
+        obj = m(compiler.module_name, *tree[1:], **opts)
+    except HyTypeError as e:
+        if e.expression is None:
+            e.expression = tree
+        raise
+    except Exception as e:
+        msg = "expanding `" + str(tree[0]) + "': " + repr(e)
+        raise HyMacroExpansionError(tree, msg)
+    replace_hy_obj(obj, tree)
+    return obj
```
vodik: On the topic of immutablity, it seems like we count on `replace_hy_obj` for its side-effects, but it also returns the replaced obj.

Might be easier for future work if we start using its return value: `return replace_hy_obj(obj, tree)`

Kodiologist:
Exactly.

Kodiologist:
I've changed this to a more direct approach that quits macro-expansion as soon as there's nothing left to expand instead of doing a comparison.

PR #1578, opened 2018-04-07
Add `returns` to function ast generation

vodik:
The `returns` entry was added as part of PEP 3107 to store a functions annotation for its return value, which there currently is no support for in Hy.

That said, we should try and always emit the same AST that Python would, so lets emit it anyways.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -2012,6 +2012,7 @@ class HyASTCompiler(object):
                     args=args,
                     body=body.stmts,
                     decorator_list=[],
+                    returns=None,
                     docstring=(None if docstring is None else
                         str_type(docstring)))
 


Kodiologist:
Let's cross this bridge when we come to it.

> we should try and always emit the same AST that Python would

That seems like a tall order.

vodik:
I swear you said that on another PR... Oh well, no strong feelings either way, figured it doesn't hurt. Feel free to close it if you want then.

Kodiologist:
I did say that Hy should be able to produce any given Python construct. For example, Hy should have a way to produce `lambda` expressions and decorators even though these aren't semantically necessary. Making sure that the complete AST tree matches what Python would produce for the equivalent code goes a lot further, because of issues like the management of Hy's temporary variables, and backwards-compatible AST changes that happens between Python versions but we don't attempt to track in Hy. I bet that in some cases we'd even have to reproduce some of Python's bugs.

PR #1579, opened 2018-04-08
Fix a bug where the compiler doesn't properly handle unquote-splice

vodik:
Added test coverage as well

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -51,6 +51,7 @@ Bug Fixes
 * Fixed a REPL crash caused by illegle unicode escape string inputs
 * `NaN` can no longer create an infinite loop during macro-expansion
 * Fixed a bug that caused `try` to drop expressions
+* Fixed a bug where the compiler didn't properly compile `unquote-splice`
 
 Misc. Improvements
 ----------------------------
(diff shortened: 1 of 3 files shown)


vodik:
@Kodiologist probably because its ambiguous: `(quote @)` a.k.a. matmul

Kodiologist:
Ah, all right, then.

vodik:
@Kodiologist hmmm, wonder if it makes sense to list the expression we failed to unquote though.

Kodiologist:
Don't our fancy error messages already do that, by pointing to the part of the code where you tried to unquote something?

vodik:
Yeah, its a good point. Its just harder to interpret with `~@`. But yeah, I rest my case.

gilch:
```
=> `(~@"mu")
HyExpression([
  'm',
  'u'])
=> `(~@ "mu")
HyExpression([
  <function hyx_Xcommercial_atX at 0x000001FDC787A730>,
  HyString('mu')])
=> `(~ @ "mu")
HyExpression([
  HySymbol('~'),
  HySymbol('@'),
  HyString('mu')])
```
Interesting that three interpretations are possible since `~` and `@` are also both operators (invert and matmul). Be careful with spacing when using these. Probably should be documented better.

PR #1580, opened 2018-04-08
Burninate HyCons

### LINKED ISSUE #568: cons doesn't always promote lists to HyObjects
```
=> (setv c (cons {"a" 1} []))
c = cons({u'a': 1L, }, [])
=> (type (car c))
type(c[0L])
<class 'hy.models.dict.HyDict'>
=> (get (car c) "a")
c[0L][u'a']
  File "/home/maresp/workbench/cbf/venv/lib/python2.7/site-packages/hy/models/list.py", line 43, in __getitem__
    ret = super(HyList, self).__getitem__(item)
TypeError: list indices must be integers, not unicode
=>
```

The cons is because I'm not sure how to get/construct a HyDict otherwise

### LINKED ISSUE #1576: Burninate HyCons
This is an idea that's been mentioned briefly in other contexts ([e.g.](https://github.com/hylang/hy/pull/1246#issuecomment-286272390)) but hasn't gotten its own issue. I want to remove cons cells because:

1. They don't seem to provide anything you couldn't get with lists, tuples, quasiquoting, HyExpression, or the like and a helper function or two.
2. They're warty. To my knowledge, the below features are all intentional.
    1. Dotted lists use an infix syntax, unlike everything else.
    2. Dotted lists aren't legal at the top level.
    3. The HyCons constructor autoboxes, unlike all other Hy model constructors.
    4. `cons` may not actually return a cons cell. For example, `(cons 1 [2 3 4])` returns `['1 2 3 4]`. (The single quote on the first value is from wart #iii.)

From a deep dive into issues and PRs, I found two explanations for why we have HyCons:

1. "We need cons support to be called a Lisp." (#183)
2. @algernon has [found it useful](https://github.com/hylang/hy/issues/240#issuecomment-131219593) for writing algernon/adderall and hylang/hydiomatic.

I disagree with (1) because we've successfully represented Lisp data structures with Python data structures and other Hy model objects like HyExpression and HyList.

As for (2), I'm not sure I understand the explanation or example, perhaps because I'm not familiar with these libraries, but it looks like `` `(~?fn ~?name ~?params . ~?body) `` could be written `` `(~fn ~?name ~?params ~@?body) ``. Note that we've already broken support for these two libraries in other ways (neither has been updated since April 2016).

How do we feel about this?

Kodiologist:
- Closes #568
- Closes #1576


gilch:
@hylang/core this is a major change. If you have objections, speak up.

Kodiologist:
That's technically enough to merge, but let's give it a grace period of a few days just in case.

Kodiologist:
Okay, very good.

PR #1586, opened 2018-04-16
Fix a bug that caused `try` to drop expressions

### LINKED ISSUE #1584: Multi expression "try" form discards some of them
When using multiple expressions in the body of a `try` form, those that are function calls seem to be discarded by the compiler.  Given the following code sample:

``` hy
(setv val 0)
(print (try
         (+= val 1)
         (print "Check")
         val
         (except [Exception] -1)))
```

Executing with Hy 0.14.0 under Python 3.5.2 prints `1` but **not** `Check`.  Running `hy2py3` prints the following code:

``` python
val = 0
try:
    val += 1
    _hy_anon_var_1 = val
except Exception:
    _hy_anon_var_1 = -1
print(_hy_anon_var_1)
```

Using a `do` form solves the issue.  However, the documentation seems to suggest that multiple expressions are allowed in the `try` body; as well as #1427.

Kodiologist:
Fixes #1584.


etanol:
Thanks for that, much appreciated :-)

Issue #1573, opened 2018-04-05
Pattern-matching compiler


Kodiologist:
The compiler currently inspects and manipulates trees of Hy models by hand. The next big Hy thing I want to do is replace that code with a more concise and hopefully less buggy pattern-matching approach. I'm currently thinking of some sort of parser-combinator approach that looks like this, using the `.` special form as an example:

    . -> [EXPR("invocant"),
        ANY("keys", OR(SYM("attr"), SB(EXPR("key"))))]

You could then write `tree.invocant` to get the first argument, or `tree.keys[1].key` to get a key that's in a square-bracketed third argument.

Whether I write a little parser-combinator library for the purpose or write some code to use an existing library, the resulting code will likely be useful for Hy users for writing macros, so I'll give it a public interface.

With this done, #1542 will be easier.

vodik:
On this note, and something I think I'll distract myself with - we should probably leverage `astor` and `ast.parse` to write more tests that assert we generate the same ast as a known python expression:

PR #1587, opened 2018-04-16
Ban `(except)`

Kodiologist:


Kodiologist:
Yes, I forgot that's okay here because I'm working on #1573 where I have to be a bit more careful with boolification.

Issue #1411, opened 2017-08-31
Rethink fn arguments list


gilch:
Hy currently has `&optional`, `&key`, `&kwargs`, `&rest`, and `&kwonly`.

Python does not *seem* this complicated. It only has `*` and `**` to remember. Clojure only has `&`, but it can use that with its dict destructure to [emulate kwargs](https://stackoverflow.com/questions/717963/clojure-keyword-arguments#3969693). It's an elegant solution, but I don't think it's good enough for Hy, which needs to interoperate closely with Python.

To be fair, Hy is not any worse than Common Lisp's `&allow-other-keys &environment &rest &aux &key &whole &body &optional`.

Is there any way we can simplify this? Any simplification here would be a very deep change, and bound to be controversial. But it's easier to build advanced macros on simpler, more consistent special forms. The time for breaking changes is sooner, not later.

I think the special form should follow Python as closely as possible. I don't know that I ever use `&key`, but that kind of thing probably belongs in a macro built on top of `fn`, not in the special form itself.

If we re-do `#*` and `#**` to create their own Hy models, as discussed in #1366, then we could also use these in the arguments list. This would make Hy more consistent with Python, and with its own other features that use that syntax.

But that would only take care of `&rest` and `&kwargs`. Using `#*`, we can't distinguish `&rest` from `&kwonly` without commas. And how do we indicate default arguments without an infix `=`? It can't just be a list without `&optional`, or it looks like the destructuring argument bind we still have from Python2. It's also nice to be able to get a `None` default without typing it explicitly.

So how else could this work?
### Option 1
Maybe from this:
```Hy
(defn foo [a &optional ["b" 1] [c :c] d &rest args &kwonly x [y 2] &kwargs kwargs])
```
To this:
```Hy
(defn foo [a  :b 1  :c ':c  :d None  #* args  x  :y 2  #** kwargs])
```
And the equivalent Python.
```Python
def foo(a, b=1, c=HyKeyword("c"), d=None, *args, x, y=2, **kwargs): pass
```
Here we're using :keyword pairs to stand for `=`. This makes it look like a Hy function call, which is more consistent and easier to remember. Note that we have to quote keywords we want to pass as data, so they aren't interpreted as a kwarg name. But function calls work the same way. (And given the strange way our keywords work currently #1352, it would actually expand to `c='\ufdd0:c'`. We might want to change this.)

[...]

PR #1588, opened 2018-04-19
Remove &key

Kodiologist:
It's redundant with `&optional`.


gilch:
Related to #1411.

I have no objection to removing `&key`, since I never use it and it makes the `let` macro more complicated than necessary. If we include this functionality at all, it should be in a macro in terms of `&optional`, not built into the compiler.

PR #1603, opened 2018-05-03
Fix bug in let's handle-dot.

### LINKED ISSUE #1602: bug in let handle-dot for . DSL
```
=> (let [spam (fn[])  a 1] (setv spam.a 42) (print (. spam a)))

  ...
NameError: name 'spam' is not defined
```
The first symbol after the `.` isn't getting expanded. I've got a fix already.

gilch:
Fixes #1602. The first element after the `.` should be expanded normally, even if it's a symbol.

### DIFF
--- a/hy/contrib/walk.hy
+++ b/hy/contrib/walk.hy
@@ -136,12 +136,14 @@ Arguments without a header are under None.
                                     :quote-level (+ self.quote-level x))))
 
   (defn handle-dot [self]
-    `(. ~@(walk (fn [form]
+    `(. ~(self.expand-symbols (first (self.tail)))
+        ~@(walk (fn [form]
                   (if (symbol? form)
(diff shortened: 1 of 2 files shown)


hy/contrib/walk.hy
```diff
@@ -136,12 +136,14 @@ Arguments without a header are under None.
...
-    `(. ~@(walk (fn [form]
+    `(. ~(self.expand-symbols (first (self.tail)))
+        ~@(walk (fn [form]
                   (if (symbol? form)
                       form  ; don't expand attrs
                       (self.expand-symbols form)))
                 identity
-                (self.tail))))
+                (cut (self.tail)
+                     1))))
```
Kodiologist: Any reason for the newline and indent here? You have plenty of room for the `1` on the previous line.

PR #1606, opened 2018-05-10
Fixes #1605: remove macros ap-pipe and ap-compose

### LINKED ISSUE #1605: the arguments order of ap-compose is counter-intuitive
Of all the functional ([programming paradigm](https://en.wikipedia.org/wiki/Function_composition_(computer_science)) and/or [mathematics](https://en.wikipedia.org/wiki/Function_composition)) literatures and conventions (**_unscientific guesswork by myself_**, think textbooks teaching programming/lambda calculus, existing libraries), `compose` works from right to left.

Provided `compose` is the function composition operator, 
`((compose f  g) x)` should means `(f (g x))`

I understand that `ap-compose` is by no means a `compose` function (or a macro, for that matter) in the traditional functional programming sense as it takes forms not functions, but the name suggests otherwise, this is **counter-intuitive** at best and **error-prone** at worst to someone already familiar to functional programming (that is, 90% of hy's user base, okay, I am also guessing here, but you get the point).

Thus, I suggest, either

#### Option 1
Changing the name of the macro into `ap-pipe-fn`, not only the order is in line with `ap-pipe`, but also emphasize there is a function being created, the only true difference between the two.

#### Option 2
Reversing the order of the arguments, and name it `ap-compose-fn`.
In this case, we have `ap-pipe` but no `ap-compose`, and `ap-compose-fn` without the counterpart, `ap-pipe-fn`.  This feels incomplete and the choices are arbitrary, we should also add `ap-compose` and `ap-pipe-fn`.

I can make a PR if needed, albeit it is quite a small change.

pyx:
Anaphoric macros do not work well with point-free style programming, in
which case both threading macros and `comp` are more adequate.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -11,6 +11,9 @@ Removals
 * `&key` is no longer special in lambda lists. Use `&optional` instead.
 * Tuple unpacking is no longer built into special forms for function
   definition (`fn` etc.)
+* Macros `ap-pipe` and `ap-compose` have been removed.
+  Anaphoric macros do not work well with point-free style programming,
+  in which case both threading macros and `comp` are more adequate.
 
 Other Breaking Changes
 ------------------------------
--- a/docs/extra/anaphoric.rst
+++ b/docs/extra/anaphoric.rst
@@ -197,38 +197,6 @@ first element instead. This exposes the element being iterated as
    45
 
 
-.. _ap-pipe:
-
-ap-pipe
-=========
-
-Usage ``(ap-pipe value form1 form2 ...)``
-
-Applies several forms in series to a value from left to right. The special variable ``ìt`` in each form is replaced by the result of the previous form.
-
-.. code-block:: hy
-
-   => (ap-pipe 3 (+ it 1) (/ 5 it))
-   1.25
-   => (ap-pipe [4 5 6 7] (list (rest it)) (len it))
-   3
-
-
-.. _ap-compose:
-
-ap-compose
-=========
-
-Usage ``(ap-compose form1 form2 ...)``
-
-Returns a function which applies several forms in series from left to right. The special variable ``ìt`` in each form is replaced by the result of the previous form.
-
-.. code-block:: hy
-
-   => (setv op (ap-compose (+ it 1) (* it 3)))
-   => (op 2)
-   9
-
 .. _#%
 
 #%
--- a/hy/extra/anaphoric.hy
+++ b/hy/extra/anaphoric.hy
@@ -100,17 +100,6 @@
      acc))
 
 
-(defmacro ap-pipe [var &rest forms]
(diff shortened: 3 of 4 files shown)


pyx:
How can I do that in github?  I know how to squash commits in comman line, but..

Kodiologist:
Just make your changes at the command line and say `git push -f` to replace this PR with your local branch.

pyx:
Thanks, I am actually using mercurial as git client, I will try.

Kodiologist:
I've never used Mercurial before, but the feature you're looking for in Git terminology is a force push. So if you Google "How do I force-push in Mercurial?" or something, that might get your answer.

pyx:
I tried, it is in fact more complicated, since the sha1 hashing scheme is different, I combine the changes and HG's git bridge loses track of hashes.
No worries, I will clone with git and do the squash.

pyx:
Done, my git-fu is rusty...

kirbyfan64:
Note: I used to try to use Mercurial to manage my Git repos too (my very first PR to Hy was done through Mercurial). In the end, though, I just gave up: I felt like I was stuck with all of Mercurial's cons (e.g. slow clone speed, smaller community) with none of Git's advantages.

pyx:
Excuse me if I go off-topic a bit..

@kirbyfan64 I use Mercurial exclusively, that's how I keep my opensource projects in sync. on both github and bitbucket (hg repos.), it's been doing well.

I had to resort to do this commit with git due to the way Hg-Git (the extension) works,  it keeps an internal mapping of hashes between the two, once the ancestral path changed, they are not in sync anymore since both git and mercurial calculate the hash all the way up to root from current commit with different algorithm.

What you observed was true, cloning git repositories with hg is time consuming, for one, it has to recalculate all the hashes at least once. Also IIRC, cloning via http/https protocol (back then, the only way to clone a git repos with mercurial besides ssh/scp) was inefficient in git (even with git itself), something about extra round-trip, it is greatly improved now. And it's true that Mercurial is losing market share big time, at least in OOS circle, a huge factor is github.com, not because of something inferior.

I am a strong believer of "diversity is beneficial". I even wrote a blog post about why I keep using Mercurial, pointing out the advantages[1] against git.

Git could use a better interface, many commands do more than one thing, take `rebase` for example, it does rebasing, merging, history editing, commit squashing, etc. depends on the situation and how you invoke it.

I am not saying Mercurial is better than git, or the other way around, they are just different beneath the UI/terminology similarities[2].

I will try using Mercurial until I cannot. :smile: 

[1]: Mercurial comes with two powerful functional (yes, as in functional programming) query languages (revset and fileset) to search history, phase, changeset evolution, etc.
[2]: as further as even how they record everthing, e.g. git is snapshot based, mercurial is changeset delta based

Issue #1612, opened 2018-05-20
import syntax improvements


vodik:
We've used to have three forms for importing:

```clj
(import foo)          ; import foo
(import [foo])        ; import foo
(import [foo [bar]])  ; from foo import bar
```

We've currently dropped the second form, but that leaves a weird inconsistency:

```clj
(import foo)          ; import foo
(import [foo [bar]])  ; from foo import bar
```

The inner list is no longer necessary. We're now free to change the syntax to be less nested:

```clj
(import foo)          ; import foo
(import [foo bar])    ; from foo import bar
```

And this is close to how Clojure does it (though with expressions instead of lists):

```clj
(import (java.util Date Calendar)
        (java.net URI ServerSocket)
        java.sql.DriverManager)
```

gilch:
If you drop the inner list like that, I think we should switch to `()` like Clojure. Semantically, `()` means the head is special, and it's highlighted as such, while the tail has the arguments, while `[]` means either the elements are equal status or it contains assignment pairs. I think we used nested `[]` before because this assignment hint holds for many of Hy's forms, with `setv` as a notable exception. But what's weird about import is the assignment targets are on the right, which breaks the `[]` pattern anyway.

gilch:
We should also make `require` consistent with `import` if we change it.

Kodiologist:
Fortunately, that's easy now that `import` and `require` are parsed with the same code.

gilch:
Clojure's approach could perhaps be simplified further, like
```
(import
  bacon
  (foo.bar
    (baz quux norlf)
    spam
    eggs))
```
Which would be like
```
import bacon
from foo.bar.baz import quux, norlf
from foo.bar import spam, eggs
```
So an expression sets the prefix for all of its children.

But this potentially breaks the "preferably only one obvious way to do it" rule, since you can now use either a `.` or a new expression.
```
(import
  bacon
  (foo
    (bar
      (baz quux norlf)
      spam
      eggs)))
```
Removing the `.` separator is an option. We could even nest in more `()` to replace the `.` prefix for relative imports. But for common, simple cases, it gets ugly.
```
;; from .foo.bar.baz import quux
(import [.foo.bar.baz [quux]])  ; old way
(import (.foo.bar.baz quux))  ; proposed
(import ((foo (bar (baz quux)))))  ; no dots. How bad?
```
Maybe it's not worth the trouble. It's not to hard to build macros on top of special forms though.

gilch:
[...]

PR #1614, opened 2018-05-21
Simplify import statements

vodik:
Removes an unnecessary level of indentation, and changes them from using lists to expressions. See for details #1612

Nice work with the pattern matching compiler, its really easy to make the change.

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -152,7 +152,7 @@ point per form via the name instead of always the first or last argument.
   'Sir Joseph Cooke Verco'
 
   ;; more convoluted example to load web page and retrieve data from it
-  => (import [urllib.request [urlopen]])
+  => (import urllib.request [urlopen])
   => (as-> (urlopen "http://docs.hylang.org/en/stable/") it
   ...      (.read it)
   ...      (.decode it "utf-8")
@@ -1100,12 +1100,12 @@ that ``import`` can be used.
     ;; Import from a module
     ;;
     ;; Python: from os.path import exists, isdir, isfile
-    (import [os.path [exists isdir isfile]])
+    (import os.path [exists isdir isfile])
 
     ;; Import with an alias
     ;;
     ;; Python: import sys as systest
-    (import [sys :as systest])
+    (import sys :as systest)
 
     ;; You can list as many imports as you like of different types.
     ;;
@@ -1113,16 +1113,16 @@ that ``import`` can be used.
     ;; from tests.resources import kwtest, function_with_a_dash
     ;; from os.path import exists, isdir as is_dir, isfile as is_file
     ;; import sys as systest
-    (import [tests.resources [kwtest function-with-a-dash]]
-            [os.path [exists
-                      isdir :as dir?
-                      isfile :as file?]]
-            [sys :as systest])
+    (import tests.resources [kwtest function-with-a-dash])
+            os.path [exists
+                     isdir :as dir?
+                     isfile :as file?]
+            sys :as systest)
 
     ;; Import all module functions into current namespace
     ;;
     ;; Python: from sys import *
-    (import [sys [*]])
+    (import sys [*])
 
 
 fn
@@ -1349,16 +1349,16 @@ The following are all equivalent ways to call a macro named ``foo`` in the modul
     (require mymodule)
     (mymodule.foo 1)
 
-    (require [mymodule :as M])
+    (require mymodule :as M)
     (M.foo 1)
 
-    (require [mymodule [foo]])
+    (require mymodule [foo])
     (foo 1)
 
-    (require [mymodule [*]])
+    (require mymodule [*])
     (foo 1)
 
-    (require [mymodule [foo :as bar]])
+    (require mymodule [foo :as bar])
     (bar 1)
 
 Macros that call macros
@@ -1381,7 +1381,7 @@ And then, in your main program, you write:
 
 .. code-block:: clj
 
-    (require [mymodule [foo]])
+    (require mymodule [foo])
 
     (print (mymodule.foo 3))
 
@@ -1389,8 +1389,8 @@ Running this raises ``NameError: name 'repexpr' is not defined``, even though
 writing ``(print (foo 3))`` in ``mymodule`` works fine. The trouble is that your
 main program doesn't have the macro ``repexpr`` available, since it wasn't
 imported (and imported under exactly that name, as opposed to a qualified name).
-You could do ``(require [mymodule [*]])`` or ``(require [mymodule [foo
-repexpr]])``, but a less error-prone approach is to change the definition of
+You could do ``(require mymodule [*])`` or ``(require mymodule [foo
+repexpr])``, but a less error-prone approach is to change the definition of
 ``foo`` to require whatever sub-macros it needs:
 
 .. code-block:: clj
@@ -1400,8 +1400,8 @@ repexpr]])``, but a less error-prone approach is to change the definition of
         (require mymodule)
         (mymodule.repexpr ~n (input "Gimme some input: "))))
 
-It's wise to use ``(require mymodule)`` here rather than ``(require [mymodule
-[repexpr]])`` to avoid accidentally shadowing a function named ``repexpr`` in
+It's wise to use ``(require mymodule)`` here rather than ``(require mymodule
+[repexpr])`` to avoid accidentally shadowing a function named ``repexpr`` in
 the main program.
 
 Qualified macro names
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -35,7 +35,12 @@ Other Breaking Changes
   made into its own object type.
 * `HySymbol` no longer inherits from `HyString`.
 * `(except)` is no longer allowed. Use `(except [])` instead.
-* `(import [foo])` is no longer allowed. Use `(import foo)` instead.
+* The `import` and `require` expressions have been reworked. They now takes
+  arguments as expressions instead of lists
+
+  * `(import [foo])` is no longer allowed. Use `(import foo)` instead.
+  * `(import [foo :as bar])` is no longer allowed. Use `(import foo :as bar)`.
+  * `(import [foo [bar]])` is no longer allowed. Use `(import foo [bar])`.
 
 New Features
 ------------------------------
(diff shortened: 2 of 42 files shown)


Kodiologist:
Thanks!

I think we should stick with square brackets. Special forms tend to use parentheses only for inner elements that look like special forms themselves, such as `(except …)` and `(else …)`. Otherwise, they use square brackets for grouping.

Also, are you sure that `(import foo :as bar)` is better than `(import [foo :as bar])`? With the former syntax, related elements among the arguments of `import` are no longer grouped.

vodik:
> I think we should stick with square brackets. 

No strong opinion either way. I'll wait till we have consensus before spending the effort fixing that up.

> Also, are you sure that `(import foo :as bar)` is better than `(import [foo :as bar])`? With `(import foo :as bar)`, related elements among the arguments of import are no longer grouped.

I was a little skeptical of the change at first, but I do think its an improvement. Your concern is fair, but it just took a little but of use to learn how to read it quickly, and most of the issues here can be addressed with style guidelines.

Here's where I see there's an issue with writing `(import foo :as bar)` is that we introduce two different internal inconsistencies with the syntax:

- We write `(import [foo :as bar])`
- And we write `(import [foo bar :as baz])`
- We don't write `(import [foo [bar :as baz]])`

We either should put them in on both cases or in neither case (leaning towards neither).

The other important thing to note is that if we force someone to write `(import [foo :as bar])`, they might not understand what's going on behind the scene and assume that something like `(import [foo :as bar baz])` will work to effectively be `import foo as bar; from foo import baz`, but that won't work. By removing those inter braces, its no longer possible to make that mistake.

As for reading it:

I think its a matter of style. Doing this kind of convention is actually pretty clear imho.

```clj
(import (os.path exists
                 isdir :as dir?
                 isfile :as file?)
        sys :as systest
        re)
```

Kodiologist:
Okay, we can lose the extra brackets for `:as`.

gilch:
> Special forms tend to use parentheses only for inner elements that look like special forms themselves

That's not really true of macros though, even in Hy. To the end-user, the distinction is kind of fuzzy. And Clojure has special forms that break this "rule" (even though `import` is a macro in Clojure), like `letfn`.

Kodiologist:
> That's not really true of macros though, even in Hy. To the end-user, the distinction is kind of fuzzy.

You mean we have core macros that use parentheses for grouping? We should probably fix that. Although, most macros that expect groups won't distinguish between square brackets and parents; you need some code to check for that explicitly. Unless you're using model patterns, but that's not a thing yet.

gilch:
`->`, `->>`, `as->`, and `doto` all use `()` grouping. Like this new `import`, and unlike `try`, the possible expression heads are not drawn from a fixed set of pseudo-"special forms".

gilch:
Argh, OK, think about [how this gets _indented_](http://p.hagelb.org/import-indent.html),
```clj
;; brackets imply that all entries should be indented as peers:
  (:import [java.io File]
           [org.apache.maven.artifact.ant
            Authentication DependenciesTask RemoteRepository]))

;; while parens imply that the first element is special and only the
;; ones after it should be indented as peers:
  (:import (java.io File)
           (org.apache.maven.artifact.ant Authentication DependenciesTask
                                          RemoteRepository)))

Generated by Phil Hagelberg using scpaste at Wed Nov 24 20:58:39 2010. PST. (raw)
```
The head is _special_, not a peer. If we drop the nested brackets, we need to change the remaining outer level to parentheses.

Kodiologist:
We already disagree how to indent Lisp, so I'm not sure what rhetorical strategy you were aiming for there.

But, I will admit that parens do suggest more than square brackets that the head is special and not a peer. I can only reply that the rule that a parenthesized expression is like a function call (it's a special form, or it's a macro call, or it becomes one of these things, or the like), whereas brackets are used for everything else, seems more salient to me.

However! I think I like your original proposal, #851, best. If we're going to break the world to remove unnecessary punctuation from `import` and `require`, we might as go the whole way. The two guys who seemed opposed to that are no longer active on Hy. @vodik, what do you think of it?

gilch:
> We already disagree how to indent Lisp

You refuse to learn the standard and won't let your editor do it for you either, you mean. Maybe you enjoy counting brackets. As a meditative exercise. Or maybe you feel the need to write code no-one else can read. For job security.

> However! I think I like your original proposal, #851, best.

My proposed syntax from #851, which drops the outer brackets instead of the inner ones. The new pattern matching compiler should be able to parse it just as easily.
```clj
(import os.path [isdir :as dir?
                 isfile :as file?
                 exists]
        sys :as systest
        re)
```
This version does need the square brackets to indent properly, not parentheses. It also means that as you type, you don't have to even think about the brackets until you need them. With the status quo syntax, I'm frequently frustrated when typing imports at the repl because I have to back up and insert more brackets. I think it's slightly nicer that @vodik's current proposal in that regard, but the current proposal is closer to Clojure's `import` macro, so there's that.

Kodiologist:
> You refuse to learn the standard and won't let your editor do it for you either, you mean. Maybe you enjoy counting brackets. As a meditative exercise. Or maybe you feel the need to write code no-one else can read. For job security.

I don't accuse you of indenting code in the way you like for some stupid or nefarious purpose, so please don't do that to me. It's evident (from when you threatened a "hostile fork", from when you unilaterally closed one of my PRs an hour after I'd opened it, and from this) that indentation is a sensitive matter for you. I don't know why that is, but it's not appropriate to take it out on me or anybody else.

gilch:
> it's not appropriate to take it out on me

Stop playing the victim, Kodi. Real talk. I can understand your initial mistakes borne of ignorance, but your continued refusal to format your code in a readable way demonstrates that you don't care about wasting our time if it saves you a little. Your hostility toward the rest of the team deserves a response in kind, because you won't care unless someone makes you, by calling you on it. Spinning your personal laziness as a disagreement with me personally was pure spite. Enough of this.

kirbyfan64:
Okay, okay, I think we've all had our fun, but just to be clear: indentation is purely, at the end of the day, nothing other than opinion, and I wouldn't consider it something worth seriously debating outside of academic circles.  At minimum, it's not worth accusations that will likely be forgetting by next week.

kirbyfan64:
We seem to have agreed that #851 is the nicest syntax. Does anyone else (at least, anyone who's actually working on Hy right now) want to object?

paultag:
> Stop playing the victim, Kodi. Real talk. I can understand your initial mistakes borne of ignorance, but your continued refusal to format your code in a readable way demonstrates that you don't care about wasting our time if it saves you a little. Your hostility toward the rest of the team deserves a response in kind, because you won't care unless someone makes you, by calling you on it. Spinning your personal laziness as a disagreement with me personally was pure spite. Enough of this.

@gilch I understand there's some amount of disagreement here, and I'll admit that I'm not super active these days, but shit like this is so completely out of line.

I understand you perceive there to be a lot of hostility, and I understand you know that this behavior is unacceptable and used it to try to make a point, but it's 100% not OK. We're all here to work *together* on something fun. I don't understand why we're at each other's necks.

I don't care about who started what or who disagrees about what, but no one should be writing crap like that about **indentation**. Readability is important, but getting personal is beyond rational.

Let's make sure that this doesn't happen again, I'm incredibly *incredibly* sad to read conversations like this in PRs and discussions.

As a community, we've never had to take any steps beyond reminders like this to fix behavior, so I'm fairly confident that we can cut this out without any extra actions needing to be taken.

That being said, I'm going to start keeping an eye on comments on all PRs for a while, I'm really unhappy to read this.

Kodiologist:
Well said, guys.

> We seem to have agreed that #851 is the nicest syntax. Does anyone else (at least, anyone who's actually working on Hy right now) want to object?

It looks like we're all cool with it. My guess is that Simon would have no objection, but he just hasn't had the chance to return to this PR. Maybe I'll pick it up myself if he doesn't.

vodik:
:+1: on looking nice.

vodik:
Done.

My only concern with the new syntax, after playing with it, is that maybe this is a little visually ambiguous if the `import` causes a particular name to be imported or not:

```clj
(import sys           ; imports sys
        sys [stdin])  ; imports stdin, not sys - the second expression changes the meaning of the first one
```

And while I can live with it (so I don't protest *too* hard but I am officially voting for my original syntax), my previous iteration I think was immune to this problem. And I think its mostly a style issue anyways, as its only really weird when presented like this: `(import sys sys [stdin])` - the double `sys` might throw newbies off in this form.

Kodiologist:
Presumably you would write it in one line as

    (import sys  sys [stdin])

or so, using double spaces to separate pairs.

gilch:
How about dropping the space for the one-line version `(import sys sys[stdin])` as a matter of style? This is similar to the dict lookups with the dot dsl. `(. spam['eggs'])`, which also look a little more like Python without the space. It even looks OK for multiple from imports if you think of it as a tuple lookup in the module dict like,
`(import sys sys[stdin stdout stderr])`, especially if we do #1481.

I already approved the first version that dropped the inner brackets (but only because it used `()` for the outer brackets), but I'd approve my original proposal from #851 too.

Kodiologist:
> How about dropping the space for the one-line version `(import sys sys[stdin])` as a matter of style?

That makes sense, too.

> I'd approve my original proposal from #851 too.

That's what Simon has done now, right?

docs/language/api.rst
```diff
@@ -1100,29 +1100,29 @@ that ``import`` can be used.
...
     ;; import sys as systest
-    (import [tests.resources [kwtest function-with-a-dash]]
-            [os.path [exists
-                      isdir :as dir?
-                      isfile :as file?]]
-            [sys :as systest])
+    (import tests.resources [kwtest function-with-a-dash])
+            (os.path exists
+                     isdir :as dir?
+                     isfile :as file?)
```
Kodiologist: I think you missed some parentheses here.

Kodiologist:
@vodik, could you rebase onto the new master?

vodik:
@gilch like it. Okay, I'll do another pass where we make all the imports have a consistent style then, if we agree on this.

Kodiologist:
Looks like you hit an internal PyPy bug! Nice.

kirbyfan64:
Might be a random issue... Out of curiosity, I'll try restarting it just to see.

Kodiologist:
@vodik, how are we doing here?

Kodiologist:
I'm closing this because of the lack of follow-up. Please reopen if you want to pick this up again.

PR #1616, opened 2018-05-21
Streamline auto-promotion and position spoofing

Kodiologist:


vodik:
So something that's always been bothering me about this stuff is the following snippet doesn't work as expected:

```clj
(eval '(do
   (import pdb)
   (pdb.set-trace)))
```

I mean, it works, but there's no reason is has to be detached from where in the file it was defined. We throw out that information and spoof it to make it function (always says line 1, column 1):

```
> <eval>(1)<module>()->None
(pdb)
```

I'm bringing this up as I think we're getting to a position where the line numbers can properly propagate, so I want to see what other people thing. Getting that working, though, would only be a partial fix - we'd need to be okay with hauling around filename as well (maybe even the module spec, somehow).

Kodiologist:
I've never used the debugger. I don't know much about it.

vodik:
Well, its nothing strictly about the debugger. The problem is the stack trace is useless. But I don't think it has to be the case.

PR #1626, opened 2018-05-30
Fancier `for` and comprehensions

### LINKED ISSUE #588: list-comp only iterates over the last expression when it is passed a do

### LINKED ISSUE #1476: New async forms may need let tests.
The `let` macro (really, the symbol macro machinery it's built upon) needs to correctly account for *every special form* to work properly. When Hy's special forms change, `let` might need an update. #1471 introduced new special forms, which `let` would probably treat as a default case. This may or may not work, but I'm not entirely sure how the new forms are supposed to work.

Kodiologist:
(This PR depends on #1625. I'll rebase this one after that one is merged.)

- Closes #1476.
- Closes #588.

This PR is an update of #1468. (I previously said I'd update #1437 instead, but I changed my mind again.)

`for/a` works like `for` except that the outermost generated `for` is asynchronous. Chances are this isn't exactly what users of `for/a` want, but since none of us use it, I figure we should just wait till somebody complains. I haven't added new tests for it.

To do:
- [ ] ~~flake~~ Better done as its own PR
- [x] Test that variables are scoped properly
- [x] Clean up test organization
- [x] Replace uses of the old forms with the new forms
- [x] Get rid of the old forms
- [x] Documentation
- [x] NEWS


pyx:
`lisp-comp`, `set-comp`, `genexpr`, `dict-comp` will be replaced by `lfor`, `sfor`, `gfor`, `dfor` in the future, is it correct? If so, will they be in next release?

IMHO, `for/l`, `for/s`, `for/g`, `for/d` look better (we do have `for/a` already), `for/list`, `for/set`, `for/generator`, `for/dict`, `for/async` are even better. Just my 2 cents.

Kodiologist:
> `lisp-comp`, `set-comp`, `genexpr`, `dict-comp` will be replaced by `lfor`, `sfor`, `gfor`, `dfor` in the future, is it correct?

If this gets merged, yes.

>  If so, will they be in next release?

If this gets merged, no, `list-comp` etc. will be gone, since `lfor` etc. replace them.

pyx:
Thanks. By "they", I meant the new forms (`lfor`, etc.)

Any idea about when? I am going to release a new version of my library, might as well wait for the new form.

Kodiologist:
Ideally there will be a new release of Hy this month or next month.

pyx:
Fair enough.
I was thinking since we already have `for/a`, `wait/a`, `defmacro/g` ... but to each their own

Kodiologist:
One theoretical advantage of the `lfor` names is that if the Python people add asychronous generator comprehensions, as rumor states they may, the result will have an obvious name (`gfor/a`) instead of being one of `for/a/g`, `for/g/a`, `for/ga`, or `for/ag`.

kirbyfan64:
Ehh, I'm not too sure about these names... I know everyone's going to hate this, but IMO something like `(for->list) (for->gen)` would still work nicely with `/a` but also be more readable.

Kodiologist:
I'd prefer short names, but I'll change the names if there's a consensus for a specific alternative.

gilch:
I'm happy enough with `lfor`, `sfor`, `gfor`, and `dfor` names, but I'd consider alternatives.

I'm reluctant to approve incomplete fixes if that means we don't get to complete them later. But the status quo is so broken that I'm inclined to approve this despite my previous veto, since this seems like a better starting point to fix my previous objections than what we have now. This is not finished. But maybe we should merge it anyway. 

And I'm still concerned that this will break `let`, but perhaps that will be motivation enough to finish what we've started here.

Kodiologist:
Actually, since there are no objections, I'll work on the to-do list now.

Kodiologist:
> I'm still concerned that this will break `let`

I think `let` already didn't work correctly with comprehensions, since they're not mentioned in `contrib/walk.hy`, so at least this change won't make things any worse.

Kodiologist:
Done.

Kodiologist:
I've belatedly realized that `async for` in comprehensions has in fact been implemented in Python, as of Python 3.6. I think the simplest way to implement this is to remove `for/a` and just allow `:async LVALUE ITERABLE` as a comprehension clause, so the asynchronous equivalent of

    (lfor x (range 10) x)

would be

    (lfor :async x (range 10) x)

and likewise, instead of

    (for/a [x (range 10)] (print x))

you'd write

    (for [:async x (range 10)] (print x))

Any objections?

gilch:
Clarify for me how this would work in nested loops. Can we make some of them async and some not, like
```
(lfor :async x (foo/a) :async y (bar/a) [x y])
(lfor x (range 10) :async y (bar/a) [x y])
(lfor :async x (foo/a) y (range 5) [x y])
```
?
We could also consider `:await` as the keyword. I'm not sure which fits the Python better here, since we seem to be labeling the assignment instead of the `for`. I also think the `(await)` keyword needs to work in the comprehension body too now. I'm not sure if the compiler would allow that.

Kodiologist:
Yep, nested loops work exactly like your example. I've made and pushed the changes, so feel free to try them out.

Issue #1635, opened 2018-06-09
Leading `-` should not get converted to underscores.


gilch:
If you define a converter function with a name like
```
(defn ->foo [] ...)
```
Then it becomes module-private and you can't import it with `*`. (Unless you override this with `__all__`.) This is a gotcha that Hy doesn't need.

I propose we revise the mangling rules to distinguish leading `-`s from leading `_`s. Any internal `-` will continue to mangle to `_` as now, but if they're in the lead, it becomes `XhyphenHminusX`s, or preferably something shorter like `XsubX`s if we do the short names #1577. But leading `_` won't change.

We could also distinguish non-leading `_`s from `-`s for full reversibility by converting them to `Xlow_lineX` (or `XscoreX`).

Dunder names would look weird if we had to write them like `__init--`, so trailing `_`s will use the same rule as for the leading `_`s. Then we'd write `__init__` etc.

Examples,
```
=> (mangle "->>")
'hyx_XsubXXgtXgtX'
=> (mangle "->foo")
'hyx_XsubXXgtXfoo'
=> (mangle "foo->bar")
'hyx_foo_XgtXbar'
=> (mangle "_spam-eggs")
'_spam_eggs'
=> (mangle "_spam_eggs")
'_hyx_spamXscoreXeggs'
=> (mangle "__init__")
'__init__'
=> (mangle "--init--")
'hyx_XsubXXsubXinitXsubXXsubX'
=> (mangle "_-_")
'_hyx_XsubX_
=> (mangle "-_-")
'hyx_XsubXXscoreXXsubX'
=> (mangle "__")
'__'
```
In summary, special characters get mangled to X-quoted forms except for leading `_`s, trailing `_`s, and internal `-`, which all convert to `_`.

Kodiologist:
I have to admit that I never considered the possiblity of beginning a name with `->`; when I've seen that digraph in Lisp names before, it's been in the middle of the name.

> I propose we revise the mangling rules to distinguish leading `-`s from leading `_`s. Any internal `-` will continue to mangle to `_` as now, but if they're in the lead, it becomes `XhyphenHminusX`s, or preferably something shorter like `XsubX`s if we do the short names #1577. But leading `_` won't change.

That sounds reasonable.

> We could also distinguish non-leading `_`s from `-`s for full reversibility by converting them to `Xlow_lineX` (or `XscoreX`).

[...]

PR #1636, opened 2018-06-11
Style guide update.

### LINKED ISSUE #1062: Style guide: When to use underscores?

### LINKED ISSUE #1069: Style guide: suggest an extra space where Clojure uses comma.
In an `assoc`/`map`/`let`, etc. usually Clojure uses line breaks to show pairs.

``` Clojure
{1 11
 2 22
 3 33}
```

Hy also adopts this style where applicable. More lines aren't more readable in every situation. Too much whitespace is also noise. But Clojure has an alternative of using commas when it's short enough to fit on one line:

``` Clojure
{1 11, 2 22, 3 33}
```

The pairs are still distinct, even if there are many of them:

``` Clojure
{1 9, 2 8, 3 7, 4 6, 5 5, 6 4, 7 3, 8 2, 9 1}
```

If you wrote this without commas, it could be difficult to tell key from value in the middle.

``` Clojure
{1 9 2 8 3 7 4 6 5 5 6 4 7 3 8 2 9 1}
```

This is rightly considered bad style in Clojure.

Hy can't use commas this way, and I don't really want to change that #198, but Hy could use a space instead of a comma:

``` Hy
{1 9  2 8  3 7  4 6  5 5  6 4  7 3  8 2  9 1}
```

In a fixed-width font, the pairs are still distinct. Programmers nearly always use a fixed-width font.

I suggest adding this format to Hy's style guide, and adjusting Hy's editing tools to use it as well.

gilch:
The style guide is in dire need of an update. The original contained errors, was silent on important points, was unclear about some things, and the language has evolved significantly since then.

Closes #1062 
Closes #1069 

This is a major update. I consider most of this tentative and up for discussion. But the three laws represent decades of Lisp convention (it's how Emacs does it), and are _required_ for [Parinfer](https://shaunlebron.github.io/parinfer/) to work. Parinfer's gentle learning curve compared to full-on Paredit and the way indentation-based blocks come naturally to Pythonistas make it a great fit for Hy's audience. I would _really_ hate to give that up.

I also think that the rule that ending brackets end the line dramatically improve code clarity, but I'm not sure how strict we want to be about it.

The rest is just kind of how I do things at the moment. I'd reconsider if you have good reasons. But I think this is enough of an improvement that we might want to just merge this and continue discussing further changes from there. But please, point out any bugs or typos you spot before.

Some of this also assumes changes that haven't been merged yet, like #1612. And fixing #1635 could also simplify parts.

@ekaschalk this is closely related to hy-mode. Have we left out anything important? Are there suggestions that you'd object to including in hy-mode?

@paultag this has the potential to spark heated disagreements, which is why I've avoided updating it for so long. We might need an arbiter.


Kodiologist:
Ryan, did you mean to attach those comments to particular lines of the file?

Matthew, I don't personally like the style you've advocated here, but I didn't like the previous version, either, so I see no reason to object to the changes. My only serious objection to merging this is that #1612 isn't in yet. Ultimately, my preference would be not to include any style guide in the reference manual, since style is idiosyncratic and liable to change over time. Personal blog posts, like Paul's original post, are an example of a better medium for this subject.

While you're editing the style guide, though, here are some more suggestions:

- Remove the epigraphs. I'm not against epigraphs or humor, but I don't think these quotes are good choices because they have no apparent relation to the content except for containing the word "style".
- If we're going to include that koan in the manual, let's explain what it's supposed to mean. I for one have no idea what it means.
- Describing the design goals of Hy (here listed under "The Tao of Hy") is good, but a style guide isn't the right place for it. Let's move this material to elsewhere in the manual.

ekaschalk:
Few comments.

On outlines:

1. Outlines in-general are not `;;;`-styled. Only emacs-lisp has that and
   would've been changed if not for historical compatibility (see
   `outline-regexp`). Outlines in Hy are `;; *`-styled. This is the same for
   clojure-mode and common-lisp-mode.
2. I don't think it should be mentioned or recommended explicitly as it is Emacs
   only. 

On `#_` usage:

1. This recently has syntax highlighting support. But no editing support, I see
   (un)-capturing the current/next/prev/containing-defun forms in `#_` as
   possibilities.

On indentation:

1. If indentation is being finalized, then which symbols yield special
   indentation should be finalized. I'm not sure if this list should be
   referenced in the style guide, but at a minimum it should correspond to
   Emacs's defaults (eg. `setv` isn't specially indented).

On `-` vs `_`:

1. I really don't like `__` prefixed dunders/private methods. One of my favorite
   parts of lisp not having to reach for the shift key as often. 

2. Less joking, why is the line drawn at dunders? What about imports, they
   reference specific files that are underscored. I'd prefer us to go all-in on hyphens.

3. I don't find the negative argument convincing. `(+ -foo 1)` doesn't add the
   `negated` foo, and this is imo obvious.
   
4. Perhaps fix a recommendation for unused arguments `(setv [x -] x-y)` as I
   could see `_` being a reasonable choice.

On everything else:

1. I think a very short section on mangling might be appropriate. It is relevant
   when designing hy code that is to be imported and ran within python.
2. DTRT acronym I had to look up, wasn't clear to me.
3. Not personally a fan of leading with the koan.
4. Typo, your coding style section has `(setv square? [x] ...`).
5. I don't agree with 100 char limit. I'd rather defer to PEP8 at 79 lines.

Kodiologist:
> Perhaps fix a recommendation for unused arguments `(setv [x -] x-y)` as I
could see `_` being a reasonable choice.

Careful, assigning to `-` or `_` will shadow the shadow `-` operator. So while it's handy in Python, it's probably unwise in Hy.

ekaschalk:
Ah that's a silly mistake, thanks for pointing that out. I suppose I will switch to `--` instead.

kirbyfan64:
> Ryan, did you mean to attach those comments to particular lines of the file?

WTH my mobile GitHub client is screwing with me. :neutral_face: I'll delete them and re-add them to the proper lines...

docs/style-guide.rst
```diff
@@ -51,90 +50,558 @@ into the making of Hy.
...
+
+    ;; Acceptable if the list is long. (Three isn't that long though.)
+    ;; This is better for version control line diffs.
+    [
+     (foo)
+     (bar)
+     (baz)
+     #_ /]
+
+    ;; Unacceptable and an syntax error. Lost a bracket.
```
kirbyfan64: and **a** syntax error

docs/style-guide.rst
```diff
@@ -51,90 +50,558 @@ into the making of Hy.
...
+    [(foo)
+     ;; (bar)
+     ;; (baz)]
+
+    ;; Unacceptable. Broke law #1.
```
kirbyfan64: Maybe be explicit that these few examples are for commenting out list items? Took me a few moments to realize it.

docs/style-guide.rst
```diff
@@ -51,90 +50,558 @@ into the making of Hy.
...
+
+    ;; Be consistent. Separate all groups the same way in a form.
+
+    {1 9  2 8
+     3 7  4 6  5 5}  ; Pick one or the other!
+    {1 9  2 8 3 7  4 6  5 5}  ; You forgot something.
+
+    ;; Groups of one must also be consistent.
+
+    (foo 1 2 3}  ; No need for extra spaces here.
```
kirbyfan64: ( and }?

docs/style-guide.rst
```diff
@@ -51,90 +50,558 @@ into the making of Hy.
...
+                 #* args  ; #* goes with what it unpacks.
+                 #** kwargs)
+
+    ;; Yep, those are pairs too.
+    (setv x 1
+          y 2)
+
++ Macros and special forms can have "special" arguments that are indented like function arguments.
+
++ Indent the non-special arguments (usually the body) one space past the parent bracket.
```
kirbyfan64: These two bullet points should probably be together to make it easier to see the connection.

docs/style-guide.rst
```diff
@@ -51,90 +50,558 @@ into the making of Hy.
...
+    ;; Headers should only be one line.
+    ;; This is non-header commentary, but not about a particular form.
+    ;; These can span multiple lines.
+    ;; These are separated from the next form or form comment by a blank line.
 
-   ;; Style-compliant but just states the obvious
-   (setv ind (dec x))  ; sets index to x-1
+    ;; Good.
+    (setv ind (dec x))  ; indexing starts from 0
+                                            ; margin comment continues on the next line.
```
kirbyfan64: Was the extra indentation accidental or intentional? I don't see anything here really justifying that.

docs/style-guide.rst
```diff
@@ -51,90 +50,558 @@ into the making of Hy.
...
 
+Prefer docstrings to comments where applicable--in ``defn``, ``defclass``, and at the top of the module.
 
-+ Inline comments shall be two spaces from the end of the code; they
-  must always have a space between the comment character and the start
-  of the comment. Also, try to not comment the obvious.
+The ``(comment)`` macro is still subject to the three laws.
+If you're tempted to violate them, consider discarding a string instead with ``#_``.
+
+Semicolon comments shall start with some number of semicolons
```
kirbyfan64: Maybe mention that semicolon count guidance is below?

docs/style-guide.rst
gilch: Wow, I think I need a better font. This was barely visible in my Editor. Switching to Source Code Pro.

Kodiologist: I'm a fan of DejaVu Sans Mono. Before that, I stuck to Andale Mono.

docs/style-guide.rst
gilch: So like
```
+ Macros and special forms can have "special" arguments that are indented like function arguments.
  Indent the non-special arguments (usually the body) one space past the parent bracket.
```
?

docs/style-guide.rst
gilch: Yeah we should probably discuss that a bit more. The Clojure Style Guide (linked in the thanks section at the bottom, even in the current version) uses the term "[margin comments](https://github.com/bbatsov/clojure-style-guide#one-semicolon-for-margin-comments)".

@ekaschalk Emacs likes to indent Hy's `;` comments to column 32, but it's column 40 in clojure-mode, and lisp-mode. I'm pretty sure I've seen column-aligned comments like this in Common Lisp code. I think they call them "column comments". How would you want to handle these?

Should "margin comments" be a separate idea from the "column comments"? Should we forbid continuing margin comments on an otherwise blank line in the style guide? Should we get rid of the 2-space margin comments and only recommend column comments? Something else?

docs/style-guide.rst
gilch: Too long before the example? @ekaschalk was suggesting we ditch the headers altogether, that would make this section shorter. Then we'd only have the `;` and `;;` styles.

gilch:
> Can we ditch the shalls? should sounds less intense IMO.

The current version uses "shall", so I rolled with it. "Should" works for me though.

> Kinda related to the above, one nice thing about Effective Dart is the use of all-caps signal words like DO and PREFER in a consistent fashion to make it easier to break down the guide, and it'd be especially nice in a style guide this big.

PEP-8 is using "Yes:" and "No:". I can try to make those more consistent and in all caps. Maybe "PREFERRED", "OK", and "BAD"?

gilch:
> On outlines

OK, so remove the `;;;` and `;;;;` styles from the guide altogether then?

> On #_ usage

I'm not sure I understand what you mean. I noticed clojure-mode highlights discarded forms like comments. I'm not sure if we want the editor to have a command to insert #_ for a selection. Does clojure-mode do that? Usually you won't need to insert that many. The editor can do `;;` comments already. We just have to be more careful when commenting out closing brackets. Lisp more expression-oriented than line-oriented.

> On indentation

If you want to be that complete, we should probably go over all the special forms one-by-one. And the core macros too. Maybe even extra and contrib macros. 

I hope I got the "special indent" terminology right here. So in a `defn`, the name and lambda list are special but the body is not, right?

In any macro or special form with a body (written `&rest body` in the macros), the body is not special. If the body is last (always is, I think), then the other args are special. I'm not sure about docstrings. Maybe those should be special too? But the extra indent is not really a good thing inside the string itself, so maybe no.

Your example was `setv`. In the proposed style guide so far, acceptable styles would be
```
;; extra space
(setv x 1  y 2)
;; standard function indent
(setv x 1
       y 2)
;; alternative function indent--1 space past parent bracket.
(setv
  x 1
  y 2)
```
And the second is preferred over the third, just like for functions, which never have special arguments.

In the case of `if`, I think I've seen the else clause treated as special in emacs lisp, but not in Clojure. I'm not sure which that should be.

> `(setv square? [x]`

Hah. That was in the original and I didn't notice.

ekaschalk:
> OK, so remove the ;;; and ;;;; styles from the guide altogether then?

I think so, the people gaining utility from that, wouldn't need to be told about it.

> #_ 

Our existing support is very similar to Clojure's. I was considering a binding that would, say, insert (or remove) a `#_` before the containing form at point. Simple QOL stuff, not sure if Clojure has anything like it, but I don't think that matters.

> I hope I got the "special indent" terminology right here.

Anything other than a normal indent is special. A normal indent is right now: no args on same line of form opener -> 1+ indent used, args on same line of opener -> indent at first arg.

Anything at the same depth as the `defn` is indented specially (1+ regardless if whether args on same line as `defn` or not). Indentation is entirely determined by the symbol opening the form (atm).

> If you want to be that complete, we should probably go over all the special forms one-by-one. And the core macros too. Maybe even extra and contrib macros.

I think this is a useful discussion to have. The sooner we finalize indentation, the better.

gilch:
> DTRT acronym I had to look up, wasn't clear to me.

So did I. I'll spell it out then.

> I don't agree with 100 char limit. I'd rather defer to PEP8 at 79 lines

We're still using punchcards. Le sigh. All right. We're based on Python, so let's follow all of PEP-8's rules on that point. (This will include the even shorter limits for docstrings, which could leak to Python anyway. I should have had that in there already.)

ekaschalk:
> We're still using punchcards. 

I choose my zoom level on my laptop/desktop to fit 2/4 windows open respectively side by side assuming 80 characters. I'm sure I'm not the only one who bases their layouts with that assumption.

docs/style-guide.rst
gilch: Was using IntelliJ with Python plugins (PyCharm, basically) The one it had before was just called `Monospaced`. I wonder if that's the default. The curlies looked like parens with a little dot.

Source Code Pro is the default for Spacemacs, so I'm used to it.

gilch:
I don't get the koan macro either. The epigraphs are not my fault. The current version has them, and this PR simply didn't change that. I think removing those or moving the design goals should be done in a separate PR. This one is big enough already.

gilch:
@kirbyfan64, I tried adding more consistent landmarks, including more headers and all-caps signal words to hopefully make it easier to break down. Let me know if it goes far enough.

I reworded to avoid "shall"s, and I also rewrote a lot of the commentary, but mostly kept the code examples. I also added more rules. So you might want to re-read everything.

I would still consider most of this tentative even if merged, but the community might not have even thought of some of these issues, so I think we can have a more productive discussion with a framework to start from. So I added "As in Python, we expect the style of Hy to be refined over time in practice." to the conclusion.

Some of the subsections could be reordered for clarity, but that's best saved for another commit, or it would have make the diff completely unrecognizable.

Kodiologist:
@kirbyfan64, I think the ball's in your court to assess whether you're happy with the changes.

refi64:
Ugh so super sorry, I literally never realized that this was still hanging here. @gilch are you still interested in this? If so, I'll give this a pass again for approval.

gilch:
@refi64 sure, give it another pass.

docs/style-guide.rst
```diff
@@ -51,126 +50,952 @@ into the making of Hy.
...
+    {"foo" "spam"  "bar" "eggs"  "baz" "bacon"}
+
+    ;; OK. Yep, those are pairs too.
+    (setv x 1
+          y 2)
+
+    ;; PREFERRED. This fits on one line.
+    (setv x 1  y 2)
+
+    ;; BAD. Doesn't separate groups.
```
refi64: nit: there's no preferred example for the if, only "ok" so it's not clear if there's actually a preference for something else. I'm guessing preferred would be (based on the previous examples):

```hy
(print (if (< n 0.0) "negative"
           (= n 0.0) "zero"
           ...))
```

gilch: It's been a while since I wrote this, so I don't remember exactly why there's no example for that.

Unfortunately, other Lisps don't completely agree on how to indent `if`s (The three laws always apply, of course.), so I'm not sure what the preference should be. I suppose I can pick one and be more explicit about it, as I said last year, 
> I would still consider most of this tentative even if merged, but the community might not have even thought of some of these issues, so I think we can have a more productive discussion with a framework to start from.

Emacs style dedents the else clause slightly.
```
;; OK?
(if (predicate? x)
  (bar x)
 (baz x))
```
Clojure style does not.
```
;; PREFERRED?
(if (predicate? x)
  (bar x)
  (baz x))
```
If there are any elif parts, then Hy's `if` is most similar to Clojure's `cond` or Arc's `if`.
```
;; PREFERRED?
(if (< n 0.0) "negative"
    (= n 0.0) "zero"
    (> n 0.0) "positive"
    "not a number")
```
Maybe this looks weird if there's only a then and else part:
```
;; OK?
(if (predicate? x) (bar x)
  (baz x))
```
Clojure's style guide [recommends](https://github.com/bbatsov/clojure-style-guide#short-forms-in-cond) that `cond` pairs be simple enough to fit on one line, or where not possible, separated with comments or blank lines. One could always factor out functions, I suppose, but do blocks may also be OK.

gilch:
@refi64 I fixed a small merge conflict. I'm inclined to merge this as-is and fix the nits afterwards. Doc fixes only require one approval if I recall.

PR #1646, opened 2018-06-27
Adjust Python version check for `cache_from_source`.

brandonwillard:
Python 3.3 will fail when Hy tries to use `importlib.util.cache_from_source`,
because it was actually introduced in Python 3.4.

https://docs.python.org/3/library/importlib.html#importlib.util.cache_from_source

### DIFF
--- a/hy/_compat.py
+++ b/hy/_compat.py
@@ -21,6 +21,7 @@ except ImportError:
 import sys, keyword
 
 PY3 = sys.version_info[0] >= 3
+PY34 = sys.version_info >= (3, 4)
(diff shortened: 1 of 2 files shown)


Kodiologist:
Python 3.3 is no longer supported since Hy 0.14.0. Its end-of-life for the Python development team was on 2017-09-29.

vodik:
I have no problem merging this, but just a heads up, Python 3.4 *is* the minimum version.

brandonwillard:
Ah, makes sense; I'll delete it.

PR #1647, opened 2018-06-28
Fix let rebind bug.

### LINKED ISSUE #1645: let macro (re)binding
Per @gilch's [comment](https://github.com/algernon/monaxhyd/pull/2#issuecomment-400720214) in https://github.com/algernon/monaxhyd/pull/2, there seems to be some unexpected behavior in the way `let` (re)binds its arguments:
```clojure
=> (require [hy.contrib.walk [let]])
=> (let [a 2 a 1] a)
_hyx_XsemicolonXletXvertical_lineX1270 = {}
_hyx_XsemicolonXletXvertical_lineX1270['a'] = 2
_hyx_XsemicolonXletXvertical_lineX1270['a']

2
=> None
```
The expected output is `1`.

A work-around is given in [the aforementioned comment](https://github.com/algernon/monaxhyd/pull/2#issuecomment-400720214) using nested `let`s.

gilch:
fixes #1645 

Also adds a test for the issue, and cleans up the two files.--proper special indent for let tests, and removes some outdated comments in the walk module.

### DIFF
--- a/hy/contrib/walk.hy
+++ b/hy/contrib/walk.hy
@@ -305,6 +305,7 @@ as can nested let forms.
       (macro-error bindings "let bindings must be paired"))
   (setv g!let (gensym 'let)
         replacements (OrderedDict)
+        keys []
         values [])
   (defn expander [symbol]
     (.get replacements symbol symbol))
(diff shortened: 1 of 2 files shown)


Kodiologist:
Could you spin out the indentation changes into their own commit? That would make it easier to see the substantive changes.

gilch:
The only new test is at the bottom `test-let-rebind` everything else in that file should just be whitespace.

I also removed the big comment at the bottom of walk. The rest is more substantive.

gilch:
Rebased so the final commit shows substantive changes.

gilch:
Oh it was already approved.

>  does let and its subroutines use destructive modification of models

I do not recall in enough detail. It was based on the earlier stuff in the `walk` module, which was pretty functional-looking. I think it mostly rebuilt everything as it walked over it, which should work fine with immutable models. Unless some of that building mutated the new things. It shouldn't be too hard to adapt anyway.

kirbyfan64:
@gilch Ready for merge...based on your comment, it seems like there's nothing else to really do? (Just checking first.)

gilch:
Yeah, it's just a bugfix, so I don't think we need any change in the docs. I'm not sure if it's worth adding to NEWS, but that tends to create merge confilcts.

gilch:
I'm just going to merge it. NEWS can go in #1652, if necessary.

PR #1581, opened 2018-04-09
Make keywords callable to simplify the compiler

vodik:
Quick PR to get #846 done. Just need to go over documentation.

PR #1656, opened 2018-07-08
Make HyKeyword callable

### LINKED ISSUE #846: :keywords as objects

Kodiologist:
Replaces #1581.

I believe this closes #846.


gilch:
No particular objection, since this is a superior implementation of functionality we already had, but what about making symbols callable too? That also works in Clojure. Although in Python, it might make even more sense for strings to be callable. We can't really alter the builtin `str`, but we could add that to quoted HyStrings.

Kodiologist:
I could see that, but it would probably be better in a new PR (or just a new issue, for now).

PR #1664, opened 2018-07-25
Expand `require`d macros in `macroexpand-all`

brandonwillard:
Minor change that allows `macroexpand-all` to expand `require`d macros.

See [this discussion for some context](https://github.com/algernon/adderall/pull/9#issuecomment-407560245).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -89,3 +89,4 @@
 * Simon Gomizelj <simon@vodik.xyz>
 * Yigong Wang <wang@yigo.ng>
 * Oskar Kvist <oskar.kvist@gmail.com>
+* Brandon T. Willard <brandonwillard@gmail.com>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -86,6 +86,7 @@ Misc. Improvements
 ----------------------------
 * `hy-repr` uses registered functions instead of methods.
 * `hy-repr` supports more standard types.
+* `macroexpand-all` will now expand macros introduced by a `require` in the body of a macro.
 
 0.14.0
 ==============================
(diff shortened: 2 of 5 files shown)


Kodiologist:
1. Spin out the unrelated style edits to `walk.hy` into their own commit.
2. Put the test macro into one of the existing files in `tests/resources` instead of its own file.
3. I think the test is too brittle because it depends on `require` expanding to `None`. Instead, try something like `(assert (= (last (macroexpand-all '(require-macro))) '(setv blah 1)))`.
4. Update NEWS and AUTHORS.

Otherwise, this looks good.

brandonwillard:
Ha, yeah, I forgot about those docstring edits; I'll just retract them.

brandonwillard:
Also, I wasn't sure whether or not it was desirable to have a `require` expand to `None`.  It doesn't seem to _hurt_ anything, but does produces a rather redundant expression (e.g. an unnecessary `(do None ...)`).  

We could easily filter-out those `None`s, but that would introduce topics like expression simplification/normalization, canonical forms, etc., and none of that was going on in the module already.

Kodiologist:
Right. Don't attempt to remove it for now.

brandonwillard:
Also, before I forget, that `require` in `walk` isn't side effect-free, in that it adds the `require`d macros to the calling module, which seems less than ideal.

I considered creating a temporary shallow copy of the calling module (or at least the `hy.macros._hy_macros` dicts) for a side-effect free version, but didn't because it appeared to relate to @gilch's comments on namespace enhancements and possibly other fundamental changes that would facilitate such an implementation.

brandonwillard:
Thanks; I look forward to helping everyone get Hy!

Issue #1667, opened 2018-07-31
Weird behavior of map function


falematte:
Why I do not get the values? 
![selection_999 406](https://user-images.githubusercontent.com/25055204/43466535-4430b0dc-94cf-11e8-8ed2-07325d4e1076.jpg)

gilch:
Because you're running on Python 3.

This is Hy's issue list (for bugs and complaints), not a help forum. For help with Hy, ask on our IRC channel, stackoverflow with the [hy] tag or on our mailing list.

Issue #1666, opened 2018-07-31
comp with list of function?


xuqinghan:
list_fn = [f g]
......
    ((comp list_fn) x))
TypeError: 'list' object is not callable

how to convert (comp list_fn) to (comp f g) ?

-----------------------------------------
and if use macro

```
list_expr = [HyExpression([
  HySymbol('/'),
  HyInteger(2)]), HyExpression([
  HySymbol('+'),
  HyInteger(1)])]
```

    (-> x list_expr ))
TypeError: 'list' object is not callable

what is the right way to use ((comp list_fn) x)) and (-> x list_expr ))  ?

gilch:
Did you try `#*`?

This is Hy's issue list (for bugs and complaints), not a help forum. For help with Hy, ask on our IRC channel (may take up to a day), stackoverflow with the [hy] tag or on our mailing list.

Issue #1595, opened 2018-04-25
How can I pass named optional arguments to hy functions?


DictumMortuum:
Sorry if this has been answered before, I don't really know the terms to search it.

I want to do a basic GET request with username and password. So I found [this](http://docs.python-requests.org/en/latest/user/authentication/#basic-authentication) page. 

```
>>> from requests.auth import HTTPBasicAuth
>>> requests.get('https://api.github.com/user', auth=HTTPBasicAuth('user', 'pass'))
```

I suppose that I have to do something like:

```
#! /usr/bin/env hy
(import [requests [get]])
(import [requests.auth [HTTPBasicAuth]])
(get "https://url" {:auth (HTTPBasicAuth "user" "pass")})
```
but it doesn't work. How do I pass these "auth=HTTPBasicAuth('user', 'pass')" arguments to python functions?

Kodiologist:
The GitHub issues list is for bug reports and feature requests. For support, try our mailing list, IRC channel, or Stack Overflow.

Issue #1357, opened 2017-08-03
Improving IRC experience


gilch:
Two ideas here.

First, our README points to the #hy channel on Freenode. I see new users on there asking for help pretty frequently. But I've occasionally been frustrated when someone asks an easy question and then logs off before I see it. I can't realistically answer everything instantly, even if I enable alerts. Sometimes I'm busy with other things. Sometimes they're in a distant timezone and I'm asleep.

We could mention in the README that if nobody's talking on the channel that you should wait in the channel about a day for a response.

Better yet, we could set up a server message in the channel itself explaining this. But I don't know who the admin for the channel is.

Second, it would be nice if we could evaluate code snippets to demonstrate things to the new users. This could be done with a bot. I noticed we have one of those https://github.com/hylang/hygdrop, but nobody's touched it in years, and we don't seem to have any active bots watching the channel. Just `omghy` that occasionally pops in and out.

Once we demonstrate this, users could private message the bot and try things for themselves.

It would be nice if the bot could optionally use both the latest master or the latest stable, so we could demonstrate appropriate code depending on which version the user is asking about. I don't know if github has a way to push updates like that, but it might have to be set up here in this repo. (Otherwise, the bot could check for updates once a day or something.)

We'd naturally want to sandbox it and set output length limits to prevent abuse, and timeouts to deal with infinite loops. And there should be a command to reset it.

I propose naming the new bot `cuddles`. 🦑

kirbyfan64:
TBH I think we should be using Gitter or Discord anyway, but ehh. ;)

deejoe:
@cwebber @paultag and @olasd are the admins for #hy, as best I can tell from the brief conversation I had with Freenode's ChanServ just now, e.g.:

```
/msg chanserv access #hy list
```

deejoe:
As for using whatever IM-du-jour beyond IRC, I think setting up a bridge between any given service and the IRC channel is the most inclusive approach, whose maintenance might be made scalable by advocates of that service taking on the responsibility for making whatever arrangements need to be made.

My preferred protocol in this regard is Matrix, and conveniently enough for me as its advocate, Freenode and Matrix.org already have arranged the bridge, which ```/names``` shows me several people are already using (the default nick for an IRC user coming across a Matrix bridge has an [m] postfix, making them easy to spot. This can be customized but rarely is.)

Anyway, the Matrix side of the bridge is accessible through the web manifestation of the riot.im client through this link:

[...]

PR #1668, opened 2018-08-01
Point out appropriate help forums

gilch:
I occasionally see issues like #1667, #1666, #1595, etc. that probably shouldn't be on our issue list.

I considered adding a RTFM--sometimes these questions are about simple things that are clearly spelled out in the manual, but it can be hard to search it for special characters. (It might be easier to clone and grep the repo just to find the right docs sometimes.)

Does anyone still check the mailing list? I pretty much never use it. I am watching the `[hy]` tag on Stack Overflow, and would usually get around to checking our IRC channel via Matrix within about a day.

Also helps a little with #1357.

If anyone feels they can word this more politely without losing clarity for someone in a hurry, feel free to push changes.

I'd call this a doc fix, so one approval will suffice.

### DIFF
--- a/issue_template.md
+++ b/issue_template.md
@@ -1,1 +1,7 @@
-If you're reporting a bug, make sure you can reproduce it with the very latest, bleeding-edge version of Hy from the `master` branch on GitHub. Bugs in stable versions of Hy are fixed on `master` before the fix makes it into a new stable release. You can delete this text after reading it.
+Hy's issue list is for bugs, complaints, and feature requests.
+
+For help with Hy, ask on our IRC channel (may take up to a day), Stack Overflow with the `[hy]` tag, or on our mailing list.
+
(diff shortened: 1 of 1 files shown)


kirbyfan64:
FWIW I actively monitor the mailing list.

As a side note, personally, I'd like to see some medium of message-based communication instead of IRC, since the lack of message persistence kind of hurts it a bit and bouncers are a PITA to set up. Maybe try something like [Zulip](https://zulipchat.com) or [Matrix](https://matrix.org/)? (The former is more like a mix of Slack and email, and the latter is more like IRC.)

There are other services too, of course, but @paultag would probably prefer something OSS...

gilch:
@kirbyfan64 I'm already accessing our freenode IRC via [Riot.im](https://riot.im)/Matrix so I don't have to keep a client connected all the time. Maybe we should point them to that somehow. I don't know who has admin rights to `#hy` or how to set that up.

kirbyfan64:
Matrix's IRC bridge is known to be a bit finnicky, though. Apparently they're working on an improved version, but it's not ready yet. 

There are also some niceties that native Matrix has that can't be matched by IRC, like Markdown support (for e.g. embedded code blocks). Zulip also has a neat ability to create drafts, kind of like email.

PR #1669, opened 2018-08-02
Some simple speed-ups

Kodiologist:
On my machine, with Python 3.7, running a byte-compiled program `(print  "hi")` with master takes 0.190 s, compared to 0.073 s with these changes.


Kodiologist:
If we ever want to implement reader macros, we'll probably have to ditch rply, but we'd probably also end up with a slower parser because we'd need to abstract things more.

PR #1671, opened 2018-08-14
New importer

brandonwillard:
This is an update (rebase to current master) of https://github.com/hylang/hy/pull/1518, along with implemented `hyc` byte-compilation (copy of `py_compile`), removed Python 2.x compatibility (Hy no longer supports it), and `importer` tests converted to the new `importlib`.  

(I couldn't find a good way to contribute changes to @vodik's existing PR; that's why this exists)

Also, a lot of utility code was removed in https://github.com/hylang/hy/commit/5960c9083a078a753548a37ac049e0315dda2aca, because it came directly from [`importlib._bootstrap`](https://github.com/python/cpython/blob/3.4/Lib/importlib/_bootstrap.py) and/or `importlib._bootstrap_external` and, since we were already importing from `importlib._bootstrap`, it seemed just as reasonable to use the necessary functions from their original source; however, these modules are not public-facing.  At the very least, we can have everything in working order by simply accounting for discrepancies (e.g. `importlib._bootstrap_external` appearing after 3.5) across versions.


Kodiologist:
I don't think we're ready to drop Python 2 support, sorry. The Python folks are still supporting it till 1 Jan 2020.

Kodiologist:
I should add that even if you want to drop Python 2, it's a big enough change that it should get its own PR.

brandonwillard:
Agh, yeah; well, I'm in the process of reducing the import machinery to the bare essentials in 3.x, so, once I've figured out exactly what that should look like, I'll re-add 2.7 support.

brandonwillard:
In the search for minimal integration with Python >= 3.4 `importlib`, I found this small monkey-patch sufficient for at least the basics (e.g. regular `import`s of `.hy` files) and even direct use of standard `py_compile`:
```python
import importlib

from hy.compiler import hy_compile
from hy.lex import tokenize
from hy.models import HyObject, HyExpression, HySymbol

importlib.machinery.SOURCE_SUFFIXES += ['.hy']

# Keep a copy of the original method.
_source_to_code = importlib.machinery.SourceFileLoader.source_to_code

def hy_parse(source):
    source = re.sub(r'\A#!.*', '', source)
    return HyExpression([HySymbol("do")] + tokenize(source + "\n"))

# New, patched method that considers .hy files.
def _hy_source_to_code(self, data, path, *, _optimize=-1):
    if os.path.isfile(path) and if path.rpartition('.')[-1] == 'hy':
            data = hy_compile(hy_parse(data.decode("utf-8")), path)

    return _source_to_code(self, data, path, _optimize=_optimize)

# Patch the built-in method.
importlib.machinery.SourceFileLoader.source_to_code = _hy_source_to_code
```

If this approach works across the board, we could remove the custom copy of `py_compile` and almost all existing `importlib` code in this branch.  Patches like this aren't usually ideal, but at least it's patching [a method specified in the `importlib` API/spec](https://docs.python.org/3.4/library/importlib.html#importlib.abc.InspectLoader.source_to_code) without compromising it's spec (as well as adding to a likewise exposed [`SOURCE_SUFFIXES` variable](https://docs.python.org/3.4/library/importlib.html#importlib.machinery.SOURCE_SUFFIXES)).

Since the current approach is a work-around for `FileFinder`'s inflexibilities (see [here](https://stackoverflow.com/questions/41990169/how-to-use-sys-path-hooks-for-customized-loading-of-modules)) in a way that effectively supersedes the standard import mechanisms, it's probably worth considering a patch that immediately provides the same functionality and more.

Kodiologist:
Okay, that sounds like a decent thing to try.

vodik:
Hey, really happy to see this work continued!

Sorry I haven't had the time or energy to keep up my open source work. I'll try and give you a good code review tomorrow.

On August 20, 2018 12:59:47 AM EDT, "Brandon T. Willard" <notifications@github.com> wrote:
>Closed #1671.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/hylang/hy/pull/1671#event-1796760320

-- 

Issue #1540, opened 2018-03-24
Module docstrings don't work with together with Hy core imports


Kodiologist:
Similarly to #1367, any automatic imports from core will prevent a string literal from being the first statement in a module, and hence from being the module docstring.

    $ echo '"hello world" (print (get [1 2] 0))' > test.hy
    $ hy -c '(import test) (print test.__doc__)'
    1
    hello world
    $ echo '"hello world" (print (first [1 2]))' > test.hy 
    $ hy -c '(import test) (print test.__doc__)'          
    1
    None

gilch:
We can just set the `__doc__` global and put module docstrings wherever we please. But I'd rather get rid of autoimports altogether somehow. It's causing a lot of problems and the above is just one of them. Hy's core is not discoverable with `dir`. We can't easily `import *` from a Hy module, since the Hy core we happen to use in that module comes with it.

What we really want is our own `builtins` module for Hy. But Python doesn't have an easy way to add that. And dumping our "builtin" stuff into Python's `builtins` is not nice to the Python side. So we autoimport. #1407 had some ideas on how to deal with that differently.

vodik:
> What we really want is our own builtins module for Hy. But Python doesn't have an easy way to add that.

It does... we just have to drop Python 2 :wink: See [BuiltinImporter](https://docs.python.org/3/library/importlib.html#importlib.machinery.BuiltinImporter)

gilch:
Wait, what? Does this really do what we need? Python 2.7 also [has a "BuiltinImporter" class](https://docs.python.org/2/library/imputil.html#imputil.BuiltinImporter), I checked. But it's not clear to me how to use it or what it's good for from those linked docs.

Issue #1367, opened 2017-08-05
`from __future__ import` doesn't work together with Hy core function imports


woodrush:
While this works,
```clojure
(import [__future__ [print_function]])
(setv a 10)
```
This doesn't work:
```clojure
(import [__future__ [print_function]])
(setv name 10)
```

The former compiles to the Python code (using hy2py)
```python
from __future__ import print_function
a = 10
```

While the latter compiles to
```python
from hy.core.language import name
from __future__ import print_function
name = 10
```
yielding the error
```python
# ==>
#   File "importbug.hy", line 1
#     (import [__future__ [print_function]])
#     ^
# SyntaxError: from __future__ imports must occur at the beginning of the file
```

So this is because `name` is used in the Hy core, and Hy adds  `from hy.core.language import name` at the beginning of the file, preceding `from __future__ import print_function`.

It also happens with other `__future__` module imports such as `from __future__ import braces`. It also happens with this context, where it is clear that `name` doesn't refer to `hy.core.language.name` :
```clojure
(import [__future__ [print_function]])
(defclass SomeObject [object]
  (defn __init__ [self name]
    (setv self.name name)))
```
which compiles to
```python
from hy.core.language import name
from __future__ import print_function


class SomeObject(object):

    def __init__(self, name):
        self.name = name
        return None
```

Kodiologist:
Is there any `__future__` import that's useful in Hy, though? `print_function` is on by default in Hy on Python 2; it may not show up in the `hy2py` output, but `hy2py` is intended as a debugging tool and there's currently no guarantee that it produces working code (see in particular #970).

koji-kojiro:
What @woodrush mentioned is not a problem of only `hy2py`.

```shell-session
$ hy -c "(import [__future__ [print_function]]) (def name 10)"
Traceback (most recent call last):
  File "/usr/local/bin/hy", line 9, in <module>
    load_entry_point('hy==0.12.1+101.g5bf9ecf', 'console_scripts', 'hy')()
  File "/usr/local/lib/python3.5/dist-packages/hy/cmdline.py", line 344, in hy_main
    sys.exit(cmdline_handler("hy", sys.argv))
  File "/usr/local/lib/python3.5/dist-packages/hy/cmdline.py", line 313, in cmdline_handler
    return run_command(options.command)
  File "/usr/local/lib/python3.5/dist-packages/hy/cmdline.py", line 193, in run_command
    pretty_error(import_buffer_to_module, "__main__", source)
[...]

PR #1673, opened 2018-08-27
Fix AST handling of docstrings and __future__ ordering

brandonwillard:
This closes hylang/hy#1367, closes hylang/hy#1540

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1719,7 +1719,19 @@ def hy_compile(tree, module_name, root=ast.Module, get_expr=False):
     if not get_expr:
         result += result.expr_as_stmt()
 
-    body = compiler.imports_as_stmts(tree) + result.stmts
+    body = []
+
+    # Pull out a single docstring and prepend to the resulting body.
+    if (len(result.stmts) > 0 and
+        issubclass(root, ast.Module) and
+        isinstance(result.stmts[0], ast.Expr) and
+        isinstance(result.stmts[0].value, ast.Str)):
+
+        body += [result.stmts.pop(0)]
+
+    body += sorted(compiler.imports_as_stmts(tree) + result.stmts,
+                   key=lambda a: not (isinstance(a, ast.ImportFrom) and
+                                      a.module == '__future__'))
 
     ret = root(body=body)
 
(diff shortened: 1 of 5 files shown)


hy/compiler.py
```diff
@@ -1719,7 +1736,18 @@ def hy_compile(tree, module_name, root=ast.Module, get_expr=False):
...
+    # Pull out a single docstring and prepend to the resulting body.
+    if (len(result.stmts) > 0 and
+        issubclass(root, ast.Module) and
+        isinstance(result.stmts[0], ast.Expr) and
+        isinstance(result.stmts[0].value, ast.Str)):
+
+        body += [result.stmts.pop(0)]
+
+    body += sorted(compiler.imports_as_stmts(tree) + result.stmts,
+                   key=import_order_key)
```
Kodiologist: Couldn't you just write something like `key=lambda x: not (isinstance(x, ast.ImportFrom) and x.module == '__future__')`?

tests/compilers/test_ast.py
```diff
@@ -664,3 +664,21 @@ def test_ast_bad_yield_from():
 def test_eval_generator_with_return():
     """Ensure generators with a return statement works."""
     can_eval("(fn [] (yield 1) (yield 2) (return))")
+
+
+def test_docsting_imports():
+    """Confirm that docstrings come before import statements in AST output."""
+    hy_tree = import_buffer_to_hst('"hello world"\n(print (first [1 2]))')
+    hy_ast = hy_compile(hy_tree, 'test')
+    assert hy_ast.body[0].value.s == 'hello world'
```
Kodiologist: I think it would be a better test to add a module docstring to `resources/pydemo.hy` and check for it in `test_hy2py`.

hy/compiler.py
brandonwillard: Yeah, probably.  This started from an elaborate scheme to check `ast.ImportFrom` and multi-import `ast.Import`s, then I realized only `from __future__ import ...` statements needed to be handled and cut out parts from there.

tests/compilers/test_ast.py
brandonwillard: You fine leaving that test in there, though?  I like its "directness".

Kodiologist: It uses an internal function (`import_buffer_to_hst`), and it relies on the structure of Python AST, to test for a bug that's readily visible in user space. So I'd say we should get rid of it if we don't need it.

brandonwillard: Looks like the tests in `test_ast.py` are using `can_compile` instead; I'll make the other tests here use that as well.

Otherwise, I can replace this one with an assertion to [`test_hy2py`](https://github.com/hylang/hy/blob/master/tests/test_hy2py.py#L25) that checks `m.__doc__`.

Issue #1677, opened 2018-09-04
Hy will treat a Hy file as Python if it doesn't have a ".hy" file extension


Kodiologist:
$ echo '(print 1 2)' >testcode
    $ hy testcode
    Traceback (most recent call last):
      File "/usr/local/bin/hy", line 12, in <module>
        sys.exit(hy_main())
      File "/usr/local/lib/python3.7/site-packages/hy/cmdline.py", line 369, in hy_main
        sys.exit(cmdline_handler("hy", sys.argv))
      File "/usr/local/lib/python3.7/site-packages/hy/cmdline.py", line 355, in cmdline_handler
        runpy.run_path(filename, run_name='__main__')
      File "/usr/local/lib/python3.7/runpy.py", line 261, in run_path
        code, fname = _get_code_from_file(run_name, path_name)
      File "/usr/local/lib/python3.7/site-packages/hy/importer.py", line 195, in _get_code_from_file
        code = compile(source, fname, 'exec')
      File "testcode", line 1
        (print 1 2)
               ^
    SyntaxError: invalid syntax

@brandonwillard, you might want to take a look at this one.

brandonwillard:
Ah, yes, all the Hy-specific loader logic from #1672 is based on file extensions and falls back to Python builtins otherwise.  In other words, it only enforces a choice between files recognized as Hy and Python.

Do we want to assume that extension-less files and/or unrecognized extensions are Hy?  Should we raise an exception (from the cmdline tool) when we don't find/recognize an extension?  Should we fail silently in those cases? 

We can make some fairly straightforward changes to produce most of those conditions, but we probably want to be more laissez-faire than not.  Python's choice to interpret every file as Python source seem sorta reasonable if only because it handles the entire import process &mdash; in which we are simply injecting Hy parsing capabilities.  If we attempt to take over default behaviours, we might need to do a lot more just to preserve expected functionality elsewhere (e.g. make sure we don't block other `importlib` finders and loaders).

Kodiologist:
Yeah, I think it's reasonable to expect a `.hy` file extension of Hy files that are `import`ed, but `hy SOMETHING` should always interpret `SOMETHING` as Hy—unless, perhaps, it both has a file extension and that extension is `.py` or `.pyc`.

Kodiologist:
We also presumably want shebang lines to work:

    $ printf '#!/usr/bin/env hy\n(print 1 2)\n' > myprog
    $ chmod 755 myprog                                  
[...]

PR #1678, opened 2018-09-05
Allow `runpy` to consider non-standard Hy source extensions

brandonwillard:
Adds Hy parsing of files with unrecognized source extensions to cmdline Hy.

~~The fallback is tried only after first attempting standard Python compilation.  If the attempt raises a `SyntaxError` (and the file is not clearly identified as Python source via its extension), then it tries Hy.
The reason for choosing the order Python-then-Hy is somewhat arbitrary, except for the impression that Python's parsing might be more efficient than Hy's and, thus, better as a first attempt.~~

### DIFF
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -19,7 +19,7 @@ import astor.code_gen
 import hy
 from hy.lex import LexException, PrematureEndOfInput, mangle
 from hy.compiler import HyTypeError, hy_compile
-from hy.importer import hy_eval, hy_parse
+from hy.importer import hy_eval, hy_parse, runhy
 from hy.completer import completion, Completer
 from hy.macros import macro, require
 from hy.models import HyExpression, HyString, HySymbol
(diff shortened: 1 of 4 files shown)


brandonwillard:
@Kodiologist, the [one failure on Travis](https://travis-ci.org/hylang/hy/jobs/424621643) &mdash; for `pypy3` &mdash; is really weird; do you mind retrying that check manually?  

I've seen spurious errors for other projects with `pypy3` on Travis recently, and I can never reproduce them locally.  Perhaps it's because I have `pypy` at `Python 3.5.3[pypy-6.0.0-final]`, but, still, the errors make no sense whatsoever (relative to the changes).

Kodiologist:
I don't think this is the right way to address #1677. For example, if there's a syntax error for both Python and Hy interpretations of the file, which error do you show?

> @Kodiologist, the one failure on Travis — for pypy3 — is really weird; do you mind retrying that check manually?

I don't think that's an option on Travis, but you can trivially change your top commit (e.g. `git commit --amend --reset-author` to change the timestamp) and then force-push.

brandonwillard:
In this case, the Hy error, just as if we only assumed the file was Hy.  Otherwise, do you think Python should never even be tried for unknown filetypes?

Kodiologist:
Neat! Try adding a test where the shebang line is used by running the file as an executable, too. Assuming it's possible to arrange for that in pytest.

brandonwillard:
It is possible, but wouldn't that only be testing the shell's ability to correctly run a script?

Kodiologist:
Sorry I forgot about Hy for a few days there. I'll merge this now.

Issue #1397, opened 2017-08-25
Support the debugger from the repl


gilch:
related #741 #1395 

We can't get a code listing in the debugger if the function was defined at the repl.
```Hy
=> (defn foo []
... (setv bar 42)
... (import pdb)(pdb.set-trace))
=> (foo)
--Return--
> <eval_body>(1)foo()->None
(Pdb) l
[EOF]
(Pdb) 
```
The usual Emacs workflow involves sending toplevel forms from the file buffer to the repl, so the lack of debugger support in this case is a pain.

The truth is, Python's repl can't do this either.
```
>>> def foo():
...   bar = 42
...   import pdb; pdb.set_trace()
...
>>> foo()
--Return--
> <stdin>(3)foo()->None
(Pdb) l
[EOF]
(Pdb) 
```
So maybe that's why no-one complained. But IPython can.
```
In [1]: def foo():
   ...:     bar = 42
   ...:     import pdb; pdb.set_trace()
   ...:

In [2]: foo()
--Return--
> <ipython-input-1-8e38ce8e3905>(3)foo()->None
-> import pdb; pdb.set_trace()
(Pdb) l
  1     def foo():
  2         bar = 42
  3  ->     import pdb; pdb.set_trace()
  4
[EOF]
(Pdb) bar
42
(Pdb) 
```
So I know it's possible in principle.

ekaschalk:
I also wanted to add, is it possible to have something like IPython's %pdb toggle in hy repl (when error is encountered set trace and pdb at point)? I have no idea how involved that feature would be.

gilch:
I'm not sure about that either, but IPython is actually pretty extensible. I think it would be pretty easy to create a [custom `%hy` magic](https://ipython.readthedocs.io/en/stable/config/custommagics.html) to make IPython read Hy input for just that line. You'd just have to call Hy's compiler on that line, something like `hy.eval(hy.read_str(line))` would probably work. Then you'd get `%pdb` and all the other IPython magic for free. Even without that, you can certainly import any Hy module during an IPython session.

I'm actually considering doing it the other way around: make an IHy that reads Hy by default, which would help with #1323 as well as these debugger issues. This would be quite a bit more involved, and use IPython's [custom input transformation](https://ipython.readthedocs.io/en/stable/config/inputtransforms.html) API. I'm reluctant to include this in Hy proper, at least at first. It would be a separate project. But it might be nicer to use than the default Hy repl, since we'd be able to get nice IPython features in Hy.

[...]

Issue #1395, opened 2017-08-25
hy -i should support the debugger


gilch:
The `pdb` debugger can list the current source position even when starting Python in interactive mode with the `-i` switch.
```
$ py -i foo.py
>>> foo()
--Return--
> c:\users\me\documents\github\hy\foo.py(3)foo()->None
-> import pdb;pdb.set_trace()
(Pdb) l
  1     def foo():
  2       bar = 42
  3  ->   import pdb;pdb.set_trace()
  4
[EOF]
(Pdb) bar
42
(Pdb) c
>>>
```
But this doesn't work in Hy.
```
> vi foo.hy
> cat foo.hy
(defn foo []
  (setv bar 42)
  (import pdb)(pdb.set-trace))

> hy -i foo.hy
hy 0.13.0+52.g30dd07a using CPython(v3.6.1:69c0db5) 3.6.1 on Windows
=> (foo)
--Return--
> <eval_body>(1)foo()->None
(Pdb) l
[EOF]
(Pdb)
```
Somehow the source gets lost.

waigx:
Thinking to pick up this one if possible.

It seems that we need to override this function (I might be wrong): https://github.com/python/cpython/blob/master/Lib/pdb.py#L1194

Sorry for being unfamiliar with the code base and probably easier asking you guys: have we done similar thing (override a method in Python library) before?

waigx:
Okay, probably should go `hy/hy/contrib/` folder

vodik:
@waigx let me first see if its fixed if i run it under #1518

vodik:
Nope

@waigx I think that's a red hearing. Dig into the `__code__` object first, you'll see some inconsistencies.

In python:

```
>>> foo.__code__
<code object foo at 0x7f3e158a7930, file "moo.py", line 1>
```

and in Hy:

```
=> (. foo __code__)
<code object foo at 0x7fd2d2c470c0, file "<eval_body>", line 1>
```

If you want my opinion on what's likely happening, Python imported foo.py, but Hy evalled it instead.

vodik:
If you like, I can add this into #1518 since a lot of this stuff is already getting overhauled in that PR. I'm actively working on it currently. I appreciate the enthusiasm though.

waigx:
Ah, thanks for the heads-up. I honestly just started looking into it, and not familiar with it at all.

[...]

Issue #741, opened 2015-01-02
Easier macro debugging


cwebber:
I'm currently writing a larger program in Hy.  Coding in Hy is a dream.. debugging is a headache.  The reasons are simple: line numbers are lost in the context of any macros, and of course plenty of things in Hy are macros.  In one of @paultag's demos he shows that using pdb with Hy works surprisingly great, but of course this presentation didn't show anything wrapped in a macro.

The problem is that have no idea what will come out of a macro, and even if we did, we lose track of it by the time it comes around to the AST being generated.

Is this unsolvable?  Maybe.  But other languages have done better:
- Emacs Lisp and Racket both allow for stepping through macros.  Many lisps do not provide it, and both of these do something clever to provide it.
 -  [Emacs lisp has users "instrument" macros with specifications](https://www.gnu.org/software/emacs/manual/html_node/elisp/Edebug-and-Macros.html#Edebug-and-Macros).  You can step through functions using edebug much the way you can step through things with pdb.set_trace().  It relies on the specification to figure out what is happening somehow.
 - Racket allows for some sort of [macro debugger](http://docs.racket-lang.org/macro-debugger/index.html).
- Guile and Racket both have [tracing](https://www.gnu.org/software/guile/manual/html_node/Tracing-Traps.html), and Racket even has [macro tracing](http://docs.racket-lang.org/reference/debugging.html#%28form._%28%28lib._racket%2Ftrace..rkt%29._trace-define-syntax%29%29) (whatever that is).
- Probably if we had something like define-syntax in scheme and encouraged users to use that, it would be easier to infer some of these structures.

Is it impossible for us to do this?  I don't know... I think that if emacs lisp can have a sense of where things are, probably we can.  At the very least, once we build AST, we might be able to use some information from macro expansion (or at least, certain common patterns of macro expansion using some sort of define-syntax system) to be able to debug things.

The current state of things is pretty sorry anyway.  I'd like them to become better.  Feel free to use this thread to dump ideas.

cwebber:
Alternately, if other core developers decide "there's no way forward" or that this bug won't provide a useful dialogue, feel free to close it.

cwebber:
[DrRacket seems to have even more debugging features](http://docs.racket-lang.org/drracket/debugger.html), but how do those work?  How are you able to so easily set breakpoints and step through, from a GUI even?  What sorcery is this?  I'd love to hear more from someone who knows.

cwebber:
[...]

Issue #142, opened 2013-04-19
Add support for post-mortem debugging from the REPL


agentultra:

glyh:
There's also DAP, which is backing a lot of editors like vim/emacs.

atisharma:
`pdb` post mortem seems to work fine since inspect was patched.

In the REPL,
```hy
Hy 1.3.0 (Dogs Should Be Raw) using CPython(main) 3.14.2 on Linux
» (import pdb)
» (/ 1 0)
Traceback (most recent call last):
  File "stdin-7b3ace8766f1e1cfb3ae7c01a1a61cebed24f482", line 1, in <module>
    (/ 1 0)
     ^^^^^^
ZeroDivisionError: division by zero
» (pdb.pm)
> /home/ati/stdin-7b3ace8766f1e1cfb3ae7c01a1a61cebed24f482(1)<module>()
(Pdb) p sys.last_exc
ZeroDivisionError('division by zero')
(Pdb) bt
  /home/ati/.local/share/venv/lib/python3.14/site-packages/hy/repl.py(344)runcode()
-> self.last_value = eval(code[1], self.locals)
> /home/ati/stdin-7b3ace8766f1e1cfb3ae7c01a1a61cebed24f482(1)<module>()
(Pdb) q
```
or
```hy
Hy 1.3.0 (Dogs Should Be Raw) using CPython(main) 3.14.2 on Linux
» (import sys pdb)
» (import test)
Traceback (most recent call last):
  File "stdin-68ff4477f1e971a67b00bf460a36e1194c1e5093", line 1, in <module>
    (import test)
     ^^^^^^^^^^^^
  File "/tmp/test.hy", line 3, in <module>
    (/ 1 0)
     ^^^^^^
ZeroDivisionError: division by zero
» (pdb.pm)
> /tmp/test.hy(3)<module>()
-> (/ 1 0)
(Pdb) p sys.last_exc
ZeroDivisionError('division by zero')
(Pdb) bt
  /home/ati/.local/share/venv/lib/python3.14/site-packages/hy/repl.py(343)runcode()
-> eval(code[0], self.locals)
  /tmp/stdin-68ff4477f1e971a67b00bf460a36e1194c1e5093(1)<module>()
> /tmp/test.hy(3)<module>()
-> (/ 1 0)
(Pdb) q
```

Kodiologist:
Good enough for me. This 13-year-old issue was underspecified anyway.

PR #1680, opened 2018-09-09
Enable Hy code in the debugger

brandonwillard:
With the context manager provided here, `pdb` (`ipdb` and Pdb++) sessions started from the Hy REPL can evaluate Hy code.  For example, the following is possible:
```clojure
=> (pdb.run "(setv x 1)")                                                                                                                                                                                           
[1] > <string>(1)<module>()
(Pdb++) p (+ 1 1)                                                                                                                                                                                                   
2L
None
(Pdb++) !(print "hi")                                                                                                                                                                                               
hi
(Pdb++) up                                                                                                                                                                                                          
[0] > /home/bwillard/projects/code/python/hy/hy/cmdline.py(150)_hy_bdb_run()
-> exec(cmd, globals, locals)
(Pdb++) (.keys locals)                                                                                                                                                                                              
['__builtins__', 'pdb', 'x', u'hyx_XasteriskX1', '__name__', u'hyx_XasteriskX3', u'hyx_XasteriskX2', '__doc__']
(Pdb++) c                                                                                                                                                                                                           
=> x                                                                                                                                                                                                                
1L
=> (import [hy.contrib.walk [walk postwalk]])                                                                                                                                                                            
=> (pdb.run "(postwalk identity '(+ x 2))")                                                                                                                                                                         
[1] > <string>(0)<module>()
(Pdb++) b walk
Breakpoint 1 at /home/bwillard/projects/code/python/hy/hy/contrib/walk.hy:17
(Pdb++) c                                                                                                                                                                                                           
[3] > /home/bwillard/projects/code/python/hy/hy/contrib/walk.hy(17)walk()
-> [(instance? HyExpression form)
(Pdb++) ll                                                                                                                                                                                                          
   1     ;;; Hy AST walker                                                                                                                                                                                          
   2     ;; Copyright 2018 the authors.                                                                                                                                                                             
   3     ;; This file is part of Hy, which is free software licensed under the Expat                                                                                                                                
   4     ;; license. See the LICENSE.                                                                                                                                                                               
   5                                                                                                                                                                                                                
   6     (import [hy [HyExpression HyDict]]                                                                                                                                                                         
   7             [functools [partial]]                                                                                                                                                                              
   8             [collections [OrderedDict]]                                                                                                                                                                        
   9             [hy.macros [macroexpand :as mexpand]]                                                                                                                                                              
  10             [hy.compiler [HyASTCompiler]])                                                                                                                                                                     
  11                                                                                                                                                                                                                
  12     (defn walk [inner outer form]                                                                                                                                                                              
  13       "Traverses form, an arbitrary data structure. Applies inner to each                                                                                                                                      
  14       element of form, building up a data structure of the same type.                                                                                                                                          
  15       Applies outer to the result."                                                                                                                                                                            
  16       (cond                                                                                                                                                                                                    
  17  ->    [(instance? HyExpression form)                                                                                                                                                                          
  18         (outer (HyExpression (map inner form)))]                                                                                                                                                               
  19        [(instance? HyDict form)                                                                                                                                                                                
  20         (HyDict (outer (HyExpression (map inner form))))]                                                                                                                                                      
  21        [(instance? list form)                                                                                                                                                                                  
  22         ((type form) (outer (HyExpression (map inner form))))]                                                                                                                                                 
  23        [(coll? form)                                                                                                                                                                                           
  24         (walk inner outer (list form))]                                                                                                                                                                        
  25        [True (outer form)]))                                                                                                                                                                                   
  26                                                                                                                                                                                                                
  27     ...
(Pdb++) 
```

### DIFF
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -13,17 +13,196 @@ import io
 import importlib
 import py_compile
 import runpy
-
+import traceback
 import astor.code_gen
 
 import hy
 from hy.lex import LexException, PrematureEndOfInput, mangle
 from hy.compiler import HyTypeError, hy_compile
-from hy.importer import hy_eval, hy_parse
+from hy.importer import hy_eval, hy_parse, hy_ast_compile_flags
 from hy.completer import completion, Completer
 from hy.macros import macro, require
 from hy.models import HyExpression, HyString, HySymbol
-from hy._compat import builtins, PY3, FileNotFoundError
+from hy._compat import builtins, PY3, FileNotFoundError, str_type
+
+
+class HyPdb(object):
+    """A contextmanager that patches `bdb` and `pdb` modules to make them parse
+    Hy.
+
+    In order to affect the interactive console environment when using `exec`,
+    we create custom versions of `[pb]db` methods that use specific namespace
+    dictionaries.  Also, they force use of the closure's global and local
+    values to slightly reduce the number of functions that need patches.
+    """
+
+    @staticmethod
+    def pdb_compile(src, filename='<stdin>',
+                    mode='single', module_name='cmdline'):
+        hy_tree = hy_parse(src + '\n')
+        ast_root = ast.Interactive if mode == 'single' else ast.Module
+        hy_ast = hy_compile(hy_tree, module_name,
+                            root=ast_root)
+        code = compile(hy_ast, filename, mode, hy_ast_compile_flags)
+        return code
+
+    def __init__(self, ctx_globals=None, ctx_locals=None):
+        self.ctx_globals = ctx_globals
+        self.ctx_locals = ctx_locals
+
+    def pdbpp_getval_or_undefined(self):
+        _pdb = self.pdb
+        def _pdbpp_getval_or_undefined(self, arg):
+            """This is just for `pdb++`"""
+            try:
+                code = HyPdb.pdb_compile(arg)
+                return eval(code, self.curframe.f_globals,
+                            self.curframe.f_locals)
+            except NameError:
+                return _pdb.undefined
+
+    def hy_pdb_default(self):
+        def _hy_pdb_default(self, line):
+            if line[:1] == '!': line = line[1:]
+            locals = self.curframe_locals
+            globals = self.curframe.f_globals
+            try:
+                code = HyPdb.pdb_compile(line + '\n', mode='single')
+                save_stdout = sys.stdout
+                save_stdin = sys.stdin
+                save_displayhook = sys.displayhook
+                try:
+                    sys.stdin = self.stdin
+                    sys.stdout = self.stdout
+                    sys.displayhook = self.displayhook
+                    exec(code, globals, locals)
+                finally:
+                    sys.stdout = save_stdout
+                    sys.stdin = save_stdin
+                    sys.displayhook = save_displayhook
+            except:
+                exc_info = sys.exc_info()[:2]
+                msg = traceback.format_exception_only(*exc_info)[-1].strip()
+                print('***', msg, file=self.stdout)
+        return _hy_pdb_default
+
+    def hy_pdb_getval(self):
+        def _hy_pdb_getval(self, arg):
+            try:
+                code = HyPdb.pdb_compile(arg)
+                return eval(code, self.curframe.f_globals,
+                            self.curframe_locals)
+            except:
+                t, v = sys.exc_info()[:2]
+                if isinstance(t, str):
+                    exc_type_name = t
+                else:
+                    exc_type_name = t.__name__
+
+                print('***', exc_type_name + ':', repr(v), file=self.stdout)
+                raise
+        return _hy_pdb_getval
+
+    def hy_pdb_getval_except(self):
+        _pdb = self.pdb
+        def _hy_pdb_getval_except(self, arg, frame=None):
+            try:
+                code = HyPdb.pdb_compile(arg)
+                if frame is None:
+                    return eval(code, self.curframe.f_globals, self.curframe_locals)
+                else:
+                    return eval(code, frame.f_globals, frame.f_locals)
+            except:
+                exc_info = sys.exc_info()[:2]
+                err = traceback.format_exception_only(*exc_info)[-1].strip()
+                return _pdb._rstr('** raised %s **' % err)
+        return _hy_pdb_getval_except
+
+    def hy_bdb_runeval(self):
+        ctx_globals = self.ctx_globals
+        ctx_locals = self.ctx_locals
+
+        def _hy_bdb_runeval(self, expr, globals=ctx_globals, locals=ctx_locals):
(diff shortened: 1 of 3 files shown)


brandonwillard:
After adding tests, this should at least close
- [ ] https://github.com/hylang/hy/issues/142
- [ ] https://github.com/hylang/hy/issues/1395 (I think this was already closed by #1672)

With a little extra (and perhaps tangential REPL) work, we could get #1397 in place, as well.  Also, after some upcoming macro/`require` rehauls, some of the ideas in #741 might be do-able.

hy/cmdline.py
```diff
@@ -13,17 +13,196 @@
...
+            try:
+                code = HyPdb.pdb_compile(arg)
+                return eval(code, self.curframe.f_globals,
+                            self.curframe.f_locals)
+            except NameError:
+                return _pdb.undefined
+
+    def hy_pdb_default(self):
+        def _hy_pdb_default(self, line):
+            if line[:1] == '!': line = line[1:]
```
Kodiologist: I think the condition here is equivalent to `line.startswith('!')`, which seems like a clearer way to write it.

hy/cmdline.py
```diff
@@ -13,17 +13,196 @@
...
+                return eval(code, self.curframe.f_globals,
+                            self.curframe_locals)
+            except:
+                t, v = sys.exc_info()[:2]
+                if isinstance(t, str):
+                    exc_type_name = t
+                else:
+                    exc_type_name = t.__name__
+
+                print('***', exc_type_name + ':', repr(v), file=self.stdout)
```
Kodiologist: I think you could just write `t if isinstance(t, str) else t.__name__` instead of defining `exc_type_name`.

hy/cmdline.py
```diff
@@ -13,17 +13,196 @@
...
+        def _hy_pdb_getval_except(self, arg, frame=None):
+            try:
+                code = HyPdb.pdb_compile(arg)
+                if frame is None:
+                    return eval(code, self.curframe.f_globals, self.curframe_locals)
+                else:
+                    return eval(code, frame.f_globals, frame.f_locals)
+            except:
+                exc_info = sys.exc_info()[:2]
+                err = traceback.format_exception_only(*exc_info)[-1].strip()
```
Kodiologist: Likewise, there's no need for a local `exc_info`.

hy/cmdline.py
```diff
@@ -13,17 +13,196 @@
...
+        # if hasattr(pdb, 'pdb'):
+        #     pdb.pdb.Pdb.default = _pdb_default if restore else _hy_pdb_default
+        #     pdb.pdb.Pdb._getval = _pdb_getval if restore else _hy_pdb_getval
+        #     if hasattr(pdb.pdb.Pdb, '_getval_except'):
+        #         pdb.pdb.Pdb._getval_except = _pdb_getval_except if restore else _hy_pdb_getval_except
+        # else:
+        #     pdb.Pdb.default = _pdb_default if restore else _hy_pdb_default
+        #     pdb.Pdb._getval = _pdb_getval if restore else _hy_pdb_getval
+        #     if hasattr(pdb.Pdb, '_getval_except'):
+        #         pdb.Pdb._getval_except = _pdb_getval_except if restore else _hy_pdb_getval_except
```
Kodiologist: You probably don't want to commit this big block of commented-out code.

hy/cmdline.py
brandonwillard: All these patched functions are direct copies from the `[bp]db` source (a couple have slight cross-version modifications, though), but, yeah, they could be cleaned up a bit.

Kodiologist: In that case, you should attribute the source of the copied code in the commit message.

brandonwillard:
PyPy raises exceptions during calls to `inspect.getsource`:
```clojure
hy 0.15.0+32.g4af87dc.dirty using PyPy(fdd60ed87e941677e8ea11acf9f1819466521bf2) 3.5.3 on Linux
=> (import inspect [tests.resources.bin.pdb [*]])
=> (inspect.getsource func2)
  File "/home/bwillard/projects/code/python/hy/hy/importer.py", line 140, in hy_eval
  File "/home/bwillard/apps/anaconda3/envs/hy-dev-pypy35/lib-python/3/inspect.py", line 947, in getsource
    lines, lnum = getsourcelines(object)
  File "/home/bwillard/apps/anaconda3/envs/hy-dev-pypy35/lib-python/3/inspect.py", line 939, in getsourcelines
    return getblock(lines[lnum:]), lnum + 1
  File "/home/bwillard/apps/anaconda3/envs/hy-dev-pypy35/lib-python/3/inspect.py", line 919, in getblock
    for _token in tokens:
  File "/home/bwillard/apps/anaconda3/envs/hy-dev-pypy35/lib-python/3/tokenize.py", line 597, in _tokenize
    raise TokenError("EOF in multi-line statement", (lnum, 0))
tokenize.TokenError: ('EOF in multi-line statement', (2, 0))
```
while CPython raises no exception (notice, however, that the result it gives is incomplete):
```clojure
hy 0.15.0+32.g4af87dc.dirty using CPython(default) 2.7.15 on Linux
=> (import inspect [tests.resources.bin.pdb [*]])
=> (inspect.getsource func2)
'(defn func1 [x]\n'
```
This appears to be what's causing the current test failures in Travis, because the debugger triggers calls to `inspect` during code listing.  These problems can be fixed by patching `inspect`, as well, which generally seems like a compatibility we would want.

In the course of researching and testing said patches, I noticed that `defn` forms do not completely preserve/map source line-number information (at least not in PyPy).  It's possible to work around this &mdash; to some extent &mdash; via special `inspect` logic, but a true solution leads right back to fundamental macro questions and, most likely, changes.

## Comment on Patching for Hy

On a somewhat related note, all this patching has me rethinking how we could accomplish the same compatibility and core Python code re-use with much less effort.  In quite a few instances, the necessary patches simply wrap the use of `eval`, `exec` and `compile` so that they handle Hy code.  While patching those bulit-ins directly is possible (well, for `exec`, in Python 3.x, at least), simply replacing the regular Python handling with Hy is not sufficient; however, there might be other work-arounds available that allow those functions to handle Hy source strings as a special case.

For instance, if Hy source strings were created by the Hy interpreter and loader as a special, identifiable type (e.g. `class HySource(UserString)`), then `eval`, `exec` and `compile` could be patched with only a simple condition handling those.

Overall, patching every piece of relevant code that calls `eval`, `exec` and/or `compile` sounds like an uphill battle, so it's better to address this now.

Kodiologist:
Oddly, it seems that no Travis run is showing for this PR.

> …however, there might be other work-arounds available that allow those functions to handle Hy source strings as a special case.

You're certainly welcome to try.

Kodiologist:
I'm closing this because of the lack of follow-up. Please reopen if you want to pick this up again.

Issue #1650, opened 2018-06-29
`require` and macro dependencies


brandonwillard:
What are the expectations and conventions surrounding `require`-ed macros and their dependencies?

For example, consider the following:
### In `test-module.hy`
```clojure
(require [hy.contrib.walk [let]])

(defmacro test-module-macro [a]
  `(let [b 1] (+ b ~a)))
```
### In `some-file.hy`
```clojure
(require [test-module [*]])

(test-module-macro 1)
```

In its current state, running `some-file.hy` will produce
```python
  => let([b, 1], b + 1)

  Traceback (most recent call last):
    File "/home/bwillard/projects/code/python/hy/hy/importer.py", line 199, in hy_eval
      return eval(ast_compile(expr, "<eval>", "eval"), namespace)
    File "<eval>", line 1, in <module>
  NameError: name 'let' is not defined
```
This result implies that `(require [test-module [*]])` does not account for the required module's requirements (i.e. `let` in `hy.contrib.walk`).

How should dependencies like this be handled?  More importantly, what would it take to automatically include macro dependencies?

FYI: This question started [here](https://github.com/algernon/monaxhyd/pull/2#issuecomment-400814270), where @gilch [provided a lot of important considerations and an approach](https://github.com/algernon/monaxhyd/pull/2#issuecomment-401227376).

Kodiologist:
Automatically determining which macros a piece of code depends upon is, I suspect, undecidable. But there's no reason in principle that we couldn't make `require` able to pull in macros that are themselves `require`d by the named file. #1268 is related.

Generally, macro subroutines are better written as functions than macros. Unfortunately, that's not an option when you're using a preexisting macro like `let`. You could try using `macroexpand`.

brandonwillard:
The idea of `require`-ing a module's `require`s sounds helpful (in this situation), but perhaps not scalable &mdash; among other things.

My first response to this situation involved things like `macroexpand`, and @gilch provided some potential shortcomings in [this comment](https://github.com/algernon/monaxhyd/pull/2#issuecomment-401227376).

Given that Hy seems partly inspired by Clojure, how does it handle these things?  (Unfortunately, I'm not familiar enough with Clojure to readily find that answer in [its source](https://github.com/clojure/clojure).)

Kodiologist:
Not scalable? Do you mean you expect there to be some kind of performance problem with allowing `require` to take macros that are `require`d by the named file? Why so?

kirbyfan64:
[...]

PR #1682, opened 2018-09-24
Macro processing updates and fixes

brandonwillard:
This PR changes the way macros are loaded and managed within modules/namespaces, their scope of evaluation, and how `require` statements are treated during the module bytecode loading process.

Alongside those changes, a general refactoring was started that moves toward a `HyASTCompiler` that can operate in two fashions: 1) compilation that occurs within a module and allows macros to interact with the compiler (as they are currently configured to do), and 2) compilation that does not require a concrete module context but does not allow true macro-compiler interaction.  The latter case could drastically simplify some of the module importing work-arounds that currently exist in this PR (e.g. when interacting with `runpy`).

It addresses/closes the following issues:
- [x] https://github.com/hylang/hy/issues/1268
- [x] https://github.com/hylang/hy/issues/1650
- [x] https://github.com/hylang/hy/issues/1416

# Example
## Before
```clojure
=> (with [file (open "test_mod.hy" "w")] (file.write "(require [hy.contrib.walk [let]])\n\n(defmacro test-macro [x]\n  `(let [y ~x] (print y)))\n"))
87
=> (with [file (open "test_mod.hy" "r")] (print (file.read)))
(require [hy.contrib.walk [let]])

(defmacro test-macro [x]
  `(let [y ~x] (print y)))

=> (require [test-mod [test-macro]])
=> (test-macro 1)
  File "/home/bwillard/projects/code/python/hy-master/hy/importer.py", line 141, in hy_eval
NameError: name 'let' is not defined
```
## After
```clojure
hy 0.15.0+45.ga8d6a48 using CPython(default) 3.6.6 on Linux
=> (with [file (open "test_mod.hy" "w")] (file.write "(require [hy.contrib.walk [let]])\n\n(defmacro test-macro [x]\n  `(let [y ~x] (print y)))\n"))
87
=> (with [file (open "test_mod.hy" "r")] (print (file.read)))
(require [hy.contrib.walk [let]])

(defmacro test-macro [x]
  `(let [y ~x] (print y)))

=> (require [test-mod [test-macro]])
True
=> (test-macro 1)
1
=> (import [pprint [pprint]])
=> (pprint __macros__)
{'_hyx_XgreaterHthan_signX': <function _hy_anon_var_14 at 0x7fbc2fef0a60>,
 '_hyx_XgreaterHthan_signXXgreaterHthan_signX': <function _hy_anon_var_17 at 0x7fbc2fef0b70>,
 'assoc': <function _hy_anon_var_3 at 0x7fbc2fef07b8>,
 'comment': <function _hy_anon_var_29 at 0x7fbc2fef31e0>,
 'cond': <function _hy_anon_var_13 at 0x7fbc2fef09d8>,
 'defmacro': <function _hy_anon_var_3 at 0x7fbc2fee70d0>,
 'defmain': <function _hy_anon_var_27 at 0x7fbc2fef30d0>,
 'defn': <function _hy_anon_var_11 at 0x7fbc2fee7378>,
 'deftag': <function _hy_anon_var_7 at 0x7fbc2fee7268>,
 'doc': <function _hy_anon_var_30 at 0x7fbc2fef3268>,
 'doto': <function _hy_anon_var_16 at 0x7fbc2fef0ae8>,
 'hyx_as_XgreaterHthan_signX': <function _hy_anon_var_1 at 0x7fbc2fee7598>,
 'hyx_defmacroXexclamation_markX': <function _hy_anon_var_26 at 0x7fbc2fef3048>,
 'hyx_defmacroXsolidusXgXexclamation_markX': <function _hy_anon_var_24 at 0x7fbc2fef0f28>,
 'hyx_defnXsolidusXa': <function _hy_anon_var_14 at 0x7fbc2fee7400>,
 'hyx_if': <function _hy_anon_var_4 at 0x7fbc2fee71e0>,
 'hyx_with': <function _hy_anon_var_7 at 0x7fbc2fef08c8>,
 'hyx_withXsolidusXa': <function _hy_anon_var_8 at 0x7fbc2fef0950>,
 'ideas': <function ideas_macro at 0x7fbc302698c8>,
 'if_not': <function _hy_anon_var_18 at 0x7fbc2fef0bf8>,
 'if_python2': <function _hy_anon_var_15 at 0x7fbc2fee7488>,
 'koan': <function koan_macro at 0x7fbc30269840>,
 'lif': <function _hy_anon_var_19 at 0x7fbc2fef0c80>,
 'lif_not': <function _hy_anon_var_20 at 0x7fbc2fef0d08>,
 'macro_error': <function _hy_anon_var_8 at 0x7fbc2fee72f0>,
 'test_macro': <function <lambda> at 0x7fbc2fd9f598>,
 'unless': <function _hy_anon_var_22 at 0x7fbc2fef0e18>,
 'when': <function _hy_anon_var_21 at 0x7fbc2fef0d90>,
 'with_gensyms': <function _hy_anon_var_23 at 0x7fbc2fef0ea0>}
```


Kodiologist:
Sounds pretty exciting. Want to add some tests?

brandonwillard:
Doing just that &mdash; plus further necessary additions.  I'm working somewhat incrementally on this one, so you folks can see/be involved in the process.

brandonwillard:
The changes introduced here might also help address the following issues:
- [ ] https://github.com/hylang/hy/issues/1414
- [ ] https://github.com/hylang/hy/issues/1324

(Moved from initial comment)

Kodiologist:
At a high level, how does calling `test-macro-module` in a file other than the file it was defined in work? How does Hy now know that `let` refers to the `let` that was required in the module where `test-macro-module` was defined? How would this interact with any new macro named `let` in the calling file?

tests/resources/macro_with_require.hy
```diff
@@ -0,0 +1,7 @@
+(require [hy.contrib.walk [let]])
```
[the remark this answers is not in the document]
brandonwillard: Good point; will update.

brandonwillard:
When `test-macro-module` is `macroexpand`ed (in, say, `__main__`), we get its definition namespace (i.e. module `tests.resources.macro_with_require`) from the macro function object, then we assign the definition namespace to the `HyExpression` resulting from the macro's expansion via the new `HyObject.module` field.  (Currently, this is only done for `HyExpressions` objects resulting from a macro expansion.)

When that `HyExpression` is `macroexpand`ed, we first check the compiler namespace's `__macros__` dictionary (e.g. `__main__`), then, if that fails, the definition namespace's via `HyExpression.module`.

This form of namespace/module resolution is intentionally restrictive; for instance, it doesn't propagate into sub-expressions or apply to anything else other than macro expansion.  This can easily be changed; however, I wanted to start by implementing just enough to make non-trivial macro work a little more tolerable.

Kodiologist:
Interesting; thanks for the explanation. As the system gets more mature, we'll probably also want to change the order of checks so that the macro definition module is checked first. Otherwise, a macro user can inadvertently shadow one of the macro's subroutines.

brandonwillard:
That's a simple change to make, so I'll do it now before it causes confusion down the line.

brandonwillard:
Macro expansion now checks the macro's module's namespace first, then the "local" namespace.
Also, `require`ing all will only add the macros and tags defined in the module being `require`d (and not its `require`d dependencies).

Finally, I rewrote the tests so that they use `test.resources` macros and modules exclusively.  There's also a simple test in there confirming that `require`d macros (even `require`ments of `require`d macros) can reference variables in their module namespaces and that their expansions only bind to symbols in the namespace they're expanded (e.g. locally). (Just to be safe.)

Kodiologist:
Nice. Are you done with this PR or are you planning more changes?

brandonwillard:
Not yet; just noticed that the tag macros don't resolve in the correct order (didn't update the code in that place), so I need to fix that and add another test.

Otherwise, I have some AST/source line numbering fixes to make.  For example, `hy_eval` strips line numbering and doesn't compile with the correct `filename`, so anywhere that's used &mdash; like `eval-and-compile` &mdash; the resulting bytecode un-debuggable source.  

On a related note, `hy_eval` and `hy_parse` and `ast_compile` should all probably be in `compiler.py`; currently, there's a weird back and forth between those modules that simply isn't necessary.  Also, it seems like we shouldn't have to create a whole new `HyAstCompiler` during compilation of `eval-and-compile`, which is what's currently happening through our use of `hy_eval`, so I would like to refactor that, too.

I think the two aforementioned changes could/should have their own PRs, though, so I'll end this one here (unless something more related to macros comes up).

brandonwillard:
All right, updated the last commit with the correct macro resolution order (for tags) and corresponding tests.

Kodiologist:
> Otherwise, I have some AST/source line numbering fixes to make.

Are these regressions introduced by this PR, or preexisting issues?

brandonwillard:
Those are pre-existing issues.

Kodiologist:
I tried the test I suggested at the top of #1268 with Python 3.6.6. Running `hy foo.hy` once crashes:

       File "/home/hippo/Desktop/py36-env/bin/hy", line 12, in <module>
         sys.exit(hy_main())
       File "/home/hippo/Desktop/hyenv/hy/hy/cmdline.py", line 382, in hy_main
       File "/home/hippo/Desktop/hyenv/hy/hy/cmdline.py", line 368, in cmdline_handler
         runhy.run_path(filename, run_name='__main__')
       File "/usr/lib/python3.6/runpy.py", line 263, in run_path
         pkg_name=pkg_name, script_name=fname)
       File "/usr/lib/python3.6/runpy.py", line 96, in _run_module_code
         mod_name, mod_spec, pkg_name, script_name)
       File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
         exec(code, run_globals)
       File "/tmp/foo.hy", line 1, in <module>
         (require [hy.extra.anaphoric [ap-if]])
       File "/home/hippo/Desktop/hyenv/hy/hy/macros.py", line 106, in require
         target_module = importlib.import_module(target_module)
       File "/home/hippo/Desktop/py36-env/lib/python3.6/importlib/__init__.py", line 126, in import_module
       File "<frozen importlib._bootstrap>", line 994, in _gcd_import
       File "<frozen importlib._bootstrap>", line 971, in _find_and_load
       File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
       File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
       File "<frozen importlib._bootstrap_external>", line 678, in exec_module
       File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
       File "/tmp/foo.hy", line 2, in <module>
         (print (eval '(ap-if (+ "a" "b") (+ it "c"))))
       File "/home/hippo/Desktop/hyenv/hy/hy/importer.py", line 145, in hy_eval
         return eval(ast_compile(expr, "<eval>", "eval"), globals, locals)
     NameError: name 'ap_if' is not defined

Likewise on master (in fact, that code I wrote hasn't worked, as written, since several releases ago). On master, running it again makes no difference, but with your PR, running it again prints:

    abc
    abc

Only one `abc` is expected.

So I'm not sure that your PR is fault given that the code was already broken—all your PR has done is make it a bit less broken—but perhaps it's something to think about?

tests/native_tests/native_macros.hy
```diff
@@ -329,3 +330,137 @@
     (except [e SystemExit]
       (assert (= (str e) "42"))))
   (setv --name-- oldname))
+
+#@(pytest.mark.xfail
+    (defn test-macro-module-expansions []
```
Kodiologist: I think this test is wrong: `(nonlocal-test-macro 1)` should have been expanded before `(setv module-name-var "tests.resources.macros")` was evaluated, because the former is a macro expansion while the latter is ordinary code and they're in the same file. See my comments on #1414.

brandonwillard:
The example in #1268 should work, and it does in Python 2.7, but not Python 3.x.  I'm looking at it now.

tests/native_tests/native_macros.hy
brandonwillard: Yeah, that's why it's currently an expected fail, but I'm open to making it work &mdash; e.g. with the incremental compile-and-eval approach.  If there aren't any objections to that design change, I can put that change higher up on my priorities.

By the way, an incremental approach would drastically simplify debugging!

Kodiologist: Such an incremental approach would not allow one to expand all macros before compiling to bytecode or translating to Python.

hy/_compat.py
```diff
@@ -6,6 +6,7 @@
     import __builtin__ as builtins
 except ImportError:
     import builtins  # NOQA
+
```
Kodiologist: If you're sure you want this change, put it in its own commit.

hy/cmdline.py
```diff
@@ -363,8 +376,8 @@ def cmdline_handler(scriptname, argv):
     return run_repl(spy=options.spy, output_fn=options.repl_output_fn)
 
 
-# entry point for cmd line script "hy"
 def hy_main():
+    """Entry point for cmd line script `hy`"""
```
Kodiologist: All these transforms from comments to docstrings are good, but should also get their own commit. (A lot of important stuff is going on in this commit that we don't want crowded out by minor stuff like this.)

hy/compiler.py
```diff
@@ -2,6 +2,7 @@
 # Copyright 2018 the authors.
```
Kodiologist: The refactoring diffs in this file are big enough that they should get their own PR. I'm not sure they're all improvements, though. For example, you moved `_loopers` away from where it's used.

brandonwillard: Removed them

hy/compiler.py
```diff
@@ -1484,7 +1540,8 @@ def compile_class_expression(self, expr, root, name, rest):
                 [x for pair in attrs[0] for x in pair]).replace(attrs)))
 
         for e in body:
-            e = self.compile(self._rewire_init(macroexpand(e, self)))
+            e_expanded = macroexpand(e, self.module, self)
+            e = self.compile(self._rewire_init(e_expanded))
```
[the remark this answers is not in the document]
Kodiologist: I corrected myself; see above. Sorry.

hy/compiler.py
```diff
@@ -279,32 +281,54 @@ def is_unpack(kind, x):
             and x[0] == "unpack-" + kind)
 
 
-_stdlib = {}
+class HyASTCompiler(object):
+    """A Hy-to-Python AST compiler.
 
+    It's probably more precise to call this a transpiler, though.
```
gilch: Transpliers are compilers though. And Python AST is a lower-level representation rather than human-readable source code, so "transpiler" isn't really accurate.

brandonwillard: Ha, yeah, I was debating that in my head, and ultimately sided with "transpiler".  Still, that comment was actually a personal note that shouldn't have ended up in there!

hy/importer.py
```diff
@@ -105,19 +135,23 @@ def hy_eval(hytree, namespace=None, module_name=None, ast_callback=None):
...
-        namespace = inspect.getargvalues(frame).locals
-    if module_name is None:
-        m = inspect.getmodule(inspect.stack()[1][0])
-        module_name = '__eval__' if m is None else m.__name__
+        locals = inspect.getargvalues(frame).locals
 
-    if not isinstance(module_name, string_types):
-        raise TypeError("Module name must be a string")
+    if not isinstance(locals, dict):
+        raise TypeError("Globals must be a dictionary")
```
Kodiologist: Should this be "*Locals* must be a dictionary"?

hy/importer.py
```diff
@@ -167,6 +200,54 @@ def cache_from_source(source_path):
...
+    the module here.
+
+    The function `inspect.getmodule` has a hidden-ish feature that returns
+    modules using their associated filenames (via `inspect.modulesbyfile`),
+    and, since the Loaders (and their delegate Loaders) carry a filename/path
+    associated with the parent package, we use it as a more robust attempt to
+    obtain an existing module object.
+
+    When no module object is found, a temporary, minimally sufficient module
+    object is created for the duration of the `when` body.
```
Kodiologist: Should "when" be "with"?

hy/importer.py
```diff
@@ -167,6 +200,54 @@ def cache_from_source(source_path):
...
+        module = None
+
+    if module is None:
+        tmp_mod = True
+        module = sys.modules.setdefault(loader.name,
+                                        types.ModuleType(loader.name))
+        module.__file__ = loader.path
+        module.__name__ = loader.name
+        # module.__path__
+        # module.__package__
```
Kodiologist: How about these comments?

hy/importer.py
```diff
@@ -554,7 +657,7 @@ def hyc_compile(file_or_code, cfile=None, dfile=None, doraise=False):
     py_compile.compile = hyc_compile
 
 
-# We create a separate version of runpy, "runhy", that prefers Hy source over
+# A separate version of runpy, named "runhy", that prefers Hy source over
```
Kodiologist: And this one.

hy/macros.py
```diff
@@ -48,88 +35,144 @@ def _(fn):
...
+    -------
+    out: None
+    """
+    if not inspect.ismodule(source_module):
+        source_module = importlib.import_module(source_module)
+
+    source_macros = source_module.__dict__.setdefault('__macros__', {})
+    source_tags = source_module.__dict__.setdefault('__tags__', {})
+
+    if len(source_module.__macros__) + len(source_module.__tags__) == 0:
```
Kodiologist: I think you could write this as `not source_module.__macros__  and not source_module.__tags__`. But actually, I don't see the point of this check. Shouldn't requiring a module with no macros just not pull in any macros? Or would that have some unexpected consequence?

hy/macros.py
```diff
@@ -48,88 +35,144 @@ def _(fn):
...
-        importlib.import_module(module)
+    module_macros = module.__dict__.setdefault('__macros__', {})
+    module_tags = module.__dict__.setdefault('__tags__', {})
+
+    for builtin_mod_name in builtin_macros:
+        builtin_mod = importlib.import_module(builtin_mod_name)
+        # module.__dict__[module_name] = mod
+
+        # Make sure we don't overwrite macros in the module.
+        # TODO: Is this the desired behavior?
```
Kodiologist: More possibly forgotten comments.

hy/macros.py
```diff
@@ -152,49 +195,104 @@ def empty_fn(*args, **kwargs):
...
+    Returns
+    ------
+    out: HyObject
+        Returns a mutated tree with macros expanded.
     """
-    load_macros(compiler.module_name)
+    if not inspect.ismodule(module):
+        module = importlib.import_module(module)
+
+    assert True if compiler is None else compiler.module == module
```
Kodiologist: This could be written `assert not compiler or compiler.module == module`.

Kodiologist:
What's the purpose of the `include_require` option? Would it hurt to effectively always have it on?

hy/macros.py
brandonwillard: Perhaps it's reasonable to silently allow a `(require module)` when `module` has no macros or tags, but I'm on the fence about that.  Regardless, that check short-circuits the steps that follow and provides a much more precise error when `assignments` are not just `ALL`.

hy/compiler.py
```diff
@@ -1150,9 +1168,50 @@ def compile_import_or_require(self, expr, root, entries):
...
+                                                              asty.Str(expr, s=ast_str(v))],
+                                                        ctx=ast.Load())
+                                             for k, v in assignments]
+                        names = asty.List(expr,
+                                          elts=ast_assigns_items,
+                                          ctx=ast.Load())
+                    else:
+                        names = asty.Str(expr, s=assignments)
+
+                    ast_prefix = ast_str(prefix) if prefix else prefix
```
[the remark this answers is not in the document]
brandonwillard: These change requests from the `... if ... else ...` syntax to `... and/or ...` are rather frequent; is there a Python convention/style guide I've missed, or should we establish one?
I like the latter more, but, for some reason, I find the former more conventional (maybe even "Pythonic").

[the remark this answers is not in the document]
Kodiologist: I'm an unconventional sort of fellow, and I don't think PEP 8 says anything one way or the other about this anyway, but the way I figure, there's no reason to use `if` for expressions that can be written more directly with `and` or `or`.

brandonwillard:
Ah, yeah, I wasn't sure about `include_require` either (both its name and purpose).  It's effectively being used to designate when the compiler should be run "interactively" (i.e. compiling code to be evaluated in the same "session") or "non-interactively" code for use arbitrary use by Python package/module `Loader`s (e.g. the generation of bytecode).
If anything, it currently just avoids an unnecessary `require` in the first/"interactive" case.  When the resulting Python AST is compiled to bytecode and run, an AST-generated `require` is likely redundant.  
It might not cause issues if always included in the AST output (and probably shouldn't), but, since a second `require` currently doesn't occur in Hy, I played it safe and had it skip AST generation.

Now, I just came across a situation that could change all that.  I'm currently making the example in #1211 work, and I think the answer involves an AST-generated `require` for the "interactive" case.  Commit coming soon...

brandonwillard:
The most recent commit produces `require` AST in all cases and does not raise an exception when a module contains no macros/tags.  I think the latter should be changed, since it will surely confuse people, but something needs to be done about the situation demonstrated by the new test.

In that test, the macro [`foo` is expanded](https://github.com/brandonwillard/hy/blob/macro-changes/tests/resources/bin/circular_macro_require.hy#L8) at any time when AST and/or a Python code object is desired (e.g. a simple `import` or use of `runpy`).  During that expansion the [`require` in `foo`](https://github.com/brandonwillard/hy/blob/macro-changes/tests/resources/bin/circular_macro_require.hy#L8) is evaluated, and&mdash;in this test&mdash;for the first time, so the module `circular_macro_require` exists in `sys.modules` but does not have any defined macros (i.e. in `__macros__`).  Even though we've already compiled the forms defining the macros, since the resulting AST hasn't been evaluated yet, they effectively do not exist, and the aforementioned `require` *should* fail.

The current solution is hackish, because it simply removes the error (that should actually be there), allowing the `require` to silently fail, and then leaves the rest for runtime.  The assumption is that we're fine pushing this potential error to runtime, and&mdash;in this particular case&mdash;it isn't an issue, so no worries.

Under that assumption, we might as well have every macro expansion-processed `require` fail silently.  That would would push macro issues downstream (e.g. during an expansion that uses a missing macro) and allow us to re-introduce the removed exception, but it seems like poor design.

It would be better if we could collect macros as we compiled them.  For the case in which a module uses/`require`s its own macros, that would avoid a self import, which&mdash;during runtime&mdash;produces double output (e.g. the tests prints `42` twice: once for the `require` and its import and again from the command line run).  However, collecting macros during compilation would appear to require sequential compile-and-eval (or `defmacro` special forms handled by the compiler).

Kodiologist:
> The most recent commit produces require AST in all cases and does not raise an exception when a module contains no macros/tags. I think the latter should be changed, since it will surely confuse people, but something needs to be done about the situation demonstrated by the new test.

The reason I would expect `(require foo)` to work when `foo` has no macros is because `(import foo)` works even when `foo` is empty. `require` doesn't give you a macro or something named `foo` like `import` does, but that never happens for macros; even if `foo` had a macro `bar`, `(require foo)` would create only `foo.bar`, not `foo`.

That said, it's hard to imagine a situation where you'd want to require a module with no macros, so it shouldn't matter much.

The behavior of the new `circular_macro_require.hy` is definitely weird. You say that "Even though we've already compiled the forms defining the macros, since the resulting AST hasn't been evaluated yet, they effectively do not exist…", but 

    (defmacro bar [expr]
      `(print ~expr))

    (bar 42)

works, and has worked on master for a long time. The reason it works is that `defmacro` uses `eval-and-compile` internally, so it gets evaluated immediately after the form is compiled as well as at run-time.

So, we already collect macros at compile-time. All we should need to do to make `circular_macro_require` work is to make sure macros are immediately added to their parent module after picking their definitions.

Or am I missing something?

brandonwillard:
> The reason I would expect (require foo) to work when foo has no macros is because (import foo) works even when foo is empty. 

Yeah, that's why it's more reasonable to not err for `(require foo)`, but `from foo import bar` will fail in Python and probably should when handling the analogous `(require [foo [bar]])`.  I'm changing this logic now.

Kodiologist:
Right on. We definitely want an error when trying to require by name a macro that doesn't exist.

brandonwillard:
Just put in a better solution to this specific `run[p|h]y` issue: we now check&mdash;in a module-name independent way&mdash;whether or not we're importing the same module we're compiling, and, if we are, we can safely skip that `require` and let downstream `macroexpand`s find the macros (introduced by `defmacro`'s `eval-and-compile`) in the module being compiled.

Kodiologist:
`circular_macro_require.hy` is still printing `42` twice, though, so is this the method you're settling on?

Kodiologist:
I think so; it's more straightforward to implement it like that, right?

brandonwillard:
All right, just pushed a fix that correctly avoids duplicate output due to self-`require`s.

tests/native_tests/native_macros.hy
Kodiologist: Why's this marked resolved? The test is still there.

Kodiologist:
Very good. Take a look at my older comments and add something to NEWS and I think we're good to go.

Kodiologist:
You may need to click "Load more…" above to show some comments that GitHub hid.

hy/compiler.py
```diff
@@ -1517,31 +1577,28 @@ def _rewire_init(self, expr):
...
     @special(["eval-and-compile", "eval-when-compile"], [many(FORM)])
     def compile_eval_and_compile(self, expr, root, body):
         new_expr = HyExpression([HySymbol("do").replace(expr[0])]).replace(expr)
-        if self.module_name not in self._namespaces:
-            # Initialize a compile-time namespace for this module.
-            self._namespaces[self.module_name] = {
-                'hy': hy, '__name__': self.module_name}
+
+        # TODO: This is going to create another distinct HyCompiler instance.
+        # Why do this?  Is this why the `_stdlibs` dict was global?
```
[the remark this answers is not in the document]
brandonwillard: Yeah, but I think I'm gonna put in another refactoring PR that will take care of that point.

hy/compiler.py
```diff
@@ -559,9 +580,11 @@ def compile_unpack_iterable(self, expr, root, arg):
...
 
     @special([(not PY3, "exec*")], [FORM, maybe(FORM), maybe(FORM)])
-    # Under Python 3, `exec` is a function rather than a statement type, so Hy
-    # doesn't need a special form for it.
     def compile_exec(self, expr, root, body, globals_, locals_):
+        """
+        Under Python 3, `exec` is a function rather than a statement type,
+        so Hy doesn't need a special form for it.
+        """
         return asty.Exec(
```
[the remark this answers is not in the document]
Kodiologist: Why's this conversation marked resolved? Do you feel that the deleted comment is unhelpful? I added it originally because I figured it wasn't obvious why the method was gated to `not PY3`. But whether we end up removing the comment or not, removing it seems beside the point of this commit.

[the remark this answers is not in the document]
brandonwillard: Sorry, I think I've just been confused about your request: you want it there in its original form, and not removed?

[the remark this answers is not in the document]
Kodiologist: Right. Sorry, I was sloppy.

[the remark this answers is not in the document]
brandonwillard: Ohh, looks like they just added this "insert a suggestion" feature (literally just popped up as I wrote the previous comment).  If it shows up for you, let's try it!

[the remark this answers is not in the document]
Kodiologist: "Applying suggestions on deleted lines is not supported." :(

Kodiologist:
Hey, my dude, sorry for the incredibly long and slow review process. I haven't forgotten about this. I'm going through the PR and finding more concise ways to do things, spinning off unrelated changes, etc. The end result should be substantively the same, but I'll be sure to let you take a look before merging.

Kodiologist:
Okay, take a look. The biggest change is that when the compiler needs to add a call to the `require` function (now called `requiref`) to the AST, it uses a recursive `self.compile` call instead of building the call out of raw AST nodes. Hy trees are much more concise than AST nodes.

Most of the changes occur in one giant commit now, which isn't great, but there's no longer so much of commits undoing previous commits in the same PR.

Kodiologist:
I think your comments are supposed to be attached to some lines, instead of being free-floating comments, but GitHub, in its infinite wisdom, doesn't seem to be showing the lines. Anyway, hopefully these remarks answer your questions:

"Integrate hy.inspect into hy._compat" solves the same problem that you solved by adding `from __future__ import absolute_import`, because you needed to use the real `inspect` module in places where we were using `hy.inspect`. I'm not against `absolute_import` by any means, but we shouldn't have caused this problem in the first place.

"Rename hy.macros.require to hy.macros.requiref" is to avoid `self.compile(HyExpression([HySymbol('require'), ...]))` being treated as a special form (and hence leading to an infinite loop) instead of a function call.

"Remove an unused function `get_arity`" is separate from "Integrate hy.inspect into hy._compat" because the latter just moves some code whereas the former actually deletes something.

Kodiologist:
> "Rename hy.macros.require to hy.macros.requiref" is to avoid `self.compile(HyExpression([HySymbol('require'), ...]))` being treated as a special form (and hence leading to an infinite loop) instead of a function call.

That's the collision I'm referring to, and it didn't affect your implementation that used real AST nodes, just my implementation that uses `self.compile` on a Hy tree.

Kodiologist:
> Did you remove the `absolute_import`?

Yes, because it isn't necessary now, although it doesn't hurt.

> If that solves the problem, then why are those commits needed?

I think that having those functions in their own file, and giving that file the same name as a standard module, was a mistake in the first place. That is, this is a change worth making even if not strictly necessary to get the other parts of this PR to work. It's better to fix our mistakes than work around them.

> Besides being technically more efficient

You mean faster? It's hard for me to imagine a situation in which this change would make a noticeable speed difference (and if so, that would probably point to deeper speed problems regarding starting and stopping the compiler).

> its output/result is also exceedingly more transparent.

True, in the sense that you don't need to look elsewhere in the compiler, but I'd rather have the logic for compiling a function call in only one place, even though this means you need to scroll down to `@builds(HyExpression)` to see how it works. In theory, you could argue that my version makes the code more robust to changes in how the AST represents function calls, but such changes are unlikely. It's really about conciseness.

> Also, which commits were undoing other commits?

I don't mean to say that there were commits that completely undid a previous commit, but there were a lot of changes that overwrote previous changes, such as an `include_require` parameter for the compiler being added in one commit and removed in a later one.

brandonwillard:
> You mean faster? It's hard for me to imagine a situation in which this change would make a noticeable speed difference (and if so, that would probably point to deeper speed problems regarding starting and stopping the compiler).

Also in terms of memory and debugging.  Having to wade through more function calls and stack traces to find out how and where AST was generated is already a difficulty worth mitigating.  Plus, unnecessarily sending the interpreter through other parts of the compiler only increases the chances that it will err somewhere unrelated to the actual problem.

> True, in the sense that you don't need to look elsewhere in the compiler, but I'd rather have the logic for compiling a function call in only one place, even though this means you need to scroll down to `@builds(HyExpression)` to see how it works.

If AST generation occurred exclusively within `compile_expression`, then sure, but it doesn't.  Furthermore, what's happening in `compile_expression` isn't trivial and involves other non-trivial, yet irrelevant, aspects of the compiler (e.g. the entries in `_special_form_compilers`, the loops in `_compile_collect` and `macroexpand`).  Speaking of `macroexpand`, will a user-defined `requiref` interfere with compilation of `require`s? 


I can understand the desire to handle all `ast.Call`-generating statements in one place, but&mdash;at some point&mdash;this approach starts to negatively conflate two distinct operations: *the general act of generating Python AST* (as ultimately emitted by the compiler), and *the generation of AST for Hy trees* (which involves language-specific logic affecting Python AST generation).  
The two overlap (well, one simply depends on the other) but they also account for different things and incur different costs.  In trying to make one function account for more than one operation, it necessarily ends up sharing some of the costs between all of them.  
For instance, such an approach requires special logic to avoid "collisions" (e.g. a user-defined function, macro or built-in special form should not interfere with the compiler's internal AST generation).  Similarly, performance costs are shared unless efforts are made to distinguish between the two cases (in a sense un-doing the whole approach).  
In other words, when the compiler simply needs to generate Python AST for itself, why should it have to take exclusively Hy language-related steps?


Instead of having the *compiler* generate and immediately compile a Hy tree for a `require`, why not turn `require` into a built-in/core macro?  This approach would re-use the same compiler logic *and* keep the compiler complexity from unnecessarily growing.


Anyway, these changes don't appear to be directly related to the functionality addressed by this PR.  Are they addressing a requirement of some sort, or simply design choices/preferences?

Kodiologist:
I see your point that in various senses the compiler is doing more work than it needs to, but overall, this approach seems to suffice. The various abstractions we've already implemented appear to suffice to make everything work. You could try to factor the function-call-building logic out of `compile_expression` so it can be used here without going through the whole compiler, but I doubt this would lead to meaningful speed or RAM improvements or head off any bugs. Besides, the technique of recursive compilation is more general and can also be used for much more complex cases such as `compile_while_expression` when `cond_compiled.stmts` is true. It's worth ensuring that recursive compilation works.

> Speaking of macroexpand, will a user-defined `requiref` interfere with compilation of `require`s?

I think so. In your implementation, this would happen for functions named `require` instead of `requiref`. This is a general problem related to Hy's auto-imports which will probably require (a) more mangling or prefixes, or (b) shifting to the user the burden of managing most things that we currently import automatically. I don't recommend trying to address it in this PR.

> this approach starts to negatively conflate two distinct operations: the general act of generating Python AST (as ultimately emitted by the compiler), and the generation of AST for Hy trees (which involves language-specific logic affecting Python AST generation).

Seeing as this is the Hy compiler we're talking about, I doubt we'd buy much by enforcing a distinction between those things.

> Instead of having the compiler generate and immediately compile a Hy tree for a `require`, why not turn `require` into a built-in/core macro?

Translating special forms into macros, when possible, is almost always a good idea. If you're interested and you think it can be done, go for it.

> Anyway, these changes don't appear to be directly related to the functionality addressed by this PR.

If by "these changes" you mean my use of `ret += self.compile(HyExpression([HySymbol('requiref'), ...]))`, then it has to do with the functionality addressed by this PR because this PR introduced the idea of adding a function call to `require`/`requiref` as part of the AST. Before your work in this PR, `require` didn't compile to anything.

Kodiologist:
> I do not see how that could happen in the original implementation, and, having just tried it, it doesn't.

Using the REPL makes the automatic import happen later, so the issue doesn't appear in your example. Here's how you can see it with a script file:

    (defn require [x] (print "wrong require!"))
    (require [hy.contrib.walk [let]])
    (let [x 1] (print x))

Running that yields `TypeError: require() got an unexpected keyword argument 'assignments'`.

> No, I mean the `inspect` and `requiref`-related changes. I don't understand what they're specifically addressing that wasn't already addressed by the original commits in this PR.

They're addressing the same things. I didn't add or remove features from your PR; I just changed the implementation.

> fixing the indentation is definitely warranted and not a superfluous style change.

You're free to futz with indentation if you like, but in PRs this large, you should do in a separate commit. Otherwise, the substantive change (adding `__macros__`) gets buried in the diff for a style change (reindenting).

> In contrast, unnecessarily long, compound one-liners provide nothing noticeable, yet easily hinder readability and debuggability.

I don't know what to say except that I don't see how breaking a large nested expression into several assignments aids readability. It's an extra step to the same destination. It creates a variable that leads the reader into thinking that the saved value will be used more than once, and hence the reader needs to keep another variable in mind while reading the rest of the code in scope.

To be clear, adding or removing intermediate variables shouldn't make a meaningful performance difference, since Python passes everything by reference anyway.

> Then there's this reversion that puts _stdlib back into module scope

I think your idea to make `_stdlib` an instance variable makes perfect sense, but if you want it in this PR, put it in a separate commit.

> What's the reason for these test changes?

There's no special reason for the order of the tests; that was probably just how they ended up while I was shuffling other changes. Presuming the "more informative explanation" was a comment I removed, that sounds like a mistake on my part.

Kodiologist:
> In contrast, unnecessarily long, compound one-liners provide nothing noticeable, yet easily hinder readability and debuggability.

I forgot to address the "debuggability" part. I admit that when I'm debugging or profiling a line, I take my existing code at that line and break it into a separate assignment for every intermediate value. This is very verbose, though, so I'd recommend keeping such a change temporary.

brandonwillard:
> Using the REPL makes the automatic import happen later, so the issue doesn't appear in your example. 

I just made an update that uses the fully qualified name for the `require` function&mdash;among other simplifications&mdash;and now it doesn't seem to be affected by your example.

There's still the matter involving macro shadowing.  I don't know of a simple work-around in the current state, but I imagine we could add a separate `require` function to `core/language.hy` (or change the existing one) so that it too can be called "interactively" via its fully qualified name&mdash;or simply re-imported.

This brings up a bigger question: why do we have `import` and `require` logic in the compiler?  It looks like `compile_import_or_require` could be removed and replaced by a couple Hy "builtins" that parse the list and keyword arguments, no?

Kodiologist:
> I just made an update that uses the fully qualified name for the `require` function

Oh, well done; I should've thought of that. In theory, there's an issue of shadowing the name `hy`, but it's reasonable to expect the user not to do that.

I would suggest my old name for the commit "Integrate hy.inspect into hy._compat" rather than "Remove hy.inspect" because the new name suggests all the code was actually deleted, instead of part of it being moved and part of it being deleted. Also, be sure to separate the deletion of `get_arity` into its own commit. Putting both the move and deletion into one commit makes it not so obvious that one function doesn't end up surviving the move. But in retrospect, I moved first and deleted second, when it probably makes more sense to delete first and move what's left over second.

> There's still the matter involving macro shadowing. I don't know of a simple work-around in the current state, but I imagine we could add a separate `require` function to `core/language.hy` (or change the existing one) so that it too can be called "interactively" via its fully qualified name—or simply re-imported.

I don't understand how that would help. A user-defined macro can still shadow a core function.

> This brings up a bigger question: why do we have `import` and `require` logic in the compiler?

`import`, at least, needs to generate AST nodes that no other special forms generate, so it needs to be a special form (or implemented in terms of a special form that creates those AST nodes). `require` seems more readily doable as a macro, but its unique compiler logic at this point is very simple, anyway; it calls `macro.require` and produces a function call that's compiled instead. Its other logic is parsing, which it shares with `import`.

brandonwillard:
> import, at least, needs to generate AST nodes that no other special forms generate, so it needs to be a special form (or implemented in terms of a special form that creates those AST nodes).

You're right; it's the shared parsing logic that I was thinking of, but the AST node absolutely sets it apart.

> require seems more readily doable as a macro, but its unique compiler logic at this point is very simple, anyway; it calls macro.require and produces a function call that's compiled instead. Its other logic is parsing, which it shares with import.

I suppose it's that shared logic that would be factored-out, and `compile_import_or_require` converted to a simple `compile_import` that handles just that AST node.

Kodiologist:
That sounds reasonable.

One idea I've thought about a little but not explored is making a system so that Hy macros can produce AST nodes instead of just more Hy code, and using this to replace a lot of the compiler with macros. But a lot of the special forms provide very basic features, like `setv`, that you'd need to write those macros, so this would lead down the rabbit hole to self-hosting, which I suspect is way more trouble than it's worth.

brandonwillard:
> I would suggest my old name for the commit "Integrate hy.inspect into hy._compat" rather than "Remove hy.inspect" because the new name suggests all the code was actually deleted, instead of part of it being moved and part of it being deleted.

The content of `hy.inspect` was only being used in `hy.macros`, so I turned the relevant code into helper functions specific to `hy.macros` and kept the removal of `hy.inspect`.  Are there external API expectations (e.g. Hy-dependent packages) that should drive us to more granularly track these changes and/or keep them in `hy._compat`?

Kodiologist:
Ah, whoops, I didn't notice you'd changed the target module. That's fine, since the functions are only used there. So make that "Integrate hy.inspect into hy.macros" instead.

>  Are there external API expectations (e.g. Hy-dependent packages) that should drive us to more granularly track these changes and/or keep them in `hy._compat`?

No, it's just a matter of making the history readily comprehensible.

brandonwillard:
> One idea I've thought about a little but not explored is making a system so that Hy macros can produce AST nodes instead of just more Hy code, and using this to replace a lot of the compiler with macros. But a lot of the special forms provide very basic features, like setv, that you'd need to write those macros, so this would lead down the rabbit hole to self-hosting, which I suspect is way more trouble than it's worth.

There has to be a minimal sufficiency spec for a related Lisp we can borrow (e.g. Clojure?).  

Ideally, the compiler would be super simple and highly efficient; otherwise, if it's going to do anything fancy, it should probably handle higher-level&mdash;and perhaps more Lisp-oriented&mdash;compiler optimizations (e.g. tail-call optimization).

Speaking of tail-call optimization, why does it currently exist as an optional feature and not a more general compiler optimization?

Kodiologist:
There have been several attempts. I think #1682 is the latest.

brandonwillard:
@Kodiologist, anything else needed for this one?

Kodiologist:
Sorry, I'd assumed you were working on writing `require` as a macro and were either going to change this PR a lot or replace it. I'll look at it again tomorrow.

brandonwillard:
Turning `require` into a macro would be cool, but I would rather get these fixes through first.  I have a few other things in the pipeline that would benefit greatly from the way macros and `require` work in this PR.

Kodiologist:
It looks like you've undone the separation of deleting `get_args` and moving the other functions. Is there a reason you want to keep those changes in one commit?

brandonwillard:
An explanation for those changes is given [here](https://github.com/hylang/hy/pull/1682#issuecomment-432830317).  Otherwise, `get_arity` doesn't warrant any special distinction (it isn't even used) and the changes are easily summarized by only a single commit.

More importantly, why would we want to be so granular?  Is there something special about `get_arity` that I'm not seeing?  Is there a more general convention behind this: e.g. we want separate commits for each addition and removal of a function?

Kodiologist:
Putting the functions in `hy.macros` instead of `hy._compat` is fine.

> More importantly, why would we want to be so granular?

Because the two things being accomplished in the combined commit are independent. One is deleting dead code. Another is moving code from one place to another, without changing anything else, except what's necessary to support the move. With each change as its own commit, it's clear what's going on. On the other hand, putting both the move and deletion into one commit makes it not so obvious that one function doesn't end up surviving the move.

This is the same reason why indentation changes to a large body of code should in a separate commit from functional changes to a small portion of that code. The changes are independent, and the reader will know to look at the functional-change commit much more closely than the indentation commit.

brandonwillard:
> Because the two things being accomplished in the combined commit are independent.

`get_arity`'s removal occurs as a direct result of `hy.inspect`'s removal (i.e. the focus of the commit), so those two are not independent.  Likewise, removal of `hy.inspect` necessitates the relocation of its functions used within `hy.macros` (to `hy.macros`), so those are also not independent.  
As a result, all the actions comprising the commit are clearly centered around the removal of `hy.inspect` and perfectly suited for one commit on that topic.

Kodiologist:
> removal of `hy.inspect` necessitates the relocation of its functions used within `hy.macros` (to `hy.macros`), so those are also not independent.

I'm with you there, to be clear.

> `get_arity`'s removal occurs as a direct result of `hy.inspect`'s removal (i.e. the focus of the commit), so those two are not independent.

I don't see this. You could as easily delete `get_arity` in a separate PR, which could be merged before or after any PR in which the other functions are moved, and all the individual changes would still make sense. That seems like independence par excellence.

The only reason that I deleted `get_arity` here, in the same PR as I was moving those functions, is that I noticed `get_arity` was never actually used while I was moving the functions.

brandonwillard:
> I don't see this. You could as easily delete `get_arity` in a separate PR, which could be merged before or after any PR in which the other functions are moved, and all the individual changes would still make sense. That seems like independence par excellence.

Because you *could* delete `get_arity` independent of `hy.inspect` (e.g. in a separate PR) does not negate the fact that deleting `hy.inspect` necessarily deletes `get_arity`.  Whether or not the changes *can* be made independently is not relevant, since, for that matter, every character change could justify its own commit on such "independence".

> The only reason that I deleted get_arity here, in the same PR as I was moving those functions, is that I noticed `get_arity` was never actually used while I was moving the functions.

I don't see how its irrelevance is a reason for an unnecessary commit.  The opposite seems much more appropriate: because the function isn't used, there's no reason to emphasise its removal in the Git commit history!

Also, decoupling `get_arity`'s removal from the rest of the changes serves no apparent functional purpose.  If, for instance, one wanted to revert the change and re-introduce `get_arity` in isolation, a separate commit that first removes or relocates it from `hy.inspect` wouldn't be helpful, since the former would require the reversion of `hy.inspect`'s removal&mdash;and further justify their coupling&mdash;and the latter would require another superfluous commit that subsequently deletes `get_arity` after relocating it.

If the purpose of the separate commit is simply logging, and nothing particularly functional, then it's best to mention the removal explicitly in the commit message.  Still, I don't see why `get_arity` deserves this much attention.

Kodiologist:
> deleting `hy.inspect` necessarily deletes `get_arity`

On the contrary, you could keep the function while getting rid of its original file, as I did originally.

> Whether or not the changes can be made independently is not relevant, since, for that matter, every character change could justify its own commit on such "independence".

I don't see that, either. Suppose you spun off the first change of a single character in this PR into its own commit, making it the new first commit of the PR. I'd ask you to squash it into the following commit, because on its own, the one-character change would make no sense. The change only makes sense as part of the larger change in the commit it was split off from.

> I don't see how its irrelevance is a reason for an unnecessary commit.

Commits exist to group related changes together. If two changes are irrelevant to each other, then in general, they should be in different commits.

> Also, decoupling `get_arity`'s removal from the rest of the changes serves no apparent functional purpose. 

The purpose is not function but the legibility of history. Many sequences of commits can lead to the same final code, and are hence functionally identical, but some are more intelligible than others.

> If the purpose of the separate commit is simply logging, and nothing particularly functional, then it's best to mention the removal explicitly in the commit message.

It's better to indicate a sequence of changes in the code itself, with a sequence of commits, than in prose, with the commit message.

brandonwillard:
Just to be clear, you're arguing for situation A over B, no?

For a repository with the following file in its index:
```clojure
;; some_file.hy

(defn func-1 [] 
  (print 1))

(defn func-2 [] 
  (print 2))

(defn func-3 [] 
  (print 3))
```

# Situation A
## Commit 1
```diff
--- a/some_file.hy
+++ b/some_file.hy
@@ -1,8 +1,5 @@
 ;; some_file.hy
 
-(defn func-1 [] 
-  (print 1))
-
```
## Commit 2
```diff
--- a/some_file.hy
+++ b/some_file.hy
@@ -1,7 +1,4 @@
 ;; some_file.hy
 
-(defn func-2 [] 
-  (print 2))
-
```
## Commit 3
```diff
deleted file mode 100644
index c2ff364..0000000
--- a/some_file.hy
+++ /dev/null
@@ -1,4 +0,0 @@
-;; some_file.hy
-
-(defn func-3 [] 
-  (print 3))
```

# Situation B
## Commit 1
```diff
deleted file mode 100644
index 14ce434..0000000
--- a/some_file.hy
+++ /dev/null
@@ -1,10 +0,0 @@
-;; some_file.hy
-
-(defn func-1 [] 
-  (print 1))
-
-(defn func-2 [] 
-  (print 2))
-
-(defn func-3 [] 
-  (print 3))
```

Kodiologist:
Maybe in some cases? I guess it depends on the actual situation. If an entire file is dead code, for example, then I could certainly imagine it making sense to delete the whole file in one commit. #1189 is an example I did.

Kodiologist:
Similarly, in this case, if we were just deleting `hy.inspect` instead of moving some of its code, it would presumably make sense to delete all three functions in the same commit.

brandonwillard:
All right, I may not understand it, but I've recreated [that squashed commit](https://github.com/hylang/hy/pull/1682/commits/701db83ba942a541d50973a51e4363a1e2006e61) and added it back.

Kodiologist:
I guess the core problem is that Git doesn't have a notion of moving code. It can only track additions and deletions. So when you move code, make sure that in the commit where moving happens, **only** moving happens. When you want to make less superficial changes to that moved code (which is common, because you probably moved it because you're making related changes), do it in another commit, whether before or after the move.

Kodiologist:
Woohoo!

PR #1683, opened 2018-09-30
Fix `sys.modules` for failed imports in Python 2.7

brandonwillard:
Newly imported modules with compile and/or run-time errors were not being removed from `sys.modules`.  This commit modifies the Python 2.7 loader so that it follows Python's failed-initial-import logic and fixes the issue.


Kodiologist:
Whoops, I forgot to merge this.

PR #1685, opened 2018-10-07
Allow punning keyword arguments

### LINKED ISSUE #1119: [feature] Shorthand for `:kwarg True`, `:kwarg False`, and `:kwarg kwarg`
It's common to want to set a keyword argument to a Boolean, and also common to want to set a keyword argument to the value of a variable with the same name. I propose the following syntactic sugar for keywords when they're used for keyword arguments:
- Translate `:+foo` to `:foo True`
- Translate `:!foo` to `:foo False`
- Translate `:^foo` to `:foo foo`

(One might also be tempted to add a construct for `:foo None`, but this seems like an unusual enough case that having Hy programmers learn another piece of shorthand isn't worth it.)

I included the first two translations (`:+foo`) and (`:!foo`) in [my own implementation of keyword arguments for Hy](https://github.com/Kodiologist/Kodhy/blob/88631400ce/kodhy/macros.hy#L32), which I made because Hy didn't have that feature built-in yet. I've found this sugar handy and use it frequently in practice. It's a bit like `--flag` and `--no-flag` in command-line arguments.

 I haven't previously used the last translation (`:^foo`). It's similar to a GHC extension to Haskell called ["record puns"](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#record-puns).

Kodiologist:
Closes #1119. (The other changes I suggested there are probably overkill.)

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -12,6 +12,9 @@ New Features
 * Keyword objects (not just literal keywords) can be called, as
   shorthand for `(get obj :key)`, and they accept a default value
   as a second argument.
+* There is now syntactic sugar for setting a keyword argument to
+  a variable of the same name: `(f :^arg)` is equivalent to
+  `(f :arg arg)`.
 
 Bug Fixes
 ------------------------------
--- a/docs/language/syntax.rst
+++ b/docs/language/syntax.rst
@@ -83,10 +83,20 @@ the error ``Keyword argument :foo needs a value``. To avoid this, you can quote
 the keyword, as in ``(f ':foo)``, or use it as the value of another keyword
 argument, as in ``(f :arg :foo)``.
 
-Keywords can be called like functions as shorthand for ``get``. ``(:foo obj)``
-is equivalent to ``(get obj :foo)``. An optional ``default`` argument is also
-allowed: ``(:foo obj 2)`` or ``(:foo obj :default 2)`` returns ``2`` if ``(get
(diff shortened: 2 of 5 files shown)


gilch:
I'm opposed to this approach. The core syntax of Hy must be kept very simple. Each additional special case makes macro programming difficult. We should be trying to make it simpler, not adding more burdensome features.

We can accomplish your desired syntax with a tag macro. There's no need to alter the compiler.
```Hy
(deftag ^ [form]
  `(~@(chain #*(gfor f form
                     (if (and (keyword? f)
                              (.startswith (name f) '^))
                         [(HyKeyword (cut (name f) 1))
                          (HySymbol (cut (name f) 1))]
                         [f])))))
```
For example,
```
=> (setv sep "::")
sep = '::'
None

=> #^(print 1 2 3 :^sep :end "$\n")
print(1, 2, 3, sep=sep, end='$\n')

1::2::3$
```

Kodiologist:
It's three additional lines in the compiler. Four if you count the comment explaining what it is. I'm sure we'll manage.

gilch:
It's not the additional lines to the compiler that makes it a problem. It's the extra lines required by all the macros that now have to handle a new special case that they wouldn't have had to before.

Kodiologist:
Okay, I'll redo it as a macro.

PR #1686, opened 2018-10-08
Add a macro for punning keyword arguments

### LINKED ISSUE #1119: [feature] Shorthand for `:kwarg True`, `:kwarg False`, and `:kwarg kwarg`

Kodiologist:
Closes #1119.

Replaces  #1685.

### DIFF
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -6,7 +6,7 @@
 ;;; These macros form the hy language
 ;;; They are automatically required in every module, except inside hy.core
 
-(import [hy.models [HyList HySymbol]])
+(import [hy.models [HyList HySymbol HyKeyword HyExpression]])
 
 (defmacro as-> [head name &rest rest]
   "Beginning with `head`, expand a sequence of assignments `rest` to `name`.
@@ -276,3 +276,10 @@ Such 'o!' params are available within `body` as the equivalent 'g!' symbol."
                 _hy_tag
                 [None]
                 ['~symbol])))))
+
+
+(defmacro pun [&rest body]
+  (HyExpression (reduce + (gfor x body
+    (if (and (keyword? x) (.startswith x.name "^"))
+      [(HyKeyword (cut x.name 1)) (HySymbol (cut x.name 1))]
+      [x])))))
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -12,6 +12,7 @@ New Features
 * Keyword objects (not just literal keywords) can be called, as
   shorthand for `(get obj :key)`, and they accept a default value
   as a second argument.
+* Added a new core macro `pun`.
 
 Bug Fixes
 ------------------------------
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -1325,6 +1325,20 @@ print
 .. note:: ``print`` always returns ``None``.
 
 
+pun
+---
+
+The ``pun`` macro provides a shorthand for setting a keyword argument equal to
+a variable of the same name. Any argument that's a literal keyword beginning
+with a caret (``^``) is replaced, such that ``:^foo`` becomes ``:foo foo``, and
+the result is an expression of all the arguments. For example, the following::
+
+    (pun dict :^monitor :^verbose :^cv-folds)
+
+expands to::
+
+    (dict  :monitor monitor  :verbose verbose  :cv-folds cv-folds)
+
 quasiquote
 ----------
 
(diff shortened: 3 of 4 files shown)


gilch:
Putting this in core seems premature. It's an experimental macro not even based on anything from Clojure or Common Lisp, that means it belongs in contrib. I don't imagine I'd use this enough to justify it, but if it proves popular we could then move it to extra.

hy/core/macros.hy
```diff
@@ -276,3 +276,10 @@ Such 'o!' params are available within `body` as the equivalent 'g!' symbol."
                 _hy_tag
                 [None]
                 ['~symbol])))))
+
+
+(defmacro pun [&rest body]
+  (HyExpression (reduce + (gfor x body
+    (if (and (keyword? x) (.startswith x.name "^"))
+      [(HyKeyword (cut x.name 1)) (HySymbol (cut x.name 1))]
+      [x])))))
```
gilch: This definition isn't indented properly. Make sure it passes parlinter.

Kodiologist: [I asked you to stop this.](https://github.com/hylang/hy/pull/1338#discussion_r131307803) [Twice.](https://github.com/hylang/hy/pull/1378#discussion_r132749045) [Then you insulted me about it.](https://github.com/hylang/hy/pull/1614#issuecomment-391211451) Please don't comment on my indentation again.

Kodiologist:
I wouldn't say it's experimental (it's a very simple macro), and the inclusion criteria for contrib and extra have nothing to do with Common Lisp or Clojure.

brandonwillard:
What's the inclusion criteria for `hy.core.macros`?

Kodiologist:
We never decided on any. In practice, it contains macros that are short and are of general use, and we should probably stick to that.

Kodiologist:
@gilch To be clear, are you vetoing this as-is?

Kodiologist:
It's clear that lots of time and people and thinking went into previous languages, but none of those things imply that the results are any good. People's decisions are subject to inertia and groupthink, and there's no group so large that anything important is guaranteed to be thought of by somebody in the group. How things have been done previously is not any indiciation as to what should be done.

> Are we sure we want a macro rather than a tag?

The distinction doesn't seem important. I could switch it to a tag macro if you'd prefer that.

> How do we work around it?

With mangling, as you just realized and as already tested. It's clumsy, but it's hard to see how the clumsiness will matter considering the sheer weirdness of beginning a function parameter's name with a caret, and the coincidence of wanting to use `pun` in the same call.

> Maybe `(pun spam :^foo :^bar :XhatXq q)` oh, but you've vetoed those mangle improvements #1651

To be clear, I'm not against mangling nicknames for ASCII characters, just the absence of a clear marker of whether a name has been mangled (which is currently the `hyx_` prefix).

> I don't know how well this would work in practice. And, emphatically, *neither do you*.

I'd anticipated this problem you thought of, but if you can think of any other potential problems with the macro, by all means mention them and I can fix them and add tests. It's appropriate to hold up a PR for known problems, but not for problems they might have but we haven't even thought of yet.

Kodiologist:
@gilch Have you thought of any potential problems?

Kodiologist:
@gilch So what's the word? Do you want it as a tag macro instead?

alphapapa:
Although I've known of Hy for a while, I haven't used it "in anger" much, so I'm still a tyro.  My perspective comes from using other Lisps, so for whatever it's worth:

Since `apply` was removed (which I miss, but can adapt to), the proposed syntax seems strange to me, since it would work roughly like `apply`:

```hy
(apply dict [:monitor monitor :verbose verbose :cv-folds cv-folds])
(pun dict :^monitor :^verbose :^cv-folds)
```

What if `pun` worked like the replacement for `apply`, i.e. using `#**`, something like:

```hy
(dict #** (pun monitor verbose cv-folds))
```

Issue #1510, opened 2018-02-15
Turn off syntax highlighting of hy expressions by default


vodik:
The pretty printing is nice, and having the basic syntax highlighting to emphasis opening and closing brackets for `HyExpression` and `HyList` is certainly cool.

But I'm looking at embedding `hy` support in some of my Python projects and the ANSI colour codes are leaking into error messages, making them less legible. This should be opt-in feature of the REPL, not a default.

Kodiologist:
See also #1429.

gilch:
See also #1360. There are ways to turn this off, but if it wasn't the default, many users would never notice the feature even exists. That was my rationale for making pretty opt-out instead of opt-in. We could also perhaps add a command-line switch or environment variable to turn this off.

vodik:
I think turning it on by default inside the REPL is great, but my concern is more so that it leaks outside the context and creates a something less readable. Hy objects will cross the Python boundry:

```python
>>> foo = hy.HyExpression([hy.HySymbol('foo')])
>>> repr(foo)
"\x1b[33m\x1b[22mHyExpression\x1b[39m\x1b[22m\x1b[33m\x1b[22m([\x1b[39m\x1b[22m\n  HySymbol('foo')\x1b[33m\x1b[22m])\x1b[39m\x1b[22m"
```
What happens if I'm logging exceptions with context and there's a Hy expression? Now there's noise I have to strip out:

Here's another example where it creates noise in pytest:

```
>       assert not hy.HyExpression([hy.HySymbol('foo')])
E       AssertionError: assert not HyExpression([\n  HySymbol('foo')])
E        +  where HyExpression([\n  HySymbol('foo')]) = <class 'hy.models.HyExpression'>([HySymbol('foo')])
E        +    where <class 'hy.models.HyExpression'> = <module 'hy' from '/home/simon/src/hy/hy/__init__.py'>.HyExpression
```
pytest correctly strips escape sequences, but the newline characters are literally printed and, imho, impede understanding.

And looking at #1360, @kirbyfan64 properly points out this: 

> Python reprs that are known to not round trip are conventionally contained in <>, for example:

But the escape sequences prevent round tripping from working.

gilch:
> I think turning it on by default inside the REPL is great

So, how do you want this to work?

[...]

Issue #1486, opened 2018-01-10
Ugly error message for bad `require`


minghu6:
When I require a macro not existed, It throws:
```
Traceback (most recent call last):
  File "/home/johnzhuang/.virtualenvs/hy/lib/python3.6/site-packages/hy/importer.py", line 180, in hy_eval
    _ast, expr = hy_compile(hytree, module_name, get_expr=True)
  File "/home/johnzhuang/.virtualenvs/hy/lib/python3.6/site-packages/hy/compiler.py", line 2213, in hy_compile
    result = compiler.compile(tree)
  File "/home/johnzhuang/.virtualenvs/hy/lib/python3.6/site-packages/hy/compiler.py", line 460, in compile
    ret = self.compile_atom(_type, tree)
  File "/home/johnzhuang/.virtualenvs/hy/lib/python3.6/site-packages/hy/compiler.py", line 447, in compile_atom
    else _compile_table[atom_type](self, atom))
  File "/home/johnzhuang/.virtualenvs/hy/lib/python3.6/site-packages/hy/compiler.py", line 1731, in compile_expression
    ret = self.compile_atom(fn, expression)
  File "/home/johnzhuang/.virtualenvs/hy/lib/python3.6/site-packages/hy/compiler.py", line 447, in compile_atom
    else _compile_table[atom_type](self, atom))
  File "/home/johnzhuang/.virtualenvs/hy/lib/python3.6/site-packages/hy/compiler.py", line 819, in compile_do
    return self._compile_branch(expression)
  File "/home/johnzhuang/.virtualenvs/hy/lib/python3.6/site-packages/hy/compiler.py", line 545, in _compile_branch
    return _branch(self.compile(expr) for expr in exprs)
  File "/home/johnzhuang/.virtualenvs/hy/lib/python3.6/site-packages/hy/compiler.py", line 318, in _branch
    results = list(results)
  File "/home/johnzhuang/.virtualenvs/hy/lib/python3.6/site-packages/hy/compiler.py", line 545, in <genexpr>
    return _branch(self.compile(expr) for expr in exprs)
  File "/home/johnzhuang/.virtualenvs/hy/lib/python3.6/site-packages/hy/compiler.py", line 460, in compile
    ret = self.compile_atom(_type, tree)
  File "/home/johnzhuang/.virtualenvs/hy/lib/python3.6/site-packages/hy/compiler.py", line 447, in compile_atom
    else _compile_table[atom_type](self, atom))
[...]

Issue #1429, opened 2017-09-19
Colored reprs don't play nicely with HyMacroExpansionError


Kodiologist:
=> (defmacro m [x] (raise (ValueError (repr x))))
     <function _hy_anon_var_1 at 0x7f44974d9510>
     => (m [1 2 3])
       File "<input>", line 1, column 1

       (m [1 2 3])
       ^---------^
     HyMacroExpansionError: b"expanding `m': ValueError('\\x1b[36m\\x1b[22mHyList\\x1b[39m\\x1b[22m\\x1b[36m\\x1b[22m([\\x1b[39m\\x1b[22m\\n  HyInteger(1)\\x1b[36m\\x1b[22m,\\x1b[39m\\x1b[22m\\n  HyInteger(2)\\x1b[36m\\x1b[22m,\\x1b[39m\\x1b[22m\\n  HyInteger(3)\\x1b[36m\\x1b[22m])\\x1b[39m\\x1b[22m',)"

gilch:
`(repr '[1 2 3])` is enough to show that, what does that have to do with `HyMacroExpansionError` in particular?
```
=> (repr '[1 2 3])
'\x1b[36m\x1b[22mHyList\x1b[39m\x1b[22m\x1b[36m\x1b[22m([\x1b[39m\x1b[22m\n  HyInteger(1)\x1b[36m\x1b[22m,\x1b[39m\x1b[22m\n  HyInteger(2)\x1b[36m\x1b[22m,\x1b[39m\x1b[22m\n  HyInteger(3)\x1b[36m\x1b[22m])\x1b[39m\x1b[22m'
=> (import [hy.models])
=> (with [(hy.models.pretty False)] (repr '[1 2 3]))
'HyList([HyInteger(1), HyInteger(2), HyInteger(3)])'
```

Kodiologist:
Typing `(repr '[1 2 3])` at the command line is requesting a `repr` of a `repr`, so that makes sense, but `ValueError` expects a string, so the additional layer of `repr` imposed by `HyMacroExpansionError` doesn't seem to be helping. Contrast with:

    => (raise (ValueError (repr '[1 2 3])))
    Traceback (most recent call last):
      File "/home/hippo/Desktop/hyenv/hy/hy/importer.py", line 198, in hy_eval
        eval(ast_compile(_ast, "<eval_body>", "exec"), namespace)
      File "<eval_body>", line 1, in <module>
    ValueError: HyList([
      HyInteger(1),
[...]

Issue #657, opened 2014-09-09
The lexer needs to be more informative


rcarmo:
I've just spent a long while trying to debug the following situation:
- `a.hy` had a single-quoted string (yeah, I know...)
- `b.hy` imported `a`

When trying to run `b.hy`, I got:

```
LexException: Could not identify the next token.
```

...without a single clue as to the file or line the error was actually in. So a little more information might save a lot of time...

philiparvidsson:
This needs to be fixed as it is still an issue. I did some manual "binary search" by deleting chunks of code in a source file until finding the issue, but that is not always viable either. How about at least outputting the file, row and column of the previous token!?

PR #1687, opened 2018-10-11
Generate more concise syntax and compiler errors

brandonwillard:
This PR refactors the exception/error classes and their handling, keeps Hy source strings and their corresponding file information (if any) closer to the origin of an exception (so that calling code isn't responsible for annotating exceptions), and provides minimally intrusive traceback print-out filtering via a context manager that temporarily alters `sys.excepthook` (this is automatically enabled for Hy command-line sessions).

New environment and package variables have been introduced, as well:
*  `HY_COLORED_ERRORS`, and package variable, `hy.errors._hy_colored_errors`, that enables/disables manual error coloring, and
* `HY_COLORED_AST_OBJECTS`, and package variable, `hy.models._hy_colored_ast_objects`, that enables/disables AST object coloring.

Also, source caching has been added to the Hy interpreter.  It's modeled after the approach used by IPython via `linecache`.  Now, for example, traces will show the relevant source lines for functions, objects, etc., defined exclusively in the REPL (see the example below).

- [x] #657
- [x] #1510
- [x] #1429
- [x] #1397
- [x] #1486 (the error message part)

# Notes

As mentioned above, the `sys.excepthook` provided here only hides internal frames from the traceback *print-out*, and not the actual traceback object, so, for instance, it does not affect the stack in `pdb`.

There are other, more powerful and direct approaches that involve altering and/or creating low-level traceback objects (e.g. [Jinja's](https://github.com/pallets/jinja/blob/master/jinja2/debug.py)).  These allow for a better debugging experience (e.g. without any compiler or parser frames), but they're also more complicated and distribution dependent.  In the long run, I think it's better to have custom tracebacks, so I'll keep an eye on good/better ways to do this; for example, [this approach](https://github.com/untitaker/shippai/blob/master/python/shippai/__init__.py) is interesting.

# Example

## Before
```clojure
=> (with [file (open "a.hy" "w")] (file.write "(print 'hi\""))
11
=> (import a)
  File "/home/bwillard/projects/code/python/hy-master/hy/importer.py", line 138, in hy_eval
    eval(ast_compile(_ast, "<eval_body>", "exec"), namespace)
  File "<eval_body>", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 674, in exec_module
  File "<frozen importlib._bootstrap_external>", line 781, in get_code
  File "/home/bwillard/projects/code/python/hy-master/hy/importer.py", line 228, in _hy_source_to_code
    hy_tree = hy_parse(source)
  File "/home/bwillard/projects/code/python/hy-master/hy/importer.py", line 65, in hy_parse
    return HyExpression([HySymbol("do")] + tokenize(source + "\n"))
  File "/home/bwillard/projects/code/python/hy-master/hy/lex/__init__.py", line 19, in tokenize
    return parser.parse(lexer.lex(buf))
  File "/home/bwillard/apps/anaconda3/envs/hy-master-cpython36/lib/python3.6/site-packages/rply/parser.py", line 23, in parse
    t, symstack, statestack, state
  File "/home/bwillard/apps/anaconda3/envs/hy-master-cpython36/lib/python3.6/site-packages/rply/parser.py", line 80, in _reduce_production
    value = p.func(targ)
  File "/home/bwillard/projects/code/python/hy-master/hy/lex/parser.py", line 223, in t_partial_string
    raise PrematureEndOfInput("Premature end of input")
hy.lex.exceptions.PrematureEndOfInput: LexException: Premature end of input

=> (require blah)
  File "/home/bwillard/projects/code/python/hy-master/hy/importer.py", line 120, in hy_eval
    _ast, expr = hy_compile(hytree, module_name, get_expr=True)
  File "/home/bwillard/projects/code/python/hy-master/hy/compiler.py", line 1716, in hy_compile
    result = compiler.compile(tree)
  File "/home/bwillard/projects/code/python/hy-master/hy/compiler.py", line 350, in compile
    ret = self.compile_atom(tree)
  File "/home/bwillard/projects/code/python/hy-master/hy/compiler.py", line 344, in compile_atom
    return Result() + _model_compilers[type(atom)](self, atom)
  File "/home/bwillard/projects/code/python/hy-master/hy/compiler.py", line 1584, in compile_expression
    self, expr, unmangle(sroot), *parse_tree)
  File "/home/bwillard/projects/code/python/hy-master/hy/compiler.py", line 573, in compile_do
    return self._compile_branch(body)
  File "/home/bwillard/projects/code/python/hy-master/hy/compiler.py", line 444, in _compile_branch
    ret += self.compile(exprs[-1])
  File "/home/bwillard/projects/code/python/hy-master/hy/compiler.py", line 361, in compile
    raise_empty(HyCompileError, e, sys.exc_info()[2])
  File "<string>", line 1, in raise_empty
⤷ ModuleNotFoundError: No module named 'blah'
Compilation traceback:
File "/home/bwillard/projects/code/python/hy-master/hy/compiler.py", line 350, in compile
    ret = self.compile_atom(tree)
  File "/home/bwillard/projects/code/python/hy-master/hy/compiler.py", line 344, in compile_atom
    return Result() + _model_compilers[type(atom)](self, atom)
  File "/home/bwillard/projects/code/python/hy-master/hy/compiler.py", line 1584, in compile_expression
    self, expr, unmangle(sroot), *parse_tree)
  File "/home/bwillard/projects/code/python/hy-master/hy/compiler.py", line 1154, in compile_import_or_require
    importlib.import_module(module)
  File "/home/bwillard/apps/anaconda3/envs/hy-master-cpython36/lib/python3.6/importlib/__init__.py", line 126, in import_module
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
=> (defn test [x]
... (print(+ 1 x))
... )
=> (test 'a)
  File "/home/bwillard/projects/code/python/hy-master/hy/importer.py", line 141, in hy_eval
  File "<eval_body>", line 1, in test
TypeError: unsupported operand type(s) for +: 'int' and 'HySymbol'
=> (defmacro blah [x] `(print ~@z))
<function <lambda> at 0x7f1351095730>
=> (defmacro bleh [q] `(print (blah ~q)))
<function <lambda> at 0x7f1351095840>
=> (bleh y)
  File "<input>", unknown location
HyMacroExpansionError: expanding `blah': NameError("name 'z' is not defined",)

=> (defn foo []
... (setv bar 42)
... (import pdb) (pdb.set-trace))
=> (foo)
--Return--
> <eval_body>(1)foo()->None
(Pdb) l
[EOF]
```
## After
```clojure
hy 0.15.0+45.ga8d6a48 using CPython(default) 3.6.6 on Linux
=> (with [file (open "a.hy" "w")] (file.write "(print 'hi\""))
11
=> (import a)
  File "stdin-99c0d4eb813cb56c622500ab4696f5bbb3cb3660", line 1, in <module>
    (import a)
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 674, in exec_module
  File "<frozen importlib._bootstrap_external>", line 781, in get_code
  File "/tmp/a.hy", line 1
    (print 'hi"
              ^
hy.lex.exceptions.PrematureEndOfInput: Partial string literal
=> (require blah)
  File "stdin-97db2eacebe610e14f46e56e428d68ec1c466f2f", line 1, in <module>
    (require blah)
hy.errors.HyRequireError: No module named 'blah'
=> (defn test [x]
... (print (+ 1 x))
... )
None
=> (test 'a)
  File "stdin-e199a43c50b388bff5e7901a7a1f02f5525d862d", line 1, in <module>
    (test 'a)
  File "stdin-71d0b2f8a864d8760f5a5eab4efa209694ee2714", line 2, in test
    (print (+ 1 x))
TypeError: unsupported operand type(s) for +: 'int' and 'HySymbol'
=> (defmacro blah [x] `(print ~@z))
<function <lambda> at 0x7fd6741319d8>
=> (defmacro bleh [q] `(print (blah ~q)))
<function <lambda> at 0x7fd674131268>
=> (bleh y)
  File "stdin-ff397ee5ec5c34f03703a2acccc9f62ed9b3a9ae", line 1, in <module>
    (bleh y)
  File "stdin-a0f7e3bfef673f0bae54abd3721c76c25db16c65", line 1, in <lambda>
    (defmacro blah [x] `(print ~@z))
hy.errors.HyMacroExpansionError: expanding macro blah
  NameError: name 'z' is not defined
=> (defn foo []
... (setv bar 42)
... (import pdb) (pdb.set-trace))
None
=> (foo)
--Return--
> /home/bwillard/stdin-2741f77da2a8503d126c87cd48d527a65903cc14(3)foo()->None
(Pdb) l
  1  	(defn foo []
  2  	(setv bar 42)
  3  ->	(import pdb) (pdb.set-trace))
[EOF]
```


Kodiologist:
It looks like a `PrematureEndOfInput` exception will display the whole file, or at least, a lot of it. Also, what's the arrow pointing at?

    perl -E 'say foreach "a".."z"; say "("' >/tmp/test.hy && hy /tmp/test.hy
      File "/home/hippo/Desktop/hyenv/bin/hy", line 12, in <module>
        sys.exit(hy_main())
      File "/home/hippo/Desktop/hyenv/hy/hy/cmdline.py", line 368, in hy_main
      File "/home/hippo/Desktop/hyenv/hy/hy/cmdline.py", line 354, in cmdline_handler
        runhy.run_path(filename, run_name='__main__')
      File "/usr/local/lib/python3.7/runpy.py", line 261, in run_path
        code, fname = _get_code_from_file(run_name, path_name)
      File "/home/hippo/Desktop/hyenv/hy/hy/importer.py", line 221, in _get_code_from_file
        code = _hy_code_from_file(fname, loader_type=HyLoader)
      File "/home/hippo/Desktop/hyenv/hy/hy/importer.py", line 200, in _hy_code_from_file
        code = loader.get_code(modname)
      File "<frozen importlib._bootstrap_external>", line 860, in get_code
      File "/home/hippo/Desktop/hyenv/hy/hy/importer.py", line 246, in _hy_source_to_code
        hy_tree = hy_parse(source, filename=path)
      File "/home/hippo/Desktop/hyenv/hy/hy/importer.py", line 76, in hy_parse
        reraise(type(e), e.args, None)
      File "/home/hippo/Desktop/hyenv/hy/hy/_compat.py", line 5, in reraise
        try:
      File "/tmp/test.hy", line 1
        a
    b
    c
    [ and so on ]
    y
    z
    (

        ^
    hy.lex.exceptions.PrematureEndOfInput: Premature end of input

brandonwillard:
Looks like it's pointing at the end of the file, where the paren isn't closed, but the indenting in the print-out stops after `a`.

brandonwillard:
Just pushed a fix that should take care of that print-out.

hy/cmdline.py
```diff
@@ -65,33 +68,48 @@ def __init__(self, spy=False, output_fn=None, locals=None,
...
+
+        # Sadly, this method in Python 2.7 ignores an overridden `sys.excepthook`.
+        if sys.excepthook is sys.__excepthook__:
+            super(HyREPL, self).showtraceback()
+        else:
+            sys.excepthook(sys.last_type, sys.last_value, sys.last_traceback)
+
+        self.locals[mangle("*e")] = sys.last_value
+
+    def runsource(self, source, filename='<input>', symbol='single'):
```
Kodiologist: These two methods are very similar. Is it possible to factor out the common stuff?

hy/cmdline.py
```diff
@@ -281,9 +288,6 @@ def cmdline_handler(scriptname, argv):
                              "(e.g., hy.contrib.hy-repr.hy-repr)")
     parser.add_argument("-v", "--version", action="version", version=VERSION)
 
-    parser.add_argument("--show-tracebacks", action="store_true",
-                        help="show complete tracebacks for Hy exceptions")
-
```
Kodiologist: Also emove the corresponding bit in `docs/langauge/cli.rst`.

hy/compiler.py
```diff
@@ -379,8 +386,9 @@ def _compile_collect(self, exprs, with_kwargs=False, dict_display=False,
 
             if not PY35 and oldpy_unpack and is_unpack("iterable", expr):
                 if oldpy_starargs:
-                    raise HyTypeError(expr, "Pythons < 3.5 allow only one "
-                                            "`unpack-iterable` per call")
+                    raise HyTypeError("Pythons < 3.5 allow only one "
+                                      "`unpack-iterable` per call",
+                                      (self.filename, expr, self.source))
```
Kodiologist: Having to say ` (self.filename, expr, self.source)` every time is a drag. Try a helper function or method, like

    def err(self, expr, msg):
        return HyTypeError(msg, (self.filename, expr, self.source))

hy/compiler.py
```diff
@@ -1709,10 +1722,22 @@ def hy_compile(tree, module_name, root=ast.Module, get_expr=False):
...
+                        "being promoted to one")
+
+    try:
+        mod_loader = pkgutil.get_loader(module_name)
+        if filename is None:
+            filename = mod_loader.get_filename()
+        # TODO: Lazy load this.
+        if source is None:
+            source = mod_loader.get_source()
+    except Exception:
```
Kodiologist: Again, `except Exception` might be too broad.

hy/compiler.py
```diff
@@ -1709,10 +1722,22 @@ def hy_compile(tree, module_name, root=ast.Module, get_expr=False):
...
-
-    compiler = HyASTCompiler(module_name)
+        raise TypeError("`tree` must be a HyObject or capable of "
+                        "being promoted to one")
+
+    try:
+        mod_loader = pkgutil.get_loader(module_name)
+        if filename is None:
+            filename = mod_loader.get_filename()
+        # TODO: Lazy load this.
```
Kodiologist: I should ask: are TODO comments like this meant for yourself or for future readers and editors of the codebase?

hy/errors.py
```diff
@@ -2,52 +2,88 @@
...
+class HyTypeError(HyLanguageError, TypeError):
+    def __init__(self, message, args=None):
+        """
+        Parameters
+        ----------
+        message: str
+            The exception's message.
+        args: tuple, optional
+            A tuple with the form `(filename, expression, source)`.
+        """
```
Kodiologist: Is there any reason you chose to take `filename`, `expression`, and `source` as a tuple instead of as separate parameters? (Likewise for `HySyntaxError`.)

hy/lex/parser.py
```diff
@@ -22,84 +22,85 @@
...
-            if tok == "." and type(tok) == HySymbol:
-                raise LexException("Malformed dotted list",
-                                   tok.start_line, tok.start_column)
+# def reject_spurious_dots(*items):
+#     "Reject the spurious dots from items"
+#     for list in items:
+#         for tok in list:
+#             if tok == "." and type(tok) == HySymbol:
+#                 raise LexException("Malformed dotted list",
+#                                    tok.start_line, tok.start_column)
```
Kodiologist: Lots of commented-out code here.

hy/compiler.py
brandonwillard: Yeah, I have have something like that in a correction I'm about to push.

hy/compiler.py
brandonwillard: Heh, good question: sometimes one of those, sometimes a couple, sometimes all!  
This one is for editors, though.

Kodiologist: I see. I think an idea about how the code should change in the future, like that, is better as a GitHub issue than a comment. Otherwise we're kind of tracking things we want to do in two different places.

hy/lex/parser.py
brandonwillard: Yeah, I wasn't sure where/if that was ever used, so I didn't update it and only commented it out.

Kodiologist: Oh yeah, looks like it's dead code. I vaguely remember making it obsolete when editing the lexer a while ago. Why not just add another commit to this PR, rebased to before the other commits, deleting it?

hy/compiler.py
Kodiologist: What I'm afraid of is silencing an error that would alert us of a bug. That seems worse to me than weird crashes from unanticipated but meaningless errors we should've ignored. I'd say, just list the exception types you've been able to provoke so far. We'll add more later if we need them. Does that sound reasonable?

brandonwillard: Any relevant errors showing up here would necessarily appear in more appropriate places soon after (i.e. when one of the modules is actually imported), or&mdash;best case scenario&mdash;we would completely avoidable irrelevant errors in the core functionality that follows.

I don't mind being more precise, though, so I'll make that change.

brandonwillard: Great example of my point: the current `pytest` failures are due to an exception in this exact place (and the exact circumstances I described).

`pytest` hijacks the loader mechanisms to in order to do its assertion rewriting, and the loader object it returns from `pkgutil.get_loader` does not fulfill the Python API, so the `get_filename` and `get_source` calls will fail.  Had this exception been caught, the function would've gone forward unaffected and the tests would've done their job (and passed).  

Now, since it's not reasonable to write special conditions for an exception caused exclusively by `pytest`, we have to work around this problem indirectly and in less sure-fire ways.  Furthermore, the code is still open to issues of this nature.

Kodiologist: Ouch. Would it suffice to add `TypeError` to the list?

brandonwillard: The `TypeError` is for the other situation that uses `inspect.getfile`.

Worse yet, after fixing the `pytest` issue, I hit [this independent `pkgutil.get_loader` issue](https://bugs.python.org/issue14710) in Python 3.4.0.  Thankfully, upgrading to Python 3.4.4 fixes it.

Kodiologist:
Would you please:

- Try to split the big commits into smaller logical units.
- Separate out refactoring and other code changes unrelated to the purpose of a commit (e.g., adding `object` to the base classes of `HyREPL`) into other commits.
- Generally speaking, don't make a commit that undoes or alters changes that were introduced by a previous commit in the same PR. Instead, change the previous commit itself.
- Avoid intermediate variables that are used only once (e.g., instead of `x = a + b; z = x + y`, write `z = a + b + y`).
- Remove comments with XXX, TODO, FIXME, or questions, or any other comments that do something other than describe the current state and usage of the code. If you wish, make issues out of these removed comments or post them to this PR.

Let me know when the PR is ready for review.

hy/cmdline.py
brandonwillard: These two functions are part of the class spec and called implicitly elsewhere, so I can't get rid of them, but they can be simplified.

brandonwillard:
All right, I further separated those commits by their functionality.  Unfortunately, the first one is necessarily large-ish, due to how much code it affects; however, most of the changes are simple (e.g. adding new arguments for source and filename).

Kodiologist:
It's okay. There will at least be fewer added lines once you introduce a helper for `HyTypeError`'s new parameters.

brandonwillard:
All right, added some helpers for the relevant errors.

Kodiologist:
You mean like this?

    raise HySyntaxError.from_expression(
        "Pythons < 3.5 allow only one "
        "`unpack-mapping` per call",
        expr, self.filename, self.source)

Why not something like this?

    raise self.error(expr,
        "Pythons < 3.5 allow only one "
        "`unpack-mapping` per call")

Then you don't have to pass in `self.filename` and `self.source` every time.

brandonwillard:
No, the helper functions are the additional class methods in `HySyntaxError` and `LexException`.

Sure, a method in the compiler could be added, too.

brandonwillard:
Added a `HyASTCompiler._syntax_error` method.

Kodiologist:
Much nicer.

brandonwillard:
Any updates needed?  I find myself using this branch for everyday Hy&mdash;for the new error messages and REPL source tracking&mdash;and I would like to build-out a few new updates from it (e.g. better spy output for `hy-mode` highlighting), so it would be great to get it officially out there.

Kodiologist:
Sorry, I didn't know this was ready. I'll look at it soon.

Kodiologist:
New conflicts, woo.

brandonwillard:
Agh, yeah, I'm on it.

brandonwillard:
Conflicts resolved.

Kodiologist:
Regarding my earlier point:

> Generally speaking, don't make a commit that undoes or alters changes that were introduced by a previous commit in the same PR. Instead, change the previous commit itself.

It looks like your first commit ("Retain compiled source and file information for exceptions") introduces many changes like:

    raise HyTypeError("Keyword argument {kw} needs "
                      "a value.".format(kw=expr),
                      self.filename, expr, self.source)

which, in a later commit, are then changed to use `self._syntax_error`, like:

    raise self._syntax_error(
        "Keyword argument {kw} needs "
        "a value.".format(kw=expr),
        expr)

Can you rebase to use `self._syntax_error` the first time? This should eliminate lots of lines of diff.

brandonwillard:
All right, rebased those into the first commit.

Kodiologist:
```python
_source = re.sub(r'\A#!.*', '', source)
# Offset for removed shebang (if removed).
lineno_offset = 1 if _source != source else 0
```

The `re.sub` will only blank the first line, not remove its trailing newline, so shouldn't `lineno_offset` stay at 0 in any case?

Kodiologist:
It looks like there are still some unrelated indentation changes buried in your commits, such as

```diff
     if s.startswith('hyx_'):
         s = re.sub('{0}(U)?([_a-z0-9H]+?){0}'.format(mangle_delim),
-            lambda mo:
-               chr(int(mo.group(2), base=16))
-               if mo.group(1)
-               else unicodedata.lookup(
-                   mo.group(2).replace('_', ' ').replace('H', '-').upper()),
-            s[len('hyx_'):])
+                   lambda mo:
+                   chr(int(mo.group(2), base=16))
+                   if mo.group(1)
+                   else unicodedata.lookup(
+                           mo.group(2).replace('_', ' ').replace('H', '-').upper()),
+                   s[len('hyx_'):])
     if s.startswith('is_'):
         s = s[len("is_"):] + "?"
     s = s.replace('_', '-')
```

Probably this is the fallout of an overly aggressive auto-indenting tool. You've worked plenty hard on this PR already, so I can go back and undo these myself, but in the future, remember to revert these or put them in their own commits (with `--patch` or so) before you `git commit`. (This particular example would be a bad idea to keep since it's no longer clear where the body of the `lambda` ends.)

Kodiologist:
Kind of a shame that we have to add a `state` parameter to every single parser when we rarely need it, but I guess that's an rply limitation.

brandonwillard:
> Why's this one a HyTypeError instead of a HySyntaxError? I'm not sure I understand the difference.
> 
> ```python
> if level == 0 and op in ("unquote", "unquote-splice"):
>      if len(form) != 2:
>          raise HyTypeError("`%s' needs 1 argument, got %s" % op, len(form) - 1,
>                            self.filename, form, self.source)
>      return set(), form[1], op == "unquote-splice"
> ```

Most, if not all, of the `HyTypeError`s that were changed to `HySyntaxError`s had some sort of correlate in Python (that would raise `SyntaxError`).  In this case, `TypeError`s are&mdash;usually&mdash;raised in Python when a required argument is missing.

brandonwillard:
> ```python
> _source = re.sub(r'\A#!.*', '', source)
> # Offset for removed shebang (if removed).
> lineno_offset = 1 if _source != source else 0
> ```
> 
> The `re.sub` will only blank the first line, not remove its trailing newline, so shouldn't `lineno_offset` stay at 0 in any case?

If we're not removing leading newlines later on, then, yes, I think so.  

That reminds me; I wanted to remove it entirely.  This is the kind of stuff that should be handled by the lexer.

brandonwillard:
> Kind of a shame that we have to add a `state` parameter to every single parser when we rarely need it, but I guess that's an rply limitation.

Yeah, it would be much better if the production functions accepted the state object as an optional keyword argument.

Kodiologist:
So, do you want to rebase out the introduction of `lineno_offset` and its management, since (if I understand correctly) it should always be 0?

brandonwillard:
All right, removed.

Kodiologist:
Okay, I think the first commit's done. I spun out the introduction of `_syntax_error` into its own commit, reverted the argument flip of `macro-error`, reverted some unrelated indentation changes, reverted an introduction of a local variable that was made obsolete by a later commit, etc. I'll look at the next commit tomorrow.

brandonwillard:
Breaking out that `_syntax_error` in `compiler.py` into the first commit isn't helpful; at that point it's not even simplifying the the code it's replacing.

Kodiologist:
The whole point is that nothing interesting is happening in the new commit. By making the split, the gigantic commit in which several meaningful things are happening is made less gigantic, by removing a large number of repetitive changes.

Kodiologist:
Do you see why I want to avoid situations in which a commit that does meaningful things also makes a lot of changes of whitespace, variable names, or something like that? I want to divide the signal from the noise.

brandonwillard:
You'll have to be more specific about the which exception messages you're talking about, because some tests are specifically for the exception messages.

brandonwillard:
Those changes aren't about whitespace though.

Kodiologist:
Sure, but they're cosmetic in the sense that, before `_syntax_error` does something other than raise `HyTypeError` with exactly the same arguments, its introduction has no effect. The change in logic can be easily separated from the change in calling convention. So I separated them.

> You'll have to be more specific about which tests you're talking about, because some are specifically for the exception messages.

Consider for example:

```diff
     "Ensure tokenizing \"' \" throws a LexException that can be stringified"
     # https://github.com/hylang/hy/issues/1252
-    with lexe() as e: tokenize("' ")
-    assert "Could not identify the next token" in str(e.value)
+    with lexe() as execinfo:
+        tokenize("' ")
+
+    expected = ['  File "<string>", line -1\n',
+                "    '\n",
+                '    ^\n',
+                'LexException: Could not identify the next token.\n']
+    output = traceback.format_exception_only(execinfo.type, execinfo.value)
+
+    assert output[:-1:1] == expected[:-1:1]
+    # Python 2.7 doesn't give the full exception name, so we compensate.
+    assert output[-1].endswith(expected[-1])
``` 

What's your goal for this test change? (Intuitively, I think your change is right, or at least mostly right, but I've lost track of what's going on.)

brandonwillard:
The goal is to make sure that the source information is retained, and properly used (for the descriptive print-out) throughout the course of raising and processing that exception (per the described changes) in the context of a call to `tokenize`.  

How exactly is this confusing?  That test is directly related to the PR-added and commit-described functionality (e.g. that message couldn't be produced if we weren't tracking the source string, line information, etc.)  The changes only elaborate on its original purpose, which is&mdash;in part&mdash;to make sure that a `LexException` is thrown, by including the requirement that the `LexException` show the location of the error in the source string.

Kodiologist:
Okay, that makes sense.

> How exactly is this confusing?

That I am confused by something in no way implies that it's confusing. ;)

Kodiologist:
Okay, review of the next commit's done; the edits are:

- Factored out some copied-and-pasted code into a function
- `__tb_hidden_modules` → `_tb_hidden_modules` (you don't need Python's mangling here, right?)
- Removed trailing `:1` in slices

Kodiologist:
The commit message for "Make colored output configurable" mentions `HY_FILTER_INTERNAL_ERRORS` where I think you meant `HY_COLORED_ERRORS` (right?).

It seems that `HY_FILTER_INTERNAL_ERRORS` only has an effect when set to the empty string. I think it's counterintuitive that a binary environment variable would distinguish between an empty string and absence. How about inverting the variable and calling it `HY_SHOW_INTERNAL_ERRORS` or something?

At least at the tip of the PR, there's a regression such that a return value of `None` is no longer silenced at the REPL.

Kodiologist:
I'm sorry, Brandon, I've only just noticed that you updated this. It seems that GitHub doesn't send a notification for force-pushes. I'll continue review soon.

brandonwillard:
No problem; I try to respond with a message after force-pushes, since I also don't recall ever getting such notifications.

I believe I've addressed the matters in [your last comment](https://github.com/hylang/hy/pull/1687#issuecomment-447418669), except for the REPL return value.  Is it by design that it doesn't display `None`?  

The Python REPL doesn't either, but, if I recall, Clojure and some other Lisps do.  The only immediate advantage&mdash; that I can think of right now&mdash;involves code that reads from the REPL output (e.g. `comint` in Emacs); it's often easier to handle situations in which output is guaranteed/consistent.

Regardless, it's easy to change.

Kodiologist:
> Is it by design that it doesn't display `None`?

Yeah, the non-display of `None` has to be implemented separately, since `None` doesn't `str` or `repr` to the empty string (and then there's the matter of the newline). I think it helps keep the output neater.

Kodiologist:
So do that, if you don't mind, and then I'll look again.

Kodiologist:
Thanks. I just reviewed more of these commits and made more changes and I'm getting a "Permission denied" error when I try to push. Do you have the option to accept changes from reviewers switched off?

brandonwillard:
Yes

Kodiologist:
Can you turn it back on? Otherwise, I don't think there's a way for me to cleanly rewrite history. You can make a copy first with `git branch clean-up-exceptions--backup` or so.

brandonwillard:
I'm fine with *one* reasonably sized set of final reviewer changes&mdash;which is what I thought the last one was&mdash;but a review process consisting of rebases is too confusing and time consuming.

Please describe the changes you think are necessary to merge this PR.

Kodiologist:
I could save up all my edits till I'm done reviewing the entire PR, but I thought you'd rather get them as I make them, rather than get a giant stack of them all at once after a long time.

I've pushed my current version of `clean-up-exceptions` to my own fork. If you'd prefer to wait till I've finished reviewing the whole PR (which hopefully won't be *that* much longer), I can let you know when I'm done.

Kodiologist:
Okay, all done. Sorry for the long delay. I had a busy week and I expected there'd be a lot more work to do here, but I made only minor changes to the last two commits.

Kodiologist:
If all looks good to you, replace your version of the branch with mine and we should be good to merge (pending the final Travis check).

brandonwillard:
Finally had a chance to look at your branch; give it a push.

Kodiologist:
Okay, I'll do it this afternoon. There will probably be some merge conflicts, but if those can be smoothed out without major issues and Travis has no surprises in store, I'll merge.

PR #1688, opened 2018-10-20
Fix build status badge

mbarkhau:
I'm not sure why this is currently red, maybe it references the latest build from here: https://travis-ci.org/hylang/hy/builds. 

The image url with the explicit branch is probably the better choice.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -11,7 +11,7 @@ Welcome to Hy's documentation!
 :List: `hylang-discuss <https://groups.google.com/forum/#!forum/hylang-discuss>`_
 :IRC: ``#hy`` on Freenode
 :Build status:
-    .. image:: https://secure.travis-ci.org/hylang/hy.png
+    .. image:: https://secure.travis-ci.org/hylang/hy.png?branch=master
         :alt: Travis CI
         :target: http://travis-ci.org/hylang/hy
 


Kodiologist:
Because we're currently running Travis for PRs only, rather than both PRs and pushes, the badge probably won't work in any case. If so, we should just remove it.

mbarkhau:
Ahh, I see. So this badge is actually referring to a build that is a few months old?

https://secure.travis-ci.org/hylang/hy.png?branch=master

Kodiologist:
I suspect so. But it might also be referring to the last commit that was tested before being pushed (currently https://travis-ci.org/hylang/hy/builds/435140822), which is a reasonable backup. I'm not sure how to tell.

mbarkhau:
Maybe this is helpful? https://stackoverflow.com/questions/31882306/how-to-configure-travis-ci-to-build-pull-requests-merges-to-master-w-o-redunda?rq=1

If it were me, and it's possible, I'd just let it do another build after the merge of any pull request.

Kodiologist:
The reason I turned off push builds originally was because of a longstanding Travis bug where if you accidentally push a PR branch to origin instead of your own fork, then it gets another redundant build, and if you remove the branch, the new build line sticks around in a permanent failure state. But `branches: {only: master}` could fix that. It's worth trying.

Kodiologist:
The badges in question are gone as of #1811.

PR #1691, opened 2018-11-06
Catch IndentationError in isidentifier()

jwilk:
Fixes:

```pycon
>>> from hy._compat import isidentifier
>>> isidentifier(u"  0\n 0")
  File "hy/_compat.py", line 47, in isidentifier
    tokens = list(T.generate_tokens(StringIO(x).readline))
  File "/usr/lib/python2.7/tokenize.py", line 374, in generate_tokens
    ("<tokenize>", lnum, pos, line))
  File "<tokenize>", line 2
    0
    ^
IndentationError: unindent does not match any outer indentation level
```


Kodiologist:
Did this issue lead to a user-visible bug in `mangle` or something? If so, could you add a test?

jwilk:
I doubt any real users would ever see this, but in theory, yes, it's user-visible.
I've added a test.

Kodiologist:
Unfortunately, it looks like there's another bug: `\n` shouldn't be mangling to `XUnX`. Indeed, `(unmangle (mangle "\n"))` crashes. I think you'd better either (a) fix that bug, too,  or (b) write a test for your original fix that's independent of that bug.

Kodiologist:
Looks good. Update NEWS, squash everything to one commit, and we're good to go.

Issue #1698, opened 2018-11-09
`HyREPL` calls `hy_compile` on every processed command


brandonwillard:
`HyREPL` calls `hy_compile` on every processed command.  When that happens, a `HyASTCompiler` is constructed and all the stdlib imports are triggered again and again.

We should probably create and reuse a single `HyASTCompiler` instance for the lifetime of `HyREPL`, no?

_Originally posted by @brandonwillard in https://github.com/hylang/hy/issue_comments#issuecomment-437527370_

Kodiologist:
Probably.

Issue #1697, opened 2018-11-09
`hy.core.language` import in `hy/__init__.py


brandonwillard:
While working on Hy I've run into problems involving the `hy.core.language` imports in `hy/__init__.py` on numerous occasions.  These imports are automatic and necessarily trigger Hy compilation&mdash;among other things&mdash;when loading literally anything from the package.  

If one intends to simply import the model objects from `hy.models`&mdash;for example&mdash;then `hy/core/language.hy` is compiled, which seems like a lot of unnecessary work.  Furthermore, this behavior obfuscates the true source of errors&mdash;where there are any&mdash;by triggering package-loading exceptions for functionally independent parts of the package.

It seems like we could replace the `hy.core.language` imports with their Python equivalents, or simply remove those from `__init__.py` altogether.
Is there something to this scenario that would warrant retention of `hy.core.language` within `hy/__init__.py`?

Kodiologist:
I like the way you're thinking. I've been bitten by this before. It ought to be possible to access pure-Python parts of the hy namespace without invoking the compiler.

There are currently four imports from `hy.core.language` in `__init__.py`:

- `mangle`
- `unmangle`
- `read_str`
- `read`

The first two can just be imported from where they're actually defined, `hy.lex`, instead of `hy.core.language`. The latter two could be written in Python instead of Hy.

brandonwillard:
Exactly!

brandonwillard:
Maybe put these changes together with https://github.com/hylang/hy/issues/1695 in a reorganizing PR?

Kodiologist:
That's a decent idea. I usually run the tests with `--ignore tests/test_bin.py` anyway because they're so much slower than everything else.

Kodiologist:
> In other words, would we like to have a lightweight compiler that exclusively translates Hy AST to Python AST and implicitly relies on the current state of its target module for macro and function dependencies? 

Maybe. I see your argument, but it's hard to guess how things would shake out in practice.

PR #1700, opened 2018-11-11
Reuse Hy compiler instances

brandonwillard:
Every call to `hy_eval` and `hy_compile` will create a new `HyASTCompiler` instance, which incurs Hy standard library setup costs each time.  When calls to these functions occur more than once (e.g. by use of `HyREPL` and `eval-and-compile`), the aforementioned work is performed unnecessarily for the same module.  These changes avoid that unneeded and error-prone repetition.

- [x] #1698

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -9,6 +9,8 @@ Removals
 
 New Features
 ------------------------------
+* `eval` / `hy_eval` and `hy_compile` now accept an optional `compiler` argument
+  that enables the use of an existing `HyASTCompiler` instance.
 * Keyword objects (not just literal keywords) can be called, as
   shorthand for `(get obj :key)`, and they accept a default value
   as a second argument.
--- a/docs/language/core.rst
+++ b/docs/language/core.rst
@@ -195,7 +195,9 @@ eval
 ``eval`` evaluates a quoted expression and returns the value. The optional
 second and third arguments specify the dictionary of globals to use and the
 module name. The globals dictionary defaults to ``(local)`` and the module name
-defaults to the name of the current module.
+defaults to the name of the current module.  An optional fourth keyword parameter,
+``compiler``, allows one to re-use an existing ``HyASTCompiler`` object for the
+compilation step.
 
 .. code-block:: clj
 
@@ -1403,4 +1405,3 @@ are available. Some of their names have been changed:
   - ``dropwhile`` has been changed to ``drop-while``
 
   - ``filterfalse`` has been changed to ``remove``
-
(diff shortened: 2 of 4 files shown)


brandonwillard:
In order for a compiler instance to simply [re]use itself when calling `hy_eval` (e.g. during `eval-and-compile`), we need to make the compiler "stateless".  Members like `imports` and `temp_if` attach state to a compiler instance that don't allow its immediate reuse.

To replace `imports`, my plan of attack involves more consistent tracking/use of `Result` objects and the `imports` in those.  Basically, we can&mdash;and should&mdash;keep all "state" within `Result` objects.  This looks like it will involve an interface-like change to the compiler, though; namely, that all `HyASTCompiler.compile_*` methods must return and take `Result` objects.  While that's often the case (well, the former at least), it's not always true, and this inconsistency isn't good for the compiler's design.

For instance, `HyASTCompiler.compile_symbol` doesn't return a `Result` object, but it can be made to do so, and&mdash;with that change&mdash;it can assign stdlib auto-imports to the current state of compilation (via `Result.imports`) instead of the actual compiler instance.  Likewise, with that in place, we can make `HyASTCompiler.imports_as_stmts` produce AST for a `Results`'s `imports` and return those in another `Result` object&mdash;bringing it more in-line with its adjacent `compile_*` methods.  Additionally, the logic currently in `hy_compile` could then be moved to `HyASTCompiler.imports_as_stmts` and disentangle essential compilation logic from `hy_compile`.
   
At the very least, these changes would get us closer to a compiler that better isolates its functionality and is significantly easier to extend and debug.  

Regarding `HyASTCompiler.temp_if`, I'm having a hard time understanding exactly what its functional "intentions" are and the exact scope/constraints of its role.  I can see that it's attempting to avoid the unnecessary use/creation of temporary variables in an `if*`, but how it's using the compiler state to do this isn't very clear or well contained.  I can't tell whether or not the addition of a `temp_if` field to `Result` would work, or if there's something much simpler that would avoid such an undesirable addition.

Somewhat related issues: #1542, #1482.  Also, [this comment](https://github.com/hylang/hy/issues/1697#issuecomment-437526626) in #1697.

brandonwillard:
@Kodiologist, good call; will do.

brandonwillard:
Yeah, one minute.

brandonwillard:
OK, that should do it.

Kodiologist:
Nice. One more thing: the documentation of `eval` in core.rst`. I know, it's a problem that we have both docstrings and the redundant documentation in those files.

brandonwillard:
Didn't notice that; I'll update it.

Otherwise, it seems like we should really get started on https://github.com/hylang/hy/issues/1044.

Kodiologist:
Replace "fifth" with "fourth" and I'll merge.

Yep.

Issue #1707, opened 2018-11-26
`defmain` fails with no arguments


brandonwillard:
From the command line (e.g. `hy test.hy`), the `defmain` macro fails with empty arguments (e.g. `(defmain [x] ...)` appears to work).

## Example

```clojure
;; test.hy
(defmain []
  (print "hi"))
```
```bash
> hy test.hy
Traceback (most recent call last):
  File "/home/bwillard/apps/anaconda3/envs/hy-master-cpython36/bin/hy", line 12, in <module>
    sys.exit(hy_main())
  File "/home/bwillard/projects/code/python/hy-master/hy/cmdline.py", line 381, in hy_main
    sys.exit(cmdline_handler("hy", sys.argv))
  File "/home/bwillard/projects/code/python/hy-master/hy/cmdline.py", line 367, in cmdline_handler
    runhy.run_path(filename, run_name='__main__')
  File "/home/bwillard/apps/anaconda3/envs/hy-master-cpython36/lib/python3.6/runpy.py", line 263, in run_path
    pkg_name=pkg_name, script_name=fname)
  File "/home/bwillard/apps/anaconda3/envs/hy-master-cpython36/lib/python3.6/runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "/home/bwillard/apps/anaconda3/envs/hy-master-cpython36/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/tmp/test.hy", line 1, in <module>
    (defmain []
TypeError: <lambda>() takes 0 positional arguments but 1 was given
```

_Originally posted by @brandonwillard in https://github.com/hylang/hy/issues/1706#issuecomment-441751053_

PR #1708, opened 2018-11-26
Make `defmain` args optional or variadic when empty

brandonwillard:
Closes hylang/hy#1707.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -21,6 +21,7 @@ New Features
 
 Bug Fixes
 ------------------------------
+* Fixed issue with empty arguments in `defmain`.
 * `require` now compiles to Python AST.
 * Fixed circular `require`s.
 * Fixed module reloading.
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -223,11 +223,17 @@ Such 'o!' params are available within `body` as the equivalent 'g!' symbol."
 
 
 (defmacro defmain [args &rest body]
-  "Write a function named \"main\" and do the 'if __main__' dance"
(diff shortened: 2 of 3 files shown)


Kodiologist:
Why add the `&optional`s by force? The programmer can add it if they want it.

brandonwillard:
That just avoids cryptic error messages.  We can remove it, though.

Kodiologist:
How about improving the error messages?

brandonwillard:
There are none, so we're actually talking about adding some.  Even so, the cryptic results have more to do with the likely unexpected default first `sys.argv` value and the error messages caused by it.  

Furthermore, a user specifying their own arguments, e.g. `(defmain [a b] (print a))`, might not expect the first argument to always be the executable/script name instead of the first cmdline argument.  Perhaps it would be better to make that value accessible in some other way and let the user-specified arguments correspond directly to the cmdline arguments.  
At the very least, the docs string should state that the first macro argument will always be the binary/script name.

Kodiologist:
> There are none, so we're actually talking about adding some.

There are no custom error messages for `defmain`, but if you define `(defmain [x y] …)` and call the program without arguments, you get `TypeError: <lambda>() missing 1 required positional argument: 'y'`, which is an example of the cryptic error messages you were referring to, right?

Regarding your second paragraph, yeah, it's surprising that the first argument is the executable name, and indeed that's part of why it's surprising that `(defmain [] …)` doesn't work. The documentation should be clearer about this, and perhaps we should change it.

brandonwillard:
Will do.

No, didn't work when I first tried.

PR #1720, opened 2019-01-03
Add a let macro to make hy code a little more idiomatic

kuettler:
The macro itself is very small, hardly noticeable. And yes, there will be some runtime overhead. Also, python scoping rules do not allow (simple) assignment to variables in a closure. Definitions in a "higher up" `let` will be protected for this reason. This might be surprising.

Overall, however, it feels so much better. A proper `let` makes code so much cleaner.


olasd:
The `hy.contrib.walk` module already has a `let` macro, with (more) proper let semantics than what you're providing here (it does work around some of the python scoping mess).

kuettler:
Thanks. I should have seen that.

Issue #1722, opened 2019-01-11
Better `&optional`, `&rest`, `&kwonly`, and `&kwargs` docs.


Quelklef:
The documentation for these four constructs is, in my opinion, lacking. I got tripped up not realizing that `&optional` affected not only the next argument but also all those afterwards.

Undocumented bits include:
- That `&optional` affects all succeeding arguments
- Can `&rest` be used with `&optional`?
- That `&kwonly` affects all succeeding arguments (I assume, since this is how it works in Python)
- Can an argument come after a `&kwargs` argument?
- More stuff that I haven't thought of.

The interactions between these things is not particularly simple and, though understandable after a bit, can be difficult for a beginner (in Python and Hy or in Hy but not Python, like me). Further, the error message (at least for the `&optional` interaction) is not particularly enlightening, so it's hard to figure out what's going wrong beyond just trial&error.

Kodiologist:
Recapitulating the fine details of Python's argument handling is not a good idea—we should link to the relevant parts of the Python language reference instead—but there are definitely some outright errors and omissions in the current descriptions, such as the implication that `&optional` affects a single parameter.

kirbyfan64:
Now that I think about it, should we really have `&optional`? It seems a bit awkward given how defaults work otherwise...maybe just allow someone to omit the default but still use inner brackets to imply a default of None?

(Also hey look, I'm not dead.)

Kodiologist:
That's not a bad idea. I think the status quo is just because that's how Lisps have traditionally specified optional parameters.

Quelklef:
On the thread of more concise arguments, thoughts on not having `&kwonly` and just having all keyword-only arguments be written as a keyword with `:`?

Quelklef:
So e.g.

```hylang
(defn print [&rest values &kwonly end] ...)
```

is written instead as

```hylang
(defn print [&rest values :end] ...)
```

I would be against this in Common Lisp, because "keywords" are symbols, by Hy parses keywords differently anyway.

PR #1731, opened 2019-01-17
Some documentation additions

Quelklef:
A beginner's perspective :)


Kodiologist:
The addition for `unpack-iterable` and `unpack-mapping` makes sense.

I don't think your addition regarding `&optional` really helps with #1722. The text before your addition ("Parameter is optional. The parameter can be given…") still makes it sound like `&optional` affects only one parameter. Instead of adding an explanation that you don't need a separate `&optional` for each optional parameter, we should rewrite the documentation for the ampersand symbols to be clearer and more correct, and refer to the Python language reference for details. Also, an example that uses all the ampersand symbols in one function would probably be helpful, and that example should probably include two `&optional` parameters, one with a default value and one without.

I don't think the addition for `constantly` is very helpful because this item is listed under "Core Functions" and the description doesn't imply delayed or repeated evaluation.

Quelklef:
Alright, I'll try to rewrite the current `&optional` docs. The other `&` symbols seemed OK.

As for `constantly`, I included that note because I hit a situation where I wanted to use `constantly`, but only if it had delayed evaluation. I figure it's worth mentioning for others who were wondering like I was. But I understand your position; should I remove it?

Kodiologist:
Yeah, I don't see that being a common mistake. That said, `constantly` does seem like it would be more useful as a macro, so you could change the implementation if you wanted to.

Quelklef:
Same PR? New PR?

Kodiologist:
New, because if you add nontrivial code changes to this PR, it can't take advantage of the policy that lets documentation changes get merged quickly.

Quelklef:
New `constantly` documentation should go under Built-Ins, right? What's the difference between `Hy Core/Core Functions` and `Built-Ins`?

Quelklef:
Or `Hy Core/` anything and `Built-Ins`, for that matter..

Kodiologist:
The core functions are all functions, whereas api.rst lists special forms and macros (which should probably be separated).

Kodiologist:
Looks good. Squash this PR into two commits, one for the ampersand parameters and one for the unpacking operators, and I'll merge.

Quelklef:
Well, it's now two commits, but I rewrote history. Still a git novice; was there a nicer way of squashing the PR?

Kodiologist:
Squashing is just one of several ways to rewrite history. When you can get the rewritten history that you want with any of several options (`rebase -i`, filter-branch, manually recommitting, etc.), just use whichever option you find most convenient. At any rate, you did a fine job here.

PR #1732, opened 2019-01-18
Make `constantly` a macro

Quelklef:
As per https://github.com/hylang/hy/pull/1731.

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -302,6 +302,38 @@ As shown below, only the first matching result block is executed.
     value is greater than 5
 
 
+.. _constantly:
+
+constantly
+----------
+
+.. versionadded:: 0.15.0
+
+Usage ``(constantly 42)``
+
+Create a new function that always returns the given value, regardless of
+the arguments given to it. Delays argument evaluation until the function
+is called.
+
+.. code-block:: hy
+
+   => (setv answer (constantly 42))
+   => (answer)
+   42
+   => (answer 1 2 3)
+   42
+   => (answer 1 :foo 2)
+   42
+
+.. code-block:: hy
+
+  => (setv smile (constantly (print ":)")))
+  => (smile)
+  :)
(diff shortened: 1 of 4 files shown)


gilch:
Opposed.

The current version works just like Clojure, which I consider totally valid. Changing it will confuse people coming from Clojure.

Also, this new version will re-evaluate the code every time it's called. Why not just use a `fn` at that point? Did you intend it to evaluate only once and cache the answer?

What exactly was the use case that motivated this change?

yuhan0:
I don't think this is a good idea either - to me "constantly" strongly implies that the argument is a constant, pure value being returned instead of a side-effectful procedure. 
Using a macro incurs the cost of an additional function call every time it is called, and confuses the usual call-by-value semantics of Python. 

```clojure
(defn expensive-computation []
  (time.sleep 10)
  42)

(setv xs [1 2 3 4 5])

(map (constantly 42) xs)
;; => [42, 42, 42, 42, 42]

;; The interpreter has no way of knowing if this is a pure function and caching its results --
;; would the user expect it to complete in 10 or 50 seconds?
(map (constantly (expensive-computation)) xs)
;; => [42, 42, 42, 42, 42]
```

`const` is used quite regularly in functional programming and hurting its performance in return for (questionable) side-effectful use does not seem like a good tradeoff.. 
Perhaps `do-constantly` for the macro version under a different name?

Quelklef:
> confuses the usual call-by-value semantics of Python.

I mean, as with all macros.

But I see your point about `const`. Another option besides `do-constantly` would be `continually`. It nicely implies an effectful argument and is a shorter name.

Kodiologist:
This seems fine to me. Don't forget to add tests. But, the approval of an additional core developer would be necessary to override gilch's veto, if that's the way you want to go.

Kodiologist:
I guess that's not happening.

PR #1733, opened 2019-01-20
Fix a Python 2 crash

Kodiologist:


kirbyfan64:
Always fun when downstream messes with major packages this much...

Kodiologist:
It's a weird situation. I think that what happened is that this backport was originally made upstream, and then was reverted when it realized it created problems, but Debian or Ubuntu used a commit before the revert for its package.

PR #1746, opened 2019-02-09
add the llist function

Chaformbintrano:
An equivalent of the `list` function from normal Lisps.


Kodiologist:
Couldn't this be implemented as just `(defn llist [&rest args] (list args))`?

More to the point, I don't really see the motivation. Why would one use `(llist a b c)` instead of `[a b c]`?

Chaformbintrano:
hmm, you're right

Kodiologist:
No problem. Happy Hy-ing.

Issue #1527, opened 2018-03-11
Class docstrings are a bit finicky


Kodiologist:
Besides not being documented, they're not recognized if your `defclass` form has an attribute list. The docstring needs to appear in place of the attribute list.

gilch:
For the record, I was opposed to keeping the old attribute list syntax at all. But it does work according to the indented spec: `(defclass <Name> <Base List> [docstring] [attribute list] [class body])`

For example,
```
(defclass Foo [object]
  "doc"
  [x 1]
  (defn foo [self]
    42))
```
```
class Foo(object):
    __doc__ = 'doc'
    x = 1

    def foo(self):
        return 42
```
As you can see, Hy has no problem using both a docstring and an attribute list. This keeps the statements in the same order they're compiled, so naturally the docstring must come before any assignments in the class body.

Kodiologist:
Oh, okay, I misread the compiler method. In any case, it needs to be documented.

vodik:
Off-topic, but +1 with removing the attribute list. Especially if we reintroduce `def` (or some other form) for declaring an attribute with type annotations - then the attribute list ends up also lacking useful functionality.

Kodiologist:
I forgot to close this after it was addressed by #1750 and #1789.

PR #1750, opened 2019-02-19
added documentation for defclass docstring

digikar99:
#1527 
The following works with bar.__doc__ returning "bar":
   
    (defclass bar [] "bar" [attr "value"]) 

while the following does not:

    (defclass bar [] [attr "value"] "bar")

(bar.\_\_doc__ evaluates to bar.\_\_doc__ itself.)

This seems to be the intended python-ic behaviour, if I am not wrong.

### DIFF
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -415,14 +415,16 @@ They can be used to assign multiple variables at once:
 defclass
 --------
 
-New classes are declared with ``defclass``. It can takes two optional parameters:
-a vector defining a possible super classes and another vector containing
-attributes of the new class as two item vectors.
+New classes are declared with ``defclass``. It can take three optional parameters in the following order:
+a list defining (a) possible super class(es), a string (:term:`py:docstring`) and another list containing
+attributes of the new class along with their corresponding values.
 
 .. code-block:: clj
 
     (defclass class-name [super-class-1 super-class-2]
-      [attribute value]
+      "docstring"
+      [attribute1 value1
+       attribute2 value2]
 
       (defn method [self] (print "hello!")))
 
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -239,7 +239,7 @@ as the user enters *k*.
 
 
 comment
-----
+-------
 
 The ``comment`` macro ignores its body and always expands to ``None``.
 Unlike linewise comments, the body of the ``comment`` macro must


Kodiologist:
I think you forgot a trailing backtick. Did you try rendering the file?

Kodiologist:
Looks good, but the backtick fix should be in the first commit, and the underline fix should be in its own commit (if you want it in this PR). Try

    git reset HEAD~
    git commit --amend --patch
    git commit -a

docs/language/api.rst
```diff
@@ -415,13 +415,14 @@ They can be used to assign multiple variables at once:
 defclass
 --------
 
-New classes are declared with ``defclass``. It can takes two optional parameters:
-a vector defining a possible super classes and another vector containing
+New classes are declared with ``defclass``. It can takes three optional parameters in the following order:
```
brandonwillard: `takes` should be `take`.

docs/language/api.rst
```diff
@@ -415,13 +415,14 @@ They can be used to assign multiple variables at once:
...
+a vector defining a possible super classes, a string (:term:`py:docstring`) and another vector containing
```
brandonwillard: `a vector defining a possible super classes` should probably be `a list defining the super classes`.

digikar99:
> I think you forgot a trailing backtick. Did you try rendering the file?

Yup, I had forgotten. I copied it from a few lines below, and didn't take note of the closing back-tick. Thanks for pointing that out.

No, I had not tried rendering. I will keep "build should be successful" in mind for the future pull requests.



> Looks good, but the backtick fix should be in the first commit, and the underline fix should be in its own commit (if you want it in this PR). Try
> 
> ```
> git reset HEAD~
> git commit --amend --patch
> git commit -a
> ```

I'll also keep "commit separately for each task, no matter how small that task is". Thanks for pointing out!



> I know you didn't introduce these mistakes, but, while you're changing those lines, this would be a great opportunity to fix them!

Sure, I have incorporated the suggested changes now. I have also added another attr-value example. 

Another good-first-issue I can think of is: fixing the missing syntax-highlighting (or more generally, fixing the warnings in Sphinx build process) for some of the documentation. Example sections include [`as->`](http://docs.hylang.org/en/stable/language/api.html#as). Should I create such an issue?

Kodiologist:
If you wish, sure.

Kodiologist:
Okay, I'll clean up the commit structure and merge.

Kodiologist:
@digikar99 I'm getting "Permission denied". Don't forget to turn on editing of your PR branches. (And in the future, make a new branch for each PR rather requesting a pull from `master`).

Kodiologist:
Ah, I got it. I was pushing to the wrong branch name because my local tracking branch wasn't named `master`.

PR #1752, opened 2019-03-09
Add array indexes cut tutorial

TristanCacqueray:
This change adds to the tutorial the hy way of indexing array.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -5,7 +5,6 @@ Tutorial
 .. TODO
 ..
 ..  - How do I index into arrays or dictionaries?
-..  - How do I do array ranges?  e.g. x[5:] or y[2:10]
 ..  - Blow your mind with macros!
 ..  - Where's my banana???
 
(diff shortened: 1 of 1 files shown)


Kodiologist:
I think that by "array indexes", you mean "slices". Relatedly, `get` should be mentioned before `cut`. Otherwise, it seems to be implied that `foo[3]` would be written `(cut foo 3)` in Hy, rather than `(get foo 3)`.

TristanCacqueray:
I meant slices indeed, and actually didn't knew about `cut` either as I am still discovering hy. It's a really awesome project btw, thanks :-)

Kodiologist:
Okay, good, but I would avoid the word "array" because [it means something else in Python](https://docs.python.org/3/library/array.html). How about "collection" instead?

TristanCacqueray:
Indeed. The new patch also added an assoc example.

Kodiologist:
One step forward, one step back. I'd remove `assoc` or else show that it's equivalent to `(setv (get array 1) 42)`. Otherwise, the novice reader gets the impression that `get` is only a function and doesn't have the same flexibility as Python indices.

TristanCacqueray:
Oh wow, I didn't realized you could set the value of a get expression like so :-)
Ok, let's keep this addition about get and cut since that is what i was looking for in the first place.

Kodiologist:
Cool, looks good.

Writing documentation is hard; ideally, you want an encyclopedic knowledge of the part of the language you're documenting. Writing a tutorial is even more demanding because it needs to cover a lot of the language.

PR #1759, opened 2019-03-17
Drop support for Python 3.4

Kodiologist:
[Its official end-of-life was yesterday.](https://devguide.python.org/)


brandonwillard:
Unrelated question: is there a reason we aren't using `six` for most of those compatibility functions and constants?

Kodiologist:
> Would this be a good time to add explicit 3.7 tests?

I don't think we're missing any tests, but we're not testing 3.7 on Travis for [historical reasons](https://github.com/travis-ci/travis-ci/issues/9069) that probably don't apply anymore. 

> is there a reason we aren't using `six` for most of those compatibility functions and constants?

I don't think there is. But at this point, it's probably not worth changing, because we can start removing Python 2 support after the next release of Hy.

Kodiologist:
@brandonwillard Do clarify, does your hooray emoji indicate you approve this PR?

PR #1762, opened 2019-03-22
add `list?` function to `hy.core`

TristanCacqueray:
`expression?` will test if the input is an instance of HyExpression.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -91,3 +91,4 @@
 * Oskar Kvist <oskar.kvist@gmail.com>
 * Brandon T. Willard <brandonwillard@gmail.com>
 * Andrew R. M. <nixy@nixy.moe>
+* Tristan de Cacqueray <tdecacqu@redhat.com>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -11,6 +11,7 @@ New Features
 ------------------------------
 * Format strings with embedded Hy code (e.g., `f"The sum is {(+ x y)}"`)
   are now supported, even on Pythons earlier than 3.6.
+* New list? function.
 
 Bug Fixes
 ------------------------------
--- a/docs/language/core.rst
+++ b/docs/language/core.rst
@@ -899,6 +899,24 @@ Returns the first logically-true value of ``(pred x)`` for any ``x`` in
    True
 
 
+.. _list?-fn:
+
+list?
+-----
+
+Usage: ``(list? x)``
+
+Returns ``True`` if *x* is a list.
+
+.. code-block:: hy
+
+   => (list? '(inc 41))
+   True
+
+   => (list? '42)
+   False
+
+
 .. _string?-fn:
 
 string?
--- a/hy/contrib/hy_repr.hy
+++ b/hy/contrib/hy_repr.hy
@@ -18,7 +18,7 @@
 
 (setv -registry {})
 (defn hy-repr-register [types f &optional placeholder]
-  (for [typ (if (instance? list types) types [types])]
+  (for [typ (if (list? types) types [types])]
     (setv (get -registry typ) (, f placeholder))))
 
 (setv -quoting False)
--- a/hy/contrib/walk.hy
+++ b/hy/contrib/walk.hy
@@ -19,7 +19,7 @@
     (outer (HyExpression (map inner form)))]
    [(instance? HyDict form)
     (HyDict (outer (HyExpression (map inner form))))]
-   [(instance? list form)
+   [(list? form)
     ((type form) (outer (HyExpression (map inner form))))]
    [(coll? form)
     (walk inner outer (list form))]
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -203,6 +203,9 @@ Return series of accumulated sums (or other binary function results)."
   "Check if x is float."
   (isinstance x float))
 
+(defn list? [x]
+  (isinstance x list))
+
 (defn symbol? [s]
   "Check if `s` is a symbol."
   (instance? HySymbol s))
@@ -454,7 +457,7 @@ Even objects with the __name__ magic will work."
     disassemble drop drop-last drop-while empty? eval even? every? exec first
     filter flatten float? fraction gensym group-by identity inc input instance?
     integer integer? integer-char? interleave interpose islice iterable?
-    iterate iterator? juxt keyword keyword? last macroexpand
+    iterate iterator? juxt keyword keyword? last list? macroexpand
     macroexpand-1 mangle map merge-with multicombinations name neg? none? nth
     numeric? odd? partition permutations pos? product range read read-str
     remove repeat repeatedly rest reduce second some string string? symbol?
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -213,7 +213,7 @@ Such 'o!' params are available within `body` as the equivalent 'g!' symbol."
   (defn extract-o!-sym [arg]
     (cond [(and (symbol? arg) (.startswith arg "o!"))
            arg]
-          [(and (instance? list arg) (.startswith (first arg) "o!"))
+          [(and (list? arg) (.startswith (first arg) "o!"))
            (first arg)]))
   (setv os (list (filter identity (map extract-o!-sym args)))
         gs (lfor s os (HySymbol (+ "g!" (cut s 2)))))
--- a/tests/native_tests/core.hy
+++ b/tests/native_tests/core.hy
@@ -267,6 +267,11 @@ result['y in globals'] = 'y' in globals()")
   (assert-true (symbol? 'im-symbol))
   (assert-false (symbol? (name 'im-symbol))))
 
+(defn test-list? []
+  "NATIVE: testing the list? function"
+  (assert-false (list? "hello"))
+  (assert-true (list? [1 2 3])))
+
 (defn test-gensym []
   "NATIVE: testing the gensym function"
   (import [hy.models [HySymbol]])


Kodiologist:
Looks good. I'd rename it to `expr?`, and move the changes of preexisting code from `(isinstance FOO HyExpression)` to `(expr? FOO)` into their own commit.

TristanCacqueray:
I didn't realized `symbol?` was derived from Clojure. There doesn't seem to be an `expression?` built-in in Clojure, however there is a `list?` one that would work for my use-case.

On the other hand, if I understand correctly, hy expression are not exactly python list and python list cannot be expressed as hy expression (e.g. `'(42 44)`). HyExpression just happen to be an instance of list.

Kodiologist:
Right, even a `HyList` (`` `[1 2 3] ``) is a different thing from a `list` (`[1 2 3]`).

TristanCacqueray:
Fwiw, here is the procedure where I am using `expression?` https://gitlab.com/TristanCacqueray/hy2glsl/blob/master/hy2glsl/translator.hy

Kodiologist:
For complicated model processing like that, you may be better off with [model patterns](http://docs.hylang.org/en/stable/language/model_patterns.html) than manual manipulation.

brandonwillard:
> I didn't realized `symbol?` was derived from Clojure. There doesn't seem to be an `expression?` built-in in Clojure, however there is a `list?` one that would work for my use-case.

In general, additions to core language features and built-in functions need to be judiciously chosen.  This project has been using the following guidelines to make those choices: Is it an expected or core Python and/or Lisp function/feature, or does it provide an oft required functionality that is otherwise cumbersome to re-implement locally.

Regarding `symbol?`, that's a fairly standard Lisp built-in.  As I understand it, when it comes to anything more than basic Lisp conventions, we currently lean toward Clojure.  Otherwise, if there is a core feature/function that doesn't follow these guidelines, that's not justification for adding another.  

As a matter of fact, if you find any particularly superfluous core language functions, feel free to submit an issue for their removal.

That said, `list?` does seem to fit, and I can't say that I've noticed a particularly strong need for a built-in `expression?` (or predicates for any other language model objects).

> ... python list cannot be expressed as hy expression (e.g. `'(42 44)`)...

Sorry, I don't understand that statement.

Kodiologist:
Whether a feature is included in Clojure isn't a sensible criterion for whether to include it in Hy. What made sense in their situation need not make sense in ours. In fact, we may have such a situation right here: Clojure presumably has `list?` but not `expression?` because of a lack of a distinction between lists and expressions, whereas Hy has such a distinction. More to the point, there's no good reason to believe a priori that any decision made by the designers of Clojure (or any other programming language) was actually a good decision. Let's consider features on their merits, instead of whether other people liked them enough to put them in their language.

brandonwillard:
> Whether a feature is included in Clojure isn't a sensible criterion for whether to include it in Hy.

You're confusing _guildelines_ with **strict requirements**.  For that reason, I'm not sure what your arguments are directed toward.

> Let's consider features on their merits, instead of whether other people liked them enough to put them in their language.

Better yet, let's start by not flippantly accusing people of ignoring merits&mdash;and especially not when the discussion they've introduced is explicitly about assessing the merit of additions. 

Otherwise, it seems like you're arguing against the merits of programming language conventions and communities (which sounds familiar).  If `expression?` is a language built-in that other Lisp users&mdash;including Clojure&mdash;expect, then that alone is a reason to consider it.  If it's not, then it doesn't have that form of merit and needs others.

I'll repeat (and rephrase a little): `list?` is good, but why the more narrow `expression?`?  Because `list` and `HyExpression` are different isn't enough "merit" alone, because, if it were, we could justify predicates for all the model objects&mdash;and more.

Kodiologist:
But it's not a sensible guideline, either. Like, it's just not a good reason.

> Otherwise, it seems like you're arguing against the merits of programming language conventions and communities (which sounds familiar). 

Right. The sentence right after ("If `expression?` is a language built-in that other Lisp users—including Clojure—expect, then that alone is a reason to consider it") is the kind of reasoning I'm complaining about. Otherwise we end up with features that have no practical purpose and are there only to be like other languages, like the old HyCons. And conversely, we'd be biased against potential features merely because they're novel.

TristanCacqueray:
> > ... python list cannot be expressed as hy expression (e.g. `'(42 44)`)...
> 
> Sorry, I don't understand that statement.

In Scheme for example, `'(42 44)` results in a `(42 44)` list identical to a `(list 42 44)` evaluation.
In Hy this results in `HyExpression([HyInteger(42), ...])` which seems different from `[42 44]`.

For my initial use-case (which is to explore the code as data paradigm), the difference doesn't matter and `list?` actually looks better as it doesn't imply there is a difference between code and data.

As a new Lisp apprentice I got confused by the representation of Hy model objects. I didn't meant to cause a precedence or diverge Hy core from the other Lisp :-) There may be value in a `expression?` built-in or perhaps HyExpression should be replaced with HyList, but that's beyond the scope of this proposal. Thank you for your time and comments.

brandonwillard:
We have a confusingly named `coll?` built-in that almost works, but it deals with the broader class of non-string iterables; however, I think there's room for a proper `list?`.  If you want, just change `expression?` to implement `list?`.

TristanCacqueray:
@brandonwillard the last commit (  6082e86 ) does add a proper `list?` already i think. `coll?` is too broad indeed.

Kodiologist:
Beware that `(list? '[a b c]))` and `(list? '(im-expression))` are true under the current implementation of  `HySequence` (which is the base type of `HyList` and `HyExpression`), but they're presumably going to become false eventually. `HySequence` is currently a subclass of `list`, but #1542 would probably change that.

TristanCacqueray:
That's good to know, thus when the model become immutable, if HyExpression are no longer an instance of list, then it would be a good time to revisit adding an `expression?` built-in :-)

brandonwillard:
> But it's not a sensible guideline, either. Like, it's just not a good reason.

As a **professed Lisp dialect**, the Lisp relevance of an addition or change in Hy is **necessarily** a good reason to consider something.

> Right. The sentence right after ("If expression? is a language built-in that other Lisp users—including Clojure—expect, then that alone is a reason to consider it") is the kind of reasoning I'm complaining about. Otherwise we end up with features that have no practical purpose and are there only to be like other languages, like the old HyCons. And conversely, we'd be biased against potential features merely because they're novel.

You're arguing the same non-sequitur as before, and you're not addressing a real problem.  If you think there's novelty that out-weighs valid concerns for a poorly organized and redundant core library, then the onus is on you to demonstrate it.  Do that instead of making protracted arguments against the stated purpose of the project (i.e. to provide a Lisp dialect compatible with Python).

Also, please review the `cons` discussion; if anything, it shows exactly why Lisp conventions and communities are worth considering.  In this case, such considerations help address the extremely myopic concept of "purpose" that individuals alone can have..

Kodiologist:
> That's good to know, thus when the model become immutable, if `HyExpression` are no longer an instance of list, then it would be a good time to revisit adding an expression? built-in :-)

Okay. For the time being, I'd remove the tests

     (assert-true (list? '[a b c]))
     (assert-true (list? '(im-expression)))

and the examples with quoted forms in the documentation. I also think `(assert-false (list? (name '(im-expression))` is kind of a weird test because `name` returns a string, and for strings, you already have the more straightforward  `(assert-false (list? "hello"))`.

> As a **professed Lisp dialect**, the Lisp relevance of an addition or change in Hy is **necessarily** a good reason to consider something.

I don't follow. I am my father's son, but it doesn't follow that I should take his advice. Analogously, while Hy is obviously inspired by previous Lisps, that's no reason it needs to be constrained by them; it is its own language.

> Do that instead of making protracted arguments against the stated purpose of the project (i.e. to provide a Lisp dialect compatible with Python).

But my dude, you were the one who began the discussion of this feature in terms of whether it was similar to or different from other languages, and you're continuing to argue with me about it. If you don't want to talk about it, then don't talk about it.

> Also, please review the `cons` discussion; if anything, it shows exactly why Lisp conventions and communities are worth considering. In this case, such considerations help address the extremely myopic concept of "purpose" that individuals alone can have..

How do you figure that by rereading a discussion in which I participated, I'll come to the opposite point of view that I argued there and I'm repeating here?

Kodiologist:
Hey, Tristan, sorry I didn't notice your update. Maintainers got a notification for new commits, but not for force-pushes. Could you split this into two commits, one that introduces the function, updates the documentation, etc. and one that changes existing code from `(isinstance FOO list)` to `(list? FOO)`? Then add a third commit adding yourself to AUTHORS. Then I think we're good to merge.

TristanCacqueray:
No worries, I'm actually not used to the GitHub system, this PR should have 3 commits now.

Kodiologist:
Welcome aboard!

brandonwillard:
@Kodiologist, why are you merging this without approval from all of the assigned reviewers?

Kodiologist:
That's not exactly how our merge policy works. If you'd meant to veto it, sorry; I inferred from https://github.com/hylang/hy/pull/1762#issuecomment-476456861 that you weren't against it. If I'm mistaken, I'll undo the merge.

brandonwillard:
I don't want to hold this up anymore, so&mdash;from a quick look&mdash;I suppose it's fine to keep.

Otherwise, I think we should consider renaming `coll?` and/or adding a complementary `tuple?`.  This PR would've been a good place to address those, but I suppose someone can always put in another.

Kodiologist:
Yes, it does feel weird to have `list?` without `tuple?`.

TristanCacqueray:
@Kodiologist thanks :-)

`tuple?` is now proposed here: https://github.com/hylang/hy/pull/1770

PR #1767, opened 2019-04-02
Allow Hy to run when installed in read only filesystem

nixy:
The `dir` parameter removed by this commit caused hy to try to create a
temporary file where it is installed. This caused problems for the NixOS
package as it would be installed in the Nix store which is a read-only
filesystem.

I ran the tests and most of them passed. I rolled back my branch to master and ran them again and it appears the tests that failed were also failing on master. I've used hy with these changes and am fairly confident that they don't cause any issues.

Information on the issues facing the NixOS package can be found here: https://github.com/NixOS/nixpkgs/pull/57882

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -90,3 +90,4 @@
 * Yigong Wang <wang@yigo.ng>
 * Oskar Kvist <oskar.kvist@gmail.com>
 * Brandon T. Willard <brandonwillard@gmail.com>
+* Andrew R. M. <nixy@nixy.moe>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -12,6 +12,11 @@ New Features
 * Format strings with embedded Hy code (e.g., `f"The sum is {(+ x y)}"`)
   are now supported, even on Pythons earlier than 3.6.
 
+Bug Fixes
+------------------------------
+* Fixed a crash caused by errors creating temp files during bytecode
+  compilation
+
 0.16.0
 ==============================
 
--- a/hy/importer.py
+++ b/hy/importer.py
@@ -467,9 +467,10 @@ else:
         if cfile is None:
             cfile = cache_from_source(filename)
 
-        f = tempfile.NamedTemporaryFile('wb', dir=os.path.split(cfile)[0],
-                                        delete=False)
+        f = None
         try:
+            f = tempfile.NamedTemporaryFile('wb', dir=os.path.split(cfile)[0],
+                                            delete=False)
             f.write('\0\0\0\0')
             f.write(struct.pack('<I', timestamp))
             f.write(marshal.dumps(codeobject))
@@ -489,7 +490,8 @@ else:
             os.rename(f.name, cfile)
         except OSError:
             try:
-                os.unlink(f.name)
+                if f is not None:
+                    os.unlink(f.name)
             except OSError:
                 pass
 


Kodiologist:
Looks good, but don't forget to request pulls from a feature branch instead of master.

brandonwillard:
Besides disabling bytecode compilation, it looks like you could move the `NamedTemporaryFile` line within the `try` and avoid propagating the `OSError`.

nixy:
I'd like to avoid disabling bytecode compilation altogether. Even though the location where hy itself is located might be on a read-only filesystem the file you want to actually compile might be on a filesystem that you can write to.

It would be easy to wrap the Nix package to add `-B` to all calls to hy but I'd like to avoid that if possible.

> it looks like you could move the NamedTemporaryFile line within the try and avoid propogating the OSError

I thought I would give this a shot, but it hasn't gone super well (I'm no python expert).

Once the tempfile is moved within the try block, the variable `f` is no longer in scope for the except block. Not really sure how to work around this right now, but here is what I got.

```
        try:
            f = tempfile.NamedTemporaryFile('wb', dir=os.path.split(cfile)[0],
                                            delete=False)
            f.write('\0\0\0\0')
            f.write(struct.pack('<I', timestamp))
            f.write(marshal.dumps(codeobject))
            f.flush()
            f.seek(0, 0)
            f.write(imp.get_magic())

            # Make sure it's written to disk.
            f.flush()
            os.fsync(f.fileno())
            f.close()

            # Rename won't replace an existing dest on Windows.
            if os.name == 'nt' and os.path.isfile(cfile):
                os.unlink(cfile)

            os.rename(f.name, cfile)
        except OSError:
            try:
                os.unlink(f.name)
            except OSError:
                pass
```

And here is what I get when I run it (and it fails)

```
+$ hy
  File "/nix/store/mir9lqblvnnmxhkd7i78fh523w89fwj1-hy-0.16.0/bin/.hy-wrapped", line 11, in <module>
    sys.exit(hy_main())
  File "/nix/store/mir9lqblvnnmxhkd7i78fh523w89fwj1-hy-0.16.0/lib/python2.7/site-packages/hy/cmdline.py", line 589, in hy_main
  File "/nix/store/mir9lqblvnnmxhkd7i78fh523w89fwj1-hy-0.16.0/lib/python2.7/site-packages/hy/cmdline.py", line 583, in cmdline_handler
    return run_repl(spy=options.spy, output_fn=options.repl_output_fn)
  File "/nix/store/mir9lqblvnnmxhkd7i78fh523w89fwj1-hy-0.16.0/lib/python2.7/site-packages/hy/cmdline.py", line 428, in run_repl
    hr = HyREPL(**kwargs)
  File "/nix/store/mir9lqblvnnmxhkd7i78fh523w89fwj1-hy-0.16.0/lib/python2.7/site-packages/hy/cmdline.py", line 236, in __init__
    self.hy_compiler = HyASTCompiler(self.module)
  File "/nix/store/mir9lqblvnnmxhkd7i78fh523w89fwj1-hy-0.16.0/lib/python2.7/site-packages/hy/compiler.py", line 385, in __init__
    load_macros(self.module)
  File "/nix/store/mir9lqblvnnmxhkd7i78fh523w89fwj1-hy-0.16.0/lib/python2.7/site-packages/hy/macros.py", line 233, in load_macros
    builtin_mod = importlib.import_module(builtin_mod_name)
  File "/nix/store/ih03q5vi6kj278l4jk4kh41lqyagzzdi-python-2.7.15/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/nix/store/mir9lqblvnnmxhkd7i78fh523w89fwj1-hy-0.16.0/lib/python2.7/site-packages/hy/importer.py", line 247, in load_module
    self.exec_module(mod, fullname=fullname)
  File "/nix/store/mir9lqblvnnmxhkd7i78fh523w89fwj1-hy-0.16.0/lib/python2.7/site-packages/hy/importer.py", line 212, in exec_module
    code = self.get_code(fullname)
  File "/nix/store/mir9lqblvnnmxhkd7i78fh523w89fwj1-hy-0.16.0/lib/python2.7/site-packages/hy/importer.py", line 322, in get_code
    self.code = self.byte_compile_hy(fullname)
  File "/nix/store/mir9lqblvnnmxhkd7i78fh523w89fwj1-hy-0.16.0/lib/python2.7/site-packages/hy/importer.py", line 294, in byte_compile_hy
    hyc_compile(code, module=fullname)
  File "/nix/store/mir9lqblvnnmxhkd7i78fh523w89fwj1-hy-0.16.0/lib/python2.7/site-packages/hy/importer.py", line 492, in hyc_compile
    os.unlink(f.name)
UnboundLocalError: local variable 'f' referenced before assignment
```

brandonwillard:
> Even though the location where hy itself is located might be on a read-only filesystem the file you want to actually compile might be on a filesystem that you can write to.

It shouldn't fail in that case&mdash;if it does, then that's a separate issue worth submitting!  

The current failure is likely due to a missing or outdated (relative to the source `.hy`/`.py`) `.pyc` file in the initial installation.  The wheel could be broken or the underlying filesystem could be doing something to the timestamps used by the Python bytecode checks.  

Regardless, the bytecode compilation should be failing, but, in your case, it appears to be failing at the temporary file step, and not later where the exceptions are caught and ignored.

> Once the tempfile is moved within the try block, the variable `f` is no longer in scope for the except block.

Ah, yeah, you can just add a line like `f = None` before the `try` and then guard that `os.unlink` with a `f is not None` check.

nixy:
Fixed the code to just move the temporary file creation into the try block as discussed. I think this should be good now?

> The wheel could be broken or the underlying filesystem could be doing something to the timestamps used by the Python bytecode checks.

I'm going to guess that this is Nix's doing as it builds thing in an environment where timestamps are set to a specific time to try and ensure reproducibility (sometimes timestamps get built into code).

Running stat on the pyc files where they are installed gives me this (which I believe confirms my suspicion):

```
  File: version.pyc
  Size: 167             Blocks: 8          IO Block: 4096   regular file
Device: fe02h/65026d    Inode: 2959880     Links: 1
Access: (0444/-r--r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2019-04-04 17:52:22.594010025 -0400
Modify: 1969-12-31 19:00:01.000000000 -0500
Change: 2019-04-04 17:52:22.592010019 -0400
```

Kodiologist:
Okay, how about a NEWS and AUTHORS line?

PR #1772, opened 2019-04-09
Support Python 3.8

Kodiologist:


refi64:
Everything here is a 👍 from me...except `setv:`. I like the idea, but the name just seems really...odd. 

Random thought, could we make setv automatically use `:=` in an expression context? Though then we're adding some magic behavior that people may not notice fails under older Py3 versions, unless it's guarded behind a macro or something...

Kodiologist:
I thought of `setv*`, but I figured `setv:` would be more mnemonic by matching the name of the Python operator. I considered `:=`, but that would require special-casing the identifer `:=` to not be interpreted as a keyword. Any suggestions?

I tried overloading `setv` first, but this led to printing the result of an assignment at the REPL, which I think we should avoid in case the user is assigning a very large object, like I do a lot in data analysis.

refi64:
Hmm...

`setv*` seems a bit like a lot of our intended-for-internal-use ones, so that probably wouldn't work either, and `:=` looks a bit out of place compared to everything else we've got. I'm tempted to say `setvr` (setv returning value), but that could be my midnight brain talking.

> I tried overloading `setv` first, but this led to printing the result of an assignment at the REPL, which I think we should avoid in case the user is assigning a very large object, like I do a lot in data analysis.

Ah man, tempted to say to special case it, but again that could also be midnight brain and maybe not worth the cost. I do wonder how other languages with repls & assignment-as-an-expression handle this...

Kodiologist:
I like `setvr`.

I don't think the REPL can be special-cased properly because of things like defining a function that ends with `setv` in a file and then importing that file. The `setv` has already been compiled, so it's too late to make it `=` instead of `:=` when you call it in the REPL.

But a more general issue is that peeking to see whether the `setv`'s return value is used at compile-time sounds complex and difficult to get right. The first thing I tried didn't actually involve peeking. Instead, I made every `setv` into `:=` when legal (namely, when there's exactly two arguments and the lvalue is a plain symbol). This isn't ideal because the generated Python code uses `:=` heavily where `=` would be more idiomatic. There are also going to be some subtle gotchas about scoping in comprehensions. At the end of the day, this is probably a poor road to go down to give the user the tiny convenience of being able to write `:=` with `setv` instead of `setvr` or whatever.

yuhan0:
What about `setx`? ("set expression")
It aligns with the official Python name for this feature, "Assignment Expressions", also "setvr" is really awkward to type and sounds a little more opaque.

Kodiologist:
I like `setx` best so far.

refi64:
+1 to setx.

Kodiologist:
I've changed it to `setx`.

PR #1776, opened 2019-05-19
Hy 0.18.0

Kodiologist:
With this done, we can finally start ripping out Python 2 support. Exciting, eh?

Recall that since this is a release, the two-week merge rule does not apply.


refi64:
Like the sound of Python 2, hopefully this summer I'll finally be able to get back into Hy.

PR #1777, opened 2019-05-20
Drop support for Python 2

### LINKED ISSUE #1217: Drop support for Python 2

Kodiologist:
Closes #1217.

The code has a net shrinkage of over 700 lines.


Kodiologist:
There's still a test failure on Python 3.8 that I'm pretty sure has nothing to do with this PR. I can't replicate it on my own machine, so I think it's due to a change to Python since the release of 3.8.0a4. 3.8.0b1 should be out pretty soon, so let's just wait and see.

Kodiologist:
@brandonwillard, you may be interested in looking at [the test failure](https://travis-ci.org/hylang/hy/builds/541415078) because it's from one of your tests of macro scoping, `test-macro-namespace-resolution`.

brandonwillard:
OK, will do.

Kodiologist:
I think I got it. See #1784.

PR #1782, opened 2019-05-24
Fix versionadded directive

krysros:


Kodiologist:
It would be better to just remove all the `versionadded` directives. The documentation doesn't otherwise explain how Hy has changed; it just describes the current version. The very old `versionadded` directives misleadingly suggest otherwise.

refi64:
+1 to removing them, they're pretty useful in general, but given that Hy is pre-beta there's way too much shift between versions to summarize with these. Post-1.0, yeah they should be present IMO.

Kodiologist:
I agree. Keeping track of version changes in the documentation could make sense past 1.0.

PR #1791, opened 2019-07-06
 Run statements in the second argument of `assert`

### LINKED ISSUE #1390: Statements pulled out of the second argument of (assert) aren't added to the AST
$ echo '(assert (do (f1) (f2)) (do (f3) (f4)))' |  hy2py
    f1()
    assert f2(), f4()

Given [how the Python language reference describes `assert`](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement), I guess this should compile to:

```Python
if __debug__:
    f1()
    if not f2():
        f3()
        assert not 1, f4()
```

Kodiologist:
Fixes #1390.

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -402,21 +402,11 @@ class HyASTCompiler(object):
         ret = Result()
         for module, names in self.imports.items():
             if None in names:
-                e = HyExpression([
-                        HySymbol("import"),
-                        HySymbol(module),
-                    ]).replace(expr)
-                ret += self.compile(e)
+                ret += self.compile(mkexpr('import', module).replace(expr))
             names = sorted(name for name in names if name)
             if names:
-                e = HyExpression([
-                        HySymbol("import"),
-                        HyList([
-                            HySymbol(module),
-                            HyList([HySymbol(name) for name in names])
-                        ])
-                    ]).replace(expr)
-                ret += self.compile(e)
+                ret += self.compile(mkexpr('import',
+                    mklist(module, mklist(*names))))
         self.imports = defaultdict(set)
         return ret.stmts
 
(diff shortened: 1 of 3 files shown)


hy/compiler.py
```diff
@@ -324,6 +324,18 @@ def is_unpack(kind, x):
             and x[0] == "unpack-" + kind)
 
 
+def make_hy_model(outer, x, rest):
+   return outer(
+      [HySymbol(a) if type(a) is str else
+              a[0] if type(a) is list else a
+          for a in x] +
+      (rest or []))
+def E(*items, **kwargs):
```
refi64: I can understand the rationale for having these factored out and to be concise, but I'm not a fan of single-character names for global functions...

Kodiologist: What about `HE` and `HL`, for `HyExpression` and `HyList`?

refi64: Ehh, I just don't want the compiler to have an overabundance of these short-named functions. I guess the primary reason this exists is just to let us write Hy to be compiled inside of the compiler, in which case IMO there are probably more elegant ways to handle it (I've been toying with some ideas that I'll probably file in a separate issue/rfc). Just...not a fan of these functions for now.

Kodiologist: I think only two are necessary, which doesn't seem like an overabundance to me. But to be clear, what do you need from me for this PR?

refi64: Err true 😅 Could this change just be separate from the assert fix for now?

Kodiologist: But the assert fix uses the new functions. Are you asking me to rewrite it without them?

refi64: Gah, could we just do something like `mkexpr` or similar for now, not quite as short but shorter than `HyExpression` with manually passing everything over.

Kodiologist: Done.

PR #1792, opened 2019-07-07
Rework statements in while condition

### LINKED ISSUE #1790: The condition of a `while` may be compiled twice
```
=> (defmacro m [] (print "hi") '(do (setv x 1) 0))
<function m at 0x7f09efca4620>
=> (while (m) 2)
hi
hi
```

I created this bug in #1433. Oops.

refi64:
Fixes #1790. This avoids compiling them more than once while also applying some simplification.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -12,6 +12,7 @@ Removals
 Bug Fixes
 ------------------------------
 * Statements in the second argument of `assert` are now executed.
+* Fixed the expression of a while loop that contains statements being compiled twice.
 
 0.17.0
 ==============================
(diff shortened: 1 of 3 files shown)


refi64:
@Kodiologist Okay fixed, I moved the test to test-while-loop (where it makes more sense). I got the variable test to work nicely-ish too.

refi64:
Done, also I fixed the test because it *didn't actually have a statement inside the condition* (facepalm insert here).

Kodiologist:
It's always a good idea to check not only that the test for a fix passes with the fix, but also fails without the fix.

Kodiologist:
Can you rebase now that I've introduced a merge conflict?

refi64:
@Kodiologist done.

Kodiologist:
Thanks. I'm afraid you've got a duplicate "Bug Fixes" section in NEWS.

Kodiologist:
Fixed it.

PR #1793, opened 2019-07-08
Pre-Python 3.4 cleanup (including Python 2)

refi64:
- We can just hardcode the `*3` scripts, no Python 2 left.
- Remove remaining uses of `imp`. The previous code incorrectly assumed the functions used (`spec_from_file_location` and `module_from_spec`) were unavailable prior to Python 3.6, whereas they were added in 3.4 and 3.5, respectively. (The former *was* modified to take path-like objects in 3.6, which is where the confusion may have come from).

### DIFF
--- a/hy/__main__.py
+++ b/hy/__main__.py
@@ -11,5 +11,5 @@ import sys
 
 if len(sys.argv) > 1:
     sys.argv.pop(0)
-    imp.load_source("__main__", sys.argv[0])
+    hy.importer._import_from_path('__main__', sys.argv[0])
     sys.exit(0)  # right?
--- a/hy/importer.py
+++ b/hy/importer.py
@@ -157,3 +157,11 @@ runpy = importlib.import_module('runpy')
 
 _runpy_get_code_from_file = runpy._get_code_from_file
 runpy._get_code_from_file = _get_code_from_file
+
+
+def _import_from_path(name, path):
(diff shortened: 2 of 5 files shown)


refi64:
Ah that's because they were indeed passing a path to `spec_from_file_location`, just wrapping that with str should fix the test failure.

Kodiologist:
Nice. Shall we just remove `hy3`, `hyc3`, and `hy2py3`?

refi64:
I left the `*3` overloads to sort of align with [PEP 394](https://www.python.org/dev/peps/pep-0394/), which says that e.g. scripts should be explicit in the version they use in their hashbangs, though I guess that applies a bit less to us.

Kodiologist:
It's appropriate that we use `python3` in our shebangs, but the recommendation that "The Python 3.x `idle`, `pydoc`, and `python-config` commands should likewise be available as `idle3`, `pydoc3`, and `python3-config`; Python 2.x versions as `idle2`, `pydoc2`, and `python2-config`" is presumably given because versions of these programs exists for both Pythons 2 and 3, whereas Hy is Python 3-only.

refi64:
Actually it was also to be ready for the ever-mythical Python 4, there was a discussion about Linux packaging on the Zulip a few months ago that went over it.

Kodiologist:
Okay, I guess it shouldn't really matter, anyway.

PR #1804, opened 2019-08-03
Immutable Hy models

### LINKED ISSUE #1542: Consider making Hy models immutable.

Kodiologist:
Closes #1542.

The key change here is a small one: changing `HySequence` to inherit from `tuple` instead of `list`. All the other changes are support or unrelated cleanup.

Hy models are still mutable in their attributes, which is convenient for `.replace`. I think that removing the ability to change which elements a `HySequence` contains is the main thing we need to avoid nasty surprises from model mutation.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -16,6 +16,12 @@ New Features
 * All augmented assignment operators (except `%=` and `^=`) now allow
   more than two arguments.
 
+Other Breaking Changes
+------------------------------
+* ``HySequence`` is now a subclass of ``tuple`` instead of ``list``.
+  Thus, a ``HyList`` will never be equal to a ``list``, and you can't
+  use ``.append``, ``.pop``, etc. on an expression or list.
+
 Bug Fixes
 ------------------------------
 * Statements in the second argument of `assert` are now executed.
--- a/docs/language/internals.rst
+++ b/docs/language/internals.rst
@@ -60,6 +60,10 @@ Adding a HySequence to another iterable object reuses the class of the
 left-hand-side object, a useful behavior when you want to concatenate Hy
 objects in a macro, for instance.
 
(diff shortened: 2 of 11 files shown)


Kodiologist:
It's not that bad. In fact, a lot of the uses of `replace` already assign back to the same variable like that. I did try rewriting `replace` in a side-effect-free manner and ran into surprising problems, but that was a while ago and it might be easier now.

jjoocceeee:
After this change, is there now no way to use a python equivalent of List in Hy? Or what work around would you suggest?

brandonwillard:
These changes don't affect the built-in Python list type; they simply make the list type _used to represent Hy's expressions_ immutable.  For example, the list-like object returned by quoted Hy expressions will now be immutable.

Issue #1794, opened 2019-07-08
[RFC] Annotation syntax


refi64:
I've been thinking about the best way to handle annotations for a while, especially with Python 3.8 shipping type comment parsing in the built-in ast module. The main issue is that I want the syntax to be consistent across function definitions, variables, etc.

Here's what I've currently come up with:

```hy
; VARIABLE ANNOTATIONS

(setv x &of int 1)

; GENERIC TYPE SHORTHAND

; Instead of e.g. (get List int) for List[int] which sucks
(setv x &of (of List int) 1)

; FUNCTION ANNOTATIONS
(defn myfunc [x &of int y &of str] ...)
; Should we use nested braces instead for consistency with defaults?
; Arguably yes but now we lose consistency with setv.
(defn myfunc [[x &of int 2]])  ; like def myfunc(x: int = 2)
; Or we could only require nested braces for said defaults, e.g.
(defn myfunc [x &of int y &of int [z &of int 2]])  ; def myfunc(x: int, y: int, z: int = 2)

; TYPE COMMENT ALTERNATIVES

; Type comments suck with Hy's style so let's try something else.
; On Python 3.6 we can just lift these out to variable annotations but 3.5 might be messier
(for [[x &of int y &of int] pairs] ...)
(with [(thing) x &of int] ...)

; However we still need to support a # type: ignore equivalent, maybe:
(#ignore
  ...)
```

Some random thoughts:

- This could probably be shipped in a "tech preview" state at first to give us some time to try it out and amend what's necessarily, while letting people know that a bit of potential breakage wouldn't be unexpected.
- Do we really want to support 3.5? 3.6 has already been out for a while, and this would likely be far simpler if we can restrict what needs to be compiled to type comments to just `# type: ignore`. Heck, outside of `# type: ignore`, everything shown here should would in pre-3.8 *without* 
requiring typed_ast.

Kodiologist:
I haven't really been in the loop on the whole typing annotations thing. I understand that Python itself ignores them, so what software is currently in use that uses them?

refi64:
Function and variable annotations are still present in various forms at runtime, the part that Python ignores is the *meaning* of the annotations. That's interpreted by type checking tools.

There are two reasons I think this would be great:
- It would make room for static types in Hy. Sure we could argue about whether that fits with the Lisp ideology, but it would probably be useful for many people.
[...]

PR #1810, opened 2019-08-12
Implement PEP 3107 & 526 annotations

refi64:
Ref. #1794. I'm going to split this up into more commits (partly since I ended up majorly reworking the function argument parsing code) but figured I should get the initial WIP out there for a basic overview. The syntax used is identical to the accepted proposal in #1794.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -17,6 +17,7 @@ New Features
   inline Python code.
 * All augmented assignment operators (except `%=` and `^=`) now allow
   more than two arguments.
+* PEP 3107 and PEP 526 function and variable annotations are now supported.
 
 Other Breaking Changes
 ------------------------------
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -8,6 +8,48 @@ Hy features a number of special forms that are used to help generate
 correct Python AST. The following are "special" forms, which may have
 behavior that's slightly unexpected in some situations.
 
+^
+-
+
+The ``^`` symbol is used to denote annotations in three different contexts:
+
+- Standalone variable annotations.
+- Variable annotations in a setv call.
+- Function argument annotations.
+
+They implement `PEP 526 <https://www.python.org/dev/peps/pep-0526/>`_ and
+`PEP 3107 <https://www.python.org/dev/peps/pep-3107/>`_.
+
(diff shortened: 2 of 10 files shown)


Kodiologist:
`model[0] == HySymbol("annotate*")` — Careful, this will trigger on a string literal. You need to check that `model[0]` is a symbol.

refi64:
Oof, will change with the next push.

Kodiologist:
Is this ready for review, then?

refi64:
Was just typing out that it was indeed ready for review and then got distracted force-pushing a local tweak. :sweat_smile: 

So yeah, AFAIK this should be good to go, all tests pass locally.

refi64:
Okay so I've resolved most of this and am working on docs now (will probably finish tomorrow).

> Design-wise, it seems a bit odd to me that for function parameters with default values, the annotation is put inside the square brackets, whereas there are no square brackets for other parameters. `Does (defn f [^int p1 p2 ^str p3 &optional ^str o1 ^int [o2 0]] …)` look too weird to you?

I'm indifferent tbh, I followed that syntax because it's what was originally proposed in #1794. That being said, one con of changing it would be that you have items relating to the argument being defined inside *and* outside the brackets.

Kodiologist:
Don't forget about this, my dude. You've had several great PRs that floundered near the end and didn't make it in.

Kodiologist:
Also, I would vote for putting the annotation outside the brackets for default arguments.

refi64:
I swear this will be done on Monday at the latest...

refi64:
I think I covered everything and also added the `of` macro discussed in the same issue & docs.

Kodiologist:
All looks good except that:

- The trailing-whitespace fix for `tutorial.rst` shouldn't be in the compiler commit. Put it in its own commit if you want it in this PR.
- Is `whitelisted_roots` left in by mistake?
- I'd argue against including annotations in the tutorial, which should be short and cover only the basics.
- In the reference documentation for annotations, you may want to add some links to the Python documentation that describes their semantics.

refi64:
Done.

Kodiologist:
It looks like 806b04a adds some trailing space that's then removed in the next commit. Try `git log --stat`.

PR #1811, opened 2019-08-13
Overhaul introductory documentation

Kodiologist:
This is long overdue, and the rest of the documentation remains in less-than-great shape, and we still don't test the documentation code or automatically mirror docstrings, but it's progress.

See the commit message of the big commit for more.


README.md
```diff
@@ -1,38 +1,22 @@
...
----------------
+Hy is a Lisp dialect that's embedded in Python. Since Hy transforms its Lisp
+code into Python abstract syntax tree (AST) objects, you have the whole
+beautiful world of Python at your fingertips, in Lisp form.
 
-* [Django + Lisp](https://github.com/paultag/djlisp/tree/master/djlisp)
-* [Python `sh` Fun](https://twitter.com/paultag/status/314925996442796032)
-* [Hy IRC Bot](https://github.com/hylang/hygdrop)
-* [miniKanren in Hy](https://github.com/algernon/adderall)
+To install the latest stable release of Hy, just use the command `pip3 install
```
refi64: Side note: do we really want to be recommending pip3 without --user these days? It has a tendency to break...a lot. (Unless you're in a venv I guess, but if you *are* using venvs for apps, you might as well use pipx, right?)

Kodiologist: Oh dang, does it? I haven't noticed.

Kodiologist: Is `pip3 --user` more reliable, or is it just easier to clean up a mess in your home directory?

docs/whyhy.rst
```diff
@@ -0,0 +1,142 @@
...
+the same time, Hy goes to some lengths to allow you to do typical Lisp things
+that aren't straightforward in Python. For example, Hy provides the
+aforementioned mixing of statements and expressions, :ref:`name mangling
+<mangling>` that transparently converts symbols with names like ``valid?`` to
+Python-legal identifiers, and a :ref:`let` macro to provide block-level scoping
+in place of Python's usual function-level scoping.
+
+Overall, Hy, like Common Lisp, is intended to be an unopinionated big-tent
+language that lets you do what you want. If you're interested in a more
+small-and-beautiful approach to Lisp, in the style of Scheme or Clojure, check
```
refi64: Also minor side note: not that familiar with Clojure, but does it actually count for this? (I mean, it has a for loop macro, syntax for creating Java objects and implementing interfaces via reify...)

Kodiologist: That's a good question. I think the answer is that that was the goal, but any language that sees real use (e.g., not R5RS Scheme, but Racket) inevitably accumulates lots of features over its lifetime. Python itself is probably the best example.

README.md
refi64: More reliable; since `sudo pip3` installs to the same directory as distro package managers, it has a tendency to either overwrite the distro packages with wrong versions (broke apt for me back when I was on Ubuntu, not very fun to fix), or cause issues when you then try to install the distro version of a package that pip3 already installed. At this point [Debian defaults to --user](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=725848) (though that patch [also has its own issues](https://github.com/pypa/pip/issues/4222)).

Kodiologist: Okay, good to know. I'll add in the `--user`s and merge.

docs/whyhy.rst
refi64: True, I mean I understand the Scheme comparison, I just feel like Clojure's is a bit more of a stretch.

Kodiologist: Fair enough. I'll change it.

PR #1813, opened 2019-08-16
Avoid a crash when we can't access a history file

### LINKED ISSUE #1812: REPL crashes when ~/.hy-history is inaccessible
I've been working on updating the Hy package in Debian and in running the tests, I've noticed that the `test_bin` tests appear to expect to be able to do something with `~/.hy-history` (not sure if it's trying to read or write) but I had to set `HOME` to a real directory after which they passed.  Looking at the tests, I imagine it's happening because they're interacting with the REPL via stdin so it's trying to load and/or use the history file?

Kodiologist:
Fixes #1812.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -21,6 +21,7 @@ Bug Fixes
 * Statements in the second argument of `assert` are now executed.
 * Fixed the expression of a while loop that contains statements being compiled twice.
 * `hy2py` can now handle format strings.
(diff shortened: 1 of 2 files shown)


refi64:
(Merging now as per the trivial change rule.)

Kodiologist:
Nice, I forgot we had that rule.

PR #1814, opened 2019-08-17
Fixed broken link to Graphviz

josephtoles:
The link http://www.graphviz.org/Home.php results in a 404.


Kodiologist:
Seems fine, but in the future, don't forget to make a feature branch instead of making a pull request from `master`.

PR #1817, opened 2019-09-02
Allow inline Python 

Kodiologist:
I would've sworn that we had an issue requesting this feature, but I can't find it.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -13,6 +13,8 @@ Removals
 
 New Features
 ------------------------------
+* Added special forms ``py`` to ``pys`` that allow Hy programs to include
+  inline Python code.
 * All augmented assignment operators (except `%=` and `^=`) now allow
   more than two arguments.
 
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -1406,18 +1406,6 @@ parameter will be returned.
     True
 
 
-print
------
-
-``print`` is used to output on screen. Example usage:
-
-.. code-block:: clj
-
-    (print "Hello world!")
-
-.. note:: ``print`` always returns ``None``.
-
-
 .. _quasiquote:
 
 quasiquote
--- a/docs/language/api.rst
+++ b/docs/language/api.rst
@@ -1406,6 +1406,44 @@ parameter will be returned.
     True
 
 
+.. _py-specialform:
+
+py
+--
+
+``py`` parses the given Python code at compile-time and inserts the result into
+the generated abstract syntax tree. Thus, you can mix Python code into a Hy
+program. Only a Python expression is allowed, not statements; use
+:ref:`pys-specialform` if you want to use Python statements. The value of the
+expression is returned from the ``py`` form. ::
+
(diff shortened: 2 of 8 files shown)


asemic-horizon:
This is something I didn't even knew I wanted so much. 

Please consider creating something like a Patreon. You're generating so much business value on my end, it's almost criminal that the Hy Cabal (and I know there is one) isn't getting something from it.

brandonwillard:
Does this need to be a special form?  Couldn't it be implemented as a macro instead?

Kodiologist:
Without a new supporting special form, I don't think so. Macros can't return AST objects, just Hy models or objects that can be promoted to models.

brandonwillard:
Ah, I see.  

Well, can we at least drop the quotes requirement?  It would be easier for working with embedded quotes and it might be nicer for syntax support in an editor.

Kodiologist:
You don't have to use quotation marks per se; you just need to use a string literal. So e.g. you can use a bracket string, which would be the natural choice for long blocks of code.

Avoiding the string literal would be possible by implementing this feature in the lexer and parser rather than the compiler. For example, we could have syntax like `#| 1 + 1 |`, with a corresponding new kind of Hy model for chunks of Python code. But, it's hard to see the benefit of this added complexity. I think it won't be any harder to convince e.g. Emacs to syntax-color `(py "...")` than `#| ... |`.

Kodiologist:
Maybe "complexity" is an exaggeration. It wouldn't be that hard. It's just that it wouldn't be nearly as svelte as this 8-line implementation.

Kodiologist:
> Please consider creating something like a Patreon. You're generating so much business value on my end, it's almost criminal that the Hy Cabal (and I know there is one) isn't getting something from it.

I guess the Hy Cabal is mostly me and @refi64 these days. But it changes according to who's available. The old gang I called the Four Horsemen of Hy Maintenance was me, @refi64, @gilch, and @tuturto. Then there are a bunch of other people who have contributed substantively to Hy but aren't very active anymore.

Pending #957 (which, full disclosure, I myself am not going to do anything about soon; I'm much more interested in programming than organizing), you can send money to @paultag and leave it up to him how to spend it. If you're more interested in giving gifts to developers as thanks for ~~my~~ their excellent work than funding Hy development per se, it's probably best to send money directly to the developers in question than going through an organizational account of some kind, since that avoids us having to decide which of us deserves what proportion of your gift.

gilch:
@brandonwillard You could also make a tag macro applied to a bracket string that expands to the special form. The ability to embed other languages without escaping quotes was one of the major reasons for adding bracket strings to begin with.

Kodiologist:
Oh yeah, I should've thought of that.

Kodiologist:
@refi64 Fixed. I haven't tried to make the error message show the context of the error in the Python code, though.

refi64:
IMO that's good enough; I don't think people are going to have overly large inline Python blocks, and this still has the line number which is pretty much the most important part.

PR #1821, opened 2019-09-21
get_version.py: Prefer VERSIONFILE over git-describe(1)

nmeum:
Noticed this while packaging hy for Alpine Linux. On Alpine package
builders git is installed by default and all packages are build in
subdirectories of the package git repository. This causes git_version.py
to output the version of the package repository since it doesn't consult
the VERSIONFILE first.

This commit reverse the logic checking for the VERSIONFILE first and
only falling back to git-describe if the version file doesn't exist. I personally
think that it is more logical to do it that way.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -92,3 +92,4 @@
 * Brandon T. Willard <brandonwillard@gmail.com>
 * Andrew R. M. <nixy@nixy.moe>
 * Tristan de Cacqueray <tdecacqu@redhat.com>
+* Sören Tempel <soeren@soeren-tempel.net>
--- a/get_version.py
+++ b/get_version.py
@@ -5,16 +5,19 @@ import os, subprocess, runpy
 os.chdir(os.path.split(os.path.abspath(__file__))[0])
 VERSIONFILE = os.path.join("hy", "version.py")
 
-try:
-    __version__ = (subprocess.check_output
-                   (["git", "describe", "--tags", "--dirty"])
-                   .decode('ASCII').strip()
-                   .replace('-', '+', 1).replace('-', '.'))
-    with open(VERSIONFILE, "wt") as o:
-        o.write("__version__ = {!r}\n".format(__version__))
+if "HY_VERSION" in os.environ:
+    __version__ = os.environ["HY_VERSION"]
+else:
+    try:
+        __version__ = (subprocess.check_output
+                       (["git", "describe", "--tags", "--dirty"])
+                       .decode('ASCII').strip()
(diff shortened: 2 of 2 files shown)


Kodiologist:
But then the file won't be regenerated if the commits or tags have changed, will it?

nmeum:
Oh, right. I totally forgot that `get_version` also writes the VERSIONFILE. It won't be regenerated indeed.

refi64:
Could we maybe have get_version.py skip running git if some environment variable (e.g. HY_VERSION_NO_GIT) is set?

Kodiologist:
If that would solve your problem, @nmeum, that's certainly doable for us.

nmeum:
The environment variable would be preferred over the git-describe output and the version file would be used if git-describe fails?

If so: That would solve my problem, I can also implement this and update the PR accordingly.

nmeum:
Implemented this, maybe it would be a bit easier though to simply allow overwriting both the version file and the git-describe output using a environment variable. For example, `export HY_VERSION=0.17.0`.

Kodiologist:
That's fine too, if you'd prefer. The new code you wrote looks good except there's no need to pass `VERSIONFILE` as an argument. It can be hard-coded into the function.

nmeum:
> The new code you wrote looks good except there's no need to pass `VERSIONFILE` as an argument.

Fixed.

By the way: Is there any particular reason why the `get_version` file is installed as a data asset in `setup.py`:

https://github.com/hylang/hy/blob/6f3b6ca7358eb60651262e4fbda64eb57107b9a9/setup.py#L63-L65

refi64:
Nit: this looks good now, but the commit message is currently a bit misleading...

Kodiologist:
> By the way: Is there any particular reason why the `get_version` file is installed as a data file in `setup.py`

Yes. See bd7b8bf5e19e7a3dd99d1a98f4c405fadbe34732.

So, do you want to go with this or with your proposal "to simply allow overwriting both the version file and the git-describe output using an environment variable"?

nmeum:
> So, do you want to go with this or with your proposal "to simply allow overwriting both the version file and the git-describe output using an environment variable"?

You don't seem to have a strong opinion on this. I have a slight preference towards this proposal as it seems cleaner to me. I will therefore go ahead and implement it in a second.

> Nit: this looks good now, but the commit message is currently a bit misleading...

Thanks, will rewrite it after updating the changes proposed here.

Kodiologist:
Add yourself to AUTHORS in a separate commit and this is good to merge.

Kodiologist:
Welcome to the Hyve!

PR #1824, opened 2019-10-08
Drop clint for colors in favor of colorama

### LINKED ISSUE #1820: Remove dependency on clint
While packaging hy for [Alpine Linux](https://alpinelinux.org/), I noticed that it depends on [clint](https://github.com/kennethreitz/clint). Which wasn't packaged by Alpine either. I am not a python guy but it seems to me that clint is pretty dead. The repository is archived and the last release is from 2015.

Is there any particular reason why it is still being used, is not possible to use a command line argument parser from the python3 standard library, e.g. argparse, instead?

refi64:
Closes #1820. Not particularly exciting, though I think in some aspects the code actually ends up a tad cleaner than before.

Reason for using colorama: all clint was really used for was color, and colorama is a popular, well-supported, and battle-tested library for handling this.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -31,6 +31,11 @@ Bug Fixes
 * `hy2py` can now handle format strings.
 * Fixed crashes from inaccessible history files.
 
+Misc. Improvements
+------------------------------
+* Drop the use of the long-abandoned `clint <https://github.com/kennethreitz/clint>`_ library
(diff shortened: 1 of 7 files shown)


refi64:
Hmm what section of NEWS should it be in... I guess it would technically  be under "Other Breaking Changes"?

refi64:
Done.

> Perhaps "Misc. Improvements". It's not a breaking change for Hy code, is it?

Who knows, I've probably read the daily WTF too many times.

PR #1825, opened 2019-10-10
Fix a unit test bug on slim Python Docker images

refi64:
If HyASTCompiler is given a string, it imports it and uses it as the execution environment. However, the unit tests gave HyASTCompiler the string 'test', assuming it would create a new test module, when in
reality it would import CPython's test module that is designed for internal use. Slim Docker images don't include this module, therefore the tests would fail to run.


refi64:
(I swear, forgetting to update NEWS will be the death of me.)

Going to just merge now under the trivial change rule, since this is really minor IMO and also is a dependency of something else I'm working on (using GitHub Actions).

Issue #1719, opened 2019-01-01
parse-args macro


alphapapa:
I came up with this macro, which makes `argparse` much easier to use:
```clj
#!/usr/bin/env hy3

(defmacro/g! parse-args [&rest args]
  "Return arguments parsed from sys.argv with argparse according to ARGS.

ARGS should be a list of arguments passed to argparse.ArgumentParser.add_argument."
  `(do
     (import argparse)
     (setv g!parser (argparse.ArgumentParser))
     ~@(lfor arg args
             `(.add-argument g!parser ~@arg))
     (.parse_args g!parser)))

(defmain [&rest args]
  (setv args (parse-args
               ["files" :nargs "+" :help "Files"]
               ["-v" "--verbose" :action "store_true"]))
  (when args.verbose
    (print "verbose"))
  (print (.join "\n" args.files)))
```
Used like:
```
$ script.hy -v file
verbose
file
```
Would this be useful to include in Hy?

Kodiologist:
I like it, except that it should be written as a function (since it need not be a macro), and it should also provide ways to pass arguments to the `ArgumentParser` constructor and to `parse_args`.

alphapapa:
Hi again,

Messing with some Hy again, I'm trying to convert it to a function, but it seems awkward and difficult.  The problem I'm having is that each list of argparse argument options must be passed to the `add-argument` method properly, meaning that non-keyword arguments must be passed as plain arguments, and keyword arguments as keyword arguments.  

Doing it as a macro is easy, as backquoting and splicing constructs the sexp naturally, which Hy then parses as if the sexp were written that way in the first place (the whole point of a macro, of course).

But, as far as I can tell, doing it as a function requires awkward munging of the argument lists.  I'm sure there's a way, but I haven't figured it out yet, and I'm not sure it's worth the trouble when doing it as a macro handles it naturally.  I feel like I would end up writing a lot of code that already exists in the code that provides macros.  For example, I've got this, that almost works (ignore the debugging `print`s):

```hy
(defn parse-args-fn [&rest args]
  "Return arguments parsed from sys.argv with argparse according to ARGS.

ARGS should be a list of arguments passed to argparse.ArgumentParser.add_argument."
  (import argparse)
  (setv parser (argparse.ArgumentParser))
  (lfor arg args
        :do (print "allargs" arg)
        :setv largs ((fn [&rest args &kwargs kwargs] args) arg)
[...]

PR #1830, opened 2019-10-30
Add: parse-args function

alphapapa:
I went through the steps in `CONTRIBUTING`, but please let me know if any changes are required.

Thanks for your work on Hy.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -94,3 +94,4 @@
 * Tristan de Cacqueray <tdecacqu@redhat.com>
 * Sören Tempel <soeren@soeren-tempel.net>
 * Noah Snelson <noah.snelson@protonmail.com>
+* Adam Porter <adam@alphapapa.net>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -18,6 +18,7 @@ New Features
 * All augmented assignment operators (except `%=` and `^=`) now allow
   more than two arguments.
 * PEP 3107 and PEP 526 function and variable annotations are now supported.
+* Added function ``parse-args`` which parses arguments with ``argparse``.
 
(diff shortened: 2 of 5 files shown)


Kodiologist:
Looks good. In the documentation, can you link to Python's documentation for `argparse`? It should be doable with a `py:` cross-reference, although it always takes me some trial and error to find the syntax that Sphinx wants.

refi64:
We should probably state how much is willing be be added. Having an
argument parsing is macro cool, but IMO there should be a definitive limit
as to how much extra stuff we want it to do / we're willing maintain
(whether it be "anything argparse does" or "just as-is") for the inevitable
point where someone opens and issue wanting to add a built-in nuclear
reactor as an option.

(Side note: is there much advantage here vs using the new annotation syntax
& plac? I guess that it's in stdlib could be reason enough.)

Kodiologist:
Nobody is currently threatening to make this function do anything except delegate to `argparse`. I don't see where any creeping complexity would come in.

> the new annotation syntax & plac

I don't know what that is.

brandonwillard:
Was there a discussion concerning why this is a good *core* language addition?

Kodiologist:
[Not really.](https://github.com/hylang/hy/issues/1719#issuecomment-547469703) The thing is, our system of automatic imports is broken as designed. I'm currently thinking of just making the user import the functions or groups of functions they want, for the most part. The functions currently in `hy.core.language` will be split into themed modules, or just deleted.

Kodiologist:
But at any rate, that sort of thing isn't in scope for this PR. `hy.core.language` will continue to be a grab back till that overhaul happens.

PR #1834, opened 2019-11-12
Document gotcha with unintentional recursion in `let`

### LINKED ISSUE #1829: Warn in documentation that Python's closures are late-binding
I may have missed something, but it seems that lexical closures require the use of the `let` macro, and this doesn't seem to be documented--or at least, not prominently.  This issue seems likely to be encountered by those accustomed to Lisps or Schemes, so I think it should be documented prominently.

For example, I wanted to conditionally wrap a function in another function at runtime.  This code (where `action-fn` is the function being wrapped) doesn't work, because `action-fn` is not lexically bound, so it's bound to itself, which causes infinite recursion:

```hy
(setv action-fn (fn [&rest args &kwargs kwargs]
                  (action-fn #* args #** kwargs)
                  (print-files #* args #** kwargs)))
```

But using `let` to save the original value of `action-fn` works:

```hy
(setv action-fn (let [orig-fn action-fn]
                    (orig-fn #* args #** kwargs)
                    (print-files #* args #** kwargs))))
```
If I am indeed missing something, please forgive the noise.

Thanks for your work on Hy.

Kodiologist:
Closes #1829.

### DIFF
--- a/docs/contrib/walk.rst
+++ b/docs/contrib/walk.rst
@@ -240,7 +240,8 @@ The ``let`` macro takes two parameters: a list defining *variables*
 and the *body* which gets executed. *variables* is a vector of
 variable and value pairs.
 
-``let`` executes the variable assignments one-by-one, in the order written.
+Like the ``let*`` of many other Lisps, ``let`` executes the variable
+assignments one-by-one, in the order written::
 
 .. code-block:: hy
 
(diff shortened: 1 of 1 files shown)


docs/contrib/walk.rst
```diff
@@ -249,4 +250,8 @@ variable and value pairs.
     ...  (print x y))
     5 6
 
+Unlike them, however, each ``(let …)`` form uses only one
```
alphapapa: @Kodiologist That's a great example, thanks.

PR #1839, opened 2019-11-17
Use the released version of Python 3.8 on Travis

gliptak:


Kodiologist:
Thanks for catching this. Could you rename the commit something that makes it a bit clearer what's changed, like "Use the released version of Python 3.8 on Travis", and add yourself to AUTHORS in a separate commit?

PR #1842, opened 2019-11-18
Remove hy.contrib.multi

### LINKED ISSUE #1630: hy.contrib.multi.defn doesn't respect local scope
Consider this code:

    (require [hy.contrib.multi [defn]])

    (defn outer1 []
      (defn fun
        ([] 1))
      fun)

    (defn outer2 []
      (defn fun
        ([] 2))
      fun)

    (setv f1 (outer1))
    (print (f1))
    (setv f2 (outer2))
    (print (f2))
    (print (f1))

The final line prints 2 rather than 1 as you'd expect.

### LINKED ISSUE #1838: multi.defn functions don't allow lists as arguments
Functions defined with hy.contrib.multi's defn apparently can't take lists / unhashable types as arguments.

I presume they're being stored as the keys of a dictionary, perhaps (as part of a tuple or something).

```
=> (require [hy.contrib.multi [*]])
True
=> (defn id ([x] x))
=> (id 1)
1
=> (id [])
  File "stdin-a7f4388ddf86af51cc96477cca4f89e4fa47470b", line 1, in <module>
    (id [])
  File "/home/t/.local/lib/python3.7/site-packages/hy/contrib/multi.hy", line 31, in __call__
    (when (.fn? self i args kwargs)
  File "/home/t/.local/lib/python3.7/site-packages/hy/contrib/multi.hy", line 26, in is_fn
    (.issubset (frozenset (.keys kwargs)) com)))
TypeError: unhashable type: 'list'
```

(This was on hy 0.16.0, I tried to get/build a more recent version but can't figure out how)

Kodiologist:
Closes #1630.
Closes #1838.

`defmulti` and `defmethod` are too weak to provide real multiple dispatch a la CLOS, and the even weaker `multi.defn` is buggy despite being little more than syntactic sugar for `cond`.

I intend no prejudice against later attempts to implement real multiple dispatch.


NEWS.rst
```diff
@@ -10,6 +10,7 @@ Removals
   and `defn` instead.
 * Literal keywords are no longer parsed differently in calls to functions
   with certain names.
+* `hy.contrib.multi` has been removed. Use ``if`` or ``cond`` instead.
```
brandonwillard: It's better to recommend [`functools.singledispatch`](https://docs.python.org/3.6/library/functools.html#functools.singledispatch) and/or [`multipledispatch`](https://github.com/mrocklin/multipledispatch).

Kodiologist: I'm not super-familiar with those, but they seem more like real multiple dispatch than the things we're replacing. `multi.defn` selects a function based on the number of arguments and `defmulti` is based on value, whereas `fuctools.singledispatch` and `multipledispatch` are based on type.  So it's hard to see how you would in general use `singledispatch` or `multipledispatch` to replace a use of `multi.defn` or `defmulti`.

brandonwillard: Many uses of `defn` and `defmulti` are easily covered by the type-based dispatching provided by `singledispatch` and `multipledispatch`, and&mdash;in those cases&mdash;they are much more useful recommendations than refactoring to use `if` and/or `cond`.

For instance, here's a reproduction of the `defn` example using `multipledispatch`:
```python
from multipledispatch import dispatch


def fun(a, b=None, c=None):
    return _fun(a, b, c)

@dispatch(object, type(None), type(None))
def _fun(a, b, c):
    return "a"

@dispatch(object, object, type(None))
def _fun(a, b, c):
    return "a b"

@dispatch(object, object, object)
def _fun(a, b, c):
    return "a b c"


>>> fun(1)
'a'
>>> fun(1, 2)
'a b'
>>> fun(1, 2, 3)
'a b c'
```

Kodiologist: Oh, neat, I didn't realize that you could do that with `multipledispatch`. What about the other cases?

brandonwillard: You can leave it up to the user to figure out the details for their specific problem.  Simply providing a good direction to look is more than enough.

Kodiologist: Okay.

refi64:
Nit: could we summarize why it was removed or link to the issue from NEWS? At a glance it seems like it was just removed for no real reason.

Kodiologist:
Only if you also want to include a justification for everything else in this release. For the period I've been maintaining it, my philosophy has been that NEWS is supposed to a be a concise summary of what's changed and how to deal with changes, not why the changes were made.

refi64:
Touche.

PR #1846, opened 2019-11-24
Remove extraneous material from the style guide

Kodiologist:
The quotes aren't germane, the koan is inscrutable, the design philosophy is out of date, and the conclusion is filler.

### DIFF
--- a/docs/style-guide.rst
+++ b/docs/style-guide.rst
@@ -2,48 +2,10 @@
 Hy Style Guide
 ==============
 
-   “You know, Minister, I disagree with Dumbledore on many counts…but
-   you cannot deny he’s got style…”
-   — Phineas Nigellus Black, *Harry Potter and the Order of the Phoenix*
-
 The Hy style guide intends to be a set of ground rules for the Hyve
 (yes, the Hy community prides itself in appending Hy to everything)
 to write idiomatic Hy code. Hy derives a lot from Clojure & Common
 Lisp, while always maintaining Python interoperability.
-
-
-Prelude
-=======
-
-The Tao of Hy
--------------
-
-.. code-block:: none
-
-   Ummon asked the head monk, "What sutra are you lecturing on?"
-   "The Nirvana Sutra."
-   "The Nirvana Sutra has the Four Virtues, hasn't it?"
-   "It has."
-   Ummon asked, picking up a cup, "How many virtues has this?"
-   "None at all," said the monk.
-   "But ancient people said it had, didn't they?" said Ummon.
-   "What do you think of what they said?"
-   Ummon struck the cup and asked, "You understand?"
-   "No," said the monk.
-   "Then," said Ummon, "You'd better go on with your lectures on the sutra."
-   — the (koan) macro
-
-The following illustrates a brief list of design decisions that went
-into the making of Hy.
-
-+ Look like a Lisp; do the right thing with it (e.g. dashes turn to underscores).
-+ We're still Python. Most of the internals translate 1:1 to Python internals.
-+ Use Unicode everywhere.
-+ When in doubt, defer to Python.
-+ If you're still unsure, defer to Clojure.
-+ If you're even more unsure, defer to Common Lisp.
-+ Keep in mind we're not Clojure. We're not Common Lisp. We're
-  Homoiconic Python, with extra bits that make sense.
 
 
 Layout & Indentation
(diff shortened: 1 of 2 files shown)


docs/style-guide.rst
```diff
@@ -2,50 +2,12 @@
...
-   Ummon asked, picking up a cup, "How many virtues has this?"
-   "None at all," said the monk.
-   "But ancient people said it had, didn't they?" said Ummon.
-   "What do you think of what they said?"
-   Ummon struck the cup and asked, "You understand?"
-   "No," said the monk.
-   "Then," said Ummon, "You'd better go on with your lectures on the sutra."
-   — the (koan) macro
-
-The following illustrates a brief list of design decisions that went
```
refi64: Why remove this? From what I can tell it's something that's largely still followed, especially the `We're Homoiconic Python.`

Kodiologist: I don't feel that "largely" is good enough for reference documentation. For example, "Use Unicode everywhere" is out of date because it refers to our features to close the gap between Pythons 2 and 3 that are now gone. And it's not true that we always defer to Python before other Lisps. Sometimes we imitate Python and sometimes we imitate other Lisps. And "we're Homoiconic Python, with extra bits that make sense" doesn't do justice to all of the language's semantic additions. When Hy is discussed elsewhere, I see a lot of grousing by non-users that Hy isn't "really" Lisp, and while I've never gotten an answer as to why anybody thinks that, I suspect that outsiders assume that Hy does less than it really does.

Kodiologist: [For example.](https://www.reddit.com/r/lisp/comments/6j9mus/hy_0130_released/djddh0h/)

docs/style-guide.rst
```diff
@@ -2,50 +2,12 @@
...
-into the making of Hy.
-
-+ Look like a Lisp; do the right thing with it (e.g. dashes turn to underscores).
-+ We're still Python. Most of the internals translate 1:1 to Python internals.
-+ Use Unicode everywhere.
-+ When in doubt, defer to Python.
-+ If you're still unsure, defer to Clojure.
-+ If you're even more unsure, defer to Common Lisp.
-+ Keep in mind we're not Clojure. We're not Common Lisp. We're
-  Homoiconic Python, with extra bits that make sense.
```
brandonwillard: If anything, we actually need to be **clearer** about our past and current design decisions, and this looks like a step in the opposite direction.

Kodiologist: Surely removing misleading and outdated claims is a step in the direction of being clearer.

brandonwillard: If Hy development **was never** guided by those statements, then they would be misleading, but that's definitely not the case.

Regarding Hy's present design decisions, it's much more consistent to stick with the existing and/or past ones **until we can determine and agree upon replacements**.  If we proceed with development and admittedly have no clear design principles or guidelines, then this project will be even less likely to mature into a worthwhile v1.

Kodiologist: > If Hy development was **never** guided by those statements, then they would be misleading, but that's definitely not the case.

It wouldn't be misleading if it were qualified with something like "Here's some principles that we were thinking of adhering to at some long-past point in the development of Hy, but which were never fully realized", but that is the kind of qualification that makes the reader wonder why the qualified statement is included at all.

> it's much more consistent to stick with the existing and/or past ones **until we can determine and agree upon replacements**

We don't stick to them, though. This PR is about the mere statement of the principles. Here, I'm not attempting to change our actual adherence to them one way or the other. Stripping everything out of Hy until it's "Homoiconic Python, with extra bits that make sense", for example, would be a massive undertaking, and something I wouldn't want anyway.

Kodiologist: Anyway, once this is gone, we can still bikeshed about a new statement of principles to our heart's content. At least we won't have an old statement of principles that nobody likes or follows hanging around in the meantime.

Kodiologist:
@refi64 @brandonwillard This PR is about to turn two weeks old, so are either of you guys vetoing it? You both went silent, but I don't know if it's an okay-fine kind of silence or a no-I'm-sick-of-arguing-with-you kind of silence.

refi64:
Ehh I can live without the ideology for a bit, but imo it's still worth
filing an issue for.

Kodiologist:
Okay, no problem, we can certainly do that.

PR #1857, opened 2019-12-26
Implement chained comparisons

### LINKED ISSUE #1852: Support Python's chained comparisons
In Python, `2 == 2 > 1` is true although `(2 == 2) > 1` and `2 == (2 > 1)` are both false.

To implement chained comparisons, we should probably add a special form with infix syntax, like `(compare 2 = 2 > 1)`, rather than try to detect forms like `(> (= 2 2) 1)` and give them magical semantics. It won't be very Lispy and it probably won't be used much, but at least it'll be available.

Kodiologist:
Closes #1852.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -17,6 +17,7 @@ New Features
 ------------------------------
 * Added special forms ``py`` to ``pys`` that allow Hy programs to include
   inline Python code.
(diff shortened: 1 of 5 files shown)


Kodiologist:
Yeah, it's goofy, but I don't have any better ideas.

PR #1865, opened 2020-01-07
Revamp the setup script's install command

### LINKED ISSUE #1864: setup.py install doesn't run without dependencies
Right now, if you run `setup.py install` without installing the dependencies first, then the install will fail. This is because all the Hy files are compiled *before* installation, but in order to compile them the dependencies need to be present.

Unfortunately, fixing this isn't particularly pretty. As it stands, setup.py doesn't install the dependencies *at all* thanks to this [ever-open setuptools bug](https://github.com/pypa/setuptools/issues/456). Right now, the setuptools install command does this:

- Gathers the distribution information.
- Builds a binary egg.
- Asks easy_install to install said egg, which will also install dependencies.

In order to install the dependencies prior to compiling the Hy bytecode, one of the following would need to be done (in combination for working around the issue mentioned above):

- Figure out some way to reliably iterate through the dependencies and install them first. This would probably involve copying a bit of small code from easy_install.py.
- Get easy_install to install only the dependencies first. I like this approach, but I'm trying to figure out how to do it without calling egg_info which would involve going through the manifest *twice* on install.

Of course, we could always just say the install command is broken. This isn't as insane as it seems, because right now the only reason you'd be forced to use the install command is for distro packaging, in which case the dependencies will already be installed. In all other cases, `pip install .` is superior in every way.

refi64:
Previously, the Hy files would be bytecode compiled before the
compiler's dependencies were installed. In additon, the revamped
version properly propogates the optimization level and generally is a
bit cleaner.

Fixes #1864.

This solution wasn't nearly as bad I thought. (Please forgive minor formatting issues atm, literally typed this whole thing out on a phone.)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,5 @@
 Copyright 2020 the authors.
+Portions of setup.py, copyright 2016 Jason R Coombs <jaraco@jaraco.com>.
 
 Permission is hereby granted, free of charge, to any person obtaining a
 copy of this software and associated documentation files (the "Software"),
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,11 @@
 # This file is part of Hy, which is free software licensed under the Expat
 # license. See the LICENSE.
 
-import sys, os
+import glob
+import importlib
+import inspect
+import os
+import sys
 
 from setuptools import find_packages, setup
 from setuptools.command.install import install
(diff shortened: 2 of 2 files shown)


Kodiologist:
Would it make sense to require the user to already have `pip` (chances are, they do) and explicitly rely on `pip install`? Or does that not quite work?

refi64:
Basically, if we were to rely on pip install, there would be two scenarios:

- The dependencies are installed somewhere outside of the install command.
This would suck because...they get installed outside of the install command
and would probably break everything.

- The dependencies are installed in the install command. The majority of
the code already present would still remain, except we're now using an
external tool. Not really an advantage.

Kodiologist:
I wonder if the passage of code you used as a source is large enough to be copyrightable. I guess it's pretty much on the edge. setuptools uses exactly the same license as we do, fortunately, but if copyright applies, we'll need to copy its copyright notice per the license.

Kodiologist:
It looks like you created this branch on this repo instead of your own. Watch out for that.

Kodiologist:
I've tried this out and it looks like bytecode isn't being installed. My procedure was

- Manually uninstalling the dependencies with `pip` (this is unlikely to be necessary to reproduce the problem)
- `git clean -dfx`
- `sudo -H pip3 install .`

The dependencies seem to be managed okay, though, at least when installing with `pip`.

Kodiologist:
Although, I'm seeing the same problem on master, so that's not really a problem with this PR unless you're trying to fix it.

Kodiologist:
Scratch the above two comments. I forgot that installed bytecode gets its own directory in Python 3.

refi64:
Okay so I added a clause to LICENSE that should account for that, super sorry about pushing to the wrong remote, note to self that pushing commits at 2am on an Android phone is usually going to have some sort of undesirable error...

Kodiologist:
Sounds good.

Kodiologist:
I'll go ahead and merge it, since you haven't.

PR #1868, opened 2020-01-18
Avoid evaluating quoted f-strings

refi64:
Trivial, yay or nay?

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -12,6 +12,7 @@ Bug Fixes
 ------------------------------
 * Improved support for nesting anaphoric macros by only applying
   symbol replacement where absolutely necessary.
+* Quoted f-strings are no longer evaluated prematurely.
 
 0.18.0
 ==============================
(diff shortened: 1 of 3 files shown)


Kodiologist:
Great work. Unfortunately, this change would probably not be seen as trivial.

I think the test isn't quite what we want. You're checking the stringification of the model. It would be better to test that the replacement field is evaluated at the right time, with something like:

```
(setv quoted 'f"hello {world}")
(with [(pytest.raises UnboundLocalError)]
  (eval quoted))
(setv world "goodbye")
(assert (= (eval quoted) "hello goodbye"))
```

Kodiologist:
@refi64 Do you plan on finishing this? If not, I think I'll pick it up.

refi64:
Oof forgot, I'll finish it tomorrow

refi64:
Okay finally updated!

Kodiologist:
Noice. I'll add a NEWS item and merge.

PR #1876, opened 2020-02-20
Fix #1875: use function type for :type of parse-args rather than symbol

### LINKED ISSUE #1875: parse-args requires values to be representable as Hy models
# Issue

The behaviors between `:type` keyword argument of parse-args function in hylang and `type` of argparse.ArgumentParser.add_argument method in Python are different.

* In Python: it's used for type conversion, and the value type is function, which accepts a string argument and returns the target type value. Therefore, I can define custom function to parse the string value.
* However, in hylang, the value type is symbol (i.e. `:type 'int`), and cannot use custom function.

For example, I'd like to set image size with `<width>x<height>` format, i.e. `--size 800x600`.

In Python:

```python
import argparse

def size(s):
    width, height = s.split('x')
    return int(width), int(height)

parser = argparse.ArgumentParser()
parser.add_argument('-s', type=size)
parser.parse_args(['-s', '800x600']) #=> Namespace(s=(800, 600))
```

In hylang will get an exception:

```hy
(defn size [s]
   (setv [width height] (s.split "x"))
   [(int width) (int height)])
(parse-args [["-s" :type size]] ["-s" "800x600"])
```

Exception message:

```
  File "stdin-29462993091ad78e6a5177ff794c993fe5d75f41", line 1, in <module>
    (parse-args [["-s" :type size]] ["-s" "800x600"])
  File "/Users/redraiment/Library/Python/3.7/lib/python/site-packages/hy/core/language.hy", line 404, in parse_args
    (eval `(.add-argument parser ~@arg)))
hy.errors.HyWrapperError: Don't know how to wrap <class 'function'>: <function size at 0x11094f0e0>
```

# Solution

After research, I found the root cause.

## Case 1: `:type size`

```hy
`(.add-argument parser ~@["-s" :type size])
```

The above code becomes `'(.add-argument parser "-s" :type <function size at 0x11094f0e0>)`, and `eval` couldn't works because `<function size at 0x11094f0e0>` is not function literal.

## Case 2: `:type 'size`

```hy
`(.add-argument parser ~@["-s" :type 'size])
```

The above code becomes `'(.add-argument parser "-s" :type size)`, and `eval` would raise `NameError: name 'size' is not defined` exception because `size` function not in context.

## Solution

My solution is call `add-argument` method direction with unpacking positional arguments and keywords arguments manually, rather than use `eval`.

```hy
(defn parse-args- [spec &optional args &kwargs parser-args]
  (import argparse)
  (setv parser (argparse.ArgumentParser #** parser-args))
  (for [arg spec]
    ;; (eval `(.add-argument parser ~@arg))
    (parser.add-argument
      ;; positional arguments
      #* (take-while (complement keyword?) arg)
      ;; keywords arguments
      #** (dfor [key value] (partition (drop-while (complement keyword?) arg) 2)
            [(name key) value])))
  (.parse-args parser args))
```

And it works well for me:

```hy
(parse-args- [["-s" :type size]] ["-s" "800x600"]) ;=> Namespace(s=[800, 600])
```

redraiment:
Keep the behavior same with Python.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -96,3 +96,4 @@
 * Noah Snelson <noah.snelson@protonmail.com>
 * Adam Porter <adam@alphapapa.net>
 * Gábor Lipták <gliptak@gmail.com>
+* Zepeng Zhang <redraiment@gmail.com>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,5 +1,12 @@
 .. default-role:: code
 
+Unreleased
+==============================
+
+Other Breaking Changes
+------------------------------
+* Value type of `:type` in `parse-args` spec is now a `function` instead of `HySymbol`.
+
 0.18.0
 ==============================
 
--- a/docs/language/core.rst
+++ b/docs/language/core.rst
@@ -804,7 +804,7 @@ may be a list of keyword arguments to pass to the
 .. code-block:: hy
 
     => (parse-args [["strings" :nargs "+" :help "Strings"]
-                    ["-n" "--numbers" :action "append" :type 'int :help "Numbers"]]
+                    ["-n" "--numbers" :action "append" :type int :help "Numbers"]]
                    ["a" "b" "-n" "1" "-n" "2"]
                    :description "Parse strings and numbers from args")
     Namespace(numbers=[1, 2], strings=['a', 'b'])
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -401,7 +401,10 @@ constructor."
   (import argparse)
   (setv parser (argparse.ArgumentParser #** parser-args))
   (for [arg spec]
-    (eval `(.add-argument parser ~@arg)))
+    (parser.add-argument
+      #* (take-while (complement keyword?) arg)
+      #** (dfor [key value] (partition (drop-while (complement keyword?) arg) 2)
+            [(name key) value])))
   (.parse-args parser args))
 
 (setv EXPORTS
--- a/hy/core/language.hy
+++ b/hy/core/language.hy
@@ -401,10 +401,23 @@ constructor."
   (import argparse)
   (setv parser (argparse.ArgumentParser #** parser-args))
   (for [arg spec]
-    (parser.add-argument
-      #* (take-while (complement keyword?) arg)
-      #** (dfor [key value] (partition (drop-while (complement keyword?) arg) 2)
-            [(name key) value])))
+    (setv [positional keywords _]
+          (reduce (fn [state arg]
+                    (setv [positional keywords value-of-keyword?] state)
+                    (if value-of-keyword?
+                        (do
+                          (.append (get keywords -1) arg)
+                          [positional keywords False])
+                        (if (keyword? arg)
+                            (do
+                              (.append keywords [(name arg)])
+                              [positional keywords True])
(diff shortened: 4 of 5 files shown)


Kodiologist:
This is a step in the right direction, but I don't think your implementation is quite right, because:

1. A keyword can be the argument of another keyword: `(f :x :x)` is compiled to `f(x = HyKeyword("x"))`.
2. In Hy, unlike Python, positional arguments can appear after keyword arguments.

redraiment:
> This is a step in the right direction, but I don't think your implementation is quite right, because:
> 
> 1. A keyword can be the argument of another keyword: `(f :x :x)` is compiled to `f(x = HyKeyword("x"))`.
> 2. In Hy, unlike Python, positional arguments can appear after keyword arguments.

I think the implement works well with the first case, because it just convert the first keyword `[(name key) value]`.

For the second case, I have just update the solution: scans the arguments list, takes the `<keyword> <value>` pairs as keywords arguments and the rest as positional arguments. Now, it works well with `["-s" :help "help document" "--size"]` case.

redraiment:
BTW, it looks there is a commonly `apply` function implement in parse-args:

```hy
  ...
  (defn apply [f arguments]
    (setv [positional keywords _]
          (reduce (fn [state arg]
                    (setv [positional keywords value-of-keyword?] state)
                    (if value-of-keyword?
                        (do
                          (.append (get keywords -1) arg)
                          [positional keywords False])
                        (if (keyword? arg)
                            (do
                              (.append keywords [(name arg)])
                              [positional keywords True])
                            (do
                              (.append positional arg)
                              [positional keywords False]))))
                  arguments
                  [[] [] False]))
    (f #* positional #** (dict keywords)))
  (for [arg spec]
    (apply parser.add-argument arg))
  ...
```

I know the `apply` function has been removed since 0.14.0. However, it will be useful if there is an `apply` function to handle both positional arguments and keywords arguments.

Kodiologist:
I think that could be implemented more concisely with a loop, but hey, it works. Good job. Update NEWS, squash your changes, and add yourself to AUTHORS as a separate commit, and then I think we're done.

> I know the `apply` function has been removed since 0.14.0.

The previous function of that name is different from what you've called `apply` here. The old one was equivalent to `(defn apply [f args kwargs] (f #* args #** kwargs))`, which should make it more obvious why it became redundant when we added `#*` and `#**`.

redraiment:
> Update NEWS, squash your changes, and add yourself to AUTHORS as a separate commit

Updated. Please help me to confirm~ Thanks! :-)

Kodiologist:
I squashed and added the issue URL to the test, but I can't push back my changes because of a permissions error. Do you have "Allow edits from maintainers" checked for this PR? You should. (You should also have created a new branch for this PR. Take a look at CONTRIBUTING.rst before opening your next pull request.)

Kodiologist:
@redraiment Hello?

redraiment:
> @redraiment Hello?

Sorry for replying late.
In some reason, I couldn't use my personal computer these days. I'll follow CONTRIBUTING.rst to process the PR this weekend.

Kodiologist:
Don't sweat it for this PR. Just check "Allow edits from maintainers", let me know you've done so, and I'll take care of it.

PR #1880, opened 2020-03-10
Fix misplaced `#@` tag macro

zhaqenl:
Under the [Hy versus other Lisps](https://docs.hylang.org/en/stable/whyhy.html#hy-versus-other-lisps) section, and inside the definition of the `HelloWorld` class, the `#@` tag macro is misplaced inside the parenthesis.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -96,3 +96,4 @@
 * Noah Snelson <noah.snelson@protonmail.com>
 * Adam Porter <adam@alphapapa.net>
 * Gábor Lipták <gliptak@gmail.com>
+* Raymund MARTINEZ <zhaqenl@protonmail.com>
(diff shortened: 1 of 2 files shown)


Kodiologist:
Thanks for catching this. Would you please:
- Change the commit message. `#@` is a tag macro. Reader macros aren't implemented.
- In a separate commit, add yourself to AUTHORS.

zhaqenl:
Fixed!

I may have been mislead by [this answer](https://github.com/hylang/hy/issues/952#issuecomment-145392669) under the issues, hence my usage of “reader macro”. Anyway, thanks!

Kodiologist:
No problem. A kind of macros that look like reader macros, but are actually ordinary macros with a funny syntax, were originally added to Hy under the name "reader macros". We later renamed them to "tag macros" (briefly, "hash macros") to avoid confusion, and leave open the possiblity of implementing real reader macros later while keeping these guys around.

PR #1882, opened 2020-03-15
Fixes #1875: parse-args requires values to be representable as Hy models

### LINKED ISSUE #1875: parse-args requires values to be representable as Hy models

redraiment:
* Update argument spec parse logic of parse-args function.
* Update test case of parse-args function.
* Update document of parse-args function.
* Describe the change in NEWS file.
* Add name to AUTHORS file.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -97,3 +97,4 @@
 * Adam Porter <adam@alphapapa.net>
 * Gábor Lipták <gliptak@gmail.com>
 * Raymund MARTINEZ <zhaqenl@protonmail.com>
+* Zepeng Zhang <redraiment@gmail.com>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,5 +1,13 @@
 .. default-role:: code
 
+Unreleased
+==============================
+
+Other Breaking Changes
+------------------------------
+* `parse-args` is no longer implemented with `eval`; so e.g. you should
+  now say `:type int` instead of `:type 'int`.
+
 0.18.0
 ==============================
 
--- a/docs/language/core.rst
+++ b/docs/language/core.rst
@@ -804,7 +804,7 @@ may be a list of keyword arguments to pass to the
 .. code-block:: hy
 
     => (parse-args [["strings" :nargs "+" :help "Strings"]
-                    ["-n" "--numbers" :action "append" :type 'int :help "Numbers"]]
+                    ["-n" "--numbers" :action "append" :type int :help "Numbers"]]
                    ["a" "b" "-n" "1" "-n" "2"]
                    :description "Parse strings and numbers from args")
     Namespace(numbers=[1, 2], strings=['a', 'b'])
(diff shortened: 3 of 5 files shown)


redraiment:
"Allow edits from maintainers." option has checked.

PR #1884, opened 2020-03-16
Reduce scope of symbol replacement for anaphoric macros

### LINKED ISSUE #1883: Anaphoric macros replacing 'it' symbols in 'xs' parameter
```hy
(require [hy.extra.anaphoric [*]])

(->> [1 2 3 4 5]
  (ap-filter (= (% 2 it) 0))
  (ap-map (* 2 it))
  (list)
  (print))
```
The above code causes the error `NameError: name '_hyx_GXUffffX1' is not defined` at runtime. The code that it compiles to is:
```python
import hy.macros
from hy.core.language import is_odd
hy.macros.require('hy.extra.anaphoric', None, assignments='ALL', prefix='')
n = []
for _hyx_GXUffffX1 in range(next((_hyx_GXUffffX2 for _hyx_GXUffffX2 in [2,
    4, 8, 10, 5, 6, 3] if is_odd(_hyx_GXUffffX1)), None)):
    n.append(_hyx_GXUffffX1)
print(n)
```
But what I am expecting is:
```python
import hy.macros
from hy.core.language import is_odd
hy.macros.require('hy.extra.anaphoric', None, assignments='ALL', prefix='')
n = []
for _hyx_GXUffffX1 in range(next((_hyx_GXUffffX2 for _hyx_GXUffffX2 in [2,
    4, 8, 10, 5, 6, 3] if is_odd(_hyx_GXUffffX2)), None)):
    n.append(_hyx_GXUffffX1)
print(n)
```
I have investigated the anaphoric macros:
```hy
(defmacro ap-filter [form xs]
  (rit `(gfor  it ~xs  :if ~form  it)))

(defmacro ap-map [form xs]
  (rit `(gfor  it ~xs  ~form))

(defn recur-sym-replace [d form]
  "Recursive symbol replacement."
  (cond
    [(instance? HySymbol form)
      (.get d form form)]
    [(coll? form)
      ((type form) (gfor  x form  (recur-sym-replace d x)))]
    [True
      form]))

(defn rit [form]
  "Replace `it` with a gensym throughout `form`."
  (recur-sym-replace {'it (gensym)} form))
```
I think the issue is the `rit` function call, as it recurses over all elements in the generated code
I also investigated the other anaphoric macros and they all seem to suffer from this issue, except for `ap-reduce`

eganjs:
Resolves #1883
I would encourage reviewers to look at the commit history to see if there is any merit in the an earlier more verbose approach I took.
This is my first time committing lisp code, so any feedback would be kindly received! :)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -98,3 +98,4 @@
 * Gábor Lipták <gliptak@gmail.com>
 * Raymund MARTINEZ <zhaqenl@protonmail.com>
 * Zepeng Zhang <redraiment@gmail.com>
+* Joseph Egan <joseph.s.egan@gmail.com>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -8,6 +8,11 @@ Other Breaking Changes
 * `parse-args` is no longer implemented with `eval`; so e.g. you should
   now say `:type int` instead of `:type 'int`.
 
+Bug Fixes
+------------------------------
+* Improved support for nesting anaphoric macros by only applying
+  symbol replacement where absolutely necessary.
+
 0.18.0
 ==============================
 
--- a/docs/extra/anaphoric.rst
+++ b/docs/extra/anaphoric.rst
@@ -19,9 +19,11 @@ To use these macros you need to require the ``hy.extra.anaphoric`` module like s
 
 These macros are implemented by replacing any use of the designated
 anaphoric symbols (``it``, in most cases) with a gensym. Consequently,
-it's unwise to nest these macros, or to use an affected symbol as
-something other than a variable name, as in ``(print "My favorite
-Stephen King book is" 'it)``.
+it's unwise to nest these macros where symbol replacement is happening.
+Symbol replacement typically takes place in ``body`` or ``form``
+parameters, where the output of the expression may be returned. It is also
+recommended to avoid using an affected symbol as something other than a
+variable name, as in ``(print "My favorite Stephen King book is" 'it)``.
 
 .. _ap-if:
 
--- a/hy/extra/anaphoric.hy
+++ b/hy/extra/anaphoric.hy
@@ -3,49 +3,62 @@
 ;; This file is part of Hy, which is free software licensed under the Expat
 ;; license. See the LICENSE.
 
+;;; Macro to help write anaphoric macros
+
+(defmacro rit [&rest body]
+  """Supply `it` as a gensym and R as a function to replace `it` with the
+  given gensym throughout expressions."""
+  `(do
+    (setv it (gensym))
+    (defn R [form]
+      "Replace `it` with a gensym throughout `form`."
+      (recur-sym-replace {'it it} form))
+    ~@body))
+
+
 ;;; These macros make writing functional programs more concise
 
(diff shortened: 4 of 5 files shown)


Kodiologist:
When I recently overhauled the implementation of these functions, I added this warning to the documentation:

> These macros are implemented by replacing any use of the designated anaphoric symbols (`it`, in most cases) with a gensym. Consequently, it’s unwise to nest these macros

The problem you encountered is caused by nesting. It's just that the nesting is hidden by the use of `->>`. A form such as `(->> [1 2 3 4 5] (ap-filter (= (% 2 it) 0)) (ap-map (* 2 it)))` is compiled to `(ap-map (* 2 it) (ap-filter (= (% 2 it) 0) [1 2 3 4 5]))`, so in the expansion of `ap-map`, both `it`s are replaced.

What you realized, but I didn't appreciate in my own implementation, is that there's no need to replace `it`s in some of the arguments of these macros, such as the `xs` argument of `ap-map`. Your change is effective because it causes macros in `xs`, including inner calls to these anaphoric macros, to be expanded before `ap-map` calls `rit`, so `ap-map` never sees the inner `it`. But it would be more direct to only call `rit` on the code that needs to have its `it`s replaced to begin with. Your approach leaves an unnecessary temporary variable and assignment in the generated code.

One could try something like:

```
(defn rit [symbol form]
  (recur-sym-replace {'it symbol} form))

(defmacro ap-map [form xs]
  (setv g (gensym))
  `(gfor ~(rit g 'it) ~xs ~(rit g form)))
```

or, to avoid having to manage the gensym explicitly in the definition of every macro, one could use a macro-writing macro like this:

```
(defmacro rit [&rest body]
  `(do
    (setv g (gensym))
    (defn R [form]
      (recur-sym-replace {'it g} form))
    ~@body))

(defmacro ap-map [form xs]
  (rit `(gfor ~(R 'it) ~xs ~(R form))))
```

It's important to realize that certain kinds of nesting, such as `(ap-each (range 2) (print it) (ap-each "abc" (print it)))`, are supported by neither your changes nor my proposal. So, the warning in the documentation will have to go into a bit more detail: some kinds of nesting will work as expected and some won't.

eganjs:
First off, thanks for the quick response!

> or, to avoid having to manage the gensym explicitly in the definition of every macro, one could use a macro-writing macro like this:
> 
> ```
> (defmacro rit [&rest body]
>   `(do
>     (setv g (gensym))
>     (defn R [form]
>       (recur-sym-replace {'it g} form))
>     ~@body))
> 
> (defmacro ap-map [form xs]
>   (rit `(gfor ~(R 'it) ~xs ~(R form))))
> ```

My first approach to was much like this conceptually but had boilerplate and at the time I didn't consider the possibility of a macro-writing macro. In my second attempt  I moved away from it to make my code more concise, but your proposal seems like a better approach. I'll update my code and look at the documentation.

Kodiologist:
Excellent. Would you please:

- Add a NEWS item
- Squash all your changes into one commit
- Add yourself to AUTHORS, as a separate commit

Then we should be good.

PR #1895, opened 2020-05-28
Remove imp to avoid the deprecation message

fatestigma:
Could we remove `imp` to avoid the deprecation message if we use `python -m hy`.

```
/usr/local/lib/python3.7/site-packages/hy/__main__.py:2: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
```

Also, [since pytest 5.4, node construction is also deprecated in favor of `.from_parent`](https://docs.pytest.org/en/latest/deprecations.html#node-construction-changed-to-node-from-parent), old style will generate deprecation warning every time when we run `pytest`

```
conftest.py:48
  /.../conftest.py:48: PytestDeprecationWarning: direct construction of Module has been deprecated, please use Module.from_parent
    pytest_mod = pytest.Module(path, parent)
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -99,3 +99,4 @@
 * Raymund MARTINEZ <zhaqenl@protonmail.com>
 * Zepeng Zhang <redraiment@gmail.com>
 * Joseph Egan <joseph.s.egan@gmail.com>
+* Xi Jin <xij@usc.edu>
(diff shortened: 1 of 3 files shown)


fatestigma:
Seems Travis is still using pytest-5.2.1. So I updated a bit to only use `from_parent` if it's available.

fatestigma:
Sorry about that. Just fixed the typo.

brandonwillard:
Thanks; now, you can add yourself to `AUTHORS` and I'll merge it.

PR #1896, opened 2020-05-31
Support Python 3.9

Kodiologist:


Kodiologist:
Oops, the new beta fixed https://bugs.python.org/issue40830 but is now crashing Hy for a different reason. I'm looking into it.

PR #1906, opened 2020-06-24
Fix `require` with multiple arguments

### LINKED ISSUE #1690: Hy repl doesn't open on clean install python 3.7
Did the following on a mac:

```
$ brew install python
$ python3
Python 3.7.0 ....
$ pip3 install hy
...
$ hy
-bash: /usr/local/bin/hy: /usr/local/opt/python3/bin/python3.5: bad interpreter: No such file or directory
```

When I ``cat `which hy``` I get:

```
#!/usr/local/opt/python3/bin/python3.5

# -*- coding: utf-8 -*-
import re
import sys

from hy.cmdline import hy_main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(hy_main())
```

Why is there a shebang line pointing to python 3.5? I can run ``python3 `which hy``` and everything (so far as I can tell) works just fine.

Tried with latest git (pip install git+...) and same issue.

### LINKED ISSUE #1903: `hy2py` doesn't work for `require` multiple modules
It seems that `hy2py` only generates the last module in `(require ...)`. It might be something wrong within `hy_compile` according to the generated AST.

Example source file `test.hy`:
```hy
(require
 [hy.contrib.loop [loop]]
 [hy.contrib.walk [let]])
```

Run with `hy2py -sa test.hy`:
```
HyExpression([
  HySymbol('do'),
  HyExpression([
    HySymbol('require'),
    HyList([
      HySymbol('hy.contrib.loop'),
      HyList([
        HySymbol('loop')])]),
    HyList([
      HySymbol('hy.contrib.walk'),
      HyList([
        HySymbol('let')])])])])


Module(
    body=[Import(names=[alias(name='hy.macros', asname=None)]),
        Expr(
            value=Call(
                func=Attribute(value=Attribute(value=Name(id='hy'), attr='macros'), attr='require'),
                args=[Str(s='hy.contrib.walk'), Name(id='None')],
                keywords=[
                    keyword(arg='assignments', value=List(elts=[List(elts=[Str(s='let'), Str(s='let')])])),
                    keyword(arg='prefix', value=Str(s=''))]))])


import hy.macros
hy.macros.require('hy.contrib.walk', None, assignments=[['let', 'let']],
    prefix='')
```

Kodiologist:
Fixes #1903.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -20,6 +20,7 @@ Bug Fixes
 * Fixed a regression in the production of error messages for empty
   expressions.
 * Fixed a scoping bug for code executed with `hy -c`.
+* Fixed a bug in the compilation of multiple `require`\s.
 
 0.18.0
 ==============================
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1213,6 +1213,7 @@ class HyASTCompiler(object):
                         [[HyString(k), HyString(v)] for k, v in assignments]),
                     HyKeyword('prefix'),
                     HyString(prefix)]).replace(expr))
+                ret += ret.expr_as_stmt()
 
         return ret
 
--- a/tests/compilers/test_ast.py
+++ b/tests/compilers/test_ast.py
@@ -482,7 +482,6 @@ def test_ast_unicode_vs_bytes():
     assert s('b"\\xa0"') == bytes([160])
 
 
-@pytest.mark.skipif(not PY36, reason='f-strings require Python 3.6+')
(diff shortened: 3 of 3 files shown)


brandonwillard:
Is this fix actually for #1903?

Kodiologist:
Yep, no idea how I came up with that number.

PR #1913, opened 2020-07-22
Add Homebrew install to docs

SeekingMeaning:
https://formulae.brew.sh/formula/hy


SeekingMeaning:
Sure, where would be a good place for that?

Kodiologist:
I'm against this. I don't think it's our job to keep track of repository-specific installation instructions. Besides, we want to encourage users to use PyPI or install directly from the source.

brandonwillard:
> I'm against this. I don't think it's our job to keep track of repository-specific installation instructions. Besides, we want to encourage users to use PyPI or install directly from the source.

Agreed.

SeekingMeaning:
Sure, no problem, we'll try to keep the package updated on our side.

PR #1916, opened 2020-08-08
Fix `hy.eval` from Python failing on `require`

### LINKED ISSUE #1911: Can't `hy.eval` a `require` form in Python
I'm getting an error when doing a `require` using the `hy.eval` functionality.

The version of python works when hy is loaded:

```
hy 0.18.0 using CPython(default) 3.7.5 on Darwin                                                                                                                                
=> (require  [hy.contrib.walk [let]])                                                                                                                                           
True
```  

but fails when doing the same thing in python.

```
$ python
Python 3.7.5 (default, Jun  8 2020, 17:36:22) 
[Clang 11.0.3 (clang-1103.0.32.29)] on darwin
>>> import hy
>>> hy.eval(hy.read_str("(require  [hy.contrib.walk [let]])"))
  File "/Users/chris/.pyenv/versions/3.7.5/lib/python3.7/site-packages/hy/compiler.py", line 2099, in hy_eval
    module.__dict__, locals)
  File "<string>", line 1, in <module>
  File "/Users/chris/.pyenv/versions/3.7.5/lib/python3.7/site-packages/hy/macros.py", line 170, in require
    if _same_modules(source_module, target_module):
  File "/Users/chris/.pyenv/versions/3.7.5/lib/python3.7/site-packages/hy/macros.py", line 118, in _same_modules
    target_filename = _get_filename(target_module)
  File "/Users/chris/.pyenv/versions/3.7.5/lib/python3.7/site-packages/hy/macros.py", line 109, in _get_filename
    filename = loader.get_filename()
AttributeError: type object 'BuiltinImporter' has no attribute 'get_filename'
```

homonoidian:
A tiny brute-fix for #1911. I really don't know what's the root of the problem, though, but it seems to be dynamic typing letting odd stuff in and causing havoc!

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -100,3 +100,4 @@
 * Zepeng Zhang <redraiment@gmail.com>
 * Joseph Egan <joseph.s.egan@gmail.com>
 * Xi Jin <xij@usc.edu>
+* Alexey Yurchenko <homonoidian@yandex.ru>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,5 +1,12 @@
 .. default-role:: code
 
+Unreleased
+==============================
+
+Bug Fixes
+------------------------------
+* Fixed a bug that made `hy.eval` from Python fail on `require`.
+
 0.19.0
 ==============================
 
--- a/hy/macros.py
+++ b/hy/macros.py
@@ -105,7 +105,7 @@ def _same_modules(source_module, target_module):
         try:
             if not inspect.ismodule(module):
                 loader = pkgutil.get_loader(module)
-                if loader:
+                if isinstance(loader, importlib.machinery.SourceFileLoader):
                     filename = loader.get_filename()
             else:
                 filename = inspect.getfile(module)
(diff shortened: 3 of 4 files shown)


homonoidian:
I wrote a simple test; didn't know where exactly to place it, though; so it is where it is. My first contribution, I don't know things... but hope to help.

Kodiologist:
There's no need to call Python as an external command for the test. Just use the Python code you want to test as the body of your test.

Use `tests/resources/macros.hy` instead of `let` for an example macro.

The commit message and the title of this issue should be more descriptive; "Fix #1911" is uninformative without first looking up the bug.

Kodiologist:
Thanks, good work. Would you please:
- Move the test to `tests/importer/test_importer.py`
- Rename it to remove the "bin"
- Update NEWS
- Squash your changes into one commit
- As a separate commit, add yourself to AUTHORS

homonoidian:
Thank y'all for patience with me. Hope the thing is right now.

Kodiologist:
Looks good to me. @brandonwillard?

Kodiologist:
Welcome aboard!

PR #1922, opened 2020-11-20
Contrib/pprint

### LINKED ISSUE #1837: [Feature Request] Pretty print module
I would like to have something like `pprint` or `clojure.pprint/pprint` module in Hy. More specific, is there something that would convert something like this:

```hy
{:a {:a 0 :b 1 :c 2 :d 3 :e 4} :b {:a 0 :b 1 :c 2 :d 3 :e 4} :c {:a 0 :b 1 :c 2 :d 3 :e 4} :d {:a 0 :b 1 :c 2 :d 3 :e 4} :e {:a 0 :b 1 :c 2 :d 3 :e 4}}
```

In this (this is an example, doesn't need to be exactly like below):

```hy
{:e {:e 4 :d 3 :c 2 :b 1 :a 0}
 :d {:e 4 :d 3 :c 2 :b 1 :a 0}
 :c {:e 4 :d 3 :c 2 :b 1 :a 0}
 :b {:e 4 :d 3 :c 2 :b 1 :a 0}
 :a {:e 4 :d 3 :c 2 :b 1 :a 0}}
```

I already tried `hy-repr` and `hydiomatic.utils/hypprint`, however `hy-repr` didn't seem to pretty print long maps and hydiomatic does not work in recent versions of Hy.

allison-casey:
Closes #1837.

A `hy-repr` based implementation of python's builtin `pprint` module. It lets `python.pprint` do much of the heavy lifting so the api ends up being almost identical. It doesn't support 3.8's `sort_dicts` arg though since i'm not sure how ya'll would prefer that be implemented. 

An example of some random data
```
[{"_id" "5fb83085fb4dff7db55b121c",
  "about" 'Laborum deserunt cupidatat minim adipisicing id aute ut non '
          'excepteur id est exercitation quis ullamco. Culpa sunt laboris '
          'reprehenderit quis consequat laboris non amet sunt dolor aute '
          'pariatur fugiat. Minim veniam ipsum tempor dolore proident. Lorem '
          'esse eiusmod et laboris aute proident commodo enim elit nisi '
          'commodo veniam adipisicing.',
  "address" "832 Bartlett Place, Ezel, Virgin Islands, 5591",
  "age" 28,
  "balance" "$1,869.94",
  "company" "FLEXIGEN",
  "email" "fernandez.mcclure@flexigen.info",
  "eyeColor" "brown",
  "favoriteFruit" #{:banana :apple},
  "friends" [{"id" 0  "name" "Mattie King"}
             {"id" 1  "name" "Meadows Kinney"}
             {"id" 2  "name" "Robbins Hendricks"}],
...
```


Kodiologist:
Closes #1837.

Kodiologist:
Sounds good. Let's see some tests. You don't have to completely recapitulate the tests for `hy-repr` or for Python's `pprint`—that wouldn't be useful—but we do at least want to check that a few examples look as they ought to.

> It doesn't support 3.8's `sort_dicts` arg though since i'm not sure how ya'll would prefer that be implemented.

What's the ambiguity here? Is there more than one obvious way to do it?

allison-casey:
i'll add some tests today. And the ambiguity is that because I'm subclassing python's `PrettyPrinter`, the arg doesn't actually exist in the constructor or the super class methods that are calling the hy versions of safe-repr etc. I can just make hy pprint not do anything with the arg if your own < 3.8 easy enough, but it would make the api a bit less clear in my mind.

Kodiologist:
in Python, `sort_dicts` defaults to `True` for backwards compatibility. I think we can do a similar thing, by ignoring `sort_keys` if it's true and either passing it down or raising an error message (for Pythons < 3.8) if it's false. `hy._compat` defines a constant `PY38` for this purpose.

allison-casey:
It looks like we're running into an ambiguity issue with how pprint splits long strings when their inside sequences. 
```python
pp(["hello world hello world"], width=16)
# becomes
=> ['hello world '
    'hello world']
``` 
And obviously in hy that's two separate entries in that list. How do we disambiguate that? As far as I can tell Hy doesn't have a form for the compile time string concatenation that python has
```python
("hello " "world" "!") 
=> "hello world!"
```

allison-casey:
One option would be to have strings be broken up into 
```
[(+ "hello world "
    "hello world")]
```
This way pprint output is still round-trippable for the main built in data types like it is for `python.pprint`

Kodiologist:
Yep, that makes sense. Or you could refuse to break lines inside strings. Python's `pprint`  doesn't guarantee that it can fit inside the given `width`, either.

allison-casey:
Pretty sure these tests cover most things, let me know if I missed something though. Went with `(+ "...")` for strings since strings not being split in `clojure.pprint` always annoyed me and the round-trippability of `python.pprint` has come in handy many times for me.

Kodiologist:
Thanks for your work on this. It's impressive. I haven't grokked all of it yet, but here are some points to address and then I can do another pass.

1. Setting `sort-dicts` to a false value should be an error on Pythons < 3.8. As is, the code silently treats it as if it were true.

2. The definitions of the functions `pprint` and `ppformat` can be simplified with `&rest` and `&kwargs` a la `pp`. (I don't think keeping the same signature as the corresponding Python standard functions is essential.)

3. It seems like there's a lot of separate code for `bytes` and `str`. Can they be combined? Relatedly, can the magic-number usage of 3 and 4 in the `bytes` code be eliminated? I presume that 3 and 4 are related to the leading `b` in bytes literals, but it's not obvious how.

4. Why the `(hy-repr.hy-repr-register frozenset …)` form? A `hy-repr` function for `frozenset` is already registered.

5. I think that pytest fixture definitions apply globally to the entire set of code being tested, not just the file they're defined in, so giving a fixture a short, generic name like  `a` is probably a bad idea. Anyway, it would be straightforward to write these as ordinary functions instead of fixtures. In the case of `b`, you don't even really need a function; just inline `(setv b (list (range 200)))`.

6. In place of `(assert (= ... True))` and `(assert (= ... False))`, I would write simply `(assert ...)` and `(assert (not ...))`. It's not important to check that `bool`s are produced here.

7. I don't think we need the second argument of `assert`, like `(% "expected not recursive? for %r" safe)`.

8. I think testing of `bytes` and `str` can be combined.

allison-casey:
1. done
2. done
3. `python.pprint` pretty prints strings and bytestrings different, namely attempting to chunk bytes along boundaries of length 4 and strings along word/newline boundaries. I'd prefer not to try and merge them so that if someone has to do a fix, the methods still line up with the `python.pprint` implementations. Added comments to explain exactly where the numbers come from, is that good enough?
4. my bad, forgot to remove it while I was messing around. removed
5. changed to regular functions
6. done
7. related to 3. We're checking specifically for the 4 chunk byte boundaries and bytes related indenting so it made sense to me to put it in it's own test. I added a comment to explain this. If you'd still prefer them merged let me know.

Kodiologist:
> `python.pprint` pretty prints strings and bytestrings different, namely attempting to chunk bytes along boundaries of length 4 and strings along word/newline boundaries.

Dang, I had no idea. Okay, I guess there's some sense in imitating that.

Kodiologist:
I made some edits. Tell me what you think.

> the methods still line up with the `python.pprint` implementations pretty 1-to-1.

Does that mean you copied or consulted the source code of Python's `pprint` while writing this code? I know we're imitating the interface, but deriving our code from theirs has some copyright implications (even though we're using a compatible license).

allison-casey:
ahh copyright, yes I did.

the changes you made to `-safe-repr`'s handling of sort-dicts fails on py < 3.8. the fallback to `-safe-py-repr` does not have the `sort-dicts` arg on those versions which is why that explicit conditional was there. 

Otherwise it looks fine to me

Kodiologist:
Oh, so then the way to combine the calls would be something like `(-safe-py-repr object context maxlevels level #* (if PY38 [sort-dicts] []))`.

Kodiologist:
I wish Travis wasn't so monstrously slow these days.

allison-casey:
yes it would. I personally prefer the exlicit `if` for something like this though. I usually try and reserve that kind of conditional arg unpacking for when there are several args to unpack instead of just one. 

He's trying the best he can :( 
image

Kodiologist:
Okay, looking good. Would you please:

- Add some documentation. Because we're imitating the standard Python module so closely, you don't need to write much. It should be enough to link to the corresponding part of the Python documentation and give an example or two.
- Add an item to NEWS.
- Update LICENSE with a line "Portions of hy/contrib/pprint.hy copyright 2020 Python Software Foundation, licensed under the Python Software Foundation License Version 2."
- Squish all these changes into one commit. The commit message should reference the CPython source file or files that you copied from. Since CPython has a GitHub mirror, you can reference files with a GitHub URL; just make sure it points to a specific commit instead of a branch.
- As a separate commit, add yourself to AUTHORS.

Kodiologist:
Great work. Welcome to the Hyve!

Kodiologist (approved):

PR #1923, opened 2020-11-21
Remove `__slots__`

### LINKED ISSUE #1754: HyKeyword __getstate__ w.r.t. pickle
In approximately 0.14, this worked:
```(with [f (open "test.pickle" "w")] (pickle.dump {:a 1} f))```
In 0.16 one gets:
```TypeError: a class that defines __slots__ without defining __getstate__ cannot be pickled```
because HyKeyword does not implement `__getstate__`. This is with
```hy 0.16.0 using CPython(default) 2.7.7 on Linux```

Kodiologist:
Fixes #1754.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -6,6 +6,7 @@ Unreleased
 Bug Fixes
 ------------------------------
 * Fixed a bug that made `hy.eval` from Python fail on `require`.
+* Fixed a bug that prevented pickling of keyword objects.
 
 0.19.0
 ==============================
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -179,8 +179,6 @@ class Result(object):
     The Result object is interoperable with python AST objects: when an AST
     object gets added to a Result object, it gets converted on-the-fly.
     """
-    __slots__ = ("imports", "stmts", "temp_variables",
-                 "_expr", "__used_expr")
 
     def __init__(self, *args, **kwargs):
         if args:
--- a/hy/models.py
+++ b/hy/models.py
@@ -139,8 +139,6 @@ _wrappers[type(None)] = lambda foo: HySymbol("None")
 class HyKeyword(HyObject):
     """Generic Hy Keyword object."""
 
-    __slots__ = ['name']
-
     def __init__(self, value):
         self.name = value
 
--- a/tests/native_tests/language.hy
+++ b/tests/native_tests/language.hy
@@ -7,6 +7,7 @@
         [sys :as systest]
         re
         [operator [or_]]
+        pickle
         [hy.errors [HyLanguageError]]
         pytest)
 (import sys)
@@ -1140,6 +1141,12 @@
   (assert (!= : ":"))
   (assert (= (name ':) "")))
 
+(defn test-pickling-keyword []
+  ; https://github.com/hylang/hy/issues/1754
+  (setv x :test-keyword)
+  (for [protocol (range 0 (+ pickle.HIGHEST-PROTOCOL 1))]
+    (assert (= x
+      (pickle.loads (pickle.dumps x :protocol protocol))))))
 
 (defn test-nested-if []
   "NATIVE: test nested if"


brandonwillard (changes requested):
What exactly is wrong with the use of `__slots__` here?  Besides the minor potential for reduced memory consumption, it also works as a nice check for objects that are supposed to be immutable.

Kodiologist:
It's wrong in that it's code that serves no purpose. I know that **in theory** it might save memory or time or something, but I've yet to see a case in which that mattered. In other words, it's a premature optimization.

> it also works as a nice check for objects that are supposed to be immutable

Neither of these classes are meant to be fully immutable. `HyKeyword` has a few additional attributes `start_line` etc. that are modified as part of compilation.

brandonwillard:
> It's wrong in that it's code that serves no purpose. I know that **in theory** it might save memory or time or something, but I've yet to see a case in which that mattered. In other words, it's a premature optimization.

Unless something about `__slots__` has changed recently, or you're assuming that Hy graphs are necessarily bounded in size or the number of keywords they contain or generate, there's no "theory" behind the benefits of using `__slots__`.  They _will_ save memory and improve attribute access latency.

Also, I don't see how this fits most criteria for a "premature optimization".  For instance, if you're referring to the classic Knuth quote, then I don't think you can say that the Hy model objects&mdash;of which `HyKeyword` is one&mdash;are a noncritical part of the program.  
Furthermore, the use of `__slots__` doesn't make `HyKeyword` any less easy to work with or understand.  As you mentioned in #1754, the `pickle` issue is even less relevant now, so what's the purpose and/or impetus for this change?

If anything, it would be arguably better if we moved to make _all_ the Hy model objects use `__slots__` so that the memory and attribute access advantages would be more impactful.

> Neither of these classes are meant to be fully immutable. `HyKeyword` has a few additional attributes `start_line` etc. that are modified as part of compilation.

Are they meant to have attributes added to them?  That's the relevant type of immutability.

Kodiologist:
>  there's no "theory" behind the benefits of using __slots__. They *will* save memory and improve attribute access latency.

I've yet to see an example. Unless you can demonstrate a program in which a meaningful amount of time or memory is time is saved by giving `__slots__` to `HyKeyword` and `Result`, it's theory at best.

> so what's the purpose and/or impetus for this change?

To remove code that created a bug and serves no purpose.

brandonwillard:
> I've yet to see an example. Unless you can demonstrate a program in which a meaningful amount of time or memory is time is saved by giving `__slots__` to `HyKeyword` and `Result`, it's theory at best.

Obviously, "meaningful" needs a very clear definition for your demand to even make sense, and I highly doubt your definition of meaningful will fit every Hy user and use-case.  Otherwise, a simple search brings up quite a few examples: [this](https://stackoverflow.com/a/28059785/3006474) looks like a popular one, and it has some performance measures that could just as easily be applied to Hy objects.  Unless you think our Python objects are somehow fundamentally different and avoid the same benefits, I'm not sure why you would require custom examples.

More importantly, why should we move in a direction opposite to memory and latency improvements?  This change goes against the idea of using `__slots__` for all basic Hy model objects; do you also believe that there are no "meaningful" benefits when _all_ the model objects use `__slots__`? 

> To remove code that created a bug and serves no purpose.

If you believe it's necessary for Hy to support an outdated, non-default `pickle` format&mdash;and, thus, consider this a "bug"&mdash;then just implement `__getstate__` and `__setstate__`.  Otherwise, you're saying that we should unnecessarily require more memory and CPU cycles because it's more convenient to delete a single line (for a marginal feature) than add roughly six or eight more.


Also, regarding the relevant type of immutability, aren't these Hy model objects supposed to model the Python primitives?  `HyKeyword` is one for which there isn't an existing primitive object to extend, and, if I'm not mistake, the majority&mdash;if not all&mdash;of the underlying Python primitives we're modeling do not allow attribute assignment.  With this change, we are now stating that `HyKeyword` should have this unusual property, even though it doesn't line up with the rest of the primitives and there's no real reason for the property itself (besides developer convenience with respect to support for an old `pickle` format).

Furthermore, the current Hy model objects that _are_ mutable in this way also share this disparity with the objects they're supposed to represent.  Maintaining this parity with Python is yet another reason to use `__slots__` for all the model objects.

Kodiologist:
> Otherwise, a simple search brings up quite a few examples: this looks like a popular one

That's a benchmark. I'm asking for an example. One could add any number of micro-optimizations that look better in benchmarks while having no meaningful effect on real programs.

> do you also believe that there are no "meaningful" benefits when *all* the model objects use `__slots__`?

That's harder to guess. I would say probably, but maybe in the case of a very large program (especially one's that grown a lot from macro expansion) it could start to make a difference. We'd have to try it to know for sure.

> Also, regarding the relevant type of immutability, aren't these Hy model objects supposed to model the Python primitives?

No. They model syntax, not semantics. In other words, a `HyInt` isn't supposed to correspond to an `int` so much as `ast.Constant`. It's confusing, I know, and it was that confusion that led to us doing silly things in earlier versions like insisting that `HySet`s not contain duplicates.

brandonwillard:
> No. They model syntax, not semantics. In other words, a `HyInt` isn't supposed to correspond to an `int` so much as `ast.Constant`. It's confusing, I know, and it was that confusion that led to us doing silly things in earlier versions like insisting that `HySet`s not contain duplicates.

That's not true, because **they extend the objects they're modelling**.

Kodiologist:
Anyway, point is, you're not going to allow this change, right? Even though only these two classes have `__slots__`? 'Cause even if this is merged, you could always try a new PR to add slots to everything, and it will be substantially easier to show a performance improvement from that than from these guys.

brandonwillard:
> If you believe it's necessary for Hy to support an outdated, non-default `pickle` format—and, thus, consider this a "bug"—then just implement `__getstate__` and `__setstate__`.

You're not even going to consider that?  Instead, you would rather dismiss this option and let someone else take care of it later?  If you want legacy `pickle` support, that's how you can add it _without_ unnecessarily sacrificing the use of `__slots__`.

Kodiologist:
I mean, that's what I'm going to resort to now because you just implicitly answered "no" to my question.

PR #1924, opened 2020-11-22
Add `__setstate__` and `__getstate__` to HyKeyword

### LINKED ISSUE #1754: HyKeyword __getstate__ w.r.t. pickle

Kodiologist:
Fixes #1754.

Replaces #1923, because @brandonwillard vetoed removing `__slots__`.

Pickling `Result` with pickle protocol 0 will still fail, but this class is only for internal use and we have no intention of pickling it, so that's not a problem.

### DIFF
--- a/hy/models.py
+++ b/hy/models.py
@@ -50,12 +50,12 @@ class HyObject(object):
     `abc` starting at the first column would have `start_column` 1 and
     `end_column` 3.
     """
-    __properties__ = ["module", "start_line", "end_line", "start_column",
-                      "end_column"]
+    properties = ["module", "start_line", "end_line", "start_column",
+                  "end_column"]
 
     def replace(self, other, recursive=False):
         if isinstance(other, HyObject):
-            for attr in self.__properties__:
+            for attr in self.properties:
                 if not hasattr(self, attr) and hasattr(other, attr):
                     setattr(self, attr, getattr(other, attr))
         else:
--- a/hy/models.py
+++ b/hy/models.py
@@ -176,6 +176,15 @@ class HyKeyword(HyObject):
                 raise
             return default
 
+    # __getstate__ and __setstate__ are required for Pickle protocol
+    # 0, because we have __slots__.
+    def __getstate__(self):
+        return {k: getattr(self, k)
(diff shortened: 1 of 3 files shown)


hy/models.py
```diff
@@ -176,6 +176,18 @@ def __call__(self, data, default=_sentinel):
...
             return default
 
+    # __getstate__ and __setstate__ are required for Pickle protocol
+    # 0, because we have __slots__.
+    def __getstate__(self):
+        return {k: getattr(self, k)
+            for k in (
+                'module', 'start_line', 'end_line', 'start_column',
+                'end_column', 'name')
+            if hasattr(self, k)}
```
brandonwillard, line 186: Instead of fixing those keys, you can use the `__properties__` inherited from `HyObject` and the local `__slots__`, no?

brandonwillard (changes requested):

brandonwillard:
Here's a reason to fix that syntax/semantics confusion:

Since `HyKeyword` is inheriting from a non-`__slots__` object, I believe it will have a `__dict__` attribute anyway, in which case we would be missing out on some of the space-saving advantages of using `__slots__`.  To fix that 
 we can make `HyObject` use `__slots__`; however, this will clash with all the `HyObject` + primitive subclasses.  

If the Hy model objects held references to the actual primitives, instead of actually _being_ those primitives, this problem would be easily solved.

Otherwise, we can probably just use the `collections.abc` interfaces.

hy/models.py
Kodiologist: Yeah, I think you're right. I'll add an earlier commit to rename `__properties__`, too, because the double underscores on both sides should be reserved for magic names like `__slots__`.

Kodiologist:
Looks good now?

brandonwillard (approved):

Issue #1858, opened 2019-12-28
Allow `let`-binding more general lvalues


falloutphil:
Not sure this qualifies as a bug, but with my python hat on I'd expect let to be able to assign to n symbols returned as an n-tuple from a function, using python's standard tuple unpacking idiom.

So you can do this:

```hy
  (let [p (returns-tuple foo bar)                                                                                                                                                                          
        first-element (first p)                                                                                                                                                                                              
        second-element (second p)]                                                                                                                                                                                                
    (blah blah blah)) 
```

But you can't do this:

```hy
  (let [(, first-element second-element) (returns-tuple foo bar)]                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
    (blah blah blah)) 
```

Kodiologist:
Yeah, I would say this isn't a bug; it's a feature you might expect but isn't implemented. It would be challenging to implement all of the kinds of lvalues that Python allows for assignment, because `let` would need to dissect the lvalue and determine which symbols to substitute. There are some variations we probably wouldn't want to implement anyway because they're too weird or borderline impossible, like `(let [(get x k) (f)] …)`; would that be expected to lexically bind an individual element of `x`?

Kodiologist:
You can always use regular assignment outside the binding list, though:

```
(let [first-element 1 second-element 1]
  (setv (, first-element second-element) (returns-tuple foo bar))
  (blah blah blah))
```

niitsuma:
dbind in
https://github.com/riktor/hycl/blob/master/hycl/core.hy
provides more general lvalues

Kodiologist:
@niitsuma It looks like that library hasn't been updated for changes to Hy, and its `let` doesn't have the same semantics as `hy.contrib.walk.let`; it just creates an anonymous function and sets the new names as local variables.

niitsuma:
In oder to import common lisp features, we need 
`(import [adderall.internal [cons car cdr cons? neseq?]])`
Plz consider add `HyExpression . append` to fix this.
https://github.com/hylang/hy/issues/1869

Kodiologist:
@niitsuma Like @brandonwillard said in the linked issue, the immutability of Hy model objects is intentional. He just merged a PR a week ago to update adderall.

niitsuma:
Thanks @brandonwillard s. Now I can use cons

brandonwillard:
[...]

PR #1930, opened 2020-11-25
Contrib/iterable unpacking let

allison-casey:
Fixes: #1858

Iterable unpacking assignment is crucial to a lot of python idioms and a feature I believe necessary for a `let` macro to have in order for it to become the go to assignment form. This implements it by delegating unpacking forms to `setv`, replacing the bindings with `gensym`'ed names that are then added to the `let` macro's dict in the same manner as regular name binds. It doesn't allow for the kind of mutable unpacking assignment that python allows like `hello['world'], b = [1, 2]` or `point.x, point.y = [1, 2]` which would make the macro a lot more complicated than it needs to be (as well as just being bad syntax imo). 

An example of the resulting python code:
```
(let [[head #* tail] [1 2 3 4]]
  [head tail])
```
becomes
```python
_hyx_letXUffffX177 = {}
[_hyx_headXUffffX178, *_hyx_tailXUffffX179] = [1, 2, 3, 4]
_hyx_letXUffffX177['head'] = _hyx_headXUffffX178
_hyx_letXUffffX177['tail'] = _hyx_tailXUffffX179
[_hyx_letXUffffX177['head'], _hyx_letXUffffX177['tail']]
```

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -3,16 +3,17 @@
 Unreleased
 ==============================
 
+New Features
+------------------------------
+* `let` macro now supports extended iterable unpacking syntax
+* New contrib module `pprint`, a Hy equivalent of `python.pprint`.
+
 Bug Fixes
 ------------------------------
 * Fixed a bug that made `hy.eval` from Python fail on `require`.
 * Fixed a bug that prevented pickling of keyword objects.
 * Fixed a compiler crash from `setv` with an odd number of arguments in
   `defclass`.
-
-New Features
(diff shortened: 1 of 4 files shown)


Kodiologist:
Looks good to me. Update the documentation for `let`, add a NEWS item, and squash.

> become the go to assignment form

I don't think this is going to happen; given Python's limited scoping capabilities, `let` is always going to be something of a hack.

Kodiologist (approved):

PR #1935, opened 2021-01-01
Update copyright years

Kodiologist:
It's that time again.


manisenkov:
Would actually something like `2013-present` work? Spot this in `babel`s [LICENSE file](https://github.com/babel/babel/blob/main/LICENSE#L3)

Kodiologist:
I doubt it, since copyright is finite in duration, despite Disney's best efforts.

PR #1940, opened 2021-01-16
Add clojure style destucturing

### LINKED ISSUE #1936: Destructuring function arguments
It would be nice to have support for destructuring within function arguments. For example,

```hy
(defn foo [[x y]]
  (+ x y))

(foo [1 2]) ; => 3
```

I think this would be relatively easy to add by slightly changing the definition of the `defn` macro, or maybe a new `defn+` macro. A possible implementation could be (this ignores any possible `&optional` and `&rest` variables)

```hy
(defmacro defn+ [fn-name args &rest body]
  (setv arg-list (lfor _ args (gensym)))
  `(defn ~fn-name
     ~arg-list
     (setv ~args ~arg-list)
     ~@body))
```

Associative destructuring like in clojure would also be really useful, but I can't see how to make this work nicely with python's list destructuring. As a starting point, this simple example works but is not recursive:

```hy
(defmacro destructure-assoc [pattern coll]
  `(setv ~(.values pattern) (lfor k ~(.keys pattern) (.get ~coll k None))))

(destructure-assoc {"as" [a b] "c" c} {"as" [4 5] "c" 6})
[a b c] ; => [4 5 6]
```

peaceamongworlds:
Closes #1936.

Builds on #1328.

Mainly updated old syntax and refactored some of the code, but the main ideas are essentially the same.

Also added `defn+`, `fn+`, and `let+` macros, which have clojure-style destructuring within their arguments.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -103,3 +103,4 @@
 * Alexey Yurchenko <homonoidian@yandex.ru>
 * Allie Jo Casey <allie.jo.casey@gmail.com>
 * Joshua Munn <public@elysee-munn.family>
+* Peter Andreev <appa@gmx.co.uk>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -8,6 +8,10 @@ Bug Fixes
 * Fixed a bug where AST nodes from macro expansion did not properly
   receive source location info.
 
+New Features
+------------------------------
+* New contrib module `destructure` for Clojure-style destructuring.
+
 0.20.0 (released 2021-01-25)
 ==============================
 
--- a/docs/contrib/destructure.rst
+++ b/docs/contrib/destructure.rst
@@ -1,0 +1,158 @@
+===========
+Destructure
+===========
+
+This module is heavily inspired by destructuring from Clojure and provides very similar semantics. It provides several macros that allow for destructuring within their arguments.
+
+Introduction
+============
+
+To use these macros, you need to require them like so:
+
+.. code-block:: hy
+
+    (require [hy.contrib.destructure [setv+ fn+ defn+ let+ defn/a+ fn/a+]])
+
+Destructuring allows one to easily peek inside a data structure and assign names to values within. For example,
+
+.. code-block:: hy
+
+    (setv+ {[{name :name [weapon1 weapon2] :weapons} :as all-players] :players
+            map-name :map
+            :keys [tasks-remaining tasks-completed]}
+           data)
+
+would be equivalent to
+
+.. code-block:: hy
+
+    (setv map-name (.get data ':map)
+          tasks-remaining (.get data ':tasks-remaining)
+          tasks-completed (.get data ':tasks-completed)
+          all-players (.get data ':players)
+          name (.get (nth all-players 0) ':name)
+          weapon1 (get (.get (nth all-players 0) ':weapons) 0)
+          weapon2 (get (.get (nth all-players 0) ':weapons) 1))
+
+where ``data`` might be defined by
+
+.. code-block:: hy
+
+    (setv data {:players [{:name Joe :weapons [:sword :dagger]}
+                          {:name Max :weapons [:axe :crossbow]}]
+                :map "Dungeon"
+                :tasks-remaining 4})
+
+This is similar to unpacking iterables in Python, such as ``a, *b, c = range(10)``, however it also works on dictionaries, and has several special options.
+
+Patterns
+========
+
+Dictionary Pattern
+------------------
+
+Dictionary patterns are specified using dictionaries, where the keys corresponds to the symbols which are to be bound, and the values correspond to which key needs to be looked up in the expression for the given symbol.
+
+.. code-block:: hy
+
+    (setv+ {a :a b "b" c (, 1 0)} {:a 1 "b" 2 (, 1 0) 3})
+    [a b c] ; => [1 2 3]
+
+The keys can also be one of the following 4 special options: ``:or``, ``:as``, ``:keys``, ``:strs``.
+
+- ``:or`` takes a dictionary of default values.
+- ``:as`` takes a variable name which is bound to the entire expression.
+- ``:keys`` takes a list of variable names which are looked up as keywords in the expression.
+- ``:strs`` is the same as ``:keys`` but uses strings instead.
+
+The ordering of the special options and the variable names doesn't matter, however each special option can be used at most once.
+
+.. code-block:: hy
+
+    (setv+ {:keys [a b] :strs [c d] :or {b 2 d 4} :as full} {:a 1 :b 2 "c" 3})
+    [a b c d full] ; => [1 2 3 4 {:a 1 :b 2 "c" 3}]
+
+Variables which are not found in the expression are set to ``None`` if no default value is specified.
+
+List Pattern
+------------
+
+List patterns are specified using lists. The nth symbol in the pattern is bound to the nth value in the expression, or ``None`` if the expression has fewer than n values.
+
+There are 2 special options: ``:&`` and ``:as``.
+
+- ``:&`` takes a pattern which is bound to the rest of the expression. This pattern can be anything, including a dictionary, which allows for keyword arguments.
+- ``:as`` takes a variable name which is bound to the entire expression.
+
+If the special options are present, they must be last, with ``:&`` preceding ``:as`` if both are present.
+
+.. code-block:: hy
+
+    (setv+ [a b :& rest :as full] (range 5))
+    [a b rest full] ; => [0 1 [2 3 4] [0 1 2 3 4]]
+
+    (setv+ [a b :& {:keys [c d] :or {c 3}}] [1 2 :d 4 :e 5]
+    [a b c d] ; => [1 2 3 4]
+
+Note that this pattern calls ``list`` on the expression before binding the variables, and hence cannot be used with infinite iterators.
+
+Iterator Pattern
+----------------
+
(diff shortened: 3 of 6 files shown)


Kodiologist:
`main` and `defmain` shouldn't be necessary in a test file. `pytest` should be able to find your tests and run them.

peaceamongworlds:
I think all the tests issues are resolved now.

Kodiologist:
Thanks for your work on this. It looks like it's coming along.

It looks to me like the syntax for dictionary destructuring is still backwards. For example, to get the number 3 from `{"a" {"b" {"c" 3}}}`, you write `(=: {{{x "c"} "b"} "a"} …)`. Wouldn't it make more sense the other way around?

It's not clear that `fn+` (or `defn+`) works properly with `&optional`, `&rest`, etc. You should probably test that.

Test how `:keys` works with mangling.

Test what happens when the destructuring pattern fails to match the rvalue, as in `(:= [x] [])`.

Describe what `ifp` does in a docstring or comment.

"This might not be necessary in the future, see hylang/hy#846" — This issue is closed.

In the tests, there are places where you use `=:` where `setv` would suffice, like the definition of `D`.

The subroutine `vals@` is only used once, so I would delete it and just write `(lfor k "abcd" (get D k))` in place of `(vals@ D "abcd")`.

In the assertion `(assert (= (, a b (tuple the-rest)) (, "a" "b" (, "c" "d" "e" "f" "g"))))`, I think it would be better to test the type of `the-rest` rather than convert it to a tuple. It should be a string in this case, I think.

Don't forget to update the copyright year.

Remove all TODO, FIXME, and similar comments. Plans for Hy should go in the issue list, not the code.

peaceamongworlds:
> It looks to me like the syntax for dictionary destructuring is still backwards. For example, to get the number 3 from `{"a" {"b" {"c" 3}}}`, you write `(=: {{{x "c"} "b"} "a"} …)`. Wouldn't it make more sense the other way around?

I think that this is a matter of convention. Since I was trying to copy the way clojure does destructuring, it makes sense to me to copy this as well.

> It's not clear that `fn+` (or `defn+`) works properly with `&optional`, `&rest`, etc. You should probably test that.
> 
> Test how `:keys` works with mangling.

Added a test for this. It seems to just work.
```hy
(=: {:keys [a? -b? c->d]} {:a? 1 :-b? 2 :c->d 3})
[a? -b? c->d] ; => [1 2 3]
```

> Test what happens when the destructuring pattern fails to match the rvalue, as in `(:= [x] [])`.

Sorry I should have added more documentation. All the 'patterns' return `None` if they don't find the variable, so in this case `x` would be `None`. However if the shape of your 'pattern' doesn't match the shape of the data, then you'll get an error.

> Describe what `ifp` does in a docstring or comment.
> "This might not be necessary in the future, see #846" — This issue is closed.
> The subroutine `vals@` is only used once, so I would delete it and just write `(lfor k "abcd" (get D k))` in place of `(vals@ D "abcd")`.
> Don't forget to update the copyright year.
> Remove all TODO, FIXME, and similar comments. Plans for Hy should go in the issue list, not the code.

Done

> In the assertion (assert (= (, a b (tuple the-rest)) (, "a" "b" (, "c" "d" "e" "f" "g")))), I think it would be better to test the type of
> the-rest rather than convert it to a tuple. It should be a string in this case, I think.

I made both list destructuring and iterator destructuring convert their arguments into a list and iterator respectively, so the rest argument will be a list or an iterator, not a string. I think that this makes more sense this way as it allows iterable to be passed to both. So something like `(=: [a :& rest] (filter even? (range 10)))` will fail because filter objects aren't subscriptable.

> It's not clear that fn+ (or defn+) works properly with &optional, &rest, etc. You should probably test that.

I added keyword destructuring to lists, which allows the `:&` argument in lists to be destructured as either a list or a dictionary. So something like this should work:

```hy
(defn+ foo [[a b] {:keys [c d]} :& {:strs [e f]}]
    "bar foo"
    [a b c d e f])

(foo [1 2] {:c 3 :d 4} "e" 5 "f" 6) ; => [1 2 3 4 5 6]
```
Thus you don't really need `&rest` etc. If you do add them, they won't be themselves recursively destructable like the other arguments, hence I thought this made more sense.

Kodiologist:
Good work!

> In the tests, there are places where you use `=:` where `setv` would suffice, like the definition of `D`.

I think you forgot this one.

Test `strs` with mangling, too. It looks like `(=: {:strs [hello?]} {"is_hello" 1})` fails to match `hello?`, whereas `(=: {:strs [hello?]} {"hello?" 1})` works; is this intentional?

While it may be true that [the special symbols like `&optional`](https://docs.hylang.org/en/stable/language/api.html#defn) aren't necessary to write functions with `fn+`, users may still try to use them, perhaps by mistake, so test what should happen. For example, does `(fn+ [&optional [a b]] …)` unpack a tuple, use `b` as the default value of `a`, raise an error, or something else?

peaceamongworlds:
> > In the tests, there are places where you use `=:` where `setv` would suffice, like the definition of `D`.
> 
> I think you forgot this one.

Fixed

> Test `strs` with mangling, too. It looks like `(=: {:strs [hello?]} {"is_hello" 1})` fails to match `hello?`, whereas `(=: {:strs [hello?]} {"hello?" 1})` works; is this intentional?

I didn't think of this, so currently, there the macro doesn't do any explicit mangling. What should be the desirable behaviour? Unmangle everything?

> While it may be true that [the special symbols like `&optional`](https://docs.hylang.org/en/stable/language/api.html#defn) aren't necessary to write functions with `fn+`, users may still try to use them, perhaps by mistake, so test what should happen. For example, does `(fn+ [&optional [a b]] …)` unpack a tuple, use `b` as the default value of `a`, raise an error, or something else?

Currently all the special symbols are just treated like any other variable name. So `(fn+ [&optional [a b]] ...)` expects 2 variables, and it will attempt to destructure the second one. Should I explicitly disallow their use, or just specify in the documentation that they aren't special?

Kodiologist:
> I didn't think of this, so currently, there the macro doesn't do any explicit mangling. What should be the desirable behaviour? Unmangle everything?

Hmm. It's not obvious. Since the introduction of "late mangling" in #1517, the symbol `hello?` really does stringify to "hello?"; mangling happens only at compile-time and only for certain symbols. I'd say let's keep this behavior for now, with a test that makes it explicit, and if we have to change it later, we can.

> Should I explicitly disallow their use, or just specify in the documentation that they aren't special?

I think the latter makes sense. And you can add a test that shows that all the would-be special symbols are treated as ordinary in this context.

peaceamongworlds:
> > I didn't think of this, so currently, there the macro doesn't do any explicit mangling. What should be the desirable behaviour? Unmangle everything?
> 
> Hmm. It's not obvious. Since the introduction of "late mangling" in #1517, the symbol `hello?` really does stringify to "hello?"; mangling happens only at compile-time and only for certain symbols. I'd say let's keep this behavior for now, with a test that makes it explicit, and if we have to change it later, we can.
> 
> > Should I explicitly disallow their use, or just specify in the documentation that they aren't special?
> 
> I think the latter makes sense. And you can add a test that shows that all the would-be special symbols are treated as ordinary in this context.

Done

Kodiologist:
Okay, looking good. Here's what's left:

- Write a documentation page for the exported macros. Since we don't currently have a way to synchronize docstrings with the manual, remove the docstrings. (You can still have docstrings for internal stuff that isn't in the manual, like `ifp`.)
- Add an item to NEWS.
- Squash your changes. (This PR doesn't make any changes to old code; it just adds new code; so I don't think we need to preserve the process.)
- In a separate commit, add yourself to AUTHORS.

peaceamongworlds:
I decided to rename ``=:`` to ``setv+``. That seems more consistent with the other macro names, and I just like it more than ``=:``.

peaceamongworlds:
> Okay, looking good. Here's what's left:
> 
>     * Write a documentation page for the exported macros. Since we don't currently have a way to synchronize docstrings with the manual, remove the docstrings. (You can still have docstrings for internal stuff that isn't in the manual, like `ifp`.)
> 
>     * Add an item to NEWS.
> 
>     * Squash your changes. (This PR doesn't make any changes to old code; it just adds new code; so I don't think we need to preserve the process.)
> 
>     * In a separate commit, add yourself to AUTHORS.

I think that's all done now.

Kodiologist:
I think you need to add a link to your new page from `index.rst` for it to be navigated to. Otherwise, looks good. Thanks for your contributions, and welcome aboard!

Kodiologist (approved):

peaceamongworlds:
> I think you need to add a link to your new page from `index.rst` for it to be navigated to. Otherwise, looks good. Thanks for your contributions, and welcome aboard!

Thanks! I've added it.

allison-casey:
This makes me unseemly happy! Thank you peaceamongworlds!

allison-casey:
By the way, should we also have `defn/a+` and `fn/a+` variants?

Kodiologist:
If you like.

peaceamongworlds:
> By the way, should we also have `defn/a+` and `fn/a+` variants?

I've added them now.

allison-casey:
It doesn't look like the async variants don't actually expand to use the `defn/a` and `fn/a` forms

peaceamongworlds:
> It doesn't look like the async variants actually expand to use the `defn/a` and `fn/a` forms

That was silly of me. It's fixed now.

allison-casey:
You dropped this :crown:

gilch:
This also closes #1326.

PR #1947, opened 2021-01-23
Documents *{1,3} and *e

### LINKED ISSUE #1943: Document repl builtins *e, *i
The repl commands `*e, *1, *2 ...` don't seem to be documented. This became an issue when attempting to get [ob-hy](https://github.com/brantou/ob-hy) to behave (it was relying on `_` to return the result of last evaluation to emacs). It looks like this was changed in #1341. I'd like to submit a pull request for this, but would appreciate suggestions on where it should go :).

jams2:
Closes #1943 

Added `repl.rst` after `cli.rst`, and documented the recent results and recent error values. I also gave a (very) brief overview of the REPL itself, as I thought it should be there - let me know if it's at all erroneous.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -102,3 +102,4 @@
 * Xi Jin <xij@usc.edu>
 * Alexey Yurchenko <homonoidian@yandex.ru>
 * Allie Jo Casey <allie.jo.casey@gmail.com>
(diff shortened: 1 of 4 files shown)


Kodiologist:
In the future, make a new branch of each of your PRs instead of requesting a pull from master.

Issue #1948, opened 2021-01-24
Add support for "=" syntax in  f-strings


jams2:
Since 3.8, Python has had "[equal sign syntax](https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals)" in f-strings, which can be useful for debugging. For example:

```
In [2]: the_string = "foo"

In [9]: f"{the_string=}"
Out[12]: "the_string='foo'"

In [13]: f"{the_string=:5}"
Out[16]: 'the_string=foo  '

In [19]: from datetime import datetime; the_object=datetime.now()

In [29]: f"{the_object=!r}"
Out[32]: 'the_object=datetime.datetime(2021, 1, 24, 21, 12, 23, 550361)'

In [33]: f"{the_object=!s}"
Out[33]: 'the_object=2021-01-24 21:12:23.550361'

In [34]:
```

It would be great to have this in Hy. It looks like this could be handled in `_format_string` in `hy.compiler`, after the handling for format specs. The simple case (`f"{x=}"`) could potentially be handled in place, with the case of a trailing format specification being handled recursively.

Would you consider adding this in spite of it being for Python >=3.8? This is cool - since the parsing is handled by Hy, I think we could make this feature compatible for Hy running on older versions of Python. An example (excuse the clumsy use of pickle to ship the tree around different interpreters, I'm new to `ast`):

```
Python 3.9.1 (default, Dec 13 2020, 11:55:53)

In [97]: tree = ast.parse("print(f'{foo=}')")

In [121]: pickle.dump(tree, open("/home/joshua/test.pickle", "wb"))
```

```
Python 3.6.12 (default, Jan 24 2021, 16:44:51)

>>> tree = pickle.load(open("/home/joshua/test.pickle", "rb"))
>>> tree
<_ast.Module object at 0x7fae44ac09e8>
>>> compile(tree, filename="string", mode="exec")
<code object <module> at 0x7fae44981270, file "string", line 1>
>>> eval(compile(tree, filename="string", mode="exec"))
foo='bar'
```

I'm happy to work on this if it's desirable, and happy for any pointers also.

Kodiologist:
I'd love to have this, and you're more than welcome to take a crack at it. How do you imagine the syntax working, exactly?

> I think we could make this feature compatible for Hy running on older versions of Python

[...]

PR #1950, opened 2021-01-25
Walkable f-strings

### LINKED ISSUE #1843: `hy.contrib.walk` cannot see inside f-strings, fails to replace contained references
The AST walker in `hy.contrib.walk` doesn't see inside of an f-string, so the `let` macro built on it will fail when the inlined code references a let-bound variable:

```
hy unknown using CPython(default) 3.7.5 on Darwin
=> (require [hy.contrib.walk [*]])
True
=> (let [a 1] f"{a}")
  File "stdin-15ddbd2ca7b71de7866cc45806e8764b30264dcd", line 1, in <module>
    (let [a 1] f"{a}")
NameError: name 'a' is not defined
```

"unknown" is "master as of the time of submission"

### LINKED ISSUE #1938: hy.extra.anaphoric macros don't descend into f-strings
f-string works in these functions: 

```
(defn p [x] f"->{(+ 3 x)}")

=> (p 4)
'->7'

((fn [x] f"->{(+ 3 x)}") 4)
'->7'

```
But not in anonymous fonction defined with #%

```
(require [hy.extra.anaphoric [*]])

(#%(f"->{(+ 3 %1)}") 4)
stdin-4625cfffcbb8b22a8d640f29313ec5e7b2169a0f:1: SyntaxWarning: 'str' object is not callable; perhaps you missed a comma?
  (#%(f"->{(+ 3 %1)}") 4)
  File "stdin-4625cfffcbb8b22a8d640f29313ec5e7b2169a0f", line 1, in <module>
    (#%(f"->{(+ 3 %1)}") 4)
TypeError: <lambda>() takes 0 positional arguments but 1 was given


```

 hy --version

hy 0.19.0

scauligi:
Hopefully not too intrusive of a PR, but it's a breaking change.
Introduces two new `HySequence` nodes: `HyFComponent` and `HyFString`.
f-strings are now parsed as `HyFString`s, allowing routines like `walk` and `recur-sym-replace` to descend into expressions within f-strings.

This fixes #1843 and fixes #1938, and I added some tests to ensure no regressions for either.
This also tangentially affects #1948, as the relevant f-string format parsing has been relocated to `lex/parser.py`.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -104,3 +104,4 @@
 * Allie Jo Casey <allie.jo.casey@gmail.com>
 * Joshua Munn <public@elysee-munn.family>
 * Peter Andreev <appa@gmx.co.uk>
+* Sunjay Cauligi <scauligi@eng.ucsd.edu>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -3,15 +3,21 @@
 Unreleased
 ==============================
 
+Breaking Changes
+------------------------------
+* f-strings are now parsed as a separate `HyFString` node,
+  which is a collection of `HyString` and `HyFComponent` nodes.
+
+New Features
+------------------------------
+* New contrib module `destructure` for Clojure-style destructuring.
+
 Bug Fixes
 ------------------------------
 * Fixed a bug where AST nodes from macro expansion did not properly
   receive source location info.
 * Fixed bug in `smacrolet` by replacing `module-name` argument with `&name`.
-
-New Features
-------------------------------
-* New contrib module `destructure` for Clojure-style destructuring.
+* Fixed expressions within f-strings being inaccessible to macros.
 
 0.20.0 (released 2021-01-25)
 ==============================
--- a/hy/__init__.py
+++ b/hy/__init__.py
@@ -15,7 +15,7 @@ def _initialize_env_var(env_var, default_val):
     return res
 
 
-from hy.models import HyExpression, HyInteger, HyKeyword, HyComplex, HyString, HyBytes, HySymbol, HyFloat, HyDict, HyList, HySet  # NOQA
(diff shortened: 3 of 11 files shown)


jams2:
Heh, I was in the middle of investigating the feasibility of doing this for #1948

Kodiologist:
I like this, but it needs a little cleanup. You have a few instances where you're adding three newlines in a row, or a newline that's far from the code you're adding. The commit structure is also not ideal because you add code that you later remove in the same PR, like a bunch of asserts. It would be okay to squash everything into one commit, but better would be to have the big copy-paste from `compiler.py` to `lexer.py` in its own commit, so the reader can see in the diff between commits how you changed that code.

scauligi:
Alright! I'll rebase the commits into cleaner chunks.

scauligi:
This set of commits should be much cleaner, without the adding/removing asserts and extraneous newlines.
I split out some notable changes (moving `_format_string` and renaming `STR` to `LSTR`) into their own commits for easier diff reading.

Kodiologist:
Beautiful. A few more things:

- The compiler never expects a `STR`, by your new definition, just an `LSTR`. So I think the redefinition and the introduction of `LSTR` are unnecessary. Perhaps you were worried about the indirect backwards-incompatible change to `STR` from f-strings no longer being  `HyString`s, but I think this is fine: usage of model patterns outside of Hy itself is limited so far, and ultimately I think it was a bug that `STR` matched something that's not really a string literal.
- In `test-walking-update`, save the big new form to a variable instead of writing it out twice.
- Prepend each of your new tests against a now-fixed bug with a comment with the URL of the corresponding GitHub issue. Likewise, mention the fixed bugs in NEWS (actually, you could probably describe it as a single bug there).
- Put your addition to AUTHORS in its own commit.

scauligi:
Awesome, thanks for the comments/code review! I've rebased again with these new changes.

Kodiologist (approved):

PR #1951, opened 2021-01-26
Fixes #1944: macro-expanded ast nodes had bad lineno/col info

### LINKED ISSUE #1944: Compiler crash from invalid use of `#@`
```hy
$ > hy
hy 0.19.0+36.g14b17afd using CPython(default) 3.9.1 on Linux
=> (#@ a
... b)
  File "/home/peter/Programming/hy/hy-source/venv/bin/hy", line 12, in <module>
    sys.exit(hy_main())
  File "/usr/lib/python3.9/code.py", line 232, in interact
    more = self.push(line)
  File "/usr/lib/python3.9/code.py", line 258, in push
    more = self.runsource(source, self.filename)
  File "/home/peter/Programming/hy/hy-source/hy/hy/errors.py", line 62, in __init__
    self.compute_lineinfo(expression, filename, source, lineno, colno)
  File "/home/peter/Programming/hy/hy-source/hy/hy/errors.py", line 94, in compute_lineinfo
    self.arrow_offset = len(self.text[0])
⤷ Traceback (most recent call last):
  File "/home/peter/Programming/hy/hy-source/hy/hy/compiler.py", line 435, in compile
    ret = self.compile_atom(tree)
  File "/home/peter/Programming/hy/hy-source/hy/hy/compiler.py", line 429, in compile_atom
    return Result() + _model_compilers[type(atom)](self, atom)
  File "/home/peter/Programming/hy/hy-source/hy/hy/compiler.py", line 1778, in compile_expression
    return Result() + build_method(
  File "/home/peter/Programming/hy/hy-source/hy/hy/compiler.py", line 945, in compile_decorate_expression
    raise self._syntax_error(args[-1],
  File "/home/peter/Programming/hy/hy-source/hy/hy/compiler.py", line 454, in _syntax_error
    return HySyntaxError(message, expr, self.filename, self.source)
  File "/home/peter/Programming/hy/hy-source/hy/hy/errors.py", line 62, in __init__
    self.compute_lineinfo(expression, filename, source, lineno, colno)
  File "/home/peter/Programming/hy/hy-source/hy/hy/errors.py", line 94, in compute_lineinfo
    self.arrow_offset = len(self.text[0])
IndexError: string index out of range
```
Only seems to happen with the `#@` tag macro. Note the split line - doesn't happen otherwise.

scauligi:
Fixes #1944.

The AST nodes coming out of the `#@` expansion had a line/column number of 0 because of how "default" HyObject properties were being handled, causing an index underflow.
This change patches HyObjects to handle default properties better (and also changes the default info to 1 to prevent any future underflow bugs).

`wrap_value()` in `models.py` initialized `HyObject`s with 0'd out source locations, causing `HyObject.replace()` to not update these objects with the proper source locations later as it looks like they're already set.
I added a `__getattr__` method to `HyObject` so that it can return a default value for these properties if they're not yet set, while still allowing them to be overridden if they have not truly been initialized.

### DIFF
--- a/hy/models.py
+++ b/hy/models.py
@@ -50,7 +50,7 @@ class HyObject(object):
     `abc` starting at the first column would have `start_column` 1 and
     `end_column` 3.
     """
-    properties = ["module", "start_line", "end_line", "start_column",
+    properties = ["module", "_start_line", "end_line", "_start_column",
                   "end_column"]
 
     def replace(self, other, recursive=False):
@@ -62,6 +62,22 @@ class HyObject(object):
             raise TypeError("Can't replace a non Hy object '{}' with a Hy object '{}'".format(repr(other), repr(self)))
 
         return self
+
+    @property
+    def start_line(self):
+        return getattr(self, "_start_line", 1)
+
+    @start_line.setter
+    def start_line(self, value):
+        self._start_line = value
+
+    @property
+    def start_column(self):
+        return getattr(self, "_start_column", 1)
+
+    @start_column.setter
+    def start_column(self, value):
+        self._start_column = value
 
     def __repr__(self):
         return "%s(%s)" % (self.__class__.__name__, super(HyObject, self).__repr__())
@@ -84,10 +100,6 @@ def wrap_value(x):
         raise HyWrapperError("Don't know how to wrap {!r}: {!r}".format(type(x), x))
     if isinstance(x, HyObject):
         new = new.replace(x, recursive=False)
-    if not hasattr(new, "start_column"):
-        new.start_column = 0
-    if not hasattr(new, "start_line"):
-        new.start_line = 0
     return new
 
 
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,5 +1,13 @@
 .. default-role:: code
 
+Unreleased
+==============================
+
+Bug Fixes
+------------------------------
+* Fixed a bug where AST nodes from macro expansion did not properly
+  receive source location info.
+
 0.20.0 (released 2021-01-25)
 ==============================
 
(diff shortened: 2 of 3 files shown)


Kodiologist:
We don't still need the `assert attr in self.__dict__`, right? Also, you might add a comment here briefly hinting why we have to check `__dict__` attributes instead of using `hasattr`.

Instead of the explicit `raise AttributeError(f"{type(self).__name__ !r} has no attribute {attr !r}")`, is it possible to fall back to `object.__getattr__` via `super`?

scauligi:
Ah crap, the assert was left over from testing. I'll remove that and rebase, as well as adding comments for why we need to check `__dict__` explicitly.
re using `super()` instead of an explicit exception: when I originally tried that I was getting "superclass has no attribute __getattr__" exceptions, but that doesn't seem to be the case anymore. I was probably doing something else wrong that has since been fixed. I'll change this to fall back via `super()`.

Kodiologist (approved):

hy/models.py
```diff
@@ -56,13 +56,22 @@ class HyObject(object):
...
             raise TypeError("Can't replace a non Hy object '{}' with a Hy object '{}'".format(repr(other), repr(self)))
 
         return self
 
+    def __getattr__(self, attr):
+        if attr in ("start_line", "start_column"):
+            return 1
+        else:
+            return super().__getattr__(attr)
+
```
brandonwillard, line 74: Couldn't a `property` be used for these two?  That would be less intense than overriding `__getattr__`.

brandonwillard (changes requested):

hy/models.py
scauligi: That's entirely reasonable, and I'm somewhat embarrassed for not having thought about that.
Updated and rebased, thanks for the catch!

brandonwillard (approved):

PR #1952, opened 2021-01-26
Refactor/docs overhaul

### LINKED ISSUE #1044: set up Sphinx autodoc to put our docstrings in our reference docs.

### LINKED ISSUE #1423: Macros made with defmacro/g! lose their docstring
#1415 can read these. We might pick some other way to do it, like #1416 but they shouldn't get dropped.
```Python
=> (defmacro/g! foo [] "docstring/g!" 3)
<function _hy_anon_fn_1 at 0x00000290BC012EA0>
=> (help (. hy macros _hy_macros [__name__] ['foo]))
Help on function _hy_anon_fn_1:

_hy_anon_fn_1()

=> (defmacro bar [] "docstring" 4)
<function _hy_anon_fn_1 at 0x00000290BBFD8D90>
=> (help (. hy macros _hy_macros [__name__] ['bar]))
Help on function _hy_anon_fn_1:

_hy_anon_fn_1()
    docstring
```

### LINKED ISSUE #1424: Docstring comes out in expansion of defmacro!
Macros defined with `defmacro!` put too much in the expansion. This isn't obvious unless you have `--spy` on or use `disassemble`. I also see a lot of `None`s for some reason. It seems to be the result of each form in the `defmacro!` definition, instead of just the final one.
```Python
=> (defmacro! foo [] "doc" 42)
from hy.core.language import interleave
import hy

def _hy_anon_fn_1():
    None
    return HyExpression(((([] + [HySymbol('do')]) + [HyExpression((([] + [HySymbol('setv')]) + list((interleave([], []) or []))))]) + list((['doc', 42] or []))))
hy.macros.macro('foo')(_hy_anon_fn_1)
<function _hy_anon_fn_1 at 0x000001E6E05F4F28>
=> (foo)
None
'doc'
42
42
=> (disassemble '(foo))
from hy.core.language import disassemble
disassemble(HyExpression(([] + [HySymbol('foo')])))
"Module(body=[Expr(value=Name(id='None')), Expr(value=Str(s='doc')), Expr(value=Num(n=42))])"
```

### LINKED ISSUE #1845: wishlist: a unified index of symbols in the hy distribution, perhaps also contrib
In the current documentation the set of symbols I can reference in core hy is split into multiple distinct sections in the left-hand-side index: "Built-Ins", "Hy Core", "Extra Modules Index", and "Contributor Modules Index".  The "Hy Core" is further subdivided into "only one open at a time" lists of core, sequence, and itertools functions.

Especially with the built-in/core split this makes it more difficult than I'd like to navigate the system: I don't especially care if `instance?` is a built-in, or part of the core library, or supplied by itertools, when I just wanted to know what the function signature was.  

(OK, that one is kind of a gimmie when it comes to figuring out which section, but definitely things like `filter` or `name` are way harder to guess.)

Anyway, it'd be great if there was a single index that listed all those items so that I could scan the list to find what I'm looking for, click it, and get the right docs immediately.  In the Common Lisp HyperSpec (CLHS) this is their symbol index – it only covers symbols, as in functions, global vars, etc, and not, eg, syntax discussion, or anything like that.

I am not familiar enough with Python documentation generators to know how/if this can be done, and I'm unfortunately not in a position for a couple of months due to USA related work issues to be able to contribute to make this happen.  Sorry.

### LINKED ISSUE #1915: Document that identifiers can't start with `^`
Instead of submitting a potentially ambiguous or incomplete narrative, I'm going to post  a console log that demonstrates the problem.

```
xheimlich@xheimlich:~$ python3 -m venv hy-repr-bug
xheimlich@xheimlich:~$ source hy-repr-bug/bin/activate
(hy-repr-bug) xheimlich@xheimlich:~$ python --version
Python 3.6.9

(hy-repr-bug) xheimlich@xheimlich:~$ cat test.hy
(defmacro ^fix-keys [fields] `(+ ["ad"] ~fields))

(hy-repr-bug) xheimlich@xheimlich:~$ pip install hy==0.17.0
Collecting hy==0.17.0
  Using cached https://files.pythonhosted.org/packages/17/67/c7d078103e2b986374be547bb7dcd2e14e398bc844277cbcd8f38dd36b6c/hy-0.17.0-py2.py3-none-any.whl
Collecting astor>=0.8 (from hy==0.17.0)
  Using cached https://files.pythonhosted.org/packages/c3/88/97eef84f48fa04fbd6750e62dcceafba6c63c81b7ac1420856c8dcc0a3f9/astor-0.8.1-py2.py3-none-any.whl
Collecting funcparserlib>=0.3.6 (from hy==0.17.0)
Collecting clint>=0.4 (from hy==0.17.0)
Collecting rply>=0.7.7 (from hy==0.17.0)
  Using cached https://files.pythonhosted.org/packages/e7/82/388e0845abda4bba5ba041aa32e411b3eed56f518c96c6565559e713a5ed/rply-0.7.7-py2.py3-none-any.whl
Collecting args (from clint>=0.4->hy==0.17.0)
Collecting appdirs (from rply>=0.7.7->hy==0.17.0)
  Using cached https://files.pythonhosted.org/packages/3b/00/2344469e2084fb287c2e0b57b72910309874c3245463acd6cf5e3db69324/appdirs-1.4.4-py2.py3-none-any.whl
Installing collected packages: astor, funcparserlib, args, clint, appdirs, rply, hy
Successfully installed appdirs-1.4.4 args-0.1.0 astor-0.8.1 clint-0.5.1 funcparserlib-0.3.6 hy-0.17.0 rply-0.7.7

(hy-repr-bug) xheimlich@xheimlich:~$ hy -i test.hy
<function hyx_Xcircumflex_accentXfix_keys at 0x7f44a0771620>
hy 0.17.0 using CPython(default) 3.6.9 on Linux
=> (^fix-keys ["2"])
['ad', '2']
=> 
now exiting HyREPL...
```
That's the expected behavior. Then:
```
(hy-repr-bug) xheimlich@xheimlich:~$ pip install hy==0.19.0
Collecting hy==0.19.0
  Using cached https://files.pythonhosted.org/packages/9d/7d/24decc923f804e7b46118ab95e380473cffe6403829569c7d2f2d9a2aad2/hy-0.19.0-py2.py3-none-any.whl
Collecting colorama (from hy==0.19.0)
  Using cached https://files.pythonhosted.org/packages/c9/dc/45cdef1b4d119eb96316b3117e6d5708a08029992b2fee2c143c7a0a5cc5/colorama-0.4.3-py2.py3-none-any.whl
Requirement already satisfied: rply>=0.7.7 in ./hy-repr-bug/lib/python3.6/site-packages (from hy==0.19.0)
Requirement already satisfied: astor>=0.8 in ./hy-repr-bug/lib/python3.6/site-packages (from hy==0.19.0)
Requirement already satisfied: funcparserlib>=0.3.6 in ./hy-repr-bug/lib/python3.6/site-packages (from hy==0.19.0)
Requirement already satisfied: appdirs in ./hy-repr-bug/lib/python3.6/site-packages (from rply>=0.7.7->hy==0.19.0)
Installing collected packages: colorama, hy
  Found existing installation: hy 0.17.0
    Uninstalling hy-0.17.0:
      Successfully uninstalled hy-0.17.0
Successfully installed colorama-0.4.3 hy-0.19.0
(hy-repr-bug) xheimlich@xheimlich:~$ hy -i test.hy
  File "(defmacro ^fix-keys [fields] `(+ ["ad"] ~fields))
-ff46b2fbd8655b569bbd82c0325f4e33cb4e367d", line 1, in <module>
  File "/usr/lib/python3.6/contextlib.py", line 99, in __exit__
    self.gen.throw(type, value, traceback)
  File "/home/xheimlich/hy-repr-bug/lib/python3.6/site-packages/hy/core/bootstrap.hy", line 18, in defmacro
    (. (type name) __name__))
  File "(defmacro ^fix-keys [fields] `(+ ["ad"] ~fields))
", line 1
    (defmacro ^fix-keys [fields] `(+ ["ad"] ~fields))
    ^-----------------------------------------------^
expanding macro defmacro
  NameError: name 'name' is not defined


hy 0.19.0 using CPython(default) 3.6.9 on Linux
=> 
```

I didn't snip pip output in case some dependencies are being installed wrong for whatever reason, although as you can see it's a fresh environment.

allison-casey:
Complete overhaul of the Hy docs to unify the documentation for the core api in one place. Uses my [Sphinx Hy Domain extension](https://github.com/allison-casey/sphinxcontrib-hydomain) (do we want to bring this under hylang while were at it? It's not *complete*, but it has what we need for right now) to autogenerate the docs for almost all the core library. Existing method docs have been migrated into the docstrings of their respective methods/modules. All method documentation is now unified under `api.rst` and navigable from `cheatsheet.rst` (I need help defining the categories for this). Closes #1845, Closes #1915, Closes #1044 

Only core lib code changes, were adding docstring pass through to `defmacro/g!` and `defmacro!` which closes closes #1423 and closes #1424

I'm sure I've missed some things and there might be some bugs in `hydomain` I haven't noticed, but at this point I think it needs some more eyes on it because I'm starting to go blind to it lol.


Kodiologist:
I'll be honest, this is intimidating to review, and I don't even plan to review sphinxcontrib-hydomain. (Which, yeah, I agree, go ahead and move it to the hylang organization.) Thanks for doing this. Any suggestions for what I should look at in what order?

It may have occurred to you during this process that Hy's core is a mess. All the functions and macros are in one automatically polluting namespace instead of being organized sanely, and some are just stupid: for example, `instance?` has its arguments in the opposite order from `isinstance`. I do plan to clean things up eventually, with the user explicitly importing things from `hy.core` or something (or standard modules, in the case of things we're currently just re-exporting) in place of auto-import; auto-import should only be used when necessary (e.g., for Hy model objects when constructing forms with a backtick) and only affect a single user-visible name, like `hy`. But look how many years I've been working on Hy and I still haven't done this. I guess the point is, don't get too attached to any new documentation you write for some of these guys.

My vague idea about how to document the operators was to document the special forms and their shadowed versions in the same place, with a paragraph describing how shadowing works and how shadowed versions differ from the real operators (viz., no short-circuiting, and obviously it's a function call with the associated overhead—I think that's about it). There could be a table with operators as the rows and the effects on 0 arguments, 1 argument, 2 arguments, and more than 2 arguments as the columns; Python syntax might be a good way to concisely describe the semantics corresponding to each cell. For example, for Hy's `-`, 0 arguments aren't allowed, 1 argument means `-x`, 2 arguments mean `a - b`, and more than 2 arguments mean `x1 - x2 - x3 - x4…`.

allison-casey:
I understand, this one is pretty big. Honestly, I would just try building the docs and and peruse the cheatsheet and api. Thre isn't much actual documentation or code change to look through besides the defmacro! variants. 

And yeah, I've noticed. A lot of what I want to do with Hy right now is clean up the code base and build out the tooling to make using Hy an easier more viable experience, and this felt like a good first step. I haven't actually written much documentation for this pull request, which was kind of the point. I've only unified the existing documentation with their respective docstrings. So going forward it doesn't really matter what you end up doing with the functions, the documentation will travel with them. I think it would be worth opening a new issue to discuss all that we'd want to do to clean up the code base.

let me mock up an example of what I think you're getting at.

Kodiologist:
This is a big improvement over the old status quo, no doubt about that.

Why do square brackets appear in lambda lists in headers, such as `(await [obj])`? My first guess was that they would mean optional arguments, but `await`'s argument is mandatory.

The cheat sheet could be handy, but it could also be a maintenance burden. #1845, at least, is most readily satisfied by the Sphinx index; don't forget to link to that from `index.rst`.

The code snippet with `[head #* tail]` in the docstring for `let` is highlighted as if `#` was a comment character. Is this a bug in the Hy domain?

allison-casey:
leaving the lambda list brackets in was just how I thought would best encompass all the different ways functions can be defined. Especially when you add typings into the mix. A typed function could either look like.

```hy
(some-fn ^int [^str hello ^str world])
```
or without the brackets
```hy
(some-fn ^int ^str hello ^str world])
```
Which seemed confusing to me.

The cheatsheet is mostly auto generated from a json spec that looks like 
```json
["Collections", ["hy.core.language.first",
                        "hy.core.language.rest"]]]
```
So it's not too bad, especially if we make it a pre build step in the sphinx ``Makefile``. It's been a big help for me in other languages ([clojure](https://clojure.org/api/cheatsheet)) and libraries ([funcy](https://funcy.readthedocs.io/en/stable/cheatsheet.html)) that have them. If this were the case do you think it would still be a maintenance burden? If you think so I'll drop it in favour of a general index. 

I think sphinx is just assuming that the code snippets are actually python. Let me see if there's a way to change that.

allison-casey:
So it looks like the sphinx ``>>>`` code block is unchangeably set to python. I'll go through and use regular code blocks for everything.

allison-casey:
So I did a few things

- made `clojure` the default syntax highlighting
- unified all the `code-block` directives to use `::`
  - We're still gonna get `could not lex` errors because not all Hy is valid Clojure but until we update the Hy `Pygments` lexer it's the best we're gonna get. 
- moved all `>>>` diretives to `::` and `=>` forms
- deleted the legacy api documentation files now that they're all in `api.rst`
- fixed all the broken cross references

Kodiologist:
Cool. I see your point regarding type annotations, but I think the result is a confusing hybrid between the syntax for calling operators and the syntax for defining them. Definitions begin with `defn` or `defmacro` or something, whereas calls don't include annotations and don't bracket the arguments. Perhaps you can use something like `(some-fn ^int ^str hello ^str world)` if you use a symbol or color to distinguish the type annotation for the return value from everything else. After all, type annotations are rare (and nonexistent for special operators and macros), so we should reserve the extra punctuation for them.

Automatically generating the cheatsheet makes sense. Can you make it part of the documentation build process (ideally through `conf.py` rather than Make) instead of committing it?

Kodiologist:
You mention doctest in some of your commit messages. Is that working? That would be really cool (although this is a very big PR already). How do I run them?

allison-casey:
That's a good idea. I should be able to muck around with how its presented. I'll give it a go after i track down this macro role bug in hydomain. 

Sure! Currently it's just a hy file i run before i build the docs with `gen_cheatsheet && make clean && make html`. I'll port it over to try porting it over to `conf.py`

I had been using the sphinx doctest role `>>>` for the examples code blocks since they're the most readable, but unfortunately sphinx doctest is hardcoded to work only with python. I might take a look at it later as I continue building out the tooling.

allison-casey:
Alright I brought the autogen cheatsheet code into `conf.py` and it seems to be working fine. Also, I don't think we're going to be able to have the type annotations work the way you were thinking.

This would work fine: 

`(some-fn ^int ^str hello ^str world)`

But there's no way to disambiguate a function with a return type but no type for the first argument:

`(some-fn ^int hello ^str world)`
versus
`(some-fn ^int [hello ^str world])`

This only works if the brackets are there to delineate the return annotation from the first argument.

Kodiologist:
I don't follow. I suggested that "you use a symbol or color to distinguish the type annotation for the return value from everything else". For example, suppose only the return value were enclosed with `(returns …)`, so you'd get things like

    (some-fn (returns ^int) ^str hello ^str world)
    (some-fn (returns ^int) hello ^str world)
    (some-fn ^int hello ^str world)

That's unambiguous, right?

allison-casey:
I understand now. We could do the explicit `(some-fn (returns ^int) hello world)` option. In my mind though it's a bigger departure from the actual syntax of function definitions than the original. Python docs in sphinx follow the exact function *definition* syntax of the object and not the *calling* syntax. 
```
.. py:function some-fn(hello, world) -> int
```
is the python equivalent of 
```
.. hy:function:: (some-fn ^int [hello world])
```

So while keeping the bracketed lambda list is not typical for lisp docs it is the equivalent syntax for sphinx python docs and doesn't involve introducing new syntax that's only relevant when writing documentation.

Do you think introducing a `(returns ^int)` annotation syntax is worth the departure the established python syntax?

Kodiologist:
I've only been paying attention to display in the Sphinx-generated HTML and I haven't looked at input. Sorry for any confusion. I don't think the input syntax matters too much because we only really need to write it ourselves for special forms; it will be generated automatically from introspection in the case of functions and macros.

Display like `(some-fn hello world) -> ^int` or `(some-fn hello world) → ^int` sounds fine to me, yes.

allison-casey:
Glad we got that sorted out :). Anyway,  I pushed a new hydomain that munges the output properly. Just redownload it and give it a go.

Kodiologist:
I did `pip install --upgrade git+https://github.com/allison-casey/sphinxcontrib-hydomain.git; make docs` and I'm still seeing square brackets in the output.

allison-casey:
i think you have to do `pip install --upgrade --force-reinstall git+https://github.com/allison-casey/sphinxcontrib-hydomain.git` to get the new version since I haven't done an actual versioned release yet

Kodiologist:
Ah right, I remember having to do that for `astor` shenanigans a while back. Thanks. It looks good.

allison-casey:
cool! I'll rebase off the current master then. Should I squash the commits for this one?

Kodiologist:
Yeah, I'll take a look once you've reorganized the commits.

allison-casey:
Should I add anything to news since this doesn't add anything to the code base?

Kodiologist:
I guess there's no need to, yeah. NEWS is for user-visible changes to code.

Kodiologist:
You're still working on reorganizing the new mega-commit into separate commits, right?

allison-casey:
Oh lol I thought we wanted a mega commit. How do you want this organized? One commit for docstrings, one for rst changes, and one for the defmacro! changes sound good?

Kodiologist:
I would say the most important point is the one I raised in #1044:

> Just make sure the massive copy-paste from the docs files into docstrings are in a separate commit from the more interesting code or documentation changes. My philosophy is that changes that create a lot diff lines, but don't involve any real programming that a reader would want to review, should be separated, by commits, from the real programming.

Similarly, when you do a big find-and-replace, that can be in a separate commit as well.

It may be easier to start with your original set of 50 commits or so (which is probably still in your local reflog) and squash those systematically than to break up the new mega-commit.

allison-casey:
How does this look @Kodiologist ?

Kodiologist:
Thanks, that looks a lot better. I'll get back to you once I've reviewed it.

Kodiologist:
Looks nice.

I think you need to migrate the new destructuring documentation.

I think all the quotation-mark changes in `conf.py` are unnecessary and distracting. If you're sure you want them in this PR, put them in their own commit. If I were you, I'd just undo them.

`cheatsheet_spec` is huge; I'd put it in its own file, either as JSON or Python (either should work fine).

I'd write `role = ":ref:" if spec.get("internal", False) else role` as `if spec.get("internal"): role = ":ref:"`.

In `format_refs` and `format_row`, all the temporary variables are used only once and can be inlined. `format_refs` itself is called only once and can also be inlined.

In `format_table`, you can inline temporary variables and also use a single big expression instead of defining `table = []` and building it up imperatively over 7 statements. I'd use a temporary variable for the repeated chunk `"=" * longest_cat_name + " " + "=" * 25`. But actually, I don't think you need to compute `longest_cat_name` at all, because the output ReST markup is only for machines. Just use a big number, like 1,000.

allison-casey:
> I think you need to migrate the new destructuring documentation.
Done

> quotation-mark changes in conf.py are unnecessary and distracting
Woops, my bad. Looks like one of my pre save hooks went off. I undid them 

> cheatsheet_spec is huge; I'd put it in its own file, either as JSON or Python (either should work fine).
Done

Inlined the single use variables and simplified some of the syntax

Kodiologist:
So reST is picky enough that the divider actually has to be the width of the widest headline? It can't just be really long? Shame.

Kodiologist:
`docs.py` doesn't run right now, by the way. (Travis doesn't build the documentation.)

allison-casey:
where is `docs.py`?

Kodiologist:
Shoot, I meant `conf.py`, in the `docs` directory.

allison-casey:
woops! don't know how i missed those bugs. Fixed

Kodiologist:
Great. Fold those changes into the earlier ones, to avoid the diff noise of e.g. the quote changes, and I think we're good. Sorry this has taken so long.

allison-casey:
Alright, I think these commits are cleaner and make more sense. Let me know if there's anything else!

Kodiologist (approved):

allison-casey:
@Kodiologist It looks like this pr incidentally closes #1423 and #1424 can you confirm this?

Kodiologist:
Yep, looks like it to me. You can add tests and NEWS items for those bugs in this PR or a new PR. Your choice.

allison-casey:
Added tests, a news item, and added the closes tag to the two issues in the first post

Kodiologist:
Don't forget the issues URLs as comments.

allison-casey:
rebasing is hard :upside_down_face: 

I think this does it though

allison-casey:
ooo baby the quintuple

PR #1953, opened 2021-01-26
Make `doc` and `#doc` work with mangling

### LINKED ISSUE #1946: ``doc`` doesn't work with names that require mangling
The current implementation of ``doc`` doesn't work with macro names that require mangling, like ``->``.

It seems that this would be easy to change by just adding a call to ``mangle`` in the definition of ``doc``:
```hy
(defmacro doc [symbol]
  `(help (.get __macros__ (mangle '~symbol) None)))
```
However, would it not make sense to combine ``help``, ``doc``, and ``#doc`` and have a single macro for looking up documentation? Especially as by default python's ``help`` isn't that helpful, showing the mangled names and opening up in a pager.

peaceamongworlds:
Closes #1946.

Adds call to `mangle` in the doc macros so that they work with names such as `->`.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -20,6 +20,7 @@ Bug Fixes
 * Fixed expressions within f-strings being inaccessible to macros.
 * Fixed symbol `J` being incorrectly parsed as a complex number
 * Fixed error handling for non-symbol macro names
+* `doc` and `#doc` now work with names that require mangling.
 
 0.20.0 (released 2021-01-25)
 ==============================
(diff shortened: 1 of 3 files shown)


Kodiologist:
Add a test to `test-doc` in `tests/native_tests/core.hy`. Also, put the NEWS item in a new "Unreleased" section.

peaceamongworlds:
> Add a test to `test-doc` in `tests/native_tests/core.hy`. Also, put the NEWS item in a new "Unreleased" section.

Done

Kodiologist (approved):
Don't forget a comment in the test with the URL https://github.com/hylang/hy/issues/1946.

allison-casey:
Will approve this as well once the url to the issue has been added to the test case.

allison-casey (approved):
added the issue comment to tests

PR #1955, opened 2021-01-28
Custom history location

### LINKED ISSUE #1833: Make hy-history location configurable
How about an environment variable like `HY_HISTORY` that allows the user to change the location of `~/.hy-history`.

datschy:
Closes #1833

Adds setting of the location of the history file to the value of the `HY_HISTORY` environment variable in the `completion` context manager, if provided.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -105,3 +105,4 @@
 * Joshua Munn <public@elysee-munn.family>
 * Peter Andreev <appa@gmx.co.uk>
 * Sunjay Cauligi <scauligi@eng.ucsd.edu>
+* David Tscheppen <david.tscheppen@gmail.com>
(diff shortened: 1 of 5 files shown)


allison-casey:
A documentation page for the repl was recently added in #1947. Can you add a short note about this new env var there

Kodiologist:
I like the feature. But I don't think a test should touch a file outside the source tree. Maybe just drop `test_history_default_location`.

datschy:
> But I don't think a test should touch a file outside the source tree. Maybe just drop `test_history_default_location`.

Yeah, I was a little torn apart because of it myself. Removed the test.

Kodiologist:
Inline `is_readline_available` and `skip_no_readline`, then squash everything but the AUTHORS addition, and we should be good to go.

Kodiologist (approved):

allison-casey (approved):

PR #1956, opened 2021-02-04
Add Python 3.9 to supported versions in setup.py

tianon:
I'm guessing this was just an oversight? :sweat_smile:

I thought about adding 3.10 here as well, but we've seen bad mojo in the past from claiming to support the alphas too soon and then seeing breaking changes before release. :grimacing:

(Happy to adjust and include 3.10 though if that's what maintainers prefer! :heart:)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
   - "3.6"
   - "3.7"
   - "3.8"
-  - "3.9-dev"
+  - "3.9"
   - pypy3
 install:
   - pip install -r requirements-travis.txt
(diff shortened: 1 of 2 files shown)


tianon:
I guess `.travis.yml` should probably be updated too -- happy to roll that into this change if you'd like!
(or a new PR if you'd prefer)

Kodiologist:
Yes, it looks like I forgot to update the classifiers in #1896.

The way Travis names its versions is confusing to me, but by all means, you can replace "3.9-dev" with "3.9" if it works.

As for Python 3.10, I haven't tested compatibility yet, but usually each new minor release of Python breaks at least one thing in Hy.

Kodiologist:
I'm merging this alone, as a trivial change.

PR #1960, opened 2021-02-08
Add handling for "=" syntax in f-strings

### LINKED ISSUE #1948: Add support for "=" syntax in  f-strings

jams2:
Closes #1948 

I didn't think it was necessary to add documentation as it's part of Python, let me know if you think otherwise.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -12,6 +12,7 @@ New Features
 ------------------------------
 * New contrib module `destructure` for Clojure-style destructuring.
 * Location of history file now configurable via environment variable `HY_HISTORY`
+* Added handling for "=" syntax in f-strings.
 
 Bug Fixes
 ------------------------------
--- a/docs/language/syntax.rst
+++ b/docs/language/syntax.rst
@@ -107,6 +107,15 @@ string, before any replacement fields are considered, so you may need to
 backslash your backslashes, and you can't comment out a closing brace or the
 string delimiter.
 
+Hy's f-strings are compatible with Python's "=" debugging syntax, subject to
+the above limitations on delimiting identifiers. For example::
+
+    => (setv foo "bar")
+    => (print f"{foo = }")
+    foo = 'bar'
+    => (print f"{foo = !s :_^7}")
+    foo = __bar__
+
 .. _syntax-keywords:
 
 keywords
--- a/hy/lex/parser.py
+++ b/hy/lex/parser.py
@@ -285,15 +285,24 @@ def _format_string(state, p, rest, allow_recursion=True):
        # Parse the first form.
        try:
            from . import parse_one_thing
-           model, item = parse_one_thing(item)
+           model, remainder = parse_one_thing(item)
+           f_expression = item[:-len(remainder)]
+           item = remainder
        except LexException:
            raise
        except ValueError as e:
            raise LexException.from_lexer("f-string: " + str(e), state, p[0])
        subnodes = [model]
 
+       # Check for '=' debugging syntax, reproduce whitespace in output
+       eq_sign_match = re.match(r'\s*=\s*', item)
+       if eq_sign_match:
+           values.append(HyString(f_expression + eq_sign_match.group()))
+           item = item[eq_sign_match.end():]
+       else:
+           item = item.lstrip()
+
        # Look for a conversion character.
-       item = item.lstrip()
        conversion = None
        if item.startswith('!'):
            conversion = item[1]
@@ -312,6 +321,9 @@ def _format_string(state, p, rest, allow_recursion=True):
            raise LexException.from_lexer(
                "f-string: trailing junk in field",
                state, p[0])
+       elif eq_sign_match and not conversion:
+           # Python has a special default conversion in this case.
+           conversion = "r"
 
        values.append(HyFComponent(subnodes, conversion=conversion))
 
--- a/tests/native_tests/language.hy
+++ b/tests/native_tests/language.hy
@@ -1320,7 +1320,34 @@ cee\"} dee" "ey bee\ncee dee"))
   (with [(pytest.raises NameError)]
     (eval quoted))
   (setv world "goodbye")
-  (assert (= (eval quoted) "hello goodbye")))
+  (assert (= (eval quoted) "hello goodbye"))
+
+  ;; '=' debugging syntax.
+  (setv foo "bar")
+  (assert (= f"{foo =}" "foo ='bar'"))
+
+  ;; Whitespace is preserved.
+  (assert (= f"xyz{  foo = }" "xyz  foo = 'bar'"))
+
+  ;; Explicit conversion is applied.
+  (assert (= f"{ foo = !s}" " foo = bar"))
+
+  ;; Format spec supercedes implicit conversion.
+  (setv  pi 3.141593  fill "_")
+  (assert (= f"{pi = :{fill}^8.2f}" "pi = __3.14__"))
+
+  ;; Format spec doesn't clobber the explicit conversion.
+  (with [(pytest.raises
+           ValueError
+           :match r"Unknown format code '?f'? for object of type 'str'")]
+    f"{pi =!s:.3f}")
+
+  ;; Nested "=" is parsed, but fails at runtime, like Python.
+  (setv width 7)
+  (with [(pytest.raises
+           ValueError
+           :match r"I|invalid format spec(?:ifier)?")]
+    f"{pi =:{fill =}^{width =}.2f}"))
 
 
 (defn test-import-syntax []


Kodiologist:
We do document f-strings, in `syntax.rst`, and that description needs to be updated a bit to account for `=`.

I think the text you put in the docstring would be more helpful as one or more appropriately located comments.

It isn't true that including both a conversion and a format specifier is necessarily a runtime error in Python, nor should it be so in Hy. For example, `f"{2+2!r:5}"` is legal Python, and it can be written in Hy as `f"{(+ 2 2) !r :5}"`.

I believe you can avoid the extra variable `implicit_conversion` by just initializing `conversion` earlier.

allison-casey:
it looks like this is failing for PyPy with
`Pattern "Unknown format code 'f' for object of type 'str'" does not match "Unknown format code f for object of type 'str'"`
It seems that the PyPy throws the same exception just without the single quotes around the `f`. I can't confirm this on my end since pypy is not liking f-strings for some reason on my box, but `pytest.raises` lets you use a regex so this should work just fine.
```clojure
;; Format spec doesn't clobber the explicit conversion.
(with [(pytest.raises
         ValueError
         :match r"Unknown format code '?f'? for object of type 'str'")]
  f"{pi =!s:.3f}")
```

jams2:
Thanks both. I'm not sure how to make this work in the same manner as Python
without the extra variable for `implicit_conversion`. Consider the
following examples:

```python
In [6]: print(py_to_ast('f"{foo=}"'))

Module(
  body=[
    Expr(
      value=JoinedStr(
        values=[
          Constant(value='foo='),
          FormattedValue(
            value=Name(id='foo', ctx=Load()),
            conversion=114)]))],
  type_ignores=[])

```

We used a "=", so the `ast.FormattedValue` gets a conversion value of
114 (repr).

```python
In [74]: print(py_to_ast('f"{foo=!s}"'))

Module(
  body=[
    Expr(
      value=JoinedStr(
        values=[
          Constant(value='foo='),
          FormattedValue(
            value=Name(id='foo', ctx=Load()),
            conversion=115)]))],
  type_ignores=[])
```

A specific conversion, so the AST reflects this (115).

```python
In [75]: print(py_to_ast('f"{foo=:.2}"'))

Module(
  body=[
    Expr(
      value=JoinedStr(
        values=[
          Constant(value='foo='),
          FormattedValue(
            value=Name(id='foo', ctx=Load()),
            conversion=-1,
            format_spec=JoinedStr(
              values=[
                Constant(value='.2')]))]))],
  type_ignores=[])
```

Here, we used "=", but also a format spec, so the FormattedValue's
conversion is set to -1.

```python
In [79]: print(py_to_ast('f"{foo=!r:.2f}"'))

Module(
  body=[
    Expr(
      value=JoinedStr(
        values=[
          Constant(value='foo='),
          FormattedValue(
            value=Name(id='foo', ctx=Load()),
            conversion=115,
            format_spec=JoinedStr(
              values=[
                Constant(value='.2f')]))]))],
  type_ignores=[])
```

In this example, we wrote a format spec, but we can't set the
conversion to -1 as that would clobber the explicitly specified
conversion. Coincidentally this would raise an error at runtime, but
as you rightly pointed out not all combinations of a conversion and a
format spec will cause this. To match Python's behaviour we have to be
able to parse any combination of "=", explicit conversion, and format
spec, bearing in mind that:

1. using "=" will set the conversion to 114 if no explicit conversion
   and no format spec; and
2. a format spec with no explicit conversion will set the conversion
   to -1.
   
The problem I'm having with using the one variable to track the
conversion is that we don't know if it was set implicitly (by the
presence of a "=") or explicitly. We could check if there was a "="
(`if eq_sign_match:...`)  but that doesn't help discern between the
automatic 114 and an explicit setting. Am I missing something?

Kodiologist:
Take a look at the attempt that I pushed. It seems to pass your tests.

allison-casey:
While I like the simplification of removing `implicit_conversion`, inlining 
```python
values.append(HyString(replacement_field_memo[:
               replacement_field_memo.find(eq_sign_match.group())
               + len(eq_sign_match.group())]))
```
Feels like an oversimplifications that makes what's happening less immediately apparent. 

```python
 debug_str_end = (
     replacement_field_memo.find(eq_sign_match.group())
     + len(eq_sign_match.group())
 )
 values.append(HyString(replacement_field_memo[:debug_str_end]))
```
I find this to be much easier to understand at a glance. Is there a reason for this kind of inlining other than conciseness?

Kodiologist:
I find it easier to digest code that's more functional and less imperative, and doesn't oblige the reader to keep track of more names than necessary. But it's a matter of taste.

jams2:
Ok cool, it took me a minute to get my head around this but it looks like you've covered all the cases. I like seeing a different angle on a problem. I agree with @allison-casey that readability suffers a little without the interim variable (I generally prefer to err on the side of being over-descriptive), but I suppose it would be more in keeping with the existing style to inline it.

allison-casey:
> I find it easier to digest code that's more functional and less imperative, and doesn't oblige the reader to keep track of more names than necessary. But it's a matter of taste.

I generally agree as well. For something like this though with two logically distinct operations i would err on the side of the readability of the intermediate variable. Up to you though, as jams2 said, it would be more in keeping with the existing style.

Kodiologist (approved):

scauligi:
Might I suggest this patch? Using `.find()` is unnecessary, and this also makes the intent (recovering the parsed form itself) a little more clear IMO.

```diff
--- a/hy/lex/parser.py
+++ b/hy/lex/parser.py
@@ -280,13 +280,14 @@ def _format_string(state, p, rest, allow_recursion=True):
        if not match:
           raise LexException.from_lexer('f-string: mismatched braces', state, p[0])
        item = rest[: match.end() - 1]
-       replacement_field_memo = item  # In case we have '=' syntax
        rest = rest[match.end() :]
 
        # Parse the first form.
        try:
            from . import parse_one_thing
-           model, item = parse_one_thing(item)
+           model, remainder = parse_one_thing(item)
+           parsed_form = item[:-len(remainder)]
+           item = remainder
        except LexException:
            raise
        except ValueError as e:
@@ -296,12 +297,9 @@ def _format_string(state, p, rest, allow_recursion=True):
        # Check for '=' debugging syntax, reproduce whitespace in output
        eq_sign_match = re.match(r'\s*=\s*', item)
        if eq_sign_match:
-           values.append(HyString(replacement_field_memo[:
-               replacement_field_memo.find(eq_sign_match.group())
-               + len(eq_sign_match.group())]))
+           values.append(HyString(parsed_form + eq_sign_match.group()))
            item = item[eq_sign_match.end():]
        else:
            item = item.lstrip()
```

jams2:
What do you think about naming `parsed_form` as `f_expression`, so it matches the grammar?

allison-casey:
> Might I suggest this patch?

I like it. Went ahead and changed it. Are we all happy with this? If so, i'll go ahead and approve and merge

allison-casey (approved):

allison-casey:
Thanks @jams2! :tada: :tada:

PR #1964, opened 2021-02-11
Policy: change the interval for uncontroversial merges to 3 days

Kodiologist:
#1508 has been a big success. Whereas perfectly good PRs once sat unmerged for months because nobody other than the author could get around to reviewing them, now an uncontroversial change can get merged in at worst 2 weeks. Still, 2 weeks is a while, and I've yet to see such a long interval produce late objections or vetos to a change. In all the cases over the past few years in which a PR was objected to or changes were requested, the objection came within a day or two of the PR's creation. So after the first few days, all the waiting does is slow down an inevitable merge. Faster merges would not only allow bug fixes and new features to become available more quickly; it would also encourage new Hy developers to make more PRs because of the warm fuzzy feeling of prompt integration of one's changes.

@hylang/core Are you guys on board with this?

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -125,7 +125,7 @@ There are two situations in which a PR is allowed to be merged:
 1. When it is approved by **two** members of Hy's core team other than the PR's
    author. Changes to the documentation, or trivial changes to code, need only
    **one** approving member.
-2. When the PR is at least **two weeks** old and **no** member of the Hy core
+2. When the PR is at least **three days** old and **no** member of the Hy core
    team has expressed disapproval of the PR in its current state. (Exception: a
    PR to create a new release is not eligible to be merged under this criterion,
    only the first one.)


allison-casey:
Wholly support this change. Open source projects need momentum to stay alive and Hy isn't active enough to maintain that momentum with such a long wait time for uncontroversial merges. There's been a wonderful uptick in PR's from new faces of late and I'd hate to see that fade because it takes so long for merges to go through.

refi64:
So question...do we actually need two approvals right now?

It's pretty evidence that Kodiologist is the main maintainer right now, given the fact that there's three or so active (I'm really not including myself at this point), I'm not sure gating new requests behind 2 is really useful, unless it's a particularly large change...

Kodiologist:
You've got a point, Ryan. What policy would you propose?

allison-casey:
Maybe bugfixes, minor refactors, contrib module additions, etc require 1 approval and substantive change to compiler code require 2? Would incentive new faces to get their feet wet with bug fixes and experiment with new features while staying more conservative on mucking with the compiler. 

Or just one approval for everything, power to the people heck yeah :hammer_and_wrench: :tada: :people_holding_hands:

Kodiologist:
Even if we reduce the two-approvers rule to one approver, I'd really like to keep the second merge route, which requires no approvers other than the author, so that PRs can get merged when there's only one person active on Hy.

Kodiologist:
Okay, since there are no objections, this is happening.

PR #1966, opened 2021-02-11
Fix: Tag macro expansion compiler crash

### LINKED ISSUE #1965: Exceptions in tag macros crash the compiler
Just ran into this. An exception raised in a tag macro will cause an internal compiler bug and crash while the same in a regular macro wont. 

```hy
(deftag a []
  (raise (ValueError)))

(defmacro b []
  (raise (ValueError)))

(b) ; => ValueError
;; (a) ; => Process Hy exited abnormally with code 1 (i'm a dumb dumb and this isn't what i mean to write :/
#a() ; => Process Hy exited abnormally with code 1 
```

allison-casey:
Closes #1965.

tag macro expansion was missing the equivalent `with macro_exceptions` context manager meaning user errors in macros crashed the compiler. Also enforced the unary constraint of `deftag` that @jams2 and I noticed would also crash the compiler if not adhered to.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -21,6 +21,7 @@ Bug Fixes
 * Fixed symbol `J` being incorrectly parsed as a complex number
 * Fixed error handling for non-symbol macro names
 * `doc` and `#doc` now work with names that require mangling.
+* Fixed exceptions in tag macros crashing compiler
 
 0.20.0 (released 2021-01-25)
 ==============================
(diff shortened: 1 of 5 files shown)


Kodiologist:
It seems to work, but `tag_exceptions` is mostly a copy of `macro_exceptions`. Try combining them.

I would put the tests in `test_lex.py`.

allison-casey:
@Kodiologist your pr #1971 makes this irrelevant. I feel like I should just close this one.

Kodiologist:
Aright.

PR #1968, opened 2021-02-12
Contrib: Multi-Index slicing macros

### LINKED ISSUE #541: Support for Multidimensional Slicing (Numpy Issue)
Numpy has support for multidimensional arrays like so:

```
>>> x = numpy.array([[x for x in range(10)] for y in range(10)])
>>> x[1:5,::2]
array([[0, 2, 4, 6, 8],
       [0, 2, 4, 6, 8],
       [0, 2, 4, 6, 8],
       [0, 2, 4, 6, 8]])
```

It's a beautiful way of handling matrices but it brings up the issue that it takes in a `tuple` as an argument for slicing, making it difficult to use one of Numpy's greatest features in Hy.

I don't think this is necessarily a bug, since it's an issue with an external library's extension of Lisp syntax, but a macro to fix this issue would be amazing.

And here I once thought Python's slicing notation was beautiful and convenient.

allison-casey:
Adds macros for easy numpy/pandas style multi-index slices. Closes #541 

Seemed odd that multi index slicing is as cumbersome as it is in Hy given how prevalent it is in numpy and pandas, especially when an elegant solution was already put forth by guanyilun in that issue so this brings that in as a contrib module with some additional changes to make it more robust and cover some other cases like pandas string index slicing. I also changed the name from ``nget`` to ``ncut`` since that seemed closer to how ``cut`` works than ``get``.

### DIFF
--- a/hy/contrib/slicing.hy
+++ b/hy/contrib/slicing.hy
@@ -1,0 +1,197 @@
+;;; Hy Multi-Index Slicing module
+;; Copyright 2021 the authors.
+;; this file is part of Hy, which is free software licensed under the Expat
+;; license. See the LICENSE.
+"Macros for elegantly slicing and dicing multi-axis and multidimensional sequences.
+
+Libraries like ``numpy`` and ``pandas`` make extensive use of python's slicing syntax
+and even extends it to allow multiple slices using tuples. This makes handling
+multi-axis (pandas) and multidimensional arrays (numpy) extremely elegant and efficient.
+Hy doesn't support Python's sugared slicing syntax (``1::-2 # => slice(1, None, 2)``).
+Which makes slicing quite cumbersome, especially when tuple's get thrown into the mix.
+Hy's ``cut`` form makes single slices easy, but anything more than that becomes
+much more difficult to construct and parse. Where python can express a multidimensional
+slice as:
+
+.. code-block:: python
+
+   >>> arr[1:-2:2,3:5,:]
+
+The equivalent in Hy would be::
+
+   => (get arr (, (slice 1 -2 2) (slice 3 5) (slice None)))
+
+which is hardly ideal. This module provides an ``ncut`` macro and a ``#:`` tag macro
+that enable python's sugared slicing form in Hy so that the previous could be expressed as::
+
+   => (ncut a 1:-1:2 3:5 :)
+
+or more manually using the tag macro as::
+
+   => (get a (, #: 1:-2:2 #: 3:5 #: :))
+"
+(eval-and-compile
+  (defn parse-colon [sym]
+    (lfor index (.split (str sym) ":")
+          (if (empty? index) None
+              (int index))))
+
+  (defn parse-indexing [sym]
+    (if
+      (and (isinstance sym HyExpression) (= (get sym 0) :))
+      `(slice ~@(cut sym 1))
+
+      (and (symbol? sym) (= sym '...))
+      'Ellipsis
+
+      (and (isinstance sym (, HyKeyword HySymbol)) (in ":" (str sym)))
+      (try `(slice ~@(parse-colon sym)) (except [ValueError] sym))
+
+      sym)))
+
+(defmacro ncut [seq key1 &rest keys]
+  "N-Dimensional ``cut`` macro with shorthand slice notation.
+
+  Libraries like ``numpy`` and ``pandas`` extend Python's sequence
+  slicing syntax to work with tuples to allow for elegant handling of
+  multidimensional arrays (numpy) and multi-axis selections (pandas).
+  A key in ``ncut`` can be any valid kind of index; specific,
+  ranged, a numpy style mask. Any library can make use of tuple based
+  slicing, so check with each lib for what is and isn't valid.
+
+  Args:
+    seq: Slicable sequence
+    key1: A valid sequence index. What is valid can change from library to
+      library.
+    *keys: Additional indices. Specifying more than one index will expand
+      to a tuple allowing multi-dimensional indexing.
+
+  Examples:
+    Single dimensional list slicing
+    ::
+
+       => (ncut (list (range 10)) 2:8:2)
+       [2 4 6]
+
+    numpy multidimensional slicing:
+    ::
+
+       => (setv a (.reshape (np.arange 36) (, 6 6)))
+       => a
+       array([[ 0,  1,  2,  3,  4,  5],
+              [ 6,  7,  8,  9, 10, 11],
+              [12, 13, 14, 15, 16, 17],
+              [18, 19, 20, 21, 22, 23],
+              [24, 25, 26, 27, 28, 29],
+              [30, 31, 32, 33, 34, 35]])
+       => (ncut a (, 0 1 2 3 4) (, 1 2 3 4 5))
+       array([ 1,  8, 15, 22, 29])
+       => (ncut a 3: (, 0 2 5))
+       array([[18, 20, 23],
+              [24, 26, 29],
+              [30, 32, 35]])
+       => (ncut a 1:-1:2 3:5)
+       array([[ 9, 10],
+              [21, 22]])
+       => (ncut a ::2 3 None)
+       array([[ 3],
+              [15],
+              [27]])
+       => (ncut a ... 0)
+       array([ 0,  6, 12, 18, 24, 30])
+
+    Because variables can have colons in Hy (eg: ``abc:def`` is a valid identifier),
+    the sugared slicing form only allows numeric literals. In order to construct slices
+    that involve names and/or function calls, the form ``(: ...)`` can be used in an
+    ``ncut`` expresion as an escape hatch to ``slice``:
+    ::
+
+       => (setv abc:def -2)
+       => (macroexpand '(ncut a abc:def (: (sum [1 2 3]) None abc:def)))
+       (get a (, abc:def (slice (sum [1 2 3]) None abc:def)))
+
+    Pandas allows extensive slicing along single or multiple axes:
+    ::
+
+       => (setv s1 (pd.Series (np.random.randn 6) :index (list \"abcdef\")))
+       => s1
+       a    0.687645
+       b   -0.598732
+       c   -1.452075
+       d   -0.442050
+       e   -0.060392
+       f    0.440574
+       dtype: float64
+
+       => (ncut s1 (: \"c\" None 2))
+       c   -1.452075
+       e   -0.060392
+       dtype: float64
+
+    ::
+
+       => (setv df (pd.DataFrame (np.random.randn 8 4)
+                                 :index (pd.date-range \"1/1/2000\" :periods 8)
+                                 :columns (list \"ABCD\")))
+       => df
+                          A         B         C         D
+       2000-01-01 -0.185291 -0.803559 -1.483985 -0.136509
+       2000-01-02 -3.290852 -0.688464  2.715168  0.750664
+       2000-01-03  0.771222 -1.170541 -1.015144  0.491510
+       2000-01-04  0.243287  0.769975  0.473460  0.407027
+       2000-01-05 -0.857291  2.395931 -0.950846  0.299086
+       2000-01-06 -0.195595  0.981791 -0.673646  0.637218
+       2000-01-07 -1.022636 -0.854971  0.603573 -1.169342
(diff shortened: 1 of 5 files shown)


Kodiologist:
This colon syntax doesn't seem to support expressions, just string and numeric literals. That seems like a big limitation. However, I may be misunderstanding something, since I'm only looking at the documentation, and the documentation only provides examples, without precisely describing what `#|` and `ncut` do. Also, I'm surprised that `ncut` requires `key1`, since empty slices are useful sometimes, right?

Kodiologist:
As an example of an approach to this problem that allows arbitrary forms in each part of the slice, my [geta](https://github.com/Kodiologist/Kodhy/blob/b299b61/kodhy/macros.hy#L211) allows things like

    (geta d (f) (: (g) (h)) :)

, which in Python would be

    d[f(), g() : h(), ]

allison-casey:
> This colon syntax doesn't seem to support expressions, just string and numeric literals. That seems like a big limitation.

so `ncut` supports expressions as single index, but you're correct that it can't make a `slice` like `(slice (sum (range 5)) None)` right now. I'll have to see what that best way to do that would be. I don't think it's possible to make the tag macro support expressions like `#| (sum [1 4]):1:-2` though. but i think i can figure something out for `ncut` i could probably build off of your `aget`

> since empty slices are useful sometimes, right?

an empty slice is equivalent to a single `:` so `#| :` and `(ncut arr :)` would slice with the empty slice. 

I'll go through and add a better description to `ncut` and `#|` to better explain exactly what they're doing

peaceamongworlds:
I've been using Kodiologist's macros for slicing numpy arrays and pandas DataFrames and they seem pretty robust.

Also, can't symbols contain colons? Surely `ncut` would not be able to handle a variable whose name contains a colon?

allison-casey:
if we change `parse-indexing` to:
```hy
(defn parse-indexing [sym]
    (if
      (in ":" (str sym)) `(slice ~@(parse-colon sym))
      (in "..." (str sym)) 'Ellipsis
      (and (instance? HyExpression sym) (= (get sym 0) :)) `(slice ~@(cut sym 1))
      sym))
```

It lets us do things like:
```hy
(ncut arr (f) (: (g) (h)) 1::-2)
;; expands to => '(get arr (, (f) (slice (g) (h)) (slice 1 None -2)))
```
Which i actually quite like. It lets us have the best of both worlds, the sugared form for slices that just use literals `1::-2` and kodi's `(: ...)` for forms that need evaluation. What do ya'll think?

Kodiologist:
> What do ya'll think?

That seems to make sense. Try adding some tests.

Allowing names in the short form, like `1:abc:4`, is pretty weird, in my opinion, not least because it looks like a variable instead of a string literal.

allison-casey:
Although the numeric parsing feels hacky going through `(eval (read-str ...))`, but it actually seems to work pretty well! Something like this:

```hy
(do
    (defn f [] 1)
    (defn g [] 2)
    (setv abc:def 12
          arr [1 2 3 4 5])
    (macroexpand '(ncut arr abc:def (: (f) (g)) 5j:4.5:1/3)))
    ;; expands to => '(get arr (, abc:def (slice (f) (g)) (slice 5j 4.5 (fraction 1 3))))
```
I haven't gone through all the edge cases, but it definitely feels like a good compromise. I'll start adding tests and see what shakes loose

allison-casey:
Alright, I like this a bit better. There's a lot of edge cases for these forms it turns out but I think i got most of them. Let me know what y'all think.

peaceamongworlds:
This doesn't allow for mixing the python slicing syntax with variables:
```hy
(setv a 1
      arr [1 2 3 4 5])
(ncut arr 1:) ; => [2 3 4 5]
(ncut arr a:) ; => NameError: name 'hyx_aXcolonX' is not defined
```
And also `(ncut arr :a:)` (i.e. equivalent of `(ncut arr :1:)`) get macroexpanded to `(get arr :a:)`, which raises `TypeError: list indices must be integers or slices, not HyKeyword`, which seems weird.

I wouldn't want there to be restrictions on what the slicing syntax works with, in which case I think it would be best to just use s-expression syntax.

allison-casey:
how would differentiate between 
```hy
(setv a 1 b 2)
(ncut arr a:b)
```
and 
```hy
(setv a:b 1)
(ncut arr a:b)
```
then? I thought the idea was that if you're going to be using names or function calls you use:
```hy
(ncut arr (: abc def))
```
instead?

Kodiologist:
Python is dynamic enough that it would be theoretically possible to, but I think we can all agree that that would be more weird than helpful.

Kodiologist:
It looks like you need to allow string literals in some cases because the dot for a float would be parsed as an attribute access. But If you're using numeric literals as one of the components of a slice, shouldn't they always be integers? I think you can just restrict the sugared form to integers.

Kodiologist:
More specifically, you can restrict it to ASCII digits. I don't think one often wants to slice something with a hexadecimal literal or whatever.

allison-casey:
they're converted to strings regardless when we split on the colon so it's more an acknowledgement that, that edge case exists than an active design choice. 

Numpy technically has behaviours for complex and floats in slices (see [here](https://github.com/numpy/numpy/blob/v1.14.1/numpy/lib/index_tricks.py#L174)) and it's not particularly difficult to support that since we already have to parse the slices from strings to ints anyways, so I don't think we should restrict it.

Kodiologist:
> it's more an acknowledgement that, that edge case exists than an active design choice.

I don't think I agree. You have some code specially to handle fractions, complex numbers, and floats (e.g., `(in "/" x) (fraction x)`), and you probably need more (e.g., to handle literal floats that use scientific notation with no decimal point, like `1e3`). If you require the sugared form to use integers, you won't need that code, you won't need to allow string literals, and you won't lose anything because it doesn't seem like non-integer indices are ever useful for NumPy, although they're allowed.

To be clear, I'm not super-convinced that the sugar is useful enough to have in the first place—I rarely index things in NumPy or pandas with numeric literals, instead of variables—but it seems silly to make it more complicated in a way that nobody will use.

allison-casey:
Oh, I thought you were talking about the fact that they're parsed as strings at all. Any ways, I took out the non interger parsing of sugared forms so that's fixed. I do use numeric literal slicing quite often, mostly for dataframes and exploring data structures so I definitely think it's worth having.

hy/contrib/slicing.hy
```diff
@@ -0,0 +1,193 @@
...
+
+   => (ncut a 1:-1:2 3:5)
+
+or more manually using the tag macro as::
+
+   => (get a (, #| 1:-2:2 #| 3:5))
+"
+(eval-and-compile
+  (defn parse-colon [sym]
+    (lfor index (.split (str sym) ":")
```
scauligi, line 35: (Not sure how much we care about this but) since `HySymbol` already inherits from `str`, the call to `str` here is unnecessary

scauligi (commented):
If we're not supporting non-numeric indices, we should add appropriate error messages. Otherwise, it's a bit confusing when a user (like me when I started playing around with this) tries `(ncut arr x:y)` and gets back `NameError: name 'hyx_xXcolonXy' is not defined`.

Alternatively, since this is already a mini-DSL, I think it's totally fair to just not support symbols with colons in them, so `(ncut arr x:y)` is equivalent to `arr[slice(x, y)]`. You could probably run everything through `hy.lex.hy-parse` to skip any and all special handling of forms, which will even handle the empty string how we would want to.

hy/contrib/slicing.hy
allison-casey: the slice `:1:2` actually gets parsed as a `HyKeyword` which doesn't have `.split` so we can actually get a `HySymbol` or a `HyKeyword` here

scauligi: Ah right, because of the initial colon. Good point, my bad!

Kodiologist: Also, the fact that `HySymbol` inherits from `str`, instead of containing a `str` as an attribute like `HyKeyword`, is an implementation detail that I wouldn't rely on.

Kodiologist:
> If we're not supporting non-numeric indices, we should add appropriate error messages. Otherwise, it's a bit confusing when a user (like me when I started playing around with this) tries `(ncut arr x:y)` and gets back `NameError: name 'hyx_xXcolonXy' is not defined`.

The macro can't check for the existence of `hyx_xXcolonXy`, because that's only determined at runtime.

> so `(ncut arr x:y)` is equivalent to `arr[slice(x, y)]`

Like supporting non-integer numeric literals in this construct, I think this will get messy.

scauligi:
This patch seems like it would suffice:
```diff
--- a/hy/contrib/slicing.hy
+++ b/hy/contrib/slicing.hy
@@ -33,8 +33,7 @@ or more manually using the tag macro as::
 (eval-and-compile
   (defn parse-colon [sym]
     (lfor index (.split (str sym) ":")
-          (if (empty? index) None
-              (int index))))
+      (hy.lex.hy-parse index)))
 
   (defn parse-indexing [sym]
     (if
```

Otherwise we should check explicitly for `index.isdecimal` (or that it's only ascii digits if we want to be more strict) and raise an appropriate exception if not.

Kodiologist:
Then we'd have a weird situation where `3:4e1` is allowed but `3:40.0` isn't because of attribute parsing. I'd recommend `.isdigit`.

scauligi:
Ah yes, dots aren't allowed in HyKeywords. With that being the case we should raise an explicit error if `.isdigit` fails.

Kodiologist:
@allison-casey Sorry, it looks like I've been neglecting this.

You need to rebase after the changes to tag macros.

I find the spacing in the definition of `parse-indexing` inconsistent and hard to read. Consider something like

```
(if
  (and (instance? HyExpression sym) (= (get sym 0) :))
    `(slice ~@(cut sym 1))
  (= sym '...)
    'Ellipsis
  (and (instance? (, HyKeyword HySymbol) sym) (in ":" (str sym)))
    (try `(slice ~@(parse-colon sym)) (except [ValueError] sym))
  True
    sym)
```

Remember that currently, `(= "..." '...)`, so you need to check that `sym` really is a symbol.

Also, I would use `isinstance` instead of `instance?`, because `instance?`'s days are numbered.

In `ncut`, why write the conditional backwards, with `if-not`, instead of forwards with `if`? (I honestly don't know why we have `if-not` when you could write `(if (not …) …)`, but that's our core for you. We already have `unless`, which doesn't have an else-else clause like `if-not`, which is confusing anyway.)

It looks like a bare `:`, as in `(ncut df :)`, is tested, but not documented.

allison-casey:
As have I. I should have time today to bring this up to spec. You're proposed changes make sense and I'll add them in the update

allison-casey:
> You need to rebase after the changes to tag macros.

Done

> I find the spacing in the definition of parse-indexing inconsistent and hard to read. Consider something like

Agreed, however, emacs `indent-region` will force the indented lines back in alignment so how about this. 
```clojure
(if
      (and (isinstance sym HyExpression) (= (get sym 0) :))
      `(slice ~@(cut sym 1))

      (and (symbol? sym) (= sym '...))
      'Ellipsis

      (and (isinstance sym (, HyKeyword HySymbol)) (in ":" (str sym)))
      (try `(slice ~@(parse-colon sym)) (except [ValueError] sym))

      sym)
```
Feels much better to me and won't get reformated unexpectedly by emacs. 

> Remember that currently, (= "..." '...), so you need to check that sym really is a symbol.

Done

> Also, I would use isinstance instead of instance?, because instance?'s days are numbered.

Done

>  why write the conditional backwards, with if-not

I find `if-not` can be useful when the primary focus of the conditional is the `not` branch. That wasn't the case here and I've fixed it.

> It looks like a bare :, as in (ncut df :), is tested, but not documented.

Done

Kodiologist (approved):

allison-casey:
Forgot to mention that I changed the tag macro to `#:` and made it so you don't need to specify the `:` when writing it via an expression so `#|(: 1 2)` becomes `#:(1 2)` which seems much cleaner.

Kodiologist:
That makes sense.

Kodiologist:
@scauligi didn't actually approve this, but it looks like you addressed his objections, so this should be clear to merge by the three-day rule.

PR #1969, opened 2021-02-12
Fix compiler crash on malformed `.` special form

scauligi:
Hy compiler would crash on a root with multiple dots in a row:
```
=> (.. "crashme")
  File "/home/scauligi/workspace/hy/venv/bin/hy", line 12, in <module>
    sys.exit(hy_main())
  File "/usr/lib/python3.9/code.py", line 232, in interact
    more = self.push(line)
  File "/usr/lib/python3.9/code.py", line 258, in push
    more = self.runsource(source, self.filename)
```

This fixes that.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -22,6 +22,7 @@ Bug Fixes
 * Fixed symbol `J` being incorrectly parsed as a complex number
 * Fixed error handling for non-symbol macro names
 * `doc` and `#doc` now work with names that require mangling.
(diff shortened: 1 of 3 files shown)


allison-casey:
Awesome! I will review this tomorrow

Kodiologist (approved):
~~Don't forget a comment with the URL of the issue.~~ There is no issue. Never mind.

allison-casey (approved):
Thanks scauligi!

Issue #1970, opened 2021-02-12
Strange behavior or `#docs` / `__macros__` and `__tags__`


scauligi:
Minimum viable test case, must be run from its own file (not from the REPL):
```hy
;; (deftag bogus [-] -)  ; Uncomment to get help for `None`
;; (macroexpand '#@ (foo (defn bar [])))  ; Uncomment to get help for `#@`
#doc @
```
Ok so this one is a little weird but I think I know why it's happening.
**TL;DR** Hy patches in `__macros__` and `__tags__` to a module only on an as-needed basis and only during compilation. As a result, you can end up with Hy modules that don't have these defined.

Observed symptoms:
- If you run the above code, the `#doc @` fails with `'__tags__' is not defined`. 
- If you uncomment the `(deftag)`, the module now defines a `__tags__`, but it's only populated with the tag `bogus` &mdash; thus the `#doc @` brings up the help page for `None`.
- If you uncomment the `(macroexpand)`, then Hy finally brings the `@` tag macro into `__tags__`, and you get the proper help page.

**Further weirdness!**
Create two files like so:
```hy
;; file requireme.hy
(defn bogus [-] -)
#@(bogus (defn foo []))

;; file runme.hy
(require requireme)
#doc @
```

When you run `runme.hy`, you get the help page for `None`.
Without changing anything, run it again &mdash; now it fails with `'__tags__' is not defined`!

On the first run the compiler patches in `__tags__` and `__macros__` for the compilation process, but every subsequent run skips compilation and loads the pyc files from the pycache &mdash; and no code was ever produced to actually create a `__tags__` in `runme.hy`.

So far I only foresee this as a weird edge case of `#doc` and don't expect this leading to any real bugs, but it does produce confusing behavior for anyone doing introspection on Hy modules.

Possible solutions:
- Always add `__macros__` and `__tags__` to every Hy module at runtime, possibly via the `import hy` statement, and pre-populate with the core macros/tags.
- Document that `#doc` has strange behavior if you use it outside the REPL (and also fix the `test-doc` test, that's how I stumbled across this).
- Something else?

allison-casey:
I've had tests fail intermittently with `'__tags__' is not defined!` as well and only fixable by `rm -r`-ing all the `__pycache__` folders. So this may actually be less of an edge case than your think. I'm not sure what the best solution is. I think this one's gonna take some digging

allison-casey:
[...]

PR #1971, opened 2021-02-13
Simplify tag macros

### LINKED ISSUE #1965: Exceptions in tag macros crash the compiler

### LINKED ISSUE #1967: Require fails to load tags with same name as macro in same module
```hy
;; foo.hy
(deftag amacro [-name] `(print f"Hello {~-name}"))
(defmacro amacro [-name] `(print f"Hello {~-name}"))
```

```hy
;; bar.hy
(require [foo [amacro]])
;; (require [foo [*]]) ; This doesn't work either

(amacro "Allie") ; => (print "Hello Allie")
#amacro "Allie" ;  'amacro' is not a defined tag amcro
```

This seems unexpected since we have `#doc` and `(doc ...)` macros that work just fine. 
This seems like a pretty easy fix, changing `require` to import both just requires changing it to:
```python
if _name not in source_module.__macros__ and _name not in source_module.__tags__:
    raise HyRequireError('Could not require name {} from {}'.format(
        _name, source_module))

if _name in source_module.__macros__:
    target_macros[alias] = source_macros[_name]
if _name in source_module.__tags__:
    target_tags[alias] = source_tags[_name]
```
But is this expected behavior? It's not documented as such as far as I can tell.

Kodiologist:
- Closes #1965.
- Closes #1967.

I've reconstrued tag macros as macros whose names begin with "#", which you can call with special syntax. This simplifies a lot of internals and should remove various inconsistencies in the behavior of tag macros. To support this, `defmacro` and `require` allow macros to be named with string literals instead of symbols. `deftag` and `#doc` are unnecessary and have been removed.

I haven't updated the docs or NEWS yet. Tell me what you think.


allison-casey:
oh heck the mad man went ahead and did it. 

I'll take a look soon hopefully. his looks pretty promising though

Kodiologist:
Right, the `#foo` in `(require [#foo a])` parses as a tag-macro call, not a symbol.

Kodiologist:
By the way, I first tried having a tag macro named `#foo` be called `foo` internally, as I'd originally proposed, but I realized that we want to use tag macros like `#+` and `#%` without shadowing `+` and `%`.

allison-casey:
tangentially affects #1970 need to test this

> Right, the #foo in (require [#foo a]) parses as a tag-macro call, not a symbol.

makes sense and since it's only relevant in `defmacro` and `require` it doesn't seem so bad. 

> we want to use tag macros like #+ and #% without shadowing + and %.

agreed, i think I like this syntax better any ways

scauligi:
Would it be worthwhile to add a `deftag` macro for backwards compatilibity / ease of use?
```hy
;; backwards compat
(defmacro deftag [name &rest forms]
  `(defmacro ~(+ "#" (str name)) ~@forms))
```

I'm also looking into if there's a way to hack up the parser so that we wouldn't need to string-quote tags for `require`.

Kodiologist:
My philosophy is that backwards-compatibility features before the release of 1.0 are premature. I'm not sure that `(deftag foo …)` is better syntax than `(defmacro "#foo" …)` anyway, because the `#` isn't included.

> I'm also looking into if there's a way to hack up the parser so that we wouldn't need to string-quote tags for `require`.

I wouldn't recommend it. It will complicate the lexer considerably for an edge case.

hy/compiler.py
```diff
@@ -1134,12 +1133,16 @@ def compile_unary_operator(self, expr, root, arg):
         return operand + asty.UnaryOp(
             expr, op=ops[root](), operand=operand.force_expr)
 
-    _symn = some(lambda x: isinstance(x, HySymbol) and "." not in x)
+    def _importlike(*name_types):
+        name = some(lambda x: isinstance(x, name_types) and "." not in x)
```
scauligi, line 1137:
```diff
-        name = some(lambda x: isinstance(x, name_types) and "." not in x)
+        name = some(lambda x: isinstance(x, name_types) and (x.startswith("#") or "." not in x))
```
otherwise you can't import e.g. `"#."`

scauligi:
Alternatively, we could single-quote a tag name a la `'#%`, so you can do e.g.:
```hy
(require [hy.extra.anaphoric ['#%]])
(defmacro '#x [x] x)
```

Here's a patch that would allow this for require, you'd have to do something similar for defmacro (and/or factor stuff out into model_patterns):
```diff
diff --git a/hy/compiler.py b/hy/compiler.py
index c7224c44..01ce10d5 100755
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1134,12 +1134,19 @@ class HyASTCompiler(object):
             expr, op=ops[root](), operand=operand.force_expr)
 
     def _importlike(*name_types):
-        name = some(lambda x: isinstance(x, name_types) and "." not in x)
+        _symn = some(lambda x: isinstance(x, HySymbol) and "." not in x)
+        _symt = some(lambda x: isinstance(x, HySymbol) and (x[0] == "#" or "." not in x))
+        _tname = _symn | pexpr(sym("quote") + _symt)
         return [many(
             SYM |
-            brackets(SYM, sym(":as"), name) |
+            brackets(SYM, sym(":as"), _symn) |
             brackets(SYM, brackets(many(
-                name + maybe(sym(":as") + name)))))]
+                _tname + maybe(sym(":as") + _tname)))))]
+
+    @staticmethod
+    def _unquote_required(kids):
+        for pair in kids:
+            yield map(lambda x: x[0] if isinstance(x, HyExpression) else x, pair)
 
     @special("import", _importlike(HySymbol))
     @special("require", _importlike(HySymbol, HyString))
@@ -1167,7 +1174,7 @@ class HyASTCompiler(object):
                         raise self._syntax_error(star,
                             "* in an import name list must be on its own")
                 else:
-                    assignments = [(k, v or k) for k, v in kids]
+                    assignments = [(k, v or k) for k, v in self._unquote_required(kids)]
 
             ast_module = ast_str(module, piecewise=True)
 
diff --git a/hy/lex/parser.py b/hy/lex/parser.py
index 6121125a..a2e74e47 100755
--- a/hy/lex/parser.py
+++ b/hy/lex/parser.py
@@ -17,7 +17,7 @@ from .lexer import lexer
 from .exceptions import LexException, PrematureEndOfInput
 
 
-pg = ParserGenerator([rule.name for rule in lexer.rules] + ['$end'])
+pg = ParserGenerator([rule.name for rule in lexer.rules] + ['$end'], [('nonassoc', ['HASHOTHER'])])
 
 
 def set_boundaries(fun):
@@ -112,6 +112,12 @@ def term_discard(state, p):
     return p[2]
 
 
+@pg.production("term : QUOTE HASHOTHER")
+@set_boundaries
+def term_quote_hash(state, p):
+    return HyExpression([HySymbol("quote"), HySymbol(p[1].value)])
+
+
 @pg.production("term : QUOTE term")
 @set_quote_boundaries
 def term_quote(state, p):

```

Note that using a straight quote will quote the tag name, but a quasiquote will quote a tag _term_. I think this can be considered a feature and not a bug.

hy/compiler.py
Kodiologist: But if a macro's name has a period in it, how do you distinguish that kind of period from the scoping kind of period?

Kodiologist:
But then you wouldn't be able to use non-identifier characters in tag macros.

hy/compiler.py
scauligi: Ah, I didn't think tag macros could be scoped.
Would we want to have tag macros like `#module.tagname`?

Kodiologist: Yes, requiring a tag macro in a namespace-qualified fashion should work just as with ordinary macros, with or without this PR.

scauligi:
I'm not sure I follow. How would one even use a tag macro with non-identifier characters?
The lexer already defines a tag name only as a "#" followed by identifier characters.

Kodiologist:
So it does. Sorry. I was probably remembering the old days when tag macros had to be only 1 character long, so you could name them nearly anything, like `(`.

I'm still not entirely enthusiastic about using a quote form in place of a string literal, but I guess that's a matter of taste.

hy/compiler.py
allison-casey: If you can't import `"#."` shouldn't it be illegal to define a macro with that name too? Currently you can `(defmacro "#." [a])` but it's only usable in that module which doesn't seem right. Also should apply to a regular `(defmacro . [a])` since `.` is a special form and shadowing it with a macro kind of breaks everything + isn't even possible when using `defn` anyways.

allison-casey:
I'm still getting the `__pycache__` error that scauligi brought up in #1970 just in a slightly different way, so it doesn't look like this will fix it outright. But it is a major simplification to the code base which is a big + plus will probably help us figure that one out too. With regard to quoting, @scauligi does the test suit run all the way through with your patch? String literal feels more explicit, but quoting feels more syntactically consistent to lisp so i'm not quite sure where i stand on that yet.

scauligi:
I haven't run my patch through any of the tests; it's also not a complete patch, just a proof-of-concept since the `require` handler would also need to be updated and we would probably want to add a function to model_patterns.py.

I have a mild preference for the quoting since it feels more lisp-y to me (importing a specific symbol vs naming with a string) but I'm happy with either direction.

allison-casey:
If it's going to be that big of a change to the core compiler code it may be worth going with string literals for now and making quoting it's on pr. That way we can focus on the removal of the concept of `__tags__` for now any any unanticipated side effects that might ha e on it's own.

hy/compiler.py
Kodiologist: It seems there's a variety of interrelated, preexisting bugs involving `require` on names that get mangled and names that get required in a fully qualified fashion. I'm going to ban periods in `defmacro`, as you suggest, and leave those additional bugs for later PRs.

Kodiologist:
Okay, the doc changes are in now.

allison-casey:
I'm noticing that this also lets us define non identifier tag macros which leads to weird behavior. `(deftag "#(" [a])` is valid but can't be called, `(defamcro "#:" [a])` is actually valid and can be called now. Do we want to do some additional checking on names like this to ensure they're actually identifiers?

Kodiologist:
You actually can call them, if you use mangled names. This is similar to the issue of how Python allows `setattr(c, "$", 3)` even though `c.$` is a syntax error.

hy/core/bootstrap.hy
```diff
@@ -31,13 +31,29 @@
...
+
+       ::
+
+          => (defmacro "#x2" [form]
+          ...  `(do ~form ~form))
+
+       ::
+
+          => (setv foo 1) #x2 (+= foo 1) foo
+          3
```
scauligi, line 45: Split to multiple lines for clarity:
```hy
=> (setv foo 1)
=> #x2 (+= foo 1)
=> foo
3
```

tests/native_tests/core.hy
```diff
@@ -710,7 +710,7 @@ result['y in globals'] = 'y' in globals()")
             f"Help on function {(mangle '<-mangle->)} in module "))
   (assert (in "a fancy docstring" out))
   (assert (empty? err))
-  #doc @
+  (doc "#@")
```
scauligi, line 713: We should just remove this line, as it causes intermittent test failures. Builtin macros are not always loaded at runtime, and this is likely going remain intentional.

tests/native_tests/native_macros.hy
```diff
@@ -403,6 +401,7 @@ in expansions."
   (assert (= "This is the local version of `nonlocal-test-macro` returning 3!"
              #test-module-tag-2 3)))
 
+#@(pytest.mark.xfail
```
scauligi, line 404: Why did this test start failing?

scauligi (commented):
There's a remaining reference to `deftag` in `docs/style-guide.rst`.

Are the changes in `tests/native_tests/language.hy` related to tag macros or just refactoring while you were in the file?

tests/native_tests/native_macros.hy
Kodiologist: See the commit message.

hy/core/bootstrap.hy
Kodiologist: Done.

tests/native_tests/core.hy
Kodiologist: I put another `defmacro` in the test instead.

Kodiologist:
> There's a remaining reference to `deftag` in `docs/style-guide.rst`.

Removed.

> Are the changes in `tests/native_tests/language.hy` related to tag macros or just refactoring while you were in the file?

The ones in their own commit, at the beginning of this PR, are just refactoring while I was in the file.

tests/native_tests/native_macros.hy
scauligi: Hmm, I think I disagree with the (xfailed) assertion on line 447, but I'll make a separate issue for that.

scauligi (approved):

allison-casey (approved):

PR #1980, opened 2021-02-25
Add macros `do-n` and `list-n`

Kodiologist:
These are syntactic sugar for `for` loops when you don't need the loop variable for anything. I thought of naming one of them `dotimes`, but I decided it would be needlessly confusing because of the differences from Common Lisp's `dotimes`.

### DIFF
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -507,6 +507,45 @@
   `(if-not ~test (do ~@body)))
 
 
+(defn _do-n [count-form body]
+  `(for [~(gensym) (range ~count-form)]
+    ~@body))
+
+
+(defmacro do-n [count-form &rest body]
+  "Execute `body` a number of times equal to `count-form` and return
+  ``None``. (To collect return values, use :hy:macro:`list-n`
+  instead.) Negative values of the count are treated as 0.
+
+  This macro is implemented as a :hy:macro:`for` loop, so you can use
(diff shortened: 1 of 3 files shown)


scauligi:
Lgtm. Is there any policy / received wisdom about what goes into the core and what doesn't?

Kodiologist:
My thinking is that core macros should be small, of broadly general use, and free of new dependencies.

hy/core/macros.hy
```diff
@@ -507,6 +507,42 @@
...
+
+
+(defmacro list-n [count-form &rest body]
+  "Like :hy:macro:`do-n`, but the results are collected into a list.
+
+  ::
+
+    => (setv n 0)
+    => (list-n 5 (+= n 1) n)
+    [1 2 3 4 5]
```
peaceamongworlds, line 536: Maybe change the variable `n` to something else in this example?

It seems that `list-n` suggests that `n` is somehow special, which could confuse someone reading this at a glance. If it was changed to `i` or something I think that it would be slightly clearer.

Kodiologist: Fair enough; changed. I also factored out some code into a subroutine.

peaceamongworlds (approved):

Kodiologist:
My last change broke something. I'm trying to figure out how.

Kodiologist:
That should do it. I'll merge after Travis runs.

PR #1601, opened 2018-05-02
Remove Botsbuildbots to get Hy installing again

Kodiologist:
Fixes #1599.

PR #1982, opened 2021-02-27
Burninate botsbuildbots

### LINKED ISSUE #1961: Remove Botsbuildbots contrib module
`Bobsbuildbots` currently isn't tested or documented in the docs and I'm not sure what it even does. As of right now it doesn't actually work because it uses the old bracketed `import` syntax:
`(import [requests])`
It also doesn't really align with what the contrib module has become of late, namely a testing ground for language features and functionality. Maybe I'm just not understanding what the point of it is, but I'd vote to remove it barring some more persuasive explanation for why we should keep it.

Kodiologist:
Closes #1961.

Third time's the charm (#1189, #1601).


allison-casey:
`botsbuildbots` was mentioned in `NEWS` before so you might as well mention it in `Removals`

Kodiologist:
It's always been undocumented and it was already broken for several releases, though.

allison-casey:
Also true :woman_shrugging:

allison-casey (approved):
Burninating the broken modules :fire:

scauligi (approved):

PR #1983, opened 2021-02-27
Prevent more ways to assign to constants

### LINKED ISSUE #1780: Detection of illegal assignments is imperfect
`(setv True 3)` is a `HySyntaxError`, but `(defn f [True] True) (f 3)` works without complaint, which is problematic because it leads to `hy2py` generating invalid Python code:

    $ echo '(defn f [True] True)' | hy2py 
    def f(True):
        return True

Kodiologist:
Fixes #1780.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -42,6 +42,7 @@ Bug Fixes
 * Fixed error handling for non-symbol macro names
 * `doc` and `#doc` now work with names that require mangling.
 * Fixed compiler crash on `.` form with empty attributes.
+* Attempts to assign to constants are now more reliably detected.
 
 0.20.0 (released 2021-01-25)
 ==============================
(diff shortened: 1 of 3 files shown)


scauligi:
Test failures on 3.6, 3.7, pypy. My guess is because of this:
> Changed in version 3.8: Class ast.Constant is now used for all constants.
> Deprecated since version 3.8: Old classes ast.Num, ast.Str, ast.Bytes, ast.NameConstant and ast.Ellipsis are still available, but they will be removed in future Python releases.

So if we still plan to support these versions then you should include these classes in the new check in `_storeize`.

Kodiologist:
Yep, thanks for the diagnosis. I'll look into it.

Kodiologist:
It looks like the assignments are detected properly. The tests of the error messages are just too picky. So I changed the tests.

scauligi (approved):

allison-casey (approved):

PR #1985, opened 2021-03-01
Feature: Add callable for macro to core

### LINKED ISSUE #867: Add reader macro for generator expressions

allison-casey:
Closes #867 

This is a pretty common idiom that python even has its own sugar for so I think it makes sense to add to core. 

This should probably be grouped next to the other comprehension forms in docs, but isn't possible right now in sphinx hydomain. Once we have the new `defn` args syntax nailed down, i'll go and push an update to `hydomain` to fix all the stuff we've broken in master and give us more control on the organization of the api autodocs.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -24,7 +24,7 @@ Breaking Changes
 
 New Features
 ------------------------------
-* New standard macros `do-n` and `list-n`
+* New standard macros `do-n`, `list-n`, and `cfor`
 * New contrib module `destructure` for Clojure-style destructuring.
 * Location of history file now configurable via environment variable `HY_HISTORY`
 * Added handling for "=" syntax in f-strings.
--- a/docs/cheatsheet.json
+++ b/docs/cheatsheet.json
@@ -111,6 +111,7 @@
           "hy.core.macros.as->",
           "hy.core.macros.->",
           "hy.core.macros.->>",
+          "hy.core.macros.cfor",
           "hy.core.macros.doto",
           "hy.core.macros.of",
           "hy.core.macros.unless",
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -772,3 +772,35 @@
 
    Use ``(help foo)`` instead for help with runtime objects."
   `(help (.get __macros__ (mangle '~symbol) None)))
+
+
+(defmacro cfor [f &rest generator]
+  #[[syntactic sugar for passing a ``generator`` expression to the callable ``f``
+
+  Its syntax is the same as :ref:`generator expression <py:genexpr>`, but takes
+  a function ``f`` that the generator will be immedietly passed to. Equivalent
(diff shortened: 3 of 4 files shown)


Kodiologist:
> The equivalent in python would be: `tuple(x for x in range(10) if is_odd(x))`

Wouldn't that just be `(tuple (gfor  x (range 10)  :if (odd? x)  x))`?

allison-casey:
yes, but iterables to single arg functions is such a common idiom that calling `gfor` every time gets pretty cumbersome. 
```clojure
(= (cfor tuple x (range 10) :if (odd? x) x)
   (tuple (gfor x (range 10) :if (odd? x) x)))
```
Python lets you drop the parens for cases like these and I think we should support that too.

Kodiologist:
I see. In that case, the documentation can be much more concise. Something like "`(cfor f …)` is shorthand for `(f (gfor …))`". Although, it's not great shorthand since all you do is save one pair of parentheses. Also, I'm not clear why you provide this as both a regular macro and a tag macro. Aren't they interchangeable?

allison-casey:
Removed tag macro and simplified the docs. I find it much easier to read, saving the paren is just icing on the cake

scauligi:
Minor nitpick: use a bracketed string `#[[ ... ]]` for the docstring for `cfor` so that you don't need to escape the quotes -- it'll make looking at the docstring in the source cleaner.

allison-casey:
> use a bracketed string `#[[ ... ]]` for the docstring for `cfor`

:sunglasses: :point_right: :point_right:

Kodiologist (approved):

Kodiologist:
`(cfor dict x "ABCD" [x True])` might not be a great example, since you can write it as `(dfor x "ABCD" [x True])`.

allison-casey:
valid, i'll replace it with something more relevant

allison-casey:
Changed the example to a more real world example

scauligi (approved):

PR #1988, opened 2021-03-09
Fix bugs with `require` and mangling

Kodiologist:
Perhaps we should be converting code like `foo.bar.baz` to `(. foo bar baz)` in the parser instead of allowing periods in literal symbols. But this seems to work okay for now.


tests/native_tests/language.hy
```diff
@@ -1420,31 +1420,48 @@ cee\"} dee" "ey bee\ncee dee"))
   (assert (= y 1)))
 
 (defn test-require []
-  "NATIVE: test requiring macros from python code"
```
scauligi, line 1423: It doesn't look like this functionality is tested elsewhere; if we're still planning on supporting macros defined in python we should still have a test for this?

scauligi:
Even with the new tests I'm still a bit unclear on what the bug(s) in question were that are being fixed. Can you provide a small example of what was going wrong?

tests/native_tests/language.hy
Kodiologist: I don't think it was ever something we wanted to support per se. The only motivation I can think of for writing those definitions in Python was to try to make the test independent of the Hy syntax for defining macros. Maybe that made more sense earlier in Hy's life, but now, it's the implementation we expect to change more than the Hy syntax.

Kodiologist:
Macros whose names needed to be mangled, including tag macros, weren't handled properly by `require`, especially in qualified forms like `T.✈` or `#T.tagmacro`.

Issue #1751, opened 2019-02-21
Preparing to have to move CI from Travis


refi64:
https://www.reddit.com/r/programming/comments/at5nrz/ibera_is_gutting_travis_ci_just_a_few_weeks_after/

Basically a ton of staff just got fired, ex employees are saying they're not sure if there's even enough left to maintain it. Maybe wait for some more official confirmation before jumping the gun, but this might be something we'd be better off preparing for in advance.

Kodiologist:
Fun. Do you have any experience with alternatives, like CircleCI?

kirbyfan64:
*Personally* I'm a huge fan of Azure Pipelines, since their free OSS tier is pretty liberal and is pretty fast.

refi64:
So I'll probably take a look at this soon-ish, considering that Travis going down for nearly 8 hours is a pretty bad future sign.

Kodiologist:
Accomplished by #2066.

PR #1991, opened 2021-03-12
Add pattern matching support for Python 3.10

peaceamongworlds:
[Pattern matching](https://www.python.org/dev/peps/pep-0636/) has been added to Python 3.10.

I've added functions to the compiler to dispatch for the `match` special form and compile it to the correct `ast.Match*` nodes.

It seems to be fully functional, but there are a few edge cases with statements within patterns and guards. I'm not sure what the best way to deal with them would be. Currently it ignores any special forms in patterns, treating everything literally, and it discards any statements from guards. Both of these behaviours can lead to slightly weird results.

This is the syntax that I've implemented:
```hy
(match test-expression case1 ... caseN)
```
where each case is of the form
```hy
[pattern :if guard body]
```
where the `:if guard` is optional for each case and `body` can have many forms. Most patterns are specified in the same way as they would be in Python, and are fully composable. The exceptions are the or and as patterns, which are specified as
```hy
(:or pattern1 ... patternN)
(:as pattern name)
```
Objects are matched in the same way as in Python with normal function calls.

Here are a few examples from the PEP tutorial translated to hy:
```py
match point:
    case Point(x, y) if x == y:
        print(f"Y=X at {x}")
    case Point(x, y):
        print(f"Not on the diagonal")
```
```hy
(match point
  [(Point x y) :if (= x y)
   (print f"Y=X at {x}")]
  [(Point x y)
   (print f"Not on the diagonal")])
```
```py
match event.get():
    case Click(position=(x, y)):
        handle_click_at(x, y)
    case KeyPress(key_name="Q") | Quit():
        game.quit()
    case KeyPress(key_name="up arrow"):
        game.go_north()
    ...
    case KeyPress():
        pass # Ignore other keystrokes
    case other_event:
        raise ValueError(f"Unrecognized event: {other_event}")
```
```hy
(match (.get event)
  [(Click :position [x y])
   (handle-click-at x y)]
  [(:or (KeyPress :key-name "Q") (Quit))
   (.quit game)]
  [(KeyPress :key-name "up arrow")
   (.go-north game)]
  ...
  [(KeyPress)
   (pass) ; Ignore other keystrokes
   ]
  [other-event
   (raise (ValueError f"Unrecognized even: {other-event}"))])
```

Note that astor doesn't seem to support Python 3.10, and, in particular, doesn't know what to do with `ast.Match`, so this breaks `hy2py` and `--spy`.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -4,7 +4,7 @@ import importlib
 import py
 import pytest
 import hy
-from hy._compat import PY3_8
+from hy._compat import PY3_8, PY3_10
 
 NATIVE_TESTS = os.path.join("", "tests", "native_tests", "")
 
@@ -12,7 +12,9 @@ _fspath_pyimport = py.path.local.pyimport
 
 
 def pytest_ignore_collect(path, config):
-    return (("py3_8_only" in path.basename and not PY3_8) or None)
+    return (("py3_8_only" in path.basename and not PY3_8)
+            or ("py3_10_only" in path.basename and not PY3_10)
+            or None)
 
 
 def pyimport_patch_mismatch(self, **kwargs):
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -15,7 +15,7 @@ from hy.errors import (HyCompileError, HyTypeError, HyLanguageError,
 
 from hy.lex import mangle, unmangle, hy_parse, parse_one_thing, LexException
 
-from hy._compat import (PY3_8, reraise)
+from hy._compat import (PY3_8, PY3_10, reraise)
 from hy.macros import require, macroexpand
 
 import textwrap
@@ -366,23 +366,7 @@ class HyASTCompiler(object):
         # compilation.
         self.module.__dict__.setdefault('__macros__', {})
 
-        self.can_use_stdlib = not self.module_name.startswith("hy.core")
-
-        self._stdlib = {}
-
         self._in_pattern = False
-
-        # Everything in core needs to be explicit (except for
-        # the core macros, which are built with the core functions).
-        if self.can_use_stdlib:
-            # Load stdlib macros into the module namespace.
-            load_macros(self.module)
-
-            # Populate _stdlib.
-            for stdlib_module in hy.core.STDLIB:
-                mod = importlib.import_module(stdlib_module)
-                for e in map(ast_str, getattr(mod, 'EXPORTS', [])):
-                    self._stdlib[e] = stdlib_module
 
     def get_anon_var(self):
         self.anon_var_count += 1
@@ -800,13 +784,13 @@ class HyASTCompiler(object):
 
         return ret
 
-    @special("match",
+    @special((PY3_10, "match"),
              [FORM,
               many(brackets(FORM + maybe(sym(":if") + FORM) + many(FORM)))])
     def compile_match(self, expr, _, subject, clauses):
         subject = self.compile(subject)
         var = self.get_anon_var()
-        name = asty.Name(expr, id=ast_str(var), ctx=ast.Store())
+        name = asty.Name(expr, id=mangle(var), ctx=ast.Store())
         initial_assign = asty.Assign(expr,
                                      targets=[name],
                                      value=asty.Constant(expr, value=None))
@@ -849,7 +833,7 @@ class HyASTCompiler(object):
                             for p, g, b
                             in itertools.islice(zip(ps, gs, bs), split_index)]
             if split_index == len(ps) - 1:
-                left_overs = None
+                left_overs = []
             else:
                 left_overs = [comp_match_clauses(ps[split_index+1:],
                                                  gs[split_index+1:],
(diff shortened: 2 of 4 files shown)


Kodiologist:
Unfortunately, the lack of support for astor and for statements are showstoppers. I would suggest closing this, getting `ast.Match` support into astor, and then trying to add statement support to this.

peaceamongworlds:
I submitted a pull request to astor to fix this, so hopefully that will work out.

>  trying to add statement support to this.

This does support statements and expressions in the test expression and in the body, and it returns the last value from the body. So something like the following does what you'd expect it to do in hy:
```hy
=> (setv y (match (do (print "hy") 1) [x (print "hy again") x]))
hy
hy again
=> y
1
```
The problem is putting statements in the pattern and guard parts.

The pattern is interpreted differently by Python from other code. Function calls in here aren't actual function calls, etc. This is more like unpacking assignments with extra syntax sugar. Furthermore, if you put any ast node in the `pattern` slot of `ast.match_case` that isn't a list, dictionary, symbol or a function call, it seems to make the interpreter crash without any warnings. So other expressions like list comprehensions aren't allowed here either.

The guard expression has to run in the same scope as the pattern, since it can use the variables that were bound there, so it can't be replaced by an anonymous function. Putting all the statements from all of the guard expressions before the entire match expression will also be an issue because they will be getting run before the assignment occurs, as well as running out of order. Each case of the pattern matching expression only gets run after all the previous ones have failed.

I don't see how these can be fixed, so I'd be happy if someone can help.

I know that I might have been a little to eager to start implementing this, given that Python 3.10 is still unstable. However, given that this is going to be added to Python now, I think it would be worth discussing the syntax that we should choose.

Kodiologist:
I see. Yeah, statements aren't going to happen in the pattern because it's not a real expression context, just a special kind of lvalue. For the guard, I think you can do something like moving the guard into an ordinary Python `if` in the body of the case, although it will take some cleverness with loops or splitting the `match` statement into parts in order to get proper fall-through to the next case when the `if` fails.

isidentical:
> Unfortunately, the lack of support for astor and for statements are showstoppers.

Would it be interesting to use [`ast.unparse()`](https://docs.python.org/3/library/ast.html#ast.unparse) for future python versions, considering it will always be sync with the current grammar?

Kodiologist:
Yes, it would be nice if we could get everything we needed with `ast.unparse` and drop the dependency on `astor`. I don't know if we can.

isidentical:
> I don't know if we can.

It is possible to have dependencies relative to python versions. I don't know much about hy's internals, though I would assume you could drop it under `3.9<` and have a wrapper unparse function (and when you are dropping 3.8 eventually, you can just drop it fully);

```py
if sys.version_info >= (3, 9):
     unparse_ast = ast.unparse
else:
     import astor
     unparse_ast = astor.to_source
```

Kodiologist:
I wasn't even thinking of the availability of the function. I was just wondering how well it would work for our purposes. Only one way to find out, I guess.

isidentical:
> I wasn't even thinking of the availability of the function. I was just wondering how well it would work for our purposes. Only one way to find out, I guess.

Do you use any other astor functionality beside unparsing?

Kodiologist:
I think so, but maybe those could be replaced, too. Anyway, this PR isn't the place to hash that out.

scauligi:
From a quick grep, it looks like we only use `astor.code_gen.to_source` and `astor.dump_tree`, which are (likely) analogous to `ast.unparse` and `ast.dump` respectively. We already have compat flags for other python versions in hy/_compat.py, so you can add one for 3.9 and then select the right unparse/ast-dump based on that (and would let us drop astor for >=3.9).

Kodiologist:
#1999 has eliminated the astor issue, in theory. What are your plans for this now, Peter?

peaceamongworlds:
> For the guard, I think you can do something like moving the guard into an ordinary Python `if` in the body of the case, although it will take some cleverness with loops or splitting the `match` statement into parts in order to get proper fall-through to the next case when the `if` fails.

I think that I gave up too early last time, but it seems like splitting the `match` statement into nested statements with and explicit `if` works. For example
```hy
$ > hy --spy --repl-output-fn=hy.contrib.pprint.pformat
import hy
hy 0.20.0+80.gb83f476f using CPython(heads/makepkg-dirty:ba251c2ae66) 3.10.0a6+ on Linux
=> (match 1
...  [0 :if (print 0) 0]
...  [x :if (do (print "x" x) False) x]
...  [y :if (do (print "y" y) True) y])
_hy_anon_var_1 = None
_hy_anon_var_2 = 1
match _hy_anon_var_2:
    case 0 if print(0):
        _hy_anon_var_1 = 0
    case x:
        print('x', x)
        if False:
            _hy_anon_var_1 = x
        else:
            match _hy_anon_var_2:
                case y:
                    print('y', y)
                    if True:
                        _hy_anon_var_1 = y
_hy_anon_var_1
x 1
y 1
1
```
I guess it's time to add documentation to everything.

peaceamongworlds:
It looks like the Python 3.10.0a5+ version used by Travis is from Feb. 2, 2021, just before the match statement was added. Hence the tests are failing.

allison-casey:
I wonder if this is another reason to migrate from travis #1751. If some other CI has better support for bleeding edge python version (and doesn't take hours to build :/) this could help justify the move

Kodiologist:
I wouldn't hold my breath looking for a free test-running service that always has the latest alpha of Python, but yes, Travis has issues.

Kodiologist:
I guess we should shelve this until it works with our testing infrastructure, and we might as well wait till the `ast` structure stabilizes, too.

Kodiologist:
Before I forget, a note for when we come back to this: ensure that the nested-match implementation has the right semantics when the match object mutates between the two matches.

PR #1999, opened 2021-03-16
Remove the dependency on astor for Pythons ≥ 3.9

Kodiologist:
Python 3.9 added `ast.unparse`, which can do most of the work. It has some bugs that I fixed in `astor`, which I've worked around here, but I've submitted a patch to CPython to fix them for real (python/cpython#24897).

### DIFF
--- a/hy/_compat.py
+++ b/hy/_compat.py
@@ -6,6 +6,7 @@ import sys
 
 PY3_7 = sys.version_info >= (3, 7)
 PY3_8 = sys.version_info >= (3, 8)
+PY3_9 = sys.version_info >= (3, 9)
 PY3_10 = sys.version_info >= (3, 10)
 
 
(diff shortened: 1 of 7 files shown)


hy/core/language.hy
```diff
@@ -226,14 +226,16 @@
...
-       astor.code-gen.to-source
-       astor.dump-tree)
-    compiled))
+  (if
+    codegen
+      (hy._compat.ast-unparse compiled)
+    hy._compat.PY3_9
+      (ast.dump compiled :indent 1)
+    True
+      (ast.dump compiled)))
```
scauligi, line 238: Isn't the last form of an `if` already implicitly an `else`?

hy/_compat.py
```diff
@@ -14,3 +15,33 @@ def reraise(exc_type, value, traceback=None):
...
+                    .replace("inf", ast._INFSTR)
+                    .replace('nan', '({}-{})'.format(ast._INFSTR, ast._INFSTR)))
+            else:
+                super().visit_Constant(node)
+
+        def visit_Set(self, node):
+            if node.elts:
+                super().visit_Set(node)
+            else:
+                self.write('{1}.__class__()')
```
scauligi, line 38: (trivial, but) is there any significant difference between the representation here vs in your upstream patch `{*()}`?

peaceamongworlds (approved):

hy/core/language.hy
Kodiologist: Yes, but I think it's easier to read with the placeholder `True`.

hy/_compat.py
Kodiologist: They should be equivalent. I've changed it to `{*()}`.

hy/_compat.py
```diff
@@ -14,3 +15,33 @@ def reraise(exc_type, value, traceback=None):
         raise value.with_traceback(traceback)
     finally:
         traceback = None
+
+
+if PY3_9:
+
+    import ast, math, cmath
```
scauligi, line 22: The `math` and `cmath` imports appear irrelevant.

hy/_compat.py
```diff
@@ -14,3 +15,33 @@ def reraise(exc_type, value, traceback=None):
...
+    class Unparser(ast._Unparser):
+      # Work around some limitations of Python's `ast.unparse`.
```
scauligi, line 24: Add a reference to https://github.com/python/cpython/pull/24897. Since it's going to be backported to 3.9, we can eventually remove this workaround once it lands.

scauligi (approved):
Aside from a few more small changes, lgtm

hy/_compat.py
Kodiologist: Yep, they're leftovers.

hy/_compat.py
Kodiologist: Done.

PR #2009, opened 2021-03-20
Fix: Add test for empty decorator call

### LINKED ISSUE #1565: #@() throws an `IndexError`
Should probably throw a nicer error message like `(with-decorator)` does

allison-casey:
closes #1565


Kodiologist:
Can't you write that as `(eval '#@())`?

allison-casey:
yup, should probably refactor existing tests to do that then as well

Kodiologist (approved):

Kodiologist:
Don't forget to clean up your commit messages after squashing, by the way.

scauligi (approved):

PR #2010, opened 2021-03-20
Fix: test eval can see locals

### LINKED ISSUE #1041: eval can only see locals

allison-casey:
closes #1041

### DIFF
--- a/tests/native_tests/language.hy
+++ b/tests/native_tests/language.hy
@@ -1149,7 +1149,11 @@
   (setv x (quote 42))
   (assert (= x (hy.eval x)))
   (assert (= 27 (hy.eval (+ (quote (*)) (* [(quote 3)] 3)))))
-  (assert (= None (hy.eval (quote (print ""))))))
+  (assert (= None (hy.eval (quote (print "")))))
+
+  ;; https://github.com/hylang/hy/issues/1041
+  (assert (is (hy.eval 're) re))
+  (assert (is ((fn [] (hy.eval 're))) re)))
 
 
 (defn test-eval-false []


Kodiologist (approved):
I would use `is` rather than `=` here. Otherwise, looks good.

allison-casey:
> I would use is rather than = here

done

scauligi (approved):

PR #2018, opened 2021-03-27
Rename `defn` to `def` and `defclass` to `class`

Kodiologist:
I haven't actually made the changes yet. I opened this PR to check for sufficient agreement before I forge ahead.

Anyway, I'm thinking this is a good time to do this because then `defn` will get renamed to `def` and lambda lists will change in the same release, so users can use which keyword is used in a Hy program as a hint as to whether the code has been updated.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -17,6 +17,8 @@ Removals
 
 Breaking Changes
 ------------------------------
+* `defn` has been renamed to `def`.
+* `defclass` has been renamed to `class`.
 * f-strings are now parsed as a separate `HyFString` node,
   which is a collection of `HyString` and `HyFComponent` nodes.
 * Calling a `HyKeyword`  now looks up by its (string) name instead by its self.


Kodiologist:
To reiterate, the motivation here is better consistency with Python. These special operators are our direct equivalents of the Python keywords, so using the same name helps the user avoid accidentally mixing up Python and Hy conventions when switching between them, as one often needs to do as a Hy programmer. Using a different name from Python makes sense sometimes, for sure, like how we use the shorter `fn` in place of the clumsy `lambda`, but there's no clear motivation for `defn` over `def` except consistency with different programming languages.

peaceamongworlds:
I don't really like this idea. As I said in our previous discussion, I just think that using `defn` and `defclass` makes a lot more sense, and that, if `def` exists, it should only be used for the most basic form of assignment, not function definition.

Kodiologist:
Okay, with you opposed, I would likely need both @allison-casey and @scauligi to agree in order to get this change. What do you guys think? You seemed supportive in #1975.

Kodiologist:
For what it's worth, we used to have `def` as a regular assignment form, equivalent to `setv`, and it caused a lot of predictable confusion.

TheCatster:
I know I'm not a contributor, but in my opinion I don't see a major reason to have Hy even closer to Python. The major reason I was attracted to it was because it was a Lisp I could continue to use even when my work needed Python. Those coming from Clojure and such would have an easier time transitioning, and as someone who just started using the language a week ago and am already fully comfortable, I think I'm an example of that. Python devs wanting to learn Lisp have a smoother transition, as they can still use their Python comforts here while learning the macros and further styles of Lisp. Those are my two cents at least!

allison-casey:
I've come around to peace's point of view on this one since we last talked about it in #1975. the `def` prefix are syntactically better constructs than python's `def`, `class`, etc forms. It's just less of an issue in python because you can't introduce you're own define forms like you can in a lisp language with macros. We even have it in our own core with `defmain`, but it's a common idiom all over other lisps and code bases. `defmulti` in clojure is a macro the creates a new define therefore it gets the `def` prefix. I do this in my own code all the time too like `defcommand` when wrapping around `discord.py` methods. So i don't think this is a good idea anymore. `setv` makes sense being different to me since it's an assignment capable of mutating the argument. `defn` and `defclass` can only introduce/overwrite a define unlike `setv`

Kodiologist:
Okay, F to this proposal.

PR #2020, opened 2021-03-27
Allow expressions such as `(require [hy.contrib [walk]])`

### LINKED ISSUE #2016: Cannot `require` module from a package
The following are all valid:
```hy
$ > hy --spy --repl-output-fn=hy.contrib.pprint.pformat
import hy
hy 0.20.0+132.g2a6b79f1 using CPython(heads/makepkg-dirty:ba251c2ae66) 3.10.0a6+ on Linux
=> (require hy.contrib.walk)
hy.macros.require('hy.contrib.walk', None, assignments='ALL', prefix='hy.contrib.walk')
None
=> (import hy.contrib.walk)
import hy.contrib.walk
None
=> (require [hy.contrib.walk :as w])
hy.macros.require('hy.contrib.walk', None, assignments='ALL', prefix='w')
None
=> (import [hy.contrib.walk :as w])
import hy.contrib.walk as w
None
=> (import [hy.contrib [walk]])
from hy.contrib import walk
None
```
However, the following fails:
```hy
=> (require [hy.contrib [walk]])
  File "stdin-fc566da3a366533698786e10deb0bc456fb8d19e", line 1, in <module>
    (require [hy.contrib [walk]])
hy.errors.HyRequireError: The module <module 'hy.contrib' from '/home/peter/Programming/hy/hy-source/hy/hy/contrib/__init__.py'> has no macros
```
As I understand, `import` and `require` should allow for the same syntax, and given that the previous `require` forms work as expected, I think that the last one should as well.

Just to clarify, I think that `(require [hy.contrib [walk]])` should have the same effect as `(require [hy.contrib.walk :as walk])`.

peaceamongworlds:
Fixes #2016.

Also gets rid of error `HyRequireError` when requiring a module with no macros, since previously this behavior was inconsistent and I don't see why this should be an error. So previously
```hy
i => (require os.path)
i => (require [os [path]])
  File "stdin-08ce82c15e0a77ce6e17125ffa2f87a3fbdcd4df", line 1, in <module>
    (require [os [path]])
hy.errors.HyRequireError: The module <module 'os' from '/usr/lib/python3.9/os.py'> has no macros or tags
```
But now the second will also not throw an error.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -61,7 +61,8 @@ Bug Fixes
 * Fixed namespace pollution caused by automatic imports of Hy builtins and macros.
 * Fixed Hy model objects sometimes not being in scope.
 * Fixed `doc` sometimes failing to find builtin macros.
-* `require` now works with relative imports.
+* `require` now works with relative imports and can import modules, such as
+  `(require [hy.contrib [walk]])`.
 
 Misc. Improvements
 ------------------------------
--- a/hy/macros.py
+++ b/hy/macros.py
@@ -156,7 +156,9 @@ def require(source_module, target_module, assignments, prefix=""):
                 source_dirs = source_module.split(".")
                 target_dirs = (getattr(target_module, "__name__", target_module)
                                .split("."))
-                while source_dirs and target_dirs and source_dirs[0] == "":
+                while (len(source_dirs) > 1
+                       and source_dirs[0] == ""
+                       and target_dirs):
                     source_dirs.pop(0)
                     target_dirs.pop()
                 package = ".".join(target_dirs + source_dirs[:-1])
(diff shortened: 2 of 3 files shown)


Kodiologist:
You know, I'd forgotten that we are now only supporting Pythons recent enough to have f-strings, and so we can use f-strings in our own Python code. Neat.

Can you use parentheses instead of backslashes for the long condition?

For tests of requiring a macro that's supposed to exist, use a test macro rather than a real one in our core like `hy.contrib.destructure.setv+`, so the test doesn't depend on unrelated features existing.

Kodiologist:
And don't forget a NEWS item.

peaceamongworlds:
> You know, I'd forgotten that we are now only supporting Pythons recent enough to have f-strings, and so we can use f-strings in our own Python code. Neat.
> 
> Can you use parentheses instead of backslashes for the long condition?
> 
> For tests of requiring a macro that's supposed to exist, use a test macro rather than a real one in our core like `hy.contrib.destructure.setv+`, so the test doesn't depend on unrelated features existing.

Done

Kodiologist (approved):

scauligi (approved):

PR #2024, opened 2021-03-29
Fix #2021: hy2py cannot import local module

### LINKED ISSUE #2021: `hy2py` cannot import local module during compile time
`eggs.hy`:
```hy
(print 42)
```
`spam.hy`:
```hy
(eval-when-compile (import eggs))
```

`hy2py` fails to compile `spam.hy`:
```bash
$ > hy2py spam.hy 
  File "/usr/bin/hy2py", line 12, in <module>
    sys.exit(hy2py_main())
  File "spam.hy", line 2, in <module>
    (import eggs))
hy.errors.HyEvalError: 
  File "[HyExpression([
  HySymbol('import'),
  HySymbol('eggs')])]", line 1
    (eval-when-compile
    ^----------------^
No module named 'eggs'
```
But `hy` itself runs fine:
```bash
$ > hy spam.hy 
42
```
Note: this seems to be  unaffected by whether `pyc` files are present for any of the files.

scauligi:
Set `sys.path` correctly in `hy2py` and `hyc`. Also refactor the path-setting code a little.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -63,6 +63,7 @@ Bug Fixes
 * Fixed `doc` sometimes failing to find builtin macros.
 * `require` now works with relative imports and can import modules, such as
   `(require [hy.contrib [walk]])`.
+* Fixed `hyc` and `hy2py` not finding relative imports
 
 Misc. Improvements
 ------------------------------
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -443,6 +443,16 @@ def ideas_macro(ETname):
 """)])
 
 
+def set_path(filename):
+    """Emulate Python cmdline behavior by setting `sys.path` relative
+    to the executed file's location."""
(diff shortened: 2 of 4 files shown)


peaceamongworlds (approved):

Kodiologist:
Instead of `sys.path.insert(0, None)`, why not put the `sys.path.pop(0)` at the bottoms of the two loops?

Why the `os.path.realpath`? There are no symlinks to resolve here, right?

scauligi:
> Instead of `sys.path.insert(0, None)`, why not put the `sys.path.pop(0)` at the bottoms of the two loops?

Good point. Fixed.

> Why the `os.path.realpath`? There are no symlinks to resolve here, right?

The original code here also used realpath; based on some quick testing I did it seems that this mimics what the Python REPL does as well.

Kodiologist (approved):
Rebase this and it's clear to merge.

PR #2026, opened 2021-03-29
Remove `with*` and make `with` treat `_` as special

### LINKED ISSUE #1993: `with` forms cannot parse multiple non-named with-exprs
Currently, `with` parses like so (simplification mine):
```
=> (with [a b c d e] ...)
with b as a:
  with d as c:
    with e:
      ...
```

As a result, there is no way of writing multiple non-named forms, nor is there any way to have a non-named form anywhere other than the end.

I propose the following instead, with a set of outer brackets for the with-exprs and inner brackets for any named with-exprs:

```
=> (with [[a b] c d [e f]] ...)
with b as a, c, d, f as e:
  ...
```

peaceamongworlds:
Fixes #1993.

`with*` is deprecated and `with` is now a special form. It also compile `_` to unnamed context managers.


hy/compiler.py
```diff
@@ -896,36 +896,45 @@ def compile_decorate_expression(self, expr, name, args):
...
+              many(FORM)])
     def compile_with_expression(self, expr, root, args, body):
-        thing, ctx = (None, args[0]) if args[1] is None else args
-        if thing is not None:
-            thing = self._storeize(thing, self.compile(thing))
-        ctx = self.compile(ctx)
+        # thing, ctx = (None, args[0]) if args[1] is None else args
+        # if thing is not None:
+        #     thing = self._storeize(thing, self.compile(thing))
+        # ctx = self.compile(ctx)
```
scauligi, line 907: Did you mean to leave these commented?

tests/native_tests/contrib/walk.hy
```diff
@@ -58,26 +58,14 @@
   ;; make sure a macro from the current module works
   (assert (= (macroexpand-all '(foo-walk))
              42))
-  (assert (= (macroexpand-all '(with [a 1]))
-             '(with* [a 1])))
+  (assert (= (macroexpand-all '(-> 1 a))
+             '(a 1)))
```
scauligi, line 62: These new tests seems unrelated to the current PR

scauligi (commented):
Can you add tests for the new `with` pattern(s)? In particular we want to make sure that using `_` in fact makes an unnamed manager instead of introducing a new `_` name.

tests/native_tests/contrib/walk.hy
peaceamongworlds: The old tests used `with`, which is no longer a macro, so I had to change them to something else.

hy/compiler.py
peaceamongworlds: No, that was silly

Kodiologist:
It looks like there was a mistake in the original model pattern for `with*`, where `FORM` was used instead of `SYM` for the new variable. Might as well fix that.

peaceamongworlds:
> It looks like there was a mistake in the original model pattern for `with*`, where `FORM` was used instead of `SYM` for the new variable. Might as well fix that.

Python allows for unpacking in `with` statements if the context manager returns a tuple or list. For example:
```python
[ins] In [134]: class Test():
           ...:     def __init__(self, x, y):
           ...:         self.x = x
           ...:         self.y = y
           ...:     def __enter__(self):
           ...:         return self.x, self.y
           ...:     def __exit__(self, type, value, traceback):
           ...:         pass
[ins] In [136]: with Test(1, 2) as (x, y):
           ...:     print(x, y)
           ...: 
1 2
```
So I think that `FORM` is correct.

Kodiologist:
Why not merge it, then? It has two approvals.

hy/compiler.py
```diff
@@ -896,36 +896,40 @@ def compile_decorate_expression(self, expr, name, args):
         fn.stmts[-1].decorator_list = decs + fn.stmts[-1].decorator_list
         return ret + fn
 
-    @special(["with*", "with/a*"],
-             [brackets(FORM, maybe(FORM)), many(FORM)])
+    @special(["with", "with/a"],
+             [(brackets(times(1, float("inf"), FORM + FORM)))
```
scauligi, line 900: minor nit: use `Inf` instead of `float("inf")` to match other patterns

docs/api.rst
```diff
@@ -1109,6 +1109,52 @@ Special Forms
...
+  manner. The archetypical example of using ``with`` is when processing files.
+  If only a single expression is supplied, or the argument is `_`, then no
+  variable is bound to the expression, as shown below.
+
+  Examples:
+    ::
+
+       => (with [arg (expr)] block)
+       => (with [(expr)] block)
+       => (with [arg1 (expr1) _ (expr2) arg3 (expr3)] block)
```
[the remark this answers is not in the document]
peaceamongworlds: Done

docs/api.rst
```diff
@@ -1109,6 +1109,52 @@ Special Forms
...
+
+  ``with`` returns the value of its last form, unless it suppresses an exception
+  (because the context manager's ``__exit__`` method returned true), in which
+  case it returns ``None``. So, the previous example could also be written::
+
+       => (print (with [f (open \"NEWS\")] (.read f)))
+           
+.. hy:function:: (with/a [#* args])
+
+  Wrap execution of `body` with/ain a context manager given as bracket `args`.
```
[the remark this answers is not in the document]
peaceamongworlds: Fixed

peaceamongworlds:
There's something weird going on with `oneplus`. There was a bug in it's implementation, see https://github.com/vlasovskikh/funcparserlib/commit/10f4017287b88fc85a67a6e13760f915a137a8e4, which has since been fixed. However, pip (and pacman for that matter) seem to install the old version which still has the bug.

Kodiologist:
I posted an issue requesting a new release of funcparserlib, so hopefully that will be taken care of soon. Failing that, you can try making Hy depend on an unreleased version from GitHub.

Kodiologist:
Or we can switch back to `times` for now. That's fair, too.

peaceamongworlds:
I think `times` is fine for now. When funcparserlib is updated, it'll be easy to update this as well.

scauligi (approved):

Issue #1992, opened 2021-03-14
The great core shakeup


Kodiologist:
Hy has a big and disorganized set of core functions and macros that get automatically  injected into `__builtins__`. This is an improvement over the old status quo, in which they were automatically imported, but still messy. I propose some radical changes to clean things up. I'll do it over the course of several PRs, and we can bikeshed about the details (e.g., the fates of individual functions and macros) when we get there.

- [ ] Remove machinery for injecting functions into `__builtins__`.
  - Thus, the only names that will be user-visible without the user `import`ing or `require`ing stuff will be `hy`, core macros, and gensyms and other temporary names.
- [x] Rename model types to remove "Hy".
- [x] Make `quote` use qualified names to pull in model objects, so `'5` will compile to `hy.models.Integer(5)` instead of `Integer(5)`.
- Reorganize core macros:
  - [x] Rename `if*` to `if`.
  - [x] Rename `annotate*` to `annotate`.
  - [x] Delete `fn*` in favor of `defn`.
  - [x] Delete `with*` and `with/a*` in favor of `with` and `with/a`,
- [x] Rename `hy.core.shadow` to something more logical like `hy.pyops` and encourage users to star-import it when they want first-class Python operators, as in `(import [hy.pyops [*]])`.
- [x] Make these functions available qualified under `hy`, as e.g. `hy.eval`: `eval` `mangle` `unmangle` `gensym` `macroexpand` `macroexpand-1` `read` `read-str` `disassemble` `repr` (formerly `hy-repr`) `repr-register` (formerly `hy-repr-register`).
- [ ] Make a new repository under `hylang`, `hyrule`, to hold a Hy library with a variety of functions and macros that are useful for Hy (and sometimes Python) but not required to implement Hy.
  - Hyrule's contribution guidelines and core team will be the same as Hy's, but it can be less stable than Hy itself and perhaps it will have a different release schedule.
  - The organization of Hyrule is to be determined, but splitting things up into modules is likely better than one big module.
  - [ ] Include all of `extra` and `contrib`, except for `hy-repr`.
  - [ ] Other macros (formerly core) to be included:
    - [ ] Control structures: `unless` `do-n` `list-n` `lif`
    - [ ] Meta-macros: `with-gensyms` `defmacro/g!`
    - [ ] Parenthesis sugar: `as->` `->` `doto` `->>`
    - [ ] Others: `assoc` `of` `defmain` `#@` `comment` `cfor`
[...]

PR #2031, opened 2021-04-01
Rename hy model types

### LINKED ISSUE #1644: Don't include the namespace in the import's name for hy models.

peaceamongworlds:
Closes #1644.

All hy models have `Hy` removed from their name, so `HyString` -> `String`, `HyList` -> `List` etc.

Also, `hy.models` has been prepended to the repr of all the models.

There are still objects which include `Hy` in their name, such as all the hy errors, `HyASTCompiler`, `HyREPL` and maybe some others. It would be nice to rename those as well, if everyone agrees.

### DIFF
--- a/docs/language/internals.rst
+++ b/docs/language/internals.rst
@@ -17,20 +17,20 @@ Hy models are a very thin layer on top of regular Python objects,
 representing Hy source code as data. Models only add source position
 information, and a handful of methods to support clean manipulation of
 Hy source code, for instance in macros. To achieve that goal, Hy models
-are mixins of a base Python class and :ref:`HyObject`.
+are mixins of a base Python class and :ref:`Object`.
 
 .. _hyobject:
 
-HyObject
+Object
 ~~~~~~~~
 
-``hy.models.HyObject`` is the base class of Hy models. It only
+``hy.models.Object`` is the base class of Hy models. It only
 implements one method, ``replace``, which replaces the source position
 of the current object with the one passed as argument. This allows us to
 keep track of the original position of expressions that get modified by
 macros, be that in the compiler or in pure hy macros.
 
-``HyObject`` is not intended to be used directly to instantiate Hy
+``Object`` is not intended to be used directly to instantiate Hy
 models, but only as a mixin for other classes.
 
 Compound Models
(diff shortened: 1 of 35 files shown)


Kodiologist:
I'm against this because I'd rather skip right to `hy.models.Integer`, as I proposed in #1992, than make that change later and break people's code again.

peaceamongworlds:
> I'm against this because I'd rather skip right to `hy.models.Integer`, as I proposed in #1992, than make that change later and break people's code again.

Are you opposed to including the models in the `hy` namespace, or changing their `__repr__` methods?

You mentioned #1644 in #1992, but that suggests making the models accessible as `hy.Model`. `quote` already uses the fully qualified name (since #1979 I think), so changing their string representation seems like a more of an aesthetic change.

Kodiologist:
> Are you opposed to including the models in the `hy` namespace, or changing their `__repr__` methods?

Both. I want `HyInteger` to be accessible as `hy.models.Integer` and I want the `repr` to reflect that.

peaceamongworlds:
Fair enough

peaceamongworlds:
I've changed it so that the hy models can only be accessed through `hy.models` and `repr` shows `hy.models` as well.

Kodiologist:
Thanks. Could you take any changes that aren't part of the massive search-and-replaces, like the NEWS line, and put them in one or more commits separate from the search-and-replaces?

docs/language/internals.rst
```diff
@@ -50,131 +50,131 @@ respectively, to ``False``.
...
 .. _hysequence:
 
-HySequence
+Sequence
 ~~~~~~~~~~
 
-``hy.models.HySequence`` is the abstract base class of "iterable" Hy
-models, such as HyExpression and HyList.
+``hy.models.Sequence`` is the abstract base class of "iterable" Hy
+models, such as hy.Expression and hy.List.
```
scauligi, line 57: Looks like there are still a few places in the docs that aren't scoped to `hy.models`, possibly other locations as well.

peaceamongworlds: I think I forgot to check everywhere, but that should be fixed now.

peaceamongworlds:
I think that should be everything. The commits are somewhat intertwined, but I've separated them out into logical sections.

Kodiologist (approved):
Well done.

scauligi (approved):

PR #2028, opened 2021-03-30
Clean up TODO and XXX comments

Kodiologist:

PR #2032, opened 2021-04-04
Release Hy 1.0a1

Kodiologist:
Typically I do a release every six months, but Hy has been moving so fast now (which I'm quite pleased about, I hasten to add—excellent work, folks) that I'd like to do an earlier release to help ease the pain of transition. Probably there will be another one around the middle of 2021.

I need two approvals from @hylang/core to do this. If I get them, I'll clean up NEWS and make the release.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,79 +1,74 @@
 .. default-role:: code
 
-Unreleased
+1.0a1 (released 2021-04-12)
 ==============================
 
 Removals
 ------------------------------
-* `name` function has been removed from core.
-  Use `(. :keyword name)` and/or explicit calls to `unmangle` instead.
+* The core function `name` has been removed.
+   Use `unmangle` or the `name` attribute of keyword objects instead.
 * `deftag` has been removed. Instead of `(deftag foo …)`,
   say `(defmacro "#foo" …)`.
 * `#doc` has been removed. Instead of `#doc @`, say `(doc "#@")`.
 * `__tags__` has been removed. Tag macros are now tracked in
   `__macros__`.
-* `eval` has been removed from core.
-  Use the fully qualified name `hy.eval` instead.
 
 Breaking Changes
 ------------------------------
-* f-strings are now parsed as a separate `HyFString` node,
-  which is a collection of `HyString` and `HyFComponent` nodes.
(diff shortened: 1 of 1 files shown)


peaceamongworlds:
Given that we might be able to complete all of the `1.0` features in the normal release timeframe, would it not make sense to hold of a new release until then? I think that we started this discussion in #2008.

Or maybe we could have a release only with the new bug fixes (looking at the recent prs: #2028, #2024, #2020 etc) and not any of the breaking features (#1975 seems like a big one)? Then we could put of all the breaking features until `1.0`, so that users didn't have to re-write their code twice.

I don't really have a strong opinion about this though, so I'm willing to be convinced either way.

Kodiologist:
> Given that we might be able to complete all of the 1.0 features in the normal release timeframe

I don't think that's realistically going to happen.

> Or maybe we could have a release only with the new bug fixes

I don't want to manage multiple lines of development before 1.0.

Kodiologist:
@allison-casey @scauligi Can I get your input here?

scauligi:
Considering the amount of major changes (and upcoming changes), I think it makes more sense to release as an alpha pre-release 1.0a1 instead of 0.21.

Kodiologist:
Wouldn't that imply something near feature-completeness, or that we'll release 1.0 soon?

allison-casey:
I dont think so, we can have as many alpha releases as we want. python3.9 had 6 alphas and 5 betas before release and we have much less to do than they do. I don't even think we even need to merge bugfixes on 0.20 just leave it as is for people to play around with something that is usable and stable.

Kodiologist:
Okay, here's a full NEWS cleanup, and a new title for the PR. @allison-casey Do you anticipate any problems getting this new version to work with hy-interpreter?

allison-casey:
shouldn't be an issue. just need to change the requirements file.

allison-casey (approved):

allison-casey:
@scauligi @peaceamongworlds is there any reason not to approve this at this point? we need two approvals for releases.

scauligi (approved):

Kodiologist:
Thanks. I'll merge this and release it tomorrow or Monday.

PR #2035, opened 2021-04-12
Fix version number of new release

Kodiologist:
Just my luck: I completed the release of 1.0a1 only to find that it will not be installed by default by `pip install hy`, and PyPI still considers 0.20.0 to be the latest version (https://pypi.org/project/hy/). This turns out to be because, quoting from pip's documentation:

> pip will only install stable versions as specified by pre-releases by default. If a version cannot be parsed as a compliant PEP 440 version then it is assumed to be a pre-release.

So we should change back to the version number I originally proposed, 0.21.0, or if you guys are still against that, something ~~PEP 440-compliant~~ that pip considers to be stable.

@allison-casey @scauligi I would be grateful if we could clean this up today.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,7 +1,9 @@
 .. default-role:: code
 
-1.0a1 (released 2021-04-12)
+0.21.0 (released ???)
 ==============================
+
+This version was also mistakenly released as 1.0a1 on 2021-04-12.
 
 Removals
 ------------------------------


allison-casey:
That was the idea though, and `1.0a1` is pep-440 compliant. It's just a pre-release compliant version number. You can install it with pip by just doing `pip install --pre hy`. Pip will install whatever the latest pre release candidate is when using `--pre` 

> Pre-releases should allow a ., -, or _ separator between the release segment and the pre-release segment. The normal form for this is without a separator. This allows versions such as 1.1.a1 or 1.1-a1 which would be normalized to 1.1a1. It should also allow a separator to be used between the pre-release signifier and the numeral. This allows versions such as 1.0a.1 which would be normalized to 1.0a1.

Kodiologist:
But then instead of just having people using the latest stable version asking us about bugs that have been fixed in master, we'll have both people using 0.20.0 and people using the latest alpha. And we'll have to have three versions of the documentation instead of two. And so on. And Hy will appear to be in complete status, to anybody who uses `pip install` and doesn't think to specifically check for prereleases, until we finally release 1.0. Surely you can't think this state of affairs is desirable.

Kodiologist:
Since this isn't getting fixed as quickly as I'd hoped, I've yanked the release from PyPI and deleted the tag from master in the meantime.

allison-casey:
Sorry, the last two days turned into a lot. And the pre-release cycle should have been explained better. The only real difference between a pre-release cycle and what @peaceamongworlds suggested with not releasing a `0.21` and just continuing to develop on master is that the alphas give us versioned public facing test releases that users can safely test out at their discretion instead of being blindsided by breaking code changes in the stable pypi releases. 

Documentation wise, there would just be `0.20` and whatever the latest alpha release is. We'd be releasing alphas anytime there's a new feature or breaking change (which there shouldn't be too many left for 1.0)

Its pretty obvious from the commits and releases that hy is actively being worked on especially if we're releasing alphas. So I don't really agree that people will think Hy is in stasis. A prelease alpha/beta/release-candidate development cycle is standard practice and you can see it in the python, typescript, clojure codebases. 

Its a tradeoff of who bears the overhead of managing breaking changes. Right now it's users who have to frustratingly update code every release because Hy is such a volatile language and no one wants to develop software against that kind of volatility. Moving to alphas as we finish up `1.0` puts it on us to educate users that bugs and features are available in pre-releases and closing issues if they haven't been shown to fail on the latest alphas.

I'd understand if that changes the calculus for people when it comes to doing alphas/betas until `1.0`. If the consensus has changed now that the tradeoffs are more explicitly stated, i'd understand and agree to going back to the way things were. I, however, still think that the alphas route is better for Hy's userbase going forward than what we're doing now.

Kodiologist:
> The only real difference between a pre-release cycle and what peaceamongworlds suggested with not releasing a 0.21 and just continuing to develop on master is that the alphas give us versioned public facing test releases that users can safely test out at their discretion instead of being blindsided by breaking code changes in the stable pypi releases.

I see. In that case I really think we should resume regular old releases. There is no timeline for 1.0 (let's be honest, it's perfectly possible that it will never be released), and I don't want Hy to be stuck on 0.20.0 until we reach this fabled promised land. Hy is beta-quality software (it works and it's useful, but it's rough around the edges) and will continue to have breaking changes in each new release until we're reasonably confident we don't have to break stuff again in the near future. Them's the breaks. People who like a particular version and want to pin it are free to do that. Releasing more frequently easies the pain by allowing codebases to be updated in smaller steps.

Kodiologist:
I hope I'm not being unreasonable, in thinking that `pip install hy` should continue to retrieve newer versions on the way to 1.0, but I do seem to be in the minority. If this PR doesn't get the two approvals it needs, I guess we'll either be releasing alphas or releasing nothing until 1.0.

Kodiologist:
@scauligi @peaceamongworlds Could I get your input here? I'd like to resolve this as soon as possible.

scauligi:
Apologies for not responding sooner; I got slammed with a paper deadline up until early this morning.

I still think we should move to prerelease numbering; we're making a deliberate decision at this stage of Hy's development to embrace more major changes (read: unstable API), and I think this is properly reflected as alpha 1.0 releases.

> puts it on us to educate users that bugs and features are available in pre-releases 

I think an easy way to do this is a final 0.x.1 patch release to add a (prominent) note on the front page of the documentation (and also the README, and also the description on PyPI) saying that we have prerelease versions available, use `pip install --pre hy` if you want the latest features. Then leave 0.x behind and move to the prerelease numbering.

> we'll have to have three versions of the documentation instead of two.

Correct me if I'm wrong, but this doesn't really incur much overhead for us, does it? Since the 0.x documentation will just remain fixed as it is.

> Hy will appear to be in complete stasis, to anybody who uses pip install and doesn't think to specifically check for prereleases

This can be solved by pushing a note in the docs, as stated above. If they're using Hy without reading the docs then we can't help them. Also as @allison-casey noted, anyone who visits the GitHub repo (say, to open an issue) would see there is still active development.

> Releasing more frequently easies the pain by allowing codebases to be updated in smaller steps.

I think this will drive people to pin Hy to a 0.x release instead; to my mind, the kind of person willing to rewrite their Hy code for rolling releases is also the kind of person willing to install `--pre hy`.

> it's perfectly possible that it will never be released

I think this is the strongest argument in favor of not switching to the prerelease schedule; once we start prereleasing, we're basically locking ourselves to a promise of a 1.0 timeline.

I'm optimistic about hitting 1.0, but also I haven't been involved for very long at all; realistically, we should have a well-defined "1.0 by DATE" target and if we don't hit all the planned changes then whatever is on master by DATE is what becomes 1.0. Until that target plan is locked down, I think it's fair to want to keep the 0.x numbering to be safe.

Kodiologist:
That is a decent point about notes in the documentation and about how three versions of the documentation "doesn't really incur much overhead for us… Since the 0.x documentation will just remain fixed as it is.".

> realistically, we should have a well-defined "1.0 by DATE" target and if we don't hit all the planned changes then whatever is on master by DATE is what becomes 1.0.

I intend to make a real effort to make 1.0 happen. Still, we're all volunteers, and we have full-time jobs, or school in your case. I'd rather never release 1.0 than name a release 1.0 when I know that more breaking changes are right around the corner. The good news is that there is no limit to the number of alphas we release.

Kodiologist:
At any rate, consensus is pretty clearly against me, so alpha releases, or at least this particular alpha release, are back on. I'm going to unyank the release, add back the tag, and then add some notes to the README etc. about alphas.

Kodiologist:
Done.

allison-casey:
I appreciate you trying an alpha release cycle Kodi. I genuinely believe that this will have us doing better by our users and answer one of the more common complaints about Hy. Maybe we can label `0.20` in the docs as "legacy" instead of "stable" to telegraph what it is better? I'll be working through the major docs overhaul after i finish the compiler refactor/annotations and we can discuss that more in there.

PR #2036, opened 2021-04-15
Remove aliases of Python standard functions

Kodiologist:


allison-casey:
And so the core overhaul has begun. We might want to keep track of exactly what we're removing so that we can migrate it over to SkyHy

Kodiologist:
I guess we could, but we're already making people explicitly import these functions, so we might as well have people import from Python standard libraries from the start instead of an intermediate step of importing from SkyHy. At any rate, for functions not already implemented elsewhere, my plan is to add them to SkyHy before deleting them from core.

PR #2039, opened 2021-04-18
Fix: hystartup affects tests

### LINKED ISSUE #2029: HYSTARTUP affects tests
Errors in `tests/test_bin.hy` if HYSTARTUP is set, and even more errors if it loads a valid `.hy` file that sets the different startup options.

Separately, it looks like there aren't any tests for the HYSTARTUP feature itself? Or at least none that I could find with a quick grep.

allison-casey:
closes #2029 also fixes macros not being able to find the hystartup module causing it to error out.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -10,6 +10,9 @@ NATIVE_TESTS = os.path.join("", "tests", "native_tests", "")
 
 _fspath_pyimport = py.path.local.pyimport
 
+# https://github.com/hylang/hy/issues/2029
+os.environ.pop("HYSTARTUP", None)
+
 
 def pytest_ignore_collect(path, config):
     return (("py3_8_only" in path.basename and not PY3_8) or None)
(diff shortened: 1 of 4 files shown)


Kodiologist:
The changes sound good, but the commit message doesn't seem to match the changes. I think that `os.environ.pop("HYSTARTUP", None)` is the only line related to the commit message. Also, you can write that as `del os.environ['HYSTARTUP']`, right?

allison-casey:
`del os.environ ['HYSTARTUP']` will `KeyError` if the key doesn't exist so the `pop` is easier than an if guard. 

Changed the message to be more generally about adding `HYSTARTUP` to tests and moved the fix for macros failing to expand in the `HYSTARTUP` module. which i forgot to mention i found while adding unit tests.

Kodiologist:
> `del os.environ ['HYSTARTUP']` will `KeyError` if the key doesn't exist

Huh, I learn a lot of things that are dumb about Python's semantics by working on Hy, but that's a new one.

Kodiologist (approved):

Issue #2014, opened 2021-03-23
Lexing bug


jeongsoolee09:
I have a function here:

```clojure
(defn repl []
  (import hw2)
  (import [importlib [reload]])
  (reload hw2)
  (import [hw2 [*]]))
```

Evaluating above in the REPL emits a `hy.lex.exceptions.PrematureEndOfInput: Premature end of input`.

jeongsoolee09:
Oops, `(import [package [*]])` is only allowed at toplevel. My bad.

Kodiologist:
I'm guessing this is another manifestation of #2004.

PR #2040, opened 2021-04-19
Properly handle/display SyntaxError in Hy REPL

### LINKED ISSUE #2004: Incorrect errors displayed in HyREPL
```hy
hy 0.20.0+89.g33ff26a3.dirty using CPython(default) 3.9.2 on Linux

=> (defn a [a a])
  File "/usr/lib/python3.9/codeop.py", line 143, in __call__
    codeob = compile(source, filename, symbol, self.flags, True)
  File "stdin-c6000d2ccd1df263b2bd313815bfae618dc57cf3", line 1
SyntaxError: duplicate argument 'a' in function definition
None

=> ; but...

=> (defn a [/])
  File "stdin-c4590d5b379ab946ccd4a96ce4e897f3dddb0494", line 1, in <module>
    (defn a [/])
  File "<stdin>", line 1
    (defn a [/])
            ^
hy.errors.HySyntaxError: positional only delimiter '/' must have an argument

=> (defn a [a a])  ; wrong error is printed below!
  File "<stdin>", line 1
    (defn a [/])
            ^
hy.errors.HySyntaxError: positional only delimiter '/' must have an argument
```

I think the error handling in the HyREPL will need a little refactoring to catch this one. The syntax error for duplicate arguments (and possibly other errors) somehow skips our try/except handling that catches most other syntax errors and simplifies stack traces, so when it gets printed it's using stale error info.

scauligi:
Fixes #2004 (and #2014).

The Hy REPL saves exceptions locally so that it can remove internal stack frames.
SyntaxErrors weren't getting saved, and so the REPL would instead
should the previous saved error.
This is now fixed.

Also removes `reraise()` from `hy._compat` since we're no longer supporting Python 2,
which simplifies some of the exception handling code throughout.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -18,6 +18,10 @@ Removals
   * From `functools`: `reduce`
   * From `fractions`: `Fraction` (formerly `fraction`)
 
+Bug Fixes
+------------------------------
+* REPL now properly displays SyntaxErrors.
+
 1.0a1 (released 2021-04-12)
 ==============================
 
--- a/hy/_compat.py
+++ b/hy/_compat.py
@@ -10,13 +10,6 @@ PY3_9 = sys.version_info >= (3, 9)
 PY3_10 = sys.version_info >= (3, 10)
 
 
-def reraise(exc_type, value, traceback=None):
-    try:
-        raise value.with_traceback(traceback)
-    finally:
-        traceback = None
-
-
 if PY3_9:
 
     import ast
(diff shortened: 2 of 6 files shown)


Kodiologist:
Nice. Try adding a test to `test_bin`, if it's easy. If it turns out to be really involved, don't worry about it.

scauligi:
@Kodiologist In addition to the other tests?

Kodiologist:
Oh yeah, there already are tests. I guess I'm blind.

Kodiologist (approved):

PR #2041, opened 2021-04-19
Refactor/compiler cleanup

allison-casey:
closes #1399 This doesn't cover everything, (namely I think that there needs to be another pass against the core logic), but I'd rather get all the quality of life improvements in this than do all the bikeshedding around that kind of deep refactor right now.

### DIFF
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -30,7 +30,7 @@ from hy.lex import hy_parse, mangle
 from contextlib import contextmanager
 from hy.lex.exceptions import PrematureEndOfInput
 from hy.compiler import (HyASTCompiler, hy_eval, hy_compile,
-                         hy_ast_compile_flags)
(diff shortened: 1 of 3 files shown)


Kodiologist:
Overall, I'm not a fan. These changes add needless temporary variables, move parentheses to hang out on their own line when grouping is already made clear by indentation, and change comments to string literals in contexts where docstrings aren't actually allowed. Removing unused imports is useful, and I'll take your word for it that the type annotations will become useful, but the rest we're better off without.

Kodiologist:
I forgot to mention: I think that adding detailed documentation for internal methods and helper functions of a class that's not supposed to be user-visible is a mistake. We don't want to maintain that.

allison-casey:
Pulling this, we can argue about syntax after 1.0

Kodiologist:
Fair enough.

PR #2042, opened 2021-04-20
Remove Toolz reimplementations

Kodiologist:


allison-casey:
if `by2s` is going to be imported into other modules, it should have a docstring imo. Otherwise looks good to me

Kodiologist:
Okay, I figured it didn't need one because it's only for internal use, but it's no trouble. Done.

allison-casey:
doing the full autodoc `args` `returns` etc would be overkill, but it's good practice to have something for those methods. 
how is the unclosed bracket string `  #[[Returns the given iterable in pairs.` not a syntax error btw? It works, but it seems like it shouldn't

allison-casey (dismissed):

Kodiologist:
Oops, you're right. I forgot to check the whole docs directory. This PR was a surprising amount of work.

allison-casey:
I can go through and fix the docs for you. `by2s` won't work since it doesn't support `fillvalue` and will drop the last entry if there's an odd number, but we can just use the `itertools.grouper` recipe for that

Kodiologist:
That would be convenient. Thanks.

allison-casey:
Docs are fixed for this and the itertools alias removals

allison-casey (approved):

PR #2050, opened 2021-05-02
Document the function return annotation

etanol:
There is one use case missing from the annotation documentation (`^`).  For example, the following Python definition:

``` python
def add1 (x: int) -> int:
    return x + 1
```

Can be translated to Hy as follows:

``` hy
(defn add1 ^int [^int x]
  (+ x 1))
```

The placement of the return type annotation does not conform to the documented rules, so it can be difficult to find or even confusing for users.  Given that function return type annotation is a very common use case, I think it should be more directly exposed through the documentation.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -39,6 +39,10 @@ Special Forms
       ; Equivalent to: def func(a: int, b: str = None, c: int = 1): ...
       (defn func [^int a ^str [b None] ^int [c 1]] ...)
 
+      ; Function return annotations have different ordering
+      (defn add1 ^int [^int x] (+ x 1))
+      (fn ^int [^int y] (+ y 2))
+
    The rules are:
 
    - The value to annotate with is the value that immediately follows the caret.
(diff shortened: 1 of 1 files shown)


Kodiologist:
Can you be a little more explicit? In between what? I would suggest wording like "it comes immediately before the lambda list (and after the name of the function, if there is one)".

scauligi:
Alternatively, should we move the return annotation to before the function name? It feels more consistent that way.

etanol:
> Alternatively, should we move the return annotation to before the function name? It feels more consistent that way.

For `defn` I don't see how, since it's a macro:

``` hy
(defn add1 ^int [^int x] (+ x 1))
;; Approximately expands to
(setv add1 (fn* ^int [^int x] (+ x 1)))
```

And the `fn*` (or `fn`) special forms don't have anywhere else to place the return annotation.  I think trying to make the annotation placement more consistent incurs into significant added complexity for the compiler.

etanol:
> Can you be a little more explicit? In between what? I would suggest wording like "it comes immediately before the lambda list (and after the name of the function, if there is one)".

I agree.  It sounds incomplete.  I have rephrased it.

scauligi:
#1992 proposes upgrading `defn` to a special form, we could do it at that time (but this should probably be saved as a discussion for that eventual PR).

In the meantime, the new doc text looks good to me 👍

scauligi (approved):

Kodiologist:
Thanks. In the future, please make a feature branch instead of requesting a pull from your own `master`.

PR #2033, opened 2021-04-06
Register `hy.models.FString` with `hy-repr` + tests.

peaceamongworlds:
Currently `hy-repr` treats f-strings as a sequence, which produces unexpected results:
```hy
=> (hy-repr 'f"a{:a}")
'\'(, "a" (, :a))'
```
This makes it so that 
```hy
=> (hy-repr 'f"a{:a}")
'\'f"a{:a}"'
```

Issue #1873, opened 2020-02-17
hy-repr probably shouldn't use the registered function of a supertype


Kodiologist:
If you make a new class `C` that inherits from a class `B` with a registered `hy-repr` function `f-B`, but don't register a function for `C`, then `hy-repr` on an instance of `C` will call `f-B`, which can be misleading: it looks like you have an instance of `B` when actually you have an instance of `C`. We should probably fall back on `repr` instead.

For example, pandas's `Timestamp` class inherits from `datetime`, so `hy-repr` on a timestamp will give you the impression that you have a plain `datetime`.

Removing this feature will mean that `hy-repr` on a `namedtuple` will no longer just work, but that's probably an okay trade-off.

Issue #1774, opened 2019-05-01
Packaging hy code with zipapp does not work


AlexanderWingard:
I'm writing this ticket in response to this [How to load Hy package in zipapp?](https://stackoverflow.com/questions/54322331/how-to-load-hy-package-in-zipapp) stackoverflow question.

I have tried reproducing this on the latest master a38cdb7ccc08b832a17bb54425baf21c9147ba03 using the files in this [gist](https://gist.github.com/AlexanderWingard/d82f00e4089ba9c760b9ae1d6c3561b1)

It works fine when not using zipapp:
```
$ python3 hyz
hello hy
```
But after zipping:
```
$ python3 -m zipapp hyz
$ python3 hyz.pyz 
Traceback (most recent call last):
  File "/home/lex/anaconda3/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/lex/anaconda3/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "hyz.pyz/__main__.py", line 2, in <module>
ModuleNotFoundError: No module named 'hyz'
```

Kodiologist:
As a workaround, you can translate Hy files to Python first with `hy2py`.

AlexanderWingard:
Ah, I was experimenting with using hyc to produce pyc files and put them inside the zip but that is not the right approach?

Kodiologist:
I don't know. You can try it. Python has had ambivalent support for running bytecode without a corresponding source file.

AlexanderWingard:
Using hy2py worked for my code, however, I really wanted to use the same trick to bring in hy.contrib.hy-repr to be able to print hy data structures nicely but did not work well for those modules.

Kodiologist:
What do you mean, "did not work well"?

AlexanderWingard:
Sorry, actually the problem was not in hy-repr but actually from it importing hy.core.language, so when trying to hy2py that file it ends up in some import loop:
```
Traceback (most recent call last):
  File "/home/lex/anaconda3/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/lex/anaconda3/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "./tmp2.pyz/__main__.py", line 2, in <module>
  File "./tmp2.pyz/hello.py", line 3, in <module>
[...]

PR #2053, opened 2021-05-05
Fix `repr` and `hy-repr` for f-strings

Kodiologist:
Supersedes #2033.


allison-casey:
Nice! Is the only thing left for `hy-repr` #1873 and #1774?

Kodiologist:
No, I also have some changes to make left over from #1550.

allison-casey:
i'll take a stab at #1873 then. I'll get a review for this in after work

hy/models.py
```diff
@@ -339,13 +341,26 @@ def replace(self, other, recursive=True):
...
+            ', conversion=' + repr(self.conversion))
+
 class FString(Sequence):
     """
     Generic Hy F-String object, for smarter f-string handling.
     Mimics ast.JoinedStr, but using String and FComponent.
     """
     def __new__(cls, s=None, brackets=None):
-        value = super().__new__(cls, s)
+        value = super().__new__(cls,
```
allison-casey, line 355: this is overly complicated and can be simplified and made more explicit with something like
```python
import operator
value = super().__new__(cls,
  # Join adjacent string nodes for the sake of equality
  # testing.
      tuple(
          reduce(operator.add, components) if is_string else next(components)
          for is_string, components in groupby(s, lambda x: isinstance(x, String))
      )
      if s else s)
```

allison-casey: Also `Sequence` is a child of `tuple` and can take a generator so the call to `tuple` can be removed and left as a generator to make 

```python
(
    reduce(operator.add, components) if is_string else next(components)
    for is_string, components in groupby(s, lambda x: isinstance(x, String))
)
```

Kodiologist: Smart using `next` like that. Done, except I kept the short variable names for the sake of line length.

allison-casey: please use the descriptive names please, it doesn't break 90 characters and it'll make this so much easier to parse and understand for the next person who has to look at this.

Kodiologist: I think 90 characters is a chonker of a line. I try to keep lines to like 70 characters past the indent. Anyway, I restored the names and added some more spacing. I hate Python's inside-out ternary operator.

Kodiologist: Good catch, thanks. Done.

allison-casey (approved):

PR #2058, opened 2021-05-09
Add: let hy cmdline args overwrite HYSTARTUP values

allison-casey:
This lets `spy` or `repl-output-fn` values set at the cmdline take precedence over those set in `HYSTARTUP`. Note though that if you set `spy` to `True` in `HYSTARTUP` there's no way to set it back to `False` since `--spy` is a truthy check and changing the cmdline args to allow it would be overkill for something you'd only really do to debug anyways.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -44,6 +44,8 @@ Bug Fixes
 * Fixed a bug in `pprint` in which `width` was ignored.
 * Corrected `repr` and `hy-repr` for f-strings.
 * `hy-repr` won't use the registerd method of a supertype anymore
+* REPL `--spy` and `--repl-output-fn` can now overwrite `HYSTARTUP` values
+  when specified as a cmdline argument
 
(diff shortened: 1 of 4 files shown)


Kodiologist:
There doesn't seem to be a test of the change to `spy`, only `output_fn`. Is that intentional?

I don't think you need `startup_spy` or `startup_output_fn`; it should suffice to just set the object attributes (or the local variables) with appropriate checks to get the desired order of precedence. I also think you don't need the `or False` in `spy or startup_spy or False`.

Add a NEWS line. I'd consider this a bugfix.

allison-casey:
Added a test

you were correct, changed 

added a NEWS item

Kodiologist:
`output_fn = output_fn`? Also, I'm not clear if the change to the default argument has any effect.

allison-casey:
don't know how missed that, fixed both

Kodiologist (approved):

PR #2059, opened 2021-05-09
Fix: hy-repr won't use the registered function of a supertype

### LINKED ISSUE #1873: hy-repr probably shouldn't use the registered function of a supertype

allison-casey:
closes #1873 namedtuple's are handled by moving the `_fields` check from the `tuple` repr function to `_base-repr` which seems to work just fine

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -43,6 +43,7 @@ Bug Fixes
 * REPL now properly displays SyntaxErrors.
 * Fixed a bug in `pprint` in which `width` was ignored.
 * Corrected `repr` and `hy-repr` for f-strings.
+* `hy-repr` won't use the registerd method of a supertype anymore
 
 .. _Toolz: https://toolz.readthedocs.io
 .. _CyToolz: https://github.com/pytoolz/cytoolz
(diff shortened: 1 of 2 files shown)


Kodiologist:
I think `(if (in (type obj) _registry) (get _registry (type obj)) [_base-repr None])` can be written `(.get _registry (type obj) [_base-repr None])`.

I think treating any unregistered object as a named tuple if it has `_fields` is a little dangerous because that's not really a reserved name. How about adding an `(isinstance x tuple)` check as well? Be careful with the wording and placement of the comment, by the way; it still says "It's a named tuple" although you've put it outside the `when` form.

allison-casey:
Done and Done

Kodiologist (approved):
Nice, thanks.

allison-casey:
i forget can i merge this now?

Kodiologist:
Nope, it needs to be either three days old or get another approval. The rules are spelled out exactly in CONTRIBUTING.

Issue #2061, opened 2021-05-10
Python 3.10: required field "lineno" missing from alias


tianon:
Python 3.10 has introduced a change that breaks Hy (feeling like a case of "This Is The Way" with Python pre-releases :trollface:).

It seems to be related to https://bugs.python.org/issue43798 (some potentially useful discussion in https://github.com/pytest-dev/pytest/issues/8539).

Here's the full backtrace (on latest master; specifically e1150cf08106a727a1a3024044710576803a8292):

```console
$ python --version
Python 3.10.0b1
$ hy
Traceback (most recent call last):
  File "/opt/hylang/hy/hy/compiler.py", line 1648, in compile_eval_and_compile
    hy_eval(new_expr + body,
  File "/opt/hylang/hy/hy/compiler.py", line 1977, in hy_eval
    eval(ast_compile(_ast, filename, "exec"),
  File "/opt/hylang/hy/hy/compiler.py", line 62, in ast_compile
    return compile(a, filename, mode, hy_ast_compile_flags)
TypeError: required field "lineno" missing from alias

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/hy", line 8, in <module>
    from hy.cmdline import hy_main
  File "/opt/hylang/hy/hy/__init__.py", line 14, in <module>
    hy.importer._inject_builtins()
  File "/opt/hylang/hy/hy/importer.py", line 178, in _inject_builtins
    core = importlib.import_module('hy.core')
  File "/usr/local/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/opt/hylang/hy/hy/core/__init__.py", line 1, in <module>
    from .shadow import *
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 879, in exec_module
  File "<frozen importlib._bootstrap_external>", line 1017, in get_code
  File "/opt/hylang/hy/hy/importer.py", line 130, in _hy_source_to_code
    data = hy_compile(hy_tree, module)
  File "/opt/hylang/hy/hy/compiler.py", line 2066, in hy_compile
[...]

PR #2062, opened 2021-05-10
Add initial GitHub Actions testing

tianon:
This mirrors `.travis.yml` fairly closely, but gives us a much newer 3.10-dev version (specifically 3.10.0-beta.1, at this time of writing, which is relevant for #2061).

Per #2061, I've set 3.10-dev to "allow failure" (since it's known-broken ATM), and I've currently got PyPy set similarly since it fails for reasons I wasn't able to chase down completely (that I think are due to how https://github.com/actions/setup-python is installing/configuring PyPy, but I'm not 100% certain).

You can see a "successful" run of this at https://github.com/tianon/hy/actions/runs/829829568. :smile:

As you can see there, it takes ~2m to run this entire matrix, and it triggered almost instantly every time I force pushed.

Relevant links:
- https://docs.github.com/en/actions/guides/building-and-testing-python
- https://github.com/actions/setup-python
- https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,55 @@
+name: Run Tests
+
+on:
+  pull_request:
+  push:
+
+jobs:
+
+  test:
+
+    strategy:
+      matrix:
+        python:
+          - '3.6'
+          - '3.7'
+          - '3.8'
+          - '3.9'
+        allow-failure: [ false ]
+        include:
+          # https://github.com/hylang/hy/issues/2061
+          - python: '3.10-dev'
+            allow-failure: true
+          # TODO figure out why the PyPy installed by https://github.com/actions/setup-python fails :/
+          - python: 'pypy-3.7'
+            allow-failure: true
+
+    name: ${{ matrix.python }}
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Kodiologist:
I like the sound of this. Would it be too much to ask you to fix the 3.10 bug and get PyPy working again as part of this PR? If so, I guess I'll try instead.

Kodiologist:
I guess so. I'll take a shot.

tianon:
Sorry, I've been trying to get back to this and haven't had a minute (and frankly fixing the 3.10 issue feels a bit over my head). :grimacing:

allison-casey:
looks like we weren't constructing an `alias` via `asty.alias` so `lineno` and `col_offset` weren't set like 3.10 expects now. also pypy works for me so maybe the changes fixed both? Anyways how do we want to do this? Merge this then merge merge my fix? I don't have github actions running on my fork atm so that might be easiest

Kodiologist:
I'm still working on this. It's mostly done. I'll make a new PR when it's ready.

Kodiologist:
Superseded.

PR #2067, opened 2021-05-16
Make `hy-repr` the default REPL output function

Kodiologist:
This is basically a reboot of #1550. `hy-repr` is now called as `hy.repr`, parallel to `hy.eval`, and it's the default output function for the REPL.

I haven't yet updated any of the documentation or NEWS. Let's make sure we're happy with the code first.


Kodiologist:
Since there are no objections, I'll work on the documentation.

Kodiologist:
Okay, done. @allison-casey This is technically clear to merge, but I want to double-check that merging is okay with you, because I just pushed a lot of documentation edits and you're our Queen of Docs.

allison-casey:
:crown: :bee:

I think we should create a `Hy` heading above `Core` in `Api` since it doesn't make sense anymore for `hy-repr` to be under the same headings as the other contributor modules. I can do that if you don't mind me editing this pr

Kodiologist:
Sure. Go ahead and merge when you're done.

allison-casey:
so currently the methods `['read', 'read_str', 'mangle', 'unmangle', 'eval', 'repr', 'repr_register']` are exposed under the root `hy` module, but all of them _except_ `repr` and `repr_register` are also injected into `builtins` currently. while i'm here can i remove those from `builtins` injection so that the canonical way to access them is through the root `hy` module?

Kodiologist:
You don't think that would be cleaner to do in a separate PR? This one is big already.

allison-casey:
fair enough, i'll finish this one up with just repr and repr-register and open a new one for the rest.

PR #2070, opened 2021-05-21
Fix: Manual is missing module names

### LINKED ISSUE #2068: The manual is missing module names
It looks like the big doc reorganization ended up omitting the actual module names of the various contrib and extra modules; for example, the section named "Walk" documents the functions and macros in `hy.contrib.walk`, but doesn't mention the name `hy.contrib.walk` or otherwise indicate how to bring the enumerated names into scope.

Pointed out in https://github.com/hylang/hy/issues/2065#issuecomment-842377526.

allison-casey:
Closes #2068 

* Added descriptions to Core and Additional modules that explain
  importing/requiring non core methods/macros
* pinned sphinx to 3.5.4 as hydomain doesn't support > 4.0 yet

made some changes to hydomain so that tag macros show up correctly and 
macros now are prefixed with `macro` so that you know that you need to `require` them
instead of `import` them at a glance

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1098,7 +1098,7 @@ Special Forms
        At end of outer loop
 
 .. hy:function:: (with [#* args])
-                   
+
   Wrap execution of `body` within a context manager given as bracket `args`.
   ``with`` is used to wrap the execution of a block within a context manager. The
   context manager can then set up the local system and tear it down in a controlled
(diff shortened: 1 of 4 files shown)


Kodiologist:
Seems good, except I would omit the note "Hy will be migrating these methods and macros do a dedicated companion library…". The manual only needs to document Hy as it is, rather than discussing plans for the future; plans for the future belong in the issues list.

allison-casey:
valid, removed

Kodiologist (approved):

PR #2073, opened 2021-05-25
Clean: make 'calling-module' & 'calling-module-name' internal to hy

allison-casey:
I didn't end up making `calling-module` private because `contrib.walk` uses it, but i did remove it from autodocs and builtins injection.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -43,6 +43,7 @@ Other Breaking Changes
 * `gensym`, `macroexpand`, `macroexpand-1`, and `disassemble` are no longer
   automatically brought into scope. Call them as `hy.gensym`, `hy.macroexpand`, etc
   or import them explicitly.
+* made `calling-module-name` private and `calling-module` internal to Hy
 
 New Features
 ------------------------------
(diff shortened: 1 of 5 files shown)


Kodiologist (approved):
Looks good. Just update NEWS.

allison-casey:
also i think we should make `cond` work like macro `if`. The implicit `do` is rarely useful in a way that makes the per branch brackets worth it.

Kodiologist:
I'm not sure why we're talking about that here, but: don't you think the brackets are nice for readability? I think it will get better results from automatic indenters, at least.

allison-casey:
just noticed it while converting over to `cond` and nested `if`'s. And no, not really. I find it becomes much noisier than the bracketless version macro `if` provided. I completely stopped using `cond` once i realized `if` did the same thing without the brackets. anyways, we can figure that out when we port over all that stuff to SkyHy and do a refactor pass anyways.

scauligi (approved):
Heyyyyy long time no see folks!
Gripes about `cond` aside, lgtm.

allison-casey:
Eyyyyy, scauligi!! Good to have you back 😊

PR #2074, opened 2021-05-26
Refactor/yeet macro error

allison-casey:
easy enough. Also looks like macro error was raising expansion errors with filename and expression backwards :/ 
`(hy.errors.HyMacroExpansionError ~reason ~filename ~expression None)`
and the init of `HyMacroExpansionError` is
```python
    def __init__(self, message, expression=None, filename=None, source=None,
                 lineno=1, colno=1):
```

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -44,6 +44,7 @@ Other Breaking Changes
   automatically brought into scope. Call them as `hy.gensym`, `hy.macroexpand`, etc
   or import them explicitly.
 * made `calling-module-name` private and `calling-module` internal to Hy
+* `macro-error` has been removed, raise typical errors instead
 
 New Features
 ------------------------------
--- a/hy/contrib/destructure.hy
+++ b/hy/contrib/destructure.hy
@@ -392,7 +392,7 @@ Iterator patterns are specified using round brackets. They are the same as list 
 (defmacro let+ [args #* body]
   "let macro with full destructuring with `args`"
   (if (odd? (len args))
-    (macro-error args "let bindings must be paired"))
+      (raise (ValueError "let bindings must be paired")))
   `(let ~(lfor [bs expr] (by2s args)
                sym (destructure bs expr)
            sym)
--- a/hy/contrib/loop.hy
+++ b/hy/contrib/loop.hy
@@ -69,7 +69,7 @@ tail-call optimization (TCO) in their Hy code.
 
 (defmacro defnr [name lambda-list #* body]
   (if (not (= (type name) hy.models.Symbol))
-    (macro-error name "defnr takes a name as first argument"))
+      (raise (TypeError "defnr takes a name as first argument")))
   `(do (require hy.contrib.loop)
        (setv ~name (hy.contrib.loop.fnr ~lambda-list ~@body))))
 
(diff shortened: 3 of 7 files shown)


Kodiologist:
Is it necessary to raise `hy.errors.HyMacroExpansionError` explicitly? I thought you could raise an ordinary error type like `ValueError` or `TypeError` and the macro-expansion process would handle it correctly.

allison-casey:
huh, apparently not :woman_shrugging: the more you know. i'll update this pr with ordinary errors

allison-casey:
Done

And letting macroexpanion handle the errros makes the work better lol. Looks like this was a win win for everyone

Kodiologist (approved):

scauligi:
Tracebacks from ordinary exception types during macro expansion are showing internal frames. I suspect the responsible code is part of the error handling spaghetting in `hy/cmdline.py`.

Using `macro-error`:
```
  File "/usr/lib/python3.9/runpy.py", line 267, in run_path
    code, fname = _get_code_from_file(run_name, path_name)
  File "<frozen importlib._bootstrap_external>", line 923, in get_code
  File "/home/scauligi/workspace/hy/hy-git/x.hy", line 1, in bad_macro
    (defmacro bad-macro [] (macro-error "test" "test"))
hy.errors.HyMacroExpansionError: test
```

Using `(raise (ValueError ...))`:
```
  File "/usr/lib/python3.9/runpy.py", line 267, in run_path
    code, fname = _get_code_from_file(run_name, path_name)
  File "<frozen importlib._bootstrap_external>", line 923, in get_code
  File "/usr/lib/python3.9/contextlib.py", line 135, in __exit__
    self.gen.throw(type, value, traceback)
  File "/home/scauligi/workspace/hy/hy-git/x.hy", line 3
    (bad-macro)
    ^---------^
expanding macro bad-macro
  ValueError: test
```

allison-casey:
so it looks like this is happening because `macroexceptions` wraps it in a macroexpansion error its adding its own stack frame to the stack before reraising the wrapped error. But since the error is technically coming from `contextlib` and not a hy internal module can't get filtered out and i'm not sure how we can change that. I'll keep digging though

allison-casey:
so it don't like it, but one solution that works is to copy the definition of `_GeneratorContextManager` from `contextlib` and create our own `contextmanager` decorator so that the exception gets raised from an internal module and gets filtered out properly. I guess this means if we can pair down the custom context manager to just whats necessary it wouldn't be too much of an addition. how do we feel about this?

allison-casey:
so it don't like it, but one solution that works is to copy the definition of `_GeneratorContextManager` from `contextlib` and create our own `contextmanager` decorator so that the exception gets raised from an internal module and gets filtered out properly. I guess this means if we can pair down the custom context manager to just whats necessary it wouldn't be too much of an addition. how do we feel about this? 

Edit: actually ended up being a lot easier than this, i pushed a fix

Kodiologist:
Seems fine to me. Maybe update the commit message for the first commit.

allison-casey:
Done and added a note in the ctx manager docstring so we don't forget why we're not using a regular ``contextmanager`` decorator

scauligi (approved):
This is awesome!

PR #2077, opened 2021-05-27
bootUnstrap

allison-casey:
Reorganize bootstrap macros as special operators

* `if*` -> `if`
* `fn*` -> `defn`
* `annotate*` -> `annotate`
* `defn/a` now special form
* bootstrap module deleted


hy/compiler.py
```diff
@@ -1614,33 +1641,39 @@ def compile_class_expression(self, expr, root, name, rest):
     def _rewire_init(self, expr):
         "Given a (setv …) form, append None to definitions of __init__."
```
scauligi, line 1680: Pedantic nitpick on docstring: This function handles more than just `setv` forms now

scauligi (commented):
(I don't know if this was discussed elsewhere, I've been gone a while...)
I'd prefer keeping `if` as-is syntax-wise allowing multiple conditions, and instead getting rid of `cond`, since `cond` then is just sugar for `if` + `do`.

scauligi:
Oh can we also do https://github.com/hylang/hy/pull/2050#issuecomment-830833239 since we're upgrading `defn`? It seems more consistent to move the return value annotation before the function name, mirroring the parameter annotations.

allison-casey:
I'd be fine with either making cond act more like clojure cond and keeping simplified if or do what you said. The do sugar just doesn't feel that useful

allison-casey:
I hadn't seen #2050. Let me mess around with the defn special form and see how I feel about it/if things explode. First impressions though do seem like it makes things more consistent

allison-casey:
also noticing that attempting to type annotate a lambda in hy creates technically executable, but syntactically illegal python code. 
```clojure
=> ((fn [^int a] a) 1)
(lambda a: int: a)(1)
1
```
```python
In [7]: (lambda a: int: a)(1)
  File "<ipython-input-7-13846469c0f2>", line 1
    (lambda a: int: a)(1)
                  ^
SyntaxError: invalid syntax
```

probably need to force function defs for anything with type annotations while we're here

scauligi:
It might be easier to disallow annotations for lambda parameters, since Python made an explicit choice to disallow them.

Otherwise we can do something like
```
_hy_tmp = lambda ...
_hy_tmp.__annotations__ = ...
```
when producing hy2py output so that it's valid Python syntax

allison-casey:
this seems to work just fine and doesn't break any of the tests.

```python
has_annotations = returns is not None or any(arg.annotation
        for arg in args.args + args.posonlyargs + args.kwonlyargs + [args.vararg, args.kwarg]
        if arg is not None)
if not force_functiondef and not body.stmts and not has_annotations:
    return ret + asty.Lambda(expr, args=args, body=body.force_expr)
```

how do we feel about this?

Kodiologist:
I think that's the right strategy, yeah. It might be smart to add a test in `test_hy2py` for this case.

allison-casey:
Done

hy/compiler.py
allison-casey: Fixed

Kodiologist:
It looks like the compiler method for `defn` has a lot of overlap with the one for `fn` and `fn/a` Can they be combined? Also, it's confusing that `compile_function_defn` is the name of the one that isn't used for `defn`.

allison-casey:
how about this?
renamed `compile_function_def` -> `compile_function_lambda` & `compile_function_defn` -> `compile_function_def`
and then moved a bunch of the ast node creation logic out into a helper methods. I like having `fn` and `defn` as separate special forms. it gives us more control over the parsing, less deep if nesting, and makes it clear the differences between generating a `defn` and an `fn`

allison-casey:
> Oh can we also do [#2050 (comment)](https://github.com/hylang/hy/pull/2050#issuecomment-830833239) since we're upgrading `defn`? It seems more consistent to move the return value annotation before the function name, mirroring the parameter annotations.

this is actually really easy it turns out now that `defn`'s a special form
```clojure
@special(["defn", "defn/a"], [OPTIONAL_ANNOTATION, SYM, lambda_list, many(FORM)])
def compile_function_def(self, expr, root, returns, name, params, body):
```
I kinda like it too. It makes it more consistent and function defs make more sense to read when folded if you have the lambda-list 
on the next line
`(defn ^int hello-world ...)`
instead of 
`(defn hello-world ^int ...)`

Kodiologist:
> how about this?
renamed `compile_function_def` -> `compile_function_lambda` & `compile_function_defn` -> `compile_function_def`
and then moved a bunch of the ast node creation logic out into a helper methods.

I still see a lot of duplicated code between the two methods, which can be moved into the common subroutine.

allison-casey:
alright this brough some more into the helper function. Dont think there's anything else that can be combined

Kodiologist:
I think the commit message "Fix: force function def on fn param annotations" is misplaced. This commit doesn't make any compiler changes. It only adds tests for that change.

Regarding `compile_macro_def`:

- I think `SYM | FORM` is redundant because `FORM` is a superset of `SYM`. Anyway, `FORM` is probably too liberal here; contrast with `argument` in `lambda_list`.
- `node` is unused.
- `ret` doesn't need to be initialized at the beginning.
- `if type(x) == Expression and x[0] == Symbol("unpack-mapping")` looks like it will crash if `x` is the empty expression. This might be preempted if you narrow the model pattern.
- It looks like most of the new model tree being passed to `self.compile` isn't getting any postion information. Is that a mistake? A `replace` or two may be called for.

allison-casey:
* changed it to parse it as a normal lambda_list and then deny kwonly and kwargs
* did a `.replace` on the defmacro expression
* removed node
* moved ret further down

Kodiologist:
Excellent. You have a leftover `ret = Result()` at the top of `compile_macro_def`, and I would see if you can avoid using all those literal integer indices in `compile_macro_def` and `compile_function_lambda`. Making programmers count is rarely a good idea. I think we should document `annotate`, and add it to `NEWS`. That should be all.

allison-casey:
> Excellent. You have a leftover `ret = Result()` at the top of `compile_macro_def`

Done

>  and I would see if you can avoid using all those literal integer indices in `compile_macro_def` and `compile_function_lambda`. Making programmers count is rarely a good idea.

Done

>  I think we should document `annotate`, and add it to `NEWS`. That should be all.

Done

Kodiologist (approved):
I think you're saying a bit more than you need to in NEWS; e.g., `fn*` was never documented, so its removal is not user-visible, and we don't provide enough support for macro introspection for the difference between special ops and builtin macros to mean much to users.

allison-casey:
made `NEWS` a little more appropriate. 

@scauligi you had some notes on this and #2074, have I addressed all of them?

scauligi:
Per https://github.com/hylang/hy/issues/1992#issuecomment-851063006 sounds like you'll do the annotation change to `defn` in a separate PR, so otherwise lgtm here.

I still would prefer we have multi-case `if` and get rid of `cond`, but that doesn't necessarily have to be in this PR.
(Although if it _were_ in the PR then `Fix: Update old references to 'if*' and macro 'if'` could be simplified to just do  `s/if*/if/`... although I guess you'd  have to do the reverse conversion from `cond` to `if` in that case)

scauligi (approved):

allison-casey:
> Per [#1992 (comment)](https://github.com/hylang/hy/issues/1992#issuecomment-851063006) sounds like you'll do the annotation change to `defn` in a separate PR, so otherwise lgtm here.
> 
> I still would prefer we have multi-case `if` and get rid of `cond`, but that doesn't necessarily have to be in this PR.
> (Although if it _were_ in the PR then `Fix: Update old references to 'if*' and macro 'if'` could be simplified to just do `s/if*/if/`... although I guess you'd have to do the reverse conversion from `cond` to `if` in that case)

yeah. i already have it ready to go so i'll open it right after this is merged. 
I think we should discuss what to do about `cond`/`if` if #1992 since i think we each have a slightly different preference we should iron out in a dedicated PR

PR #2081, opened 2021-05-31
Fix: dissallow python reserved words as parameters or func kwargs

### LINKED ISSUE #1765: A gotcha in keyword mangling
I was a bit blindsided today by a call to `requests.get` which looked like `(requests.get URL (dict :param1 1 :param2 2 :from 3))`. The results of the call were subtly wrong. It took me some digging to realize that instead of putting `&from=3` in my URL, I was putting `&hyx_from=3`, because `from` is a Python reserved word and hence needs to be mangled. In Python, `dict(from = 3)` is simply a syntax error, which is annoying, but doesn't lead to subtle bugs like the one I got here.

I can imagine a few different responses to this:

1. Live with it. Besides, I should've checked the produced URL carefully, because Web services tend not to produce an error for an unrecognized parameter, so any typo in a parameter name would've caused this sort of problem.
2. Don't use mangling to make Python reserved words legal identifiers. Instead, forbid such variable names, as in Python itself. It's cute that you can use `def` as an ordinary variable name and still produce valid Python code, but maybe it's not worth headaches like this.
3. Raise a warning in the specific case of using a reserved word as a keyword argument, because this is unlikely to do what the programmer is expecting.
4. As (3), but raise an error instead.

What do you guys think?

allison-casey:
closes #1765 

this one is actually pretty minimal and didn't break anything in any real way. One thing to note though is that while `(:from {"from" 1})` won't syntax error its still a key error since `Keyword.__call__` is defined as `return data[mangle(self.name)]` do we still think that's okay given this change?

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -49,6 +49,8 @@ Other Breaking Changes
 * `defn`, `defn/a` have been made a special forms
 * `annotate*` has been made public and renamed `annotate`
 * return annotation for `defn` has been moved to before the function name
+* Python reserved words can no longer be parameter names or
+  function call keywords
 
 New Features
 ------------------------------
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -269,6 +269,35 @@ Special Forms
            parameter-1 1
            parameter-2 2
 
+   .. _reserved_param_names:
+
+   .. note::
+
+      Parameter names cannot be Python reserved words nor can a function
+      be called with keyword arguments that are Python reserved words. This means
+      that the following will raise a `SyntaxError` as they would in Python::
+
+         (defn afunc [a b if])
+         Traceback (most recent call last):
+           File "<stdin>", line 1
+             (defn afunc [a b if])
+                              ^
+         hy.errors.HySyntaxError: parameter name cannot be Python reserved word
+
+         (dict :a 1 :from 2)
+         Traceback (most recent call last):
+           File "<stdin>", line 1
+             (dict :a 1 :from 2)
+                        ^
+         hy.errors.HySyntaxError: keyword argument cannot be Python reserved word
+
+      This only applies to parameter names and a keyword argument name. The value of
+      the parameter or keyword argument can still be a keyword of a reserved word::
+
+         => (defn test [a] a)
+         => (test :a :from)
+         :from
+
 .. hy:function:: (defn/a [name lambda-list #* body])
 
    Define `name` as a function with `lambda-list` signature and body `body`.
--- a/docs/language/syntax.rst
+++ b/docs/language/syntax.rst
@@ -127,7 +127,9 @@ the function ``f`` with the keyword argument named ``foo`` set to ``3``. Hence,
 trying to call a function on a literal keyword may fail: ``(f :foo)`` yields
 the error ``Keyword argument :foo needs a value``. To avoid this, you can quote
 the keyword, as in ``(f ':foo)``, or use it as the value of another keyword
-argument, as in ``(f :arg :foo)``.
+argument, as in ``(f :arg :foo)``. It is important to note that a keyword argument
+cannot be a Python reserved word. This will raise a ``SyntaxError`` similar to Python.
+See :ref:`defn <reserved_param_names>` for examples.
 
 Keywords can be called like functions as shorthand for ``get``. ``(:foo obj)``
 is equivalent to ``(get obj (hy.mangle "foo"))``. An optional ``default`` argument
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -30,6 +30,7 @@ import sys
 import copy
 import builtins
 import __future__
+import keyword
 
 from collections import defaultdict
 from functools import reduce
@@ -411,6 +412,11 @@ class HyASTCompiler(object):
                         expr, arg=None, value=ret.force_expr))
 
             elif with_kwargs and isinstance(expr, Keyword):
+                if keyword.iskeyword(expr.name):
+                    raise self._syntax_error(
+                        expr, "keyword argument cannot be Python reserved word"
+                    )
+
                 try:
                     value = next(exprs_iter)
                 except StopIteration:
@@ -1527,6 +1533,26 @@ class HyASTCompiler(object):
 
     def _compile_lambda_list(self, params, ret):
         posonly_parms, args_parms, rest_parms, kwonly_parms, kwargs_parms = params
+
+        py_reserved_param = next(
+            (
+                sym
+                for param in (
+                    (posonly_parms or [])
+                    + args_parms
+                    + kwonly_parms
+                    + [rest_parms, kwargs_parms]
+                )
+                if isinstance(param, tuple)
+                for sym in [param[1][0] if isinstance(param[1], List) else param[1]]
+                if keyword.iskeyword(str(sym))
+            ),
+            None,
+        )
+        if py_reserved_param:
+            raise self._syntax_error(
+                py_reserved_param, "parameter name cannot be Python reserved word"
+            )
 
         if not (posonly_parms or posonly_parms is None):
             raise self._syntax_error(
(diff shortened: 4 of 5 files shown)


scauligi:
> do we still think that's okay given this change?

I think it's fine.
The way I see it, we mangle so that `(= (:kw (dict :kw 'value)) 'value)`. But since users can't ever call `(dict :from ...)`, it's fine that `(:from {...})` won't look up the key `"from"`.

---

Do we want to do anything about e.g.
```
=> (f :2 4)
f(hyx_2=4)
```
? Should we disallow these sorts of cases or just let them happen?

---

And don't forget to make a note in NEWS!

allison-casey:
probably the right choice. just noticed it when doing the `defn` annotation change since the return ann for a function looks like `{"return" int}` which is the edge case where you can't just `(:return f.__annotation__)`, but that case is probably a lot less frequent than how we have it now. 

i think we should probably just let them happen. you cant define a parameter `2` anyways so `:2` will always type error on any function call besides `kwargs` and if your hand parsing kwargs i think it would be nice to have this `(:2 (test :2 1))` be true

Kodiologist:
I think it would be simpler and less surprising to generalize this a little bit: forbid using a Python reserved word as the text of a `hy.model.Keyword`, regardless of context. No further restrictions are needed. This means you can still have a reserved word as a parameter name; you just can't call it by name (without e.g. premangling, as in `(my-function :hyx_from 8)`).

allison-casey:
i like the idea of generalizing it so keywords can't be constructed from reserved words but i i'd still want parameter names to throw syntax errors. If you allow parameter names from reserved words but have no way to call them without mangling your just pushing the syntax error further away from where it really is since that parameter name is almost certainly the actual mistake.

Kodiologist:
All right.

allison-casey:
just realized that making keywords unable to be made from reserved words makes it impossible to create macros for special forms since we use reserved words in our special forms all over the place. i.e. `ap-last`
```clojure
(for  [~it ~xs  :if ~(R form)]
      (setv ~x ~it))
```
so we're back to function params and args i guess.

Kodiologist:
Oh, right, whoops.

```
for param in (posonly_parms or [])
+ args_parms
+ kwonly_parms
+ [rest_parms, kwargs_parms]
```

Can you indent the latter three lines to show that they're part of the `for` clause?

In ` for sym in (param[1][0] if isinstance(param[1], List) else param[1],)`, it looks like the right-hand side is always a tuple of one element, so you can remove this whole clause and inline `sym` in the following `if`.

I think `if py_reserved_param is not None` can be just `if py_reserved_param`.

Please capitalize "python" in the error message.

```
(with [e (pytest.raises HySyntaxError)]
    (hy.eval form)
    (assert (in "parameter name cannot be" e.value.msg)
```

Shouldn't the assertion be outside the `with`? The `with` will be jumped out of as soon as `hy.eval` raises `HySyntaxError`.

Finally, I think documentation and NEWS is probably called for.

allison-casey:
> Can you indent the latter three lines to show that they're part of the for clause?

Done

`sym` is used in both the if statement and as the returned value so the one elem tuple is used to prevent us from having to rewrite the `isinstance` part twice. 

> I think if py_reserved_param is not None can be just if py_reserved_param.

Done

> Please capitalize "python" in the error message.

Done

> Shouldn't the assertion be outside the `with`

Yup, fixed

Kodiologist:
> `sym` is used in both the `if` statement and as the returned value

Oh, right. Stupid Python putting the return value of the comprehension first. In that case, I'd just say use a list instead of a one-element tuple. It's slightly easier to read.

allison-casey:
done

Kodiologist:
What do you think about documentation?

allison-casey:
as a note in `defn` after the examples of all the `*` `#**` forms?

Kodiologist:
I'm not sure. I guess that makes sense. Or perhaps in the section about mangling, or the section about keywords.

allison-casey:
maybe a small note about their respective notes and then a link to the `defn` note with examples of actual function defs and calls?

Kodiologist:
Sounds reasonable.

allison-casey:
how does this look?

Kodiologist (approved):

NEWS.rst
```diff
@@ -49,6 +49,8 @@ Other Breaking Changes
 * `defn`, `defn/a` have been made a special forms
 * `annotate*` has been made public and renamed `annotate`
 * return annotation for `defn` has been moved to before the function name
+* Python reserved words can no longer be parameter names or
+  functionc all keywords
```
scauligi, line 53: typo

docs/api.rst
```diff
@@ -269,6 +269,31 @@ Special Forms
...
+      be called with keyword arguments that are Python reserved words. This means
+      that the following will raise a `SyntaxError` as they would in Python::
+
+         (defn afunc [a b if])
+             (defn afunc [a b if])
+                              ^
+         hy.errors.HySyntaxError: parameter name cannot be Python reserved word
+
+         (dict :a 1 :from 2)
+             (dict :a 1 :from 2)
```
scauligi, line 286: I think these examples are missing the prompt text / "Traceback ..." text like the example in the `#*` section has

docs/api.rst
```diff
@@ -269,6 +269,31 @@ Special Forms
...
+                        ^
+         hy.errors.HySyntaxError: keyword argument cannot be Python reserved word
+
+      This only applies to parameter names and a keyword argument name. The value of
+      the parameter or keyword argument can still be a keyword of a reserved word::
+
+         (defn test [* a] a)
+         (test :a :from)
+         => :from
```
scauligi, line 295: Prompt text (`=>`) should be on the input instead of the output

docs/language/syntax.rst
```diff
@@ -175,6 +177,11 @@ Python-legal names. The steps are as follows:
    Thus, ``green☘`` becomes ``hyx_greenXshamrockX``, ``if`` becomes
    ``hyx_if``, and ``is_-_has_dashes`` becomes ``hyx_is_XhyphenHminusX_has_dashes``.
 
+.. note::
+
+   A symbol of a reserved word (such as ``if``) will not be made legal if used as
```
scauligi, line 182: Hmm, I think the other new docs should be sufficient, putting a note here in the mangling section is more confusing than not

NEWS.rst
allison-casey: fixed

docs/api.rst
```diff
@@ -269,6 +269,31 @@ Special Forms
...
+
+         (dict :a 1 :from 2)
+             (dict :a 1 :from 2)
+                        ^
+         hy.errors.HySyntaxError: keyword argument cannot be Python reserved word
+
+      This only applies to parameter names and a keyword argument name. The value of
+      the parameter or keyword argument can still be a keyword of a reserved word::
+
+         (defn test [* a] a)
```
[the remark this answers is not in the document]
allison-casey: removed

docs/api.rst
allison-casey: added traceback text

docs/api.rst
allison-casey: :facepalm:

docs/language/syntax.rst
```diff
@@ -127,7 +127,9 @@ the function ``f`` with the keyword argument named ``foo`` set to ``3``. Hence,
 trying to call a function on a literal keyword may fail: ``(f :foo)`` yields
 the error ``Keyword argument :foo needs a value``. To avoid this, you can quote
 the keyword, as in ``(f ':foo)``, or use it as the value of another keyword
-argument, as in ``(f :arg :foo)``.
+argument, as in ``(f :arg :foo)``. It is important to note that a keyword argument
+cannot be a python reserved word. This will raise a ``SyntaxError`` similar to Python.
```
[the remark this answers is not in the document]
allison-casey: i gotchu bby

scauligi (approved):
:+1: dope

PR #2083, opened 2021-06-04
Fix equality for Hy models

### LINKED ISSUE #1363: Hy models have wrong equality

allison-casey:
closes #1363 

* models are no longer equal to equivalent python literals
* models only equal to models of the same kind
* renamed `wrap_value` -> `as_model` and exposed under `hy` module

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -51,12 +51,15 @@ Other Breaking Changes
 * return annotation for `defn` has been moved to before the function name
 * Python reserved words can no longer be parameter names or
   function call keywords
+* hy models are no longer equal to their associated Python values. `(= 1 '1)  ; => False`
 
 New Features
 ------------------------------
 * `hy-repr` is now the default REPL output function.
 * Running the module a la `python -m hy` is now equivalent to running
   the `hy` command.
+* `hy.as-model` has been added to create canonical model trees from
+  unquote spliced expressions
 
 Bug Fixes
 ------------------------------
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1482,6 +1482,7 @@ the following methods
 
 .. hy:autofunction:: hy.gensym
 
+.. hy:autofunction:: hy.as-model
 
 .. _Core:
 
--- a/docs/cheatsheet.json
+++ b/docs/cheatsheet.json
@@ -20,7 +20,8 @@
           "hy.gensym",
           "hy.macroexpand",
           "hy.macroexpand-1",
-          "hy.disassemble"
+          "hy.disassemble",
+          "hy.as-model"
         ]
       }
     ]
--- a/hy/__init__.py
+++ b/hy/__init__.py
@@ -30,7 +30,8 @@ _jit_imports = dict(
     gensym="hy.core.language",
     macroexpand="hy.core.language",
     macroexpand_1="hy.core.language",
-    disassemble="hy.core.language")
+    disassemble="hy.core.language",
+    as_model="hy.models")
 
 def __getattr__(k):
     if k not in _jit_imports:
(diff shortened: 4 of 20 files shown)


Kodiologist:
This must have been a lot of work. Thanks a lot for doing this.

You should now be able to remove several checks of the form `isinstance(…, Symbol)` that are followed up immediately with an equality test; e.g., in the definiton of `hy.model_patterns.sym`.

It looks like the change `wanted` → `Symbol(wanted)` in the commit labeled "Add: rename wrap_model -> as_model and expose under root hy module" belongs in a different commit.

> Recurisvely promote an expr `x`

The input need not be an expression.

Don't forget the hyphen in "S-expression". That's an important hyphen, especially on gay-pride month. Actually, when talking about Hy, I usually just say "expression", if only because the model type is named `Expression`. Also, I would avoid the word "unquoted". Be a little more specific, like "values that aren't Hy models".

I think your use of "Python literal" to mean "unboxed value" (i.e., "ordinary value that's not a model") in a commit message and NEWS is confusing. Literals are elements of syntax, not objects.

The last commit mssage says "Docs:", but there are no documentation changes in that commit.

Kodiologist:
What's with all the additions of `__hash__`?

Why is `Sequence.__eq__` necessary?

allison-casey:
> This must have been a lot of work. Thanks a lot for doing this.

not as long as pattern matching lol, but thats should be done soon just need to reorg the commits

> You should now be able to remove several checks of the form isinstance(…, Symbol) that are followed up immediately with an equality test; e.g., in the definiton of hy.model_patterns.sym.

done

> It looks like the change wanted → Symbol(wanted) in the commit labeled "Add: rename wrap_model -> as_model and expose under root hy module" belongs in a different commit.

Moved to correct commit

> The input need not be an expression.

rephrased as "object"

> Don't forget the hyphen in "S-expression". That's an important hyphen, especially on gay-pride month. Actually, when talking about Hy, I usually just say "expression", if only because the model type is named Expression. Also, I would avoid the word "unquoted". Be a little more specific, like "values that aren't Hy models".

Hy says gay rights :rainbow_flag: 
rephrased to be "expression", "non-hy model objects", and "associated Python values"

> What's with all the additions of __hash__?

When you overwrite the `__eq__` method of an immutable type it loses its hashability so you have to redefine it.  made `__hash__` an inherited method of `Object` to clean things up.

> Why is Sequence.__eq__ necessary?

Its not. removed

Kodiologist:
I've edited the PR to:

- Remove a stray whitespace edit.
- Capitalize "Hy" in prose.
- Remove more `Symbol` type checks.

Kodiologist (approved):

PR #2084, opened 2021-06-05
Feature/pattern matching

### LINKED ISSUE #1954: Pattern Matching
Would there be interest in adding a pattern matching macro to contrib, or would that have to go in an external library?

I have a working proof of concept right now, and it seems pretty similar to destructuring #1940, so I thought they might go well together.

allison-casey:
closes #1954 

adds `match` special form for > 3.10. This is a pretty literal translation of the python syntax so it should be easy to learn the syntax. 

had to add `end_line` & `end_column` to `Asty` and `Object` since they are required on some new nodes, this doesn't affect other nodes. seems like it makes that info more consistent with `start_line` and `start_column` anyways. 

This implementation is heavily parser based since its heavily recursive and would have been a nightmare to do by hand. As such the syntax errors are a bit obtuse, but we already knew that https://github.com/vlasovskikh/funcparserlib/issues/52

This solves the issue of arbitraty statements in the `:if` guard by lifting them up into a function. this works because all of the bind targets have already been bound into local scope by the time the function gets called.

```clojure
=> (match [0 1 2]
...     (case [0 #* x] :as z :if (as-> z $ (+ $ [3]) (len $) (= $ 4))
...         0))
def _hy_anon_var_2():
    hyx_Xdollar_signX = z
    hyx_Xdollar_signX = hyx_Xdollar_signX + [3]
    hyx_Xdollar_signX = len(hyx_Xdollar_signX)
    hyx_Xdollar_signX = hyx_Xdollar_signX == 4
    return hyx_Xdollar_signX
match [0, 1, 2]:
    case [0, *x] as z if _hy_anon_var_2():
        _hy_anon_var_1 = 0
    case _:
        _hy_anon_var_1 = None
_hy_anon_var_1
0

```

let me know if you manage to blow it up somehow :+1:

### DIFF
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -3,19 +3,18 @@
 # This file is part of Hy, which is free software licensed under the Expat
 # license. See the LICENSE.
 
-from itertools import dropwhile
+from itertools import dropwhile, takewhile
 from hy.models import (Object, Expression, Keyword, Integer, Complex,
                        String, FComponent, FString, Bytes, Symbol,
                        Float, List, Set, Dict, Sequence, as_model)
-from hy.model_patterns import (FORM, SYM, KEYWORD, STR, sym, brackets, whole,
-                               notpexpr, dolike, pexpr, times, Tag, tag, unpack)
-from funcparserlib.parser import some, many, oneplus, maybe, NoParseError, a
+from hy.model_patterns import (FORM, SYM, KEYWORD, STR, LITERAL, sym, brackets, whole,
+                               notpexpr, dolike, pexpr, times, Tag, tag, unpack, braces)
+from funcparserlib.parser import some, many, oneplus, maybe, NoParseError, a, forward_decl
 from hy.errors import (HyCompileError, HyTypeError, HyLanguageError,
                        HySyntaxError, HyEvalError, HyInternalError)
 
-from hy.lex import mangle, unmangle, hy_parse, parse_one_thing, LexException
+from hy.lex import mangle, unmangle, hy_parse, parse_one_thing, LexException, isidentifier
 
-from hy._compat import PY3_8
 from hy.macros import require, macroexpand
 
 import textwrap
@@ -631,6 +630,172 @@ class HyASTCompiler(object):
         return ret + asty.Raise(
             expr, type=ret.expr, exc=exc,
             inst=None, tback=None, cause=cause)
+
+    notsym = lambda *dissallowed: some(
+        lambda x: isinstance(x, Symbol) and str(x) not in dissallowed
+    )
+    keepsym = lambda wanted: some(lambda x: x == Symbol(wanted))
+    _pattern = forward_decl()
+    _pattern.define(
+        (
+            SYM
+            | LITERAL
+            | brackets(many(_pattern | unpack("iterable")))
+            | pexpr(keepsym("."), many(SYM))
+            | pexpr(keepsym("|"), many(_pattern))
+            | pexpr(keepsym(","), many(_pattern))
+            | braces(
+                many(LITERAL + _pattern), maybe(pvalue("unpack-mapping", SYM))
+            )
+            | pexpr(
+                notsym(".", "|", ",", "unpack-mapping", "unpack-iterable"),
+                many(_pattern),
+                many(KEYWORD + _pattern),
+            )
+        )
+        + maybe(sym(":as") + SYM)
+    )
+    match_clause = _pattern + maybe(sym(":if") + FORM)
+
+    @special(
+        ((3, 10), "match"), [FORM, many(match_clause + FORM)]
+    )
+    def compile_match_expression(self, expr, root, subject, clauses):
+        subject = self.compile(subject)
+        return_var = asty.Name(expr, id=mangle(self.get_anon_var()), ctx=ast.Store())
+
+        lifted_if_defs = []
+        match_cases = []
+        for *pattern, guard, body in clauses:
+            if guard and body == Keyword("as"):
+                raise self._syntax_error(body, ":as clause cannot come after :if guard")
+
+            body = self._compile_branch([body])
+            body += asty.Assign(pattern[0], targets=[return_var], value=body.force_expr)
+            body += body.expr_as_stmt()
+            body = body.stmts
+
+            pattern = self._compile_pattern(pattern)
+
+            if guard:
+                guard = self.compile(guard)
+                if guard.stmts:
+                    fname = self.get_anon_var()
+                    guardret = Result() + asty.FunctionDef(
+                        guard,
+                        name=fname,
+                        args=ast.arguments(
+                            args=[],
+                            varargs=None,
+                            kwarg=None,
+                            posonlyargs=[],
+                            kwonlyargs=[],
+                            kw_defaults=[],
+                            defaults=[],
+                        ),
+                        body=guard.stmts + [asty.Return(guard.expr, value=guard.expr)],
+                        decorator_list=[],
+                    )
+                    lifted_if_defs.append(guardret)
+                    guard = Result(expr=ast.parse(f"{fname}()").body[0].value)
+
+            match_cases.append(
+                ast.match_case(
+                    pattern=pattern,
+                    guard=guard.force_expr if guard else None,
+                    body=body,
+                )
+            )
+
+        returnable = Result(
+            expr=asty.Name(expr, id=return_var.id, ctx=ast.Load()),
+            temp_variables=[return_var],
+        )
+        ret = Result() + subject
+        ret += subject.expr_as_stmt()
+        ret += asty.Assign(expr, targets=[return_var], value=asty.Constant(expr, value=None))
+        if not match_cases:
+            return ret + returnable
+
+        for lifted_if in lifted_if_defs:
+            ret += lifted_if
+        ret += asty.Match(expr, subject=subject.force_expr, cases=match_cases)
+        return ret + returnable
+
+    def _compile_pattern(self, pattern):
+        value, assignment = pattern
+        if assignment is not None:
+            return asty.MatchAs(
+                value,
+                pattern=self._compile_pattern((value, None)),
+                name=mangle(self._nonconst(assignment)),
+            )
+
+        if str(value) in ("None", "True", "False"):
+            return asty.MatchSingleton(
+                value,
+                value=self.compile(value).force_expr.value,
+            )
+        elif (
+            isinstance(value, (String, Integer, Float, Complex, Bytes))
+            or isinstance(value, Symbol)
+            and "." in value
+        ):
+            return asty.MatchValue(
+                value,
+                value=self.compile(value).expr,
+            )
+        elif value == Symbol("_"):
+            return asty.MatchAs(value)
+        elif isinstance(value, Symbol):
+            return asty.MatchAs(value, name=mangle(value))
+        elif isinstance(value, Expression) and value[0] == Symbol("|"):
+            return asty.MatchOr(
+                value,
+                patterns=[self._compile_pattern(v) for v in value[1]],
+            )
+        elif isinstance(value, Expression) and value[0] == Symbol("."):
+            root, syms = value
+            dotform = mkexpr(root, *syms).replace(value)
+            return asty.MatchValue(
+                value,
+                value=self.compile(dotform).expr,
+            )
+        elif (
+            isinstance(value, List)
+            or isinstance(value, Expression)
+            and value[0] == Symbol(",")
+        ):
+            patterns = value[0] if isinstance(value, List) else value[1:][0]
+            patterns = [
+                self._compile_pattern((v, None) if is_unpack("iterable", v) else v)
+                for v in patterns
+            ]
+            return asty.MatchSequence(value, patterns=patterns)
+        elif is_unpack("iterable", value):
+            return asty.MatchStar(value, name=mangle(value[1]))
+
+        elif isinstance(value, Dict):
+            kvs, rest = value
+            keys, values = zip(*kvs) if kvs else ([], [])
+            return asty.MatchMapping(
+                value,
+                keys=[self.compile(key).expr for key in keys],
+                patterns=[self._compile_pattern(v) for v in values],
+                rest=mangle(rest) if rest else None,
+            )
+        elif isinstance(value, Expression):
+            root, args, kwargs = value
+            keywords, values = zip(*kwargs) if kwargs else ([], [])
+            return asty.MatchClass(
+                value,
+                cls=asty.Name(root, id=mangle(root), ctx=ast.Load()),
+                patterns=[self._compile_pattern(v) for v in args],
+                kwd_attrs=[kwd.name for kwd in keywords],
+                kwd_patterns=[self._compile_pattern(value) for value in values],
+            )
+        else:
+            raise self._syntax_error(value, "unsuported")
 
     @special("try",
        [many(notpexpr("except", "else", "finally")),
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -150,6 +150,12 @@ class Asty(object):
                 x, 'start_line', getattr(x, 'lineno', None)),
             col_offset=getattr(
                 x, 'start_column', getattr(x, 'col_offset', None)),
+            end_lineno=getattr(
+                x, 'end_line', getattr(x, 'end_lineno', None)
+            ),
+            end_col_offset=getattr(
+                x, 'end_column', getattr(x, 'end_col_offset', None)
+            ),
             **kwargs)))
         return getattr(Asty, name)
 asty = Asty()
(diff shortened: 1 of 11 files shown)


allison-casey:
I'm realizing now that this will break when using let since the match values aren't protected so they'll get replaced

Kodiologist:
`let` is truly the technical debt that keeps on giving.

Kodiologist:
I think the head symbol `case` isn't necessary. Every direct child form of `match` already has to be a case clause. So the syntax can be simplified to something like:

```
(match (Point -1 1)
  [(Point 0 0) "The Origin"]
  [(Point 0 :y y) f"One on the Y axis at {y}"]
  [[(Point :x 0 y1) (Point 0 :y y2)] f"Two on the Y axis at {y1}, {y2}"]
  [_ "some place off in the ether"])
```

Kodiologist:
One could even imagine an unbracketed syntax, as I know you're thinking of changing `cond`:

```
(match (Point -1 1)
  (Point 0 0) "The Origin"
  (Point 0 :y y) f"One on the Y axis at {y}"
  [(Point :x 0 y1) (Point 0 :y y2)] f"Two on the Y axis at {y1}, {y2}"
  _ "some place off in the ether")
```

In this design, only one body form is allowed (possibly preceded by `:as` and `:if` parts), so you'd use `do` if you needed more than one.

hy/compiler.py
```diff
@@ -598,6 +621,166 @@ def compile_raise_expression(self, expr, root, exc, cause):
...
+        _match_value + _assignment
+        | brackets(many(_pattern | unpack("iterable"))) + _assignment
+        | pexpr(keepsym("|"), many(_pattern)) + _assignment
+        | pexpr(keepsym(","), many(_pattern)) + _assignment
+        | curly_brackets(many(LITERAL + _pattern), maybe(pvalue("unpack-mapping", SYM))) + _assignment
+        | pexpr(
+            notsym("|", ",", "unpack-mapping", "unpack-iterable"),
+            many(_pattern),
+            many(KEYWORD + _pattern),
+        ) + _assignment
```
scauligi, line 641: The ` + _assignment` can be factored out if you stick the alternations inside a set of parentheses, no?

hy/compiler.py
```diff
@@ -598,6 +621,166 @@ def compile_raise_expression(self, expr, root, exc, cause):
...
+    )
+    match_clause = _pattern + maybe(sym(":if") + FORM) + _assignment
+
+    @special(
+        (PY3_10, "match"), [FORM, oneplus(pexpr(sym("case"), match_clause, many(FORM)))]
```
scauligi, line 646: Agreed with @Kodiologist we should drop the "case"; and a mild preference for the unbracketed version of the syntax. Also this should be `many` instead of `oneplus` to match other parts of Hy that allow vacuous constructs.

hy/compiler.py
```diff
@@ -598,6 +621,166 @@ def compile_raise_expression(self, expr, root, exc, cause):
...
+
+            if bad_assignment:
+                raise self._syntax_error(bad_assignment, ":as assignment form cannot come after an :if guard")
+
+            body = self._compile_branch(body)
+            body += asty.Assign(pattern[0], targets=[return_var], value=body.force_expr)
+            body += body.expr_as_stmt()
+            body = body.stmts
+
+            if pattern[0] == Symbol("_"):
```
scauligi, line 667: `case catchall: ...` where `catchall` is a valid symbol is allowed in match statements to name the catchall case, so this should be
```python
```

scauligi (commented):
Yeah I agree with @Kodiologist on dropping the "case" root. I don't have a strong preference on bracketed vs unbracketed but I'm leaning slightly towards unbracketed.

I can take a look at how this interacts with `let`, it might not be so bad (fingers crossed)

hy/compiler.py
allison-casey: good catch! done

hy/compiler.py
allison-casey: valid, done.

hy/compiler.py
allison-casey: fixed

allison-casey:
i added a commit with bracketless match cases. give it a try and see how it feels. If we like it i'll squash it. 
if you wanna try the bracketed form all you need to do is change 
```python
[FORM, oneplus(pexpr(sym("case"), match_clause, many(FORM)))]
```
to 
```python
[FORM, oneplus(brackets(match_clause, many(FORM)))]
```

scauligi:
Hmm, thinking on this some more, I don't think we should result in `None` for an implicit default case, it feels too much like silent failure.

We should either explicitly raise an Exception of some sort in the implicit case (similar to some other match language implementations I've seen) or just not include an implicit default at all (which reflects the equivalent Python more closely).

In the latter case, if the user tries to use/access the result of an unmatched `match`, they'll get a `NameError: name '_hy_anon_var_#' is not defined`, but the traceback still correctly points to the match form. If they don't use the result (e.g. using a partial `match` just for some side effects) then no exceptions will get thrown.

scauligi:
Sorry to step on your toes @allison-casey but I kind of sniped myself on the `let` fix. By all means edit/rearrange as you see fit.

Here's the patch for `let` (presented without tests sorry):
```diff
diff --git a/hy/contrib/walk.hy b/hy/contrib/walk.hy
index e683b82c..65e2cf31 100644
--- a/hy/contrib/walk.hy
+++ b/hy/contrib/walk.hy
@@ -320,6 +320,41 @@
                                             (if (and tail (= (len (get tail 0)) 2))
                                                 #{(get tail 0 0)}
                                                 #{})))))
+
+  (defn handle-match [self]
+    ;; protect name bindings from match patterns
+    (setv [expr #* cases] (self.tail)
+          new-expr (self.expand-symbols expr)
+          new-cases [])
+    (for [[pattern body] (by2s cases)]
+      ;; collect protected names from pattern
+      ;; before evaluating the corresponding body
+      (setv protected #{})
+      (defn handle-match-symbol [form]
+        (if (in "." form)
+          (self.expand-symbols form)
+          (do
+            (.add protected form)
+            form)))
+      (defn handle-match-call [form]
+        (setv head (get form 0))
+        (setv tail (cut form 1 None))
+        (cond
+          [(= head '.) (self.expand-symbols form)]
+          [True `(~head ~@(traverse-pattern tail))]))
+      (defn handle-pattern-form [form]
+        (cond
+          [(symbol? form) (handle-match-symbol form)]
+          [(call? form) (handle-match-call form)]
+          [(coll? form) (traverse-pattern form)]
+          [True form]))
+      (defn traverse-pattern [pattern]
+        (walk handle-pattern-form (fn [x] x) pattern))
+      (setv new-pattern (handle-pattern-form pattern))
+      (setv new-body (self.expand-symbols body (| protected self.protected)))
+      (.extend new-cases [new-pattern new-body]))
+    `(~(self.head) ~new-expr ~@new-cases))
+
   (defn handle-args-list [self]
     (setv protected #{}
           argslist [])
@@ -407,6 +442,7 @@
       [(= head 'except) (self.handle-except)]
       [(= head ".") (self.handle-dot)]
       [(= head 'defclass) (self.handle-defclass)]
+      [(= head 'match) (self.handle-match)]
       [(= head 'quasiquote) (self.+quote)]
         ;; must be checked last!
       [(in (hy.mangle head) _mangled-special) (self.handle-special-form)]
```

And also here's a quick patch for allowing `(. SYM...)` forms in match patterns to mirror `SYM.SYM...` forms:
```diff
diff --git a/hy/compiler.py b/hy/compiler.py
index 5b34b304..9f85bf93 100755
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -632,13 +632,14 @@ class HyASTCompiler(object):
         (
             _match_value
             | brackets(many(_pattern | unpack("iterable")))
+            | pexpr(keepsym("."), many(SYM))
             | pexpr(keepsym("|"), many(_pattern))
             | pexpr(keepsym(","), many(_pattern))
             | curly_brackets(
                 many(LITERAL + _pattern), maybe(pvalue("unpack-mapping", SYM))
             )
             | pexpr(
-                notsym("|", ",", "unpack-mapping", "unpack-iterable"),
+                notsym(".", "|", ",", "unpack-mapping", "unpack-iterable"),
                 many(_pattern),
                 many(KEYWORD + _pattern),
             )
@@ -735,6 +736,13 @@ class HyASTCompiler(object):
                 value,
                 value=self.compile(value).expr,
             )
+        elif isinstance(value, Expression) and value[0] == Symbol("."):
+            root, syms = value
+            dotform = mkexpr(root, *syms).replace(value)
+            return asty.MatchValue(
+                value,
+                value=self.compile(dotform).expr,
+            )
         elif isinstance(value, Symbol):
             return asty.MatchAs(
                 value, name=mangle(value)
```

Kodiologist:
> Hmm, thinking on this some more, I don't think we should result in `None` for an implicit default case, it feels too much like silent failure.

I disagree. Changing this would be a big departure from the rest of the language.

allison-casey:
a failure to match anything in a `match` statement results in a no-op so so we kind of have to return None in the context of an s-expression

allison-casey:
I pushed a commit with another potential syntax. One i'm actually quite liking having messed around with it for a bit. It replaces the `:if + FORM` syntax with wrapping the pattern in a `where` expression. This is pretty much the same as `fennel` match and i think results in a pretty clean looking expression

```clojure
(match [1 2 3 4]
  [1 _ x] :as z [x z]
  (where [_ _ x _] :as z (odd? x)) z
  (where z (= (len z) 4)) z)
```

Kodiologist:
I don't see the advantage over the more Python-like `:if`. It's weird that you add `where` in order to add a condition to the test and yet the condition form comes after the pattern.

allison-casey:
personally looks cleaner in my mind, but you're right that its not sufficiently better than python's syntax to merit creating the disparity. I can make it a macro easily anyways. Will delete it and squash bracketless

allison-casey:
implemented let binding protection + fixed `_` incorrectly being bound to a symbol

Kodiologist:
I'd call `curly_brackets` "braces" or "curlies" instead.

`_match_value` and `_assignment` are each used only once. You know what to do. Similarly, it looks like you can overwrite `target` instead of defining a new variable `subject`. "subject" is probably the better name, though.

`if guard is not None` can probably be written `if guard`.

`body=[asty.Assign(expr, targets=[return_var]`… and `elif isinstance(value, (String, Integer`… are super-long lines. Give us some newlines.

The line `print(value)` looks like a leftover debugging statement.

Add tests that statements in match bodies and `:if` clauses are executed if and only if they ought to be. For this kind of check, I often append to a list and then check the list contents afterwards.

In the test

```
(assert (is (match {}
                   {0 [1 2 {}]} 0
                   {0 [1 2 {}] 1 [[]]} 1
                   [] None)
            None))
```

I'd use something other than `None` for the last case, to distinguish from the implicit default `None`.

There's a stray newline added before `Inf = float('inf')` in `compiler.py`.

Kodiologist:
And don't forget `(assert (is (match 1) None))`.

Kodiologist:
I've tried to make the documentation shorter and more complete. Tell me what you think. And yeah, I think we might as well disallow patterns like `(, 1 2 3)` when the user could just write `[1 2 3]`. Goodness knows why Python allows both syntaxes if they do the same thing.

allison-casey:
> I'd call `curly_brackets` "braces" or "curlies" instead.

changed to `braces`

> `_match_value` and `_assignment` are each used only once. You know what to do. 

inlined

> Similarly, it looks like you can overwrite `target` instead of defining a new variable `subject`. "subject" is probably the better name, though.

renamed `target` to `subject`

> `if guard is not None` can probably be written `if guard`.

done

> `body=[asty.Assign(expr, targets=[return_var]`… and `elif isinstance(value, (String, Integer`… are super-long lines. Give us some newlines.

done

> The line `print(value)` looks like a leftover debugging statement.

deleted

> Add tests that statements in match bodies and `:if` clauses are executed if and only if they ought to be. For this kind of check, I often append to a list and then check the list contents afterwards.

done

> In the test
> 
> ```
> (assert (is (match {}
>                    {0 [1 2 {}]} 0
>                    {0 [1 2 {}] 1 [[]]} 1
>                    [] None)
>             None))
> ```
> I'd use something other than `None` for the last case, to distinguish from the implicit default `None`.

changed last clause to return `2` instead of `None`

> There's a stray newline added before `Inf = float('inf')` in `compiler.py`.

deleted

> Also, it's probably not wise to test that assignments made in failed pattern matches hang around

valid, removed

> I've tried to make the documentation shorter and more complete. Tell me what you think.

I've added an example of class matching, otherwise fine by me

> And yeah, I think we might as well disallow patterns like (, 1 2 3) when the user could just write [1 2 3]. Goodness knows why Python allows both syntaxes if they do the same thing.

Not a fan. Its a deviation from both the spec and how Hy works in other areas like tuple unpacking assignment.

Kodiologist:
> Its a deviation from… how Hy works in other areas like tuple unpacking assignment.

Good point.

Kodiologist (approved):
- Add a docstring for `braces`.
- I would use `(% x 2)` or something instead of `even?` because `even?` is going to be baleeted or moved to the spin-off library soon.

allison-casey:
>     * Add a docstring for `braces`.
> 
>     * I would use `(% x 2)` or something instead of `even?` because `even?` is going to be baleeted or moved to the spin-off library soon.

Done

@scauligi you had some notes on this, how are you feeling about the current state of the PR?

tests/native_tests/language.hy
```diff
@@ -16,7 +16,7 @@
         pytest)
 (import sys)
 
-(import [hy._compat [PY3_8]])
+(import [hy._compat [PY3_8 PY3_10]])
```
scauligi, line 19: No longer needed here?

scauligi (commented):
Edge case on default-case detection: we still need to add a(n unguarded) catch-all if the existing catch-all(s) are guarded:
```hy
=> (match 1
...  _ :if False 'unreachable)

match 1:
    case _ if False:
        _hy_anon_var_5 = hy.models.Symbol('unreachable')
_hy_anon_var_5

  File "stdin-a2b5526f71852fa367c6fbbc46d544b7841c58d8", line 1, in <module>
    (match 1
NameError: name '_hy_anon_var_5' is not defined
```

---

Is it a concern that the seemingly (in)correct expression
```hy
(match expr
  pat :if (cond pat) :as x  (, x pat))
```
parses (and can run without error! just with unexpected results) as the valid expression
```hy
(match expr
  pat :if (cond pat)  :as  ; the result of this match is just (Keyword "as")
  x  (, x pat)  ; a catch-all match bound to x
```

Kodiologist:
Rather than adding a default case, it's simpler to just initialize the output variable to `None` before the `match`.

tests/native_tests/language.hy
allison-casey: removed

allison-casey:
> Edge case on default-case detection: we still need to add a(n unguarded) catch-all if the existing catch-all(s) are guarded:
> 
> ```hy
> => (match 1
> ...  _ :if False 'unreachable)
> 
> match 1:
>     case _ if False:
>         _hy_anon_var_5 = hy.models.Symbol('unreachable')
> _hy_anon_var_5
> 
> Traceback (most recent call last):
>   File "stdin-a2b5526f71852fa367c6fbbc46d544b7841c58d8", line 1, in <module>
>     (match 1
> NameError: name '_hy_anon_var_5' is not defined
> ```

initialized the anon var to `None` and now `(match 1)` compiles to just `anon_var = None` and doesn't generate a match statement at all

> Is it a concern that the seemingly (in)correct expression
> 
> ```hy
> (match expr
>   pat :if (cond pat) :as x  (, x pat))
> ```
> 
> parses (and can run without error! just with unexpected results) as the valid expression
> 
> ```hy
> (match expr
>   pat :if (cond pat)  :as  ; the result of this match is just (Keyword "as")
>   x  (, x pat)  ; a catch-all match bound to x
> ```

made `match` raise a `HySyntaxError` if the body is `:as` and an `:if` guard exists, if you want a clause to return the keyword `:as` if now has to be quoted

```clojure
(match 1
       1 :if True :as)
  File "stdin-5322fe22a66153576430f80f7d8b627c931eda28", line 1, in <module>
    (match 1 1 :if True :as )
  File "<stdin>", line 1
    (match 1 1 :if True :as )
                        ^
hy.errors.HySyntaxError: :as clause cannot come after :if guard
```

```clojure
(match 1
       1 :if True ':as)
:as
```

Kodiologist:
> initialized the anon var to `None` and now `(match 1)` compiles to just `anon_var = None` and doesn't generate a match statement at all

You still need to evaluate the subject, though, in case it has side-effects.

allison-casey:
its still compiled and added to the Result
```python
ret = Result() + subject
ret += asty.Assign(expr, targets=[return_var], value=asty.Constant(expr, value=None))
if not match_cases:
    return ret + returnable
```

Kodiologist:
You need an `expr_as_stmt`, though:

```
$ hy --spy
import hy
hy 1.0a1+156.g8f4476f8 using CPython(default) 3.10.0b2 on Linux
=> (match (f))
_hy_anon_var_1 = None
_hy_anon_var_1
```

allison-casey:
ahhh got it. Done

scauligi (approved):
heccin' _DOPE_

PR #2091, opened 2021-06-10
Fix: Syntax error special forms on unsupported python versions

### LINKED ISSUE #2090: On Pythons too old to support a construct, create a degenerate special operator
As I implemented it, the `special` decorator in `compiler.py` can check the current version of Python, and if it fails, no special operator is created. For example, the use of `@special((PY3_8, "setx"), …)` means that on Python 3.7, `setx` is an ordinary name. This leads to unhelpful error messages, and it's confusing that the set of special operator names changes depending on the version of Python. Instead, when the version check fails, `@special` should create a degenerate special operator that unconditionally fails at compile-time with an error message saying "setx isn't supported on this version of Python" or something.

allison-casey:
closes #2090 

easy money 

attempting to `setx` on Python 3.7 will result in 
```clojure
hy 1.0a1+136.gbeee3586.dirty using PyPy(51efa818fd9b) 3.7.10 on Linux
=> (setx x 1)
  File "stdin-12a3a753791edb1e67c4fba15af57553319e7498", line 1, in <module>
    (setx x 1)
  File "<stdin>", line 1
    (setx x 1)
    ^
hy.errors.HySyntaxError: setx is not supported on Python versions before Python 3.8
```

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -62,6 +62,8 @@ New Features
   the `hy` command.
 * `hy.as-model` has been added to create canonical model trees from
   unquote spliced expressions
+* Hy will raise a `SyntaxError` if attempting to use a special form
+  on an unsupported version of Python
 
 Bug Fixes
 ------------------------------
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,8 @@
+import sys
 import os
 import importlib
+from operator import or_
+from functools import reduce
 
 import py
 import pytest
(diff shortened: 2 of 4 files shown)


Kodiologist:
Instead of `f"{root} is not supported on Python versions before Python {major}.{minor}"`, how about `f"{root} requires Python {major}.{minor} or later"`?

Kodiologist (approved):

allison-casey:
done

PR #2095, opened 2021-06-12
Remove core predicate functions

Kodiologist:
I wasn't sure which of these to delete and which to move to the spin-off library. It's one of those things where it's hard to draw the line between useful syntactic sugar and an endless parade of tiny functions. Ultimately I just decided to trash all of them.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -26,9 +26,13 @@ Removals
     (formerly `take-while`), `tee`, `zip-longest`
   * From `functools`: `reduce`
   * From `fractions`: `Fraction` (formerly `fraction`)
+
+* The following core predicate functions have been removed. Use
+  `isinstance` etc. instead.
+
+  * `empty?`, `even?`, `every?`, `float?`, `integer-char?`, `integer?`, `iterable?`, `iterator?`, `keyword?`, `list?`, `neg?`, `none?`, `numeric?`, `odd?`, `pos?`, `some`, `string?`, `symbol?`, `tuple?`, `zero?`
 * Several other core functions and macros have been removed:
 
-  * `instance?`: Use `isinstance` instead.
   * `keyword`: Use `(hy.models.Keyword (hy.unmangle …))` instead.
   * `repeatedly`: Use `toolz.iterate` instead.
   * `if-not`: Use `(if (not …) …)` instead.
--- a/conftest.py
+++ b/conftest.py
@@ -27,7 +27,7 @@ def pytest_ignore_collect(path, config):
     return reduce(
         or_,
         (name in path.basename and not condition for condition, name in versions),
-    )
+    ) or None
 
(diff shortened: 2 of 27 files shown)


allison-casey:
probably for the best

allison-casey (approved):

Kodiologist:
After sleeping on it, I think `some` should also be deleted, and `coll?`, being nontrivial and surprisingly useful, is worth keeping for the spin-off library. I'll remove some uses of `coll?` but keep it in Hy until the spin-off library exists. Does that still sound good, Allie?

allison-casey:
i would say generally err on the side of getting rid of trivial things. `some` is just a slightly more obtuse `any` whose alternative case is better served by `ap-first` anyways. In that spirit i'd say `coll?` is nontrivial enough, and relevant enough to Hy's functional style, to merit keeping for now. We can always add some of these things back in later if there's demand enough.

Kodiologist:
Yeah, the reason I wouldn't see myself using `some` is its use of a predicate function. I would more often want something like `(next (gfor …))`.

Kodiologist:
Done. I also made an unrelated fix to `conftest.py` in the first commit.

scauligi (approved):
You can use `(not expr)` in place of `(= expr 0)` as a slightly more terse replacement for `(zero? expr)`, but not a major thing.
Otherwise looks fine to me.

Many of these will reappear in the eventual standard library, yeah?

Kodiologist:
Yeah, I figured `(= x 0)` was a bit more explicit, which was the only reason to use `zero?` in place of `not` in most cases.

> Many of these will reappear in the eventual standard library, yeah?

The proposal here is that only `coll?` will. If you want to keep any of the others, now's the time to object.

scauligi:
> Yeah, I figured (= x 0) was a bit more explicit

Reasonable!

> If you want to keep any of the others...

Hmm, I think it can be useful to have the various helper predicates in the (separate) stdlib, since they seem like they'd be common enough that users would be redefining them anyway (e.g. the `zero?` predicate vs using `(fn [x] (= x 0))` every time)

Kodiologist:
Yeah, but I'm asking, which ones? All of them? Speaking for myself, I get little to no use out of these because I use comprehension forms or anaphoric macros in place of higher-order functions. They're more flexible; they're reasonably concise even if you don't already have the exact mapping function or predicate function or whatever that you need. Using higher-order functions is more concise in Haskell, but that's because Haskell makes it easy to make complex compositions of functions on the fly, so again you don't need to rely on a large preexisting library.

scauligi:
> I use comprehension forms or anaphoric macros in place of higher-order functions.

Hmm fair enough, I haven't tried using the anaphoric macros yet but I can see how they could replace needing the different predicates.

PR #2097, opened 2021-06-13
Fix: Remove __init__ suppression automagic

### LINKED ISSUE #1544: Get rid of the __init__ return suppression automagic and replace with a definit macro
This kind of automagic makes me uncomfortable.  It's also complicating the compiler and seems to be the cause of #1533.

Python has a very dynamic runtime. You can rewrite existing classes on the fly because they're objects too. You can introducing new methods and delete old ones. Metaclasses may need to do this kind of rewriting internally to function.

But the `__init__` automagic to return `None` only works in a defclass form. Not if `__init__` was declared as a function outside and the attached later.

Worse, if some metaclass other than `type` requires an `__init__` method to actually return something, it may be impossible to use in Hy.

I'm also worried this may cause issues with decorators. Does the automagic even work if `__init__` is decorated? And even worse, what if **does** work, but the decorator requires a return value in the decorated method? It would be difficult to use that in Hy.

I was opposed to this automagic from the start, but we did add this for a reason. Usually automatic returns are more convenient in Hy. But that's an error in `__init__` so it's a pain to remember to always end `__init__` methods with `None` in Hy (if the last form happens to return `None`, as many do, you might not notice the problem right away).

Explicit is better than implicit. Let's get rid of the automagic and just use a simple core `definit` macro instead, like this,
```
(defmacro definit [argslist &rest body]
  `(defn __init__ ~argslist ~@body None))
```
Which you'd use like
```
(defclass Foo []
  (definit [self x]
    (setv self.x x))
```
It returns `None` as required by the standard `type` metaclass, and `definit` is easier to type, than `def __init__` so it'll get used, instead of forgotten.

allison-casey:
closes #1544 

no magic allowed :no_entry_sign:  :mage_man: :no_entry_sign:

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -54,6 +54,7 @@ Other Breaking Changes
 * Python reserved words can no longer be parameter names or
   function call keywords
 * hy models are no longer equal to their associated Python values. `(= 1 '1)  ; => False`
+* `defclass` no longer automagically adds `None` to end of `__init__` forms
 
 New Features
 ------------------------------
--- a/hy/compiler.py
+++ b/hy/compiler.py
@@ -1856,8 +1856,7 @@ class HyASTCompiler(object):
             bodyr += self.compile(docstring).expr_as_stmt()
 
         for e in body:
-            e = self.compile(self._rewire_init(
-                macroexpand(e, self.module, self)))
+            e = self.compile(macroexpand(e, self.module, self))
             bodyr += e + e.expr_as_stmt()
 
         return bases + asty.ClassDef(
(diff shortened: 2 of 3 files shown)


Kodiologist:
Hy will absolutely be keeping some magic that's worth the weight, like letting the user treat statements as expressions, but this magic, yeah, we can probably do without.

Kodiologist (approved):

allison-casey:
can you even have a lisp with statements?

gilch:
You could certainly express a statement-oriented language in S-expressions, it would just have some restrictions on nesting the statement special forms. Given the choice, I'd rather not have the statement-expression distinction though. It would make metaprogramming more difficult.

scauligi (approved):
lgtm :+1:

allison-casey:
> You could certainly express a statement-oriented language in S-expressions, it would just have some restrictions on nesting the statement special forms. Given the choice, I'd rather not have the statement-expression distinction though. It would make metaprogramming more difficult.

That sounds kinda gross no thank you lol

PR #2098, opened 2021-06-13
bracketless cond

allison-casey:
makes `cond` act like the previously yeeted macro `if`. Also renames `ifp` to `condp` since the behavior it's building off of is now `cond` and not macro `if`

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -54,6 +54,7 @@ Other Breaking Changes
 * Python reserved words can no longer be parameter names or
   function call keywords
 * hy models are no longer equal to their associated Python values. `(= 1 '1)  ; => False`
+* brackets have been dropped from `cond` and it no longer accepts single condition branches
 
 New Features
 ------------------------------
--- a/hy/contrib/destructure.hy
+++ b/hy/contrib/destructure.hy
@@ -111,7 +111,7 @@ Iterator patterns are specified using round brackets. They are the same as list 
   [functools [reduce]]
   [hy.contrib.walk [by2s]])
 
-(defmacro! ifp [o!pred o!expr #* clauses]
+(defmacro! condp [o!pred o!expr #* clauses]
   "Takes a binary predicate ``pred``, an expression ``expr``, and a set of
   clauses. Each clause can be of the form ``cond res`` or ``cond :>> res``. For
   each clause, if ``(pred cond expr)`` evaluates to true, returns ``res`` in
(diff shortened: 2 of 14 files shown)


Kodiologist:
Maybe let's wait till it's in the spin-off library to change the syntax.

allison-casey:
valid. there is the option of making linearized if ladders a function of the `if` special form so it becomes an inherent part of the language. that would be a syntax change, but i personally have to preference. feel free to close this

Kodiologist:
I think `if` with more than three arguments was always kinda weird, and will be weirder if `cond` loses its brackets. If you allow more than three arguments, you might as well call it `cond`, and have `if` as the macro.

allison-casey:
linearized ladders with `if` always made sense to me since ternary `if` has form
```clojure
(if (condition-1) (result-1-form)
    (else-form))
```
and adding additional condition/result pairs is just duplicating whats already there
```clojure
(if (condition-1) (result-1-form)
    ...            ...
    ...            ...
    (condition-n) (result-n-form)
    (else))
```
how would you feel about making macro `if`'s ladder capability part of special form `if` and just leaving `cond` alone? Cause right now getting rid of one or the other feels like kind of a wash to me since.

* making `if` exclusively ternary is a loss of functionality and once `cond` is moved to skyhy, hy will have lost the ability to do if ladders out of the box. 
* making `cond` bracketless loses it's ability to have single form condition/result branches and its implicit `do`.

scauligi:
I like having generalized `if`, it "feels" more like python else-if ladders when I'm looking at the code; `cond` then is just sugar for general-`if` with implicit `do`.

Kodiologist:
Neither

```
(if condition-1 result-1-form
    else-form)
```

nor

```
(if condition-1 result-1-form
    ...            ...
    ...            ...
    condition-n result-n-form
    else-form)
```

ever seemed logical to me because the `else` is parallel in meaning to the then-forms, not the condition forms. That's why I write code like

```
(if (condition-1)
  result-1-form
  else-form)
```

or

```
(cond
  [condition-1 result-1-form]
  ...
  [condition-n result-n-form]
  [True        else-form])
```

For what it's worth, `if` with more than 3 arguments in Lisp doesn't traditionally allow multiple conditions, either; instead, it means

```
(if condition
    then-form
  else-form-1
  …
  else-form-n)
```

which I also think is kind of weird.

So yeah, long story short, I think that the form that has multiple conditions should be named `cond`, not `if`. I don't really care one way or the other about brackets. Losing "the ability to do if ladders out of the box" isn't a concern to me because you're not expected to use Hy without the spinoff library.

Issue #722, opened 2014-12-15
Fully functional reader macros


fjarri:
Currently Hy has some limited support of reader macros (quote, `#` etc). Is it planned (or even wanted) to add full support of reader macros similar to CL's `set-macro-character`? I think it will require either replacing the current lexer+parser pair with a custom reader in Python, or, alternatively, leaving the lexer+parser as the first stage that compiles the actual reader written in Hy.

scauligi:
Bumping this from the grave; I was fooling around and wrote a fairly straightforward reader/parser that reads one character at a time, but my initial prototype seems to actually run a bit _faster_ than the current `rply` lexer/parser.

If there's interest, I can implement the full grammar in this thing and get a PR going; having this would allow us to:
- Get rid of `rply` (which has stagnated)
- Implement constructs like `#_`, `#{...}`, etc. as proper reader macros instead of special-cased grammar rules
- Allow for user-defined reader macros...?(!)
- Have well-bracketed f-strings*, which I could not for the life of me figure out how to implement with `rply`
- Have more control over / have more informative syntax errors
- And also probably still have a (minor) performance improvement over the current lexer/parser

*Well-bracketed f-strings as in: `f"text {(expr 'with "embedded strings" (or f"other {fstrings}"))} more text"`

Kodiologist:
That definitely sounds interesting, although it would probably be better to use `funcparserlib` than write our own parser from scratch.

scauligi:
I tried a parser in `funcparserlib`* (as well as `parsita` and `parsimonious`), and they all end up slower than the current `rply` implementation. I also tried `lark`, which was just about comparable to `rply`.

The custom parser still beats all of these in terms of performance, while also easily allowing for a modifiable reader table for reader macros.

*My funcparserlib implementation is parsing by character instead of tokenizing, since early tokenization is what blocks us from better f-strings, custom readers, etc

----

Results from parsing a ~3100 line file, min of 25 runs reported
```
custom         0.1829 s
lark           0.3227 s
hy (rply)      0.3610 s
parsimonious   0.4295 s
funcparserlib  0.5004 s
parsita        0.5667 s
```

allison-casey:
hot dang scauligi this sounds really cool! do you have a fork we can look at?

Issue #1689, opened 2018-10-25
Macro name shadowing/collisions and removal


brandonwillard:
A user defined `require` macro will block the compilation of all `require` statements and there's little recourse for undoing this.  As far as I can tell, a user would have to remove the macro entry in `hy.macros._hy_macros` (or `__macros__` under #1682) and, since the macro dictionary is a Hy internal, that isn't a very reasonable recommendation.

Nearly the same situation occurs with plain functions:
```clojure
hy 0.15.0+39.gd2319dc using CPython(default) 3.6.6 on Linux
=> (defn test [] (print "test function"))
=> (defmacro test [] (print "test macro"))
<function <lambda> at 0x7fc95f8c0ea0>
=> (test)
test macro
```
Since we can't simply `(del test)` (the macro or the plain function) or redefine the function in order to regain it, this situation can easily lead to confusion and frustration.
(I imagine this might have an issue of its own already, but I didn't see one in my first pass.)
One can reset the Hy/Python session, but, when that's the easiest/best option, something needs to be fixed!

This general discord between macros, plain functions and Hy special forms like `require` is not a trivial issue, and, while we could make it easier to work around the aforementioned situations (e.g. have the compiler apply `del` to macros and tags), we should simply rethink and/or formalize the interaction between these three.

For instance, I believe I've mentioned/asked this before, but if the functions backing macros used their macro's names and were defined directly in their namespaces, then this wouldn't be an issue.

Anyway, thoughts?

Kodiologist:
One thing to keep in mind is that in at least one way, we're currently taking advantage of this: `hy.core.shadow` works by defining functions with the same names as special operators, so wherever you can't use a special form (e.g., in `(reduce + x)`, the symbol `+` can't be a special operator), the shadow function gets used instead.

It wouldn't be too unreasonable to forbid users from giving a macro the same name as a special form. That could be part of a solution, perhaps.

gilch:
In Clojure, if I recall, a special form in the function position always takes precedence over a macro. To use a macro with the same name as a special form, you'd have to use its fully qualified name; the unqualified name is assumed to mean the special form.

brandonwillard:
Out of curiosity, I made some minor edits that expose the macro functions at module level.
With those changes, one can do the following:
```clojure
hy 0.15.0+43.g31a630d using CPython(default) 3.6.6 on Linux
=> (defmacro test [] (print "macro"))
=> (test)
macro
=> test
[...]

PR #2101, opened 2021-06-17
Bespoke reader/parser for speed and extensibility

scauligi:
Very much still WIP, f-strings (among several other things) are not implemented yet.
Also needs refactoring, proper error handling, better naming, better organization, documentation, etc.
I'm opening a PR here just so that it's visible (@allison-casey https://github.com/hylang/hy/issues/722#issuecomment-863323392), but it will be a while before it's ready to be actually reviewed/merged.

But hey, user-defined reader macros!
```hy
; run this through `hy2py -s -np`

(eval-and-read
  (defn -#UPPER- [reader -]
    (.slurp_space reader)
    (assert (= "\"" (.getc reader)))
    (setv s (reader.read-rest-of-quote))
    (-> (s.upper) hy.models.String))
  (if (in (hy.mangle "&reader") (locals))
    (assoc &reader.reader-table "#UPPER" -#UPPER-)))

(defmacro "#upper" [s]
  (.upper s))

#upper "this string is a lowercase literal that gets compiled to uppercase later"
#UPPER "this string is an all-caps literal *at read time*!"
```


scauligi:
Hmm, after implementing the rest of the parser, this may have been somewhat ill-advised.
Either that or I refactored something very poorly.

```
custom         0.4069 s
hy (rply)      0.3009 s
```

scauligi:
Spoke too soon; it was due me adding a (recursive) call to `HyObject.replace` on every generated node.
Removing that (and just making sure that all generated nodes have source info) brings us back down to:
```
custom         0.2623 s
hy (rply)      0.3009 s
```
We're still in business, friendos!

allison-casey:
king :raised_hands: :crown:

Kodiologist:
You're still working on this, right?

scauligi:
Yup, it's still in the works!
I should probably push and write what I'm doing here so that folks can comment in on design choices as I'm going along.

scauligi:
I may be creeping on scope with this, but...

I have the reader yielding top-level forms as a generator now instead of tokenizing the entire file before it hits the compiler.
What does this allow you to do? Well, if you modify the reader (at compile-time), it takes effect immediately!

```hy
;; test_slicing.hy
(require slicing)

#: a:(+ 3 4):"slice slice"  ; treated as a generic tag-style macro,
                            ; limited by normal macro parsing semantics

(slicing.install-#:)        ; from here, #: is now special

#: a:(+ 3 4):"slice slice"  ; works!
```

After running through `hy2py` (spacing added for clarity):
```python
import hy
hy.macros.require('slicing', None, assignments='ALL', prefix='slicing')

hyx_Xnumber_signXXcolonX(hyx_aXcolonX)
3 + 4
hy.models.Keyword('')
'slice slice'

slice(a, 3 + 4, 'slice slice')
```

The first instance gets parsed as `(#: a:)  (+ 3 4)  :  "slice slice"`
since it's following the usual tag macro semantics, but the second
form gets handled by `#:` reader!

And of course the secret sauce:
```hy
;; slicing.hy
(defmacro install-#: []
  '(eval-when-compile
     (defn parse-node [&reader]
       (setv node
         (if (= ":" (.peekc &reader))
           None
           (.parse-one-node &reader)))
       (if (= node '...)
         'Ellipsis
         node))

     (defn read-slice [&reader key]
       (setv old-ends-ident (.copy &reader.ends-ident))
       (.add &reader.ends-ident ":")

       (setv nodes [])

       (.slurp-space &reader)
       (.append nodes (parse-node &reader))
       (while (.peek-and-getc &reader ":")
         (.append nodes (parse-node &reader)))

       (setv &reader.ends-ident old-ends-ident)
       `(slice ~@nodes))

     (assoc hy.&reader.reader-table "#:" read-slice)))
```

I can probably make this a bit easier to use, perhaps by adding a `defreader` macro or something of the sort.

----

Also managed to knock the speed up a tad by removing some more redundant stuff I was doing:
```
custom         0.2081 s
hy (rply)      0.3102 s
````

----

Also please excuse the mess of commits; I'm treating this as a dev branch rather than a PR branch at the moment, but will clean it up before the proper PR lands.

**Comment by Kodiologist** [comment]
> Well, if you modify the reader (at compile-time), it takes effect immediately!

That seems in scope to me. You can use an ordinary macro in the same file you define it, so it's nice to be able to do the same thing with a reader macro. I don't remember if other Lisps let you do this.

**Comment by scauligi** [comment]
So I think I'm pretty much ready to submit this as a PR, once I clean up the commits.

Currently, the way one modifies the reader table is by manually adding an entry to the `reader_table` dict in the currently active reader, which exposes itself as `hy.&reader` inside of `eval-*-compile` blocks (see the example I gave above).

It's quite... magic; I'd like to change this to something else but I'm not sure what a good way is of exposing the reader.

Alternatively, I could leave it as-is and create a new macro `defreader` that handles that for you.
I haven't thought too much about it, but perhaps something like
```hy
(defreader "#:" [reader]
  (do things with (.chars reader) and so on))
```

scauligi:
Oh I should probably mention, I snuck in a reader macro `#( ... )` for parsing tuples, and I also changed `#@` to a (somewhat un-lispy) reader macro that works much more like Python's decorator syntax.

Kodiologist:
> I snuck in a reader macro `#( ... )` for parsing tuples

Why use that in place of the `,` result macro?

scauligi:
`#(...)` parses to `(, ...)` since there's no model for tuples;
I tend to use tuples a lot and `#(...)` looks more like a top level structure (and is similar to Python tuple syntax).

allison-casey:
@scauligi mind if i take a crack at cleaning up the API? i'll fork and pr it into this branch. 

is there a reason that tuples are the only top level data structures that use a function position symbol? It does look a bit weird in hindsight.
```clojure
{"a": 1}
#{"a" "b"}
[1 2]
#(1 2)  ; which is a character shorter too vOv
; vs
(, 1 2)
```

scauligi:
@allison-casey By all means, that would be bomb! I'm going to be a bit busy with my defense/dissertation so I won't be too available for a while unfortunately.

Also I think you mentioned you're working through the error handling code in cmdline.py? I still haven't figured out what errors are supposed to go where, so I don't think I'm properly raising errors in the right places for the new reader, would you be able to take a look at that as well?

allison-casey:
Also, would yielding top level forms as a generator essentially turn hy's unit of compilation to top level forms like we've been talking about in #1689?

scauligi:
I believe it would, yeah, although I don't believe hy2py should be affected at all

Kodiologist:

PR #2102, opened 2021-06-19
consistent binding constructs

### LINKED ISSUE #1723: Inconsistent for syntax
I've noticed that `for` takes the iteration variable and iterable like `[var iterable]` whereas `lfor`, `dfor`, and `gfor` take it like `var iterable`:

```
(for [item iterable] body)
(lfor item iterable result)
(dfor item iterable [key val])
(gfor item iterable result)
```

while this isn't huge, it would be nice for these to be consistent, so `lfor` `dfor` and `gfor` would also look like `[item iterable]`.

allison-casey:
closes #1723 

This attempts to unify the syntax for all the paired binding constructs where the introduced bindings are intended to be used in some following `FORM` or `dolike` body (Such as `(let [...] ...)`, `(with [...] ...)`, `(except [...] ...)`, etc) by adding brackets around loopers such that they follow the form `(lfor [...] ...)`, `(sfor [...] ...)`, etc and also drops the brackets around `loop`'s bindings since they act as binding pairs and not lambda-list. 

Also does some cleanup of the comprehension compile methods since that thing had become a giant 130 line mess

### DIFF
--- a/hy/contrib/destructure.hy
+++ b/hy/contrib/destructure.hy
@@ -165,8 +165,8 @@ Iterator patterns are specified using round brackets. They are the same as list 
   "
   (setv gsyms [])
   `(do
-    (setv ~@(gfor [binds expr] (by2s pairs)
-                  sym (destructure binds expr gsyms)
+    (setv ~@(gfor [[binds expr] (by2s pairs)
+                  sym (destructure binds expr gsyms)]
               sym))
     (del ~@gsyms)))
 
@@ -179,9 +179,9 @@ Iterator patterns are specified using round brackets. They are the same as list 
         result (hy.gensym 'dict=:))
   `(do
      (setv ~result {}
-           ~@(gfor [binds expr] (by2s pairs)
+           ~@(gfor [[binds expr] (by2s pairs)
                    [k v] (by2s [#* (destructure binds expr gsyms)])
-                   syms [(if (in k gsyms) k `(get ~result '~k)) v]
+                   syms [(if (in k gsyms) k `(get ~result '~k)) v]]
                syms))
      (del ~@gsyms)
      ~result))
@@ -252,8 +252,8 @@ Iterator patterns are specified using round brackets. They are the same as list 
                    ~(if (isinstance target hy.models.Symbol)
                         (.get default target)))])
   (defn get-as [to-key targets]
-    (lfor t targets
-          sym (expand-lookup t (to-key t))
+    (lfor [t targets
+          sym (expand-lookup t (to-key t))]
       sym))
   (->> (.items binds)
        (starmap (fn [target lookup]
@@ -299,10 +299,10 @@ Iterator patterns are specified using round brackets. They are the same as list 
   (.append result `(list ~(.pop result)))
   (setv [bs magics] (find-magics binds)
         n (len bs)
-        bres (lfor [i t] (enumerate bs)
+        bres (lfor [[i t] (enumerate bs)]
                (destructure t `(.get (dict (enumerate ~dlist)) ~i) gsyms))
         err-msg "Invalid magic option :{} in list destructure"
-        mres (lfor [m t] magics
+        mres (lfor [[m t] magics]
                (ifp found m
                  ':as [t dlist]
                  ':& (destructure t (if (isinstance t hy.models.Dict)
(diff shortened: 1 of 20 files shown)


Kodiologist:
I still think brackets in forms like `(lfor  [x xs]  x)` are unnecessary.

scauligi:
They're redundant, but I do like the consistency between all the special binding forms.
@allison-casey this needs a NEWS entry, right?

allison-casey:
its generally more lispy it seems to have some kind of brackets around the binding forms at least in racket, scheme, and clojure. But so is brackets around `cond` which is present in pretty much all the other lisp languages *besides* clojure. @Kodiologist could we compromise by leaving brackets around `cond` alone and adding them back to `match` to make the conditional branch constructs consistent with other lisps i know you brought up the cond brackets/parens when we were talking about that and having looked around at other lisps i see what you were getting at better.

Kodiologist:
Hmm, well, when I get down to it, the reason I don't like adding the brackets here is because it adds punctuation that ultimately doesn't add much use. I mean, if the syntax was `(lfor [x xs] x)`, we'd presumably allow any number of body forms as in `for`, so you could say things like `(lfor [x xs] (print "Hello world!") x)`. But it's rare to want to do that; when you want multiple body forms, you're typically looping for side-effects and so you use `for` instead, which, not coincidentally, has the syntax you're proposing. So the inconsistency makes sense, although a user who could conceivably do everything with `for` or everything with `lfor` (or, of course, write their own macros instead). Horses for courses.

I should add that I'm warming up to the idea of debracketing in general. Whitespace is perfectly capable of visually grouping things. So if you want more consistency between `for` and the comprehension forms, I'd vote for dropping `for`'s brackets, so multiple body forms will need `do`.

allison-casey:
comprehensions do allow for that kind of thing with `:do`, i'd happily be gone with `:do` an instead give an implicit `do` to the body-expr of comprehensions since `:do` in a `for` is pretty much pointless. My main concern with dropping brackets around the `for` loopers is that it makes binding in loopers different from other similar binding forms i mentioned in the original pr. `with`, `let`, `except` all wrap their bindings in a bracket to be used in a following implicit `do`. 

Though if we drop the `:do` looper and giving `lfor`, `sfor`, `gfor`, and `dfor` an implicit do then _everything_ would actually be consistent. so:
```clojure
(lfor x (range 5) :do (print "hello") x)
```
would become
```clojure
(lfor [x (range 5)] (print "Hello") x)
```
which is how racket does list comps and lets use the more common implicit do idiom

Kodiologist:
> `with`, `let`, `except` all wrap their bindings in a bracket to be used in a following implicit do.

I think you're miscontruing what is essentially a concindence as the reason we have brackets. Putting aside `let`, which has always been the black sheep, `with`, `except`, and `for` all use brackets not to indicate something is bound but to separate variadic special syntax from the implicit `do`. If you don't need an implicit `do`, you don't need the brackets. Notice that the bracketed parts of `with` and `except` need not have any bindings (`with` can use an anonymous context manager and `except` can name a type of exception without binding it to a variable), and the brackets of a `for` can have things other than bindings, anyway.

scauligi:
> what is essentially a concindence as the reason we have brackets

While that is the historical reason for the brackets, it more or less also signifies binding forms at this stage in Hy's development, so I think it's worth making that an explicit choice now. I'm still strongly for bracket consistency; I still pause a half-second every time I'm writing a `*for` form to remember if I need brackets or not, and I would welcome just having brackets always even if `[lsgd]for` still only take a single a single body form with no implicit do-block.

allison-casey:
> all use brackets not to indicate something is bound but to separate variadic special syntax from the implicit do. If you don't need an implicit do, you don't need the brackets. Notice that the bracketed parts of with and except need not have any bindings 

This feels like a distinction without a difference. More so since a single `with` context manager is just a special case for binding to `_`. the fact that it compiles internally to an anonymous context manager in python is more an implementation detail than anything. In the end i agree with  scauligi that even though brackets were historically to separate the implicit do, at this point they read to people as indicating binding forms when used with a subsequent body expr and the lack of consistency in that respect makes hy's syntax more difficult to parse. 

> I would welcome just having brackets always even if [lsgd]for still only take a single a single body form with no implicit do-block.

There isn't really any consistency in lisps i can find about this. fennel and clojure only allow 1 form while racket has an implicit do so we have the latitude to do whichever we think is best. I don't have much of a preference on this point however.

Kodiologist:
Well, if consistency is the main thing that's getting you two, let's try to get rid of brackets, e.g. in `for`, instead of adding them where they aren't helpful.

scauligi:
In that case `for` would no longer have an implicit do-block? And how does that interact with the `else` clause?

Kodiologist:
Right. I guess `else` could work much as it does now, an optional expression at the very end which must begin with `else`.

Kodiologist:
Although we could switch it to use a keyword before the form itself, `:else`, so it looks more like the other parts of comprehension forms.

Kodiologist:
It doesn't look like this is going to happen, so I'm going to close this, but reopen it if that changes.

PR #2107, opened 2021-07-03
Flatten import/require syntax

### LINKED ISSUE #1612: import syntax improvements

scauligi:
Closes #1612.
Obsoletes #851 and #1614.

It's been several years now since the :-1: votes on #851, and we're revamping much of Hy's syntax anyway, so now's as good a time as any to try this (again).

Flattens `import`/`require` syntax to remove outer brackets.
Also removes brackets around `*` imports to "fix" a (very niche) edge case.

```hy
;;;; new syntax
(import foo            ; import foo
        bar [baz qux]  ; from bar import baz, qux
        twarp :as tw)  ; import twarp as tw

;; as one line, using double-space grouping style
(import foo  bar [baz qux]  twarp :as tw)

;; the "edge case"
(require my-operators *)    ; imports all macros from `my-operators`
(require my-operators [*])  ; imports just the single macro called `*`
```

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,5 +1,15 @@
 .. default-role:: code
 
+Unreleased
+==============================
+
+Other Breaking Changes
+------------------------------
+* `import` and `require` no longer need outer brackets.
+  `(import [foo [bar]])` is now `(import foo [bar])`
+  and `(import [foo :as baz])` is now `(import foo :as baz)`.
+  To import all names from a module, use `(import foo *)`.
+
 1.0a3 (released 2021-07-09)
 ==============================
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -615,12 +615,12 @@ base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
        ;; Import from a module
        ;;
        ;; Python: from os.path import exists, isdir, isfile
-       (import [os.path [exists isdir isfile]])
+       (import os.path [exists isdir isfile])
 
        ;; Import with an alias
        ;;
        ;; Python: import sys as systest
-       (import [sys :as systest])
+       (import sys :as systest)
 
        ;; You can list as many imports as you like of different types.
        ;;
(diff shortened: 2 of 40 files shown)


Kodiologist:
I'm in favor of this, but I'd like to release a new alpha soon and this will break almost every Hy program, so let's postpone it a bit.

Kodiologist:
Well done. You can inline `kids` and `prefix_name`, and you can add a NEWS entry in a new unreleased section.

Kodiologist (approved):

Kodiologist:
Here's a Python script I used to update all my `import` and `require` forms.

```python
#!/usr/bin/env python3

import sys, re

def f(m):
    operator, old = m.groups()
    chunks = ["(" + operator]
    substuff = False
    while old:
        if m := re.match(r'\s+', old):
            chunks[-1] += m[0]
        elif m := re.match(r'(?:\w|\.)+', old):
            chunks.append(m[0])
        elif m := re.match(r'''
               \[
                 ( [^\]\[]+ )
                 ( \[ [^\]]+ \] [^\]]* )?
               \]''',
               old, flags = re.VERBOSE):
            substuff = True
            chunks.append(m[1])
            if m[2]:
               chunks[-1] += '*' if re.match(r'\[\s*\*', m[2]) else m[2]
        else:
            raise(ValueError(repr(old)))
        old = old[m.end():]
    return (
        (' ' if substuff and not any('\n' in c for c in chunks) else '')
            .join(chunks) + ')')

for fpath in sys.argv[1:]:
    with open(fpath, 'rt') as o:
        code = o.read()
    
    code = re.sub('\((import|require)([^)]+)\)', f, code)

    with open(fpath, 'wt') as o:
        o.write(code)
```

PR #2110, opened 2021-07-04
better require interface

### LINKED ISSUE #1689: Macro name shadowing/collisions and removal

allison-casey:
closes #1689 

this makes the `hy.macros.require` method easier to call manually and makes a note about re-requiring macros in the api.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1108,6 +1108,23 @@ base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
       to do introspection of the current module's set of defined macros, which isn't
       really supported anyway.
 
+   .. note::
+
+      :strong:`Re-requiring shadowed macros`
+
+      Note that it's possible for a user to define a ``require`` macro that would
+      block ``require`` statements from working correctly. The core ``require`` macro
+      is syntactic sugar for the underlying ``hy.macros.require`` method which can be
+      used to re-require a shadowed ``require`` form::
+
+         => (defmacro require [x] "will block subsequent require statements")
+         => (require [hy.contrib.walk [let]])
+         "will block subsequent require statements"
+
+         => (hy.macros.require "hy.core.result_macros" ["require"])
+         => (require [hy.contrib.walk [let]])  ; will compile correctly
+
+
 .. hy:function:: (return [object])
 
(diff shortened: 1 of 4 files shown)


Kodiologist:
The documentation only discusses the very specific case of shadowing the name `require`, but you want to talk about how to recover from any core macro being shadowed, right? I'm not sure how general this technique is. `hy.macros.require` is still a function call, so in a program file, unlike the REPL, it would typically be executed too late, so you need `eval-when-compile`, and then what if `eval-when-compile` is what you shadowed? The right strategy is probably to call the `require` macro, not the function, with a fully qualified name, although that may not be fully implemented as things stand.

allison-casey:
Good point. Scauligi's new reader would likely turn Hy's unit of compilation into top level forms, since it returns a generator of tokens, which could mitigate the need for `eval-when/and-compile` guards. Let me do some more testing to find the edge cases though.

Kodiologist:

PR #2111, opened 2021-07-07
Use "environment markers" for conditional install_requires

tianon:
These are defined in PEP 508: https://www.python.org/dev/peps/pep-0508/#environment-markers

The current implementation means that installing 1.0a2 from the published wheel on Python 3.8 does not include `astor` (and thus does not work) because "the Python version that is building the wheel is different from the Python version that is installing it" (from https://hynek.me/articles/conditional-python-dependencies/).

I don't think Windows and Linux can share wheels, so `pyreadline` doesn't strictly have to be included here, but it seems more correct to use this method for both.

(Refs https://github.com/hylang/hy/commit/06f34b018e8bab6170c6a72c4a6e25201268cdb6 / https://github.com/hylang/hy/pull/1999)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,19 +61,16 @@ class Install(install):
         # https://github.com/pypa/setuptools/issues/456
         return install.run(self)
 
-install_requires = [
-    'rply>=0.7.7',
-    'funcparserlib>=0.3.6',
-    'colorama']
-if sys.version_info < (3, 9):
-    install_requires.append('astor>=0.8')
-if os.name == 'nt':
-    install_requires.append('pyreadline>=2.1')
-
 setup(
     name=PKG,
     version=__version__,
-    install_requires=install_requires,
+    install_requires=[
+        'rply>=0.7.7',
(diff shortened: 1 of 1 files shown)


tianon:
I guess at this point it might even make sense to embed this list directly in the `setup()` invocation again?  (Happy to adjust, as always :+1:)

Kodiologist:
Whoops. It's convenient that there happens to be a straightforward solution to this problem. Yes, `install_requires` can be inlined now.

tianon:
Oh, I suppose I should add that I've tested this successfully on both Python 3.8 and 3.9 :sweat_smile:

Kodiologist (approved):
Thanks for catching this so fast. I'll release 1.0a3 with just this fix once it's in.

tianon:
I was worried it would be harder but was pleased to find such a simple fix that I could contribute it myself. :smile:

scauligi (approved):

Kodiologist:
1.0a3 has been released.

PR #2115, opened 2021-07-12
Handle pyreadline failure (Windows + Python 3.10+)

### LINKED ISSUE #2114: Python 3.10 on Windows fails: AttributeError: module 'collections' has no attribute 'Callable'
Thanks to https://github.com/pyreadline/pyreadline/issues/65, Hy on Python 3.10 on Windows no longer starts, and a backtrace similar to the following is the result:

```console
  File "c:\python\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\python\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Python\Scripts\hy.exe\__main__.py", line 4, in <module>
  File "c:\python\lib\site-packages\hy\cmdline.py", line 37, in <module>
    from hy.completer import completion, Completer
  File "c:\python\lib\site-packages\hy\completer.py", line 18, in <module>
    import readline
  File "c:\python\lib\site-packages\readline.py", line 34, in <module>
    rl = Readline()
  File "c:\python\lib\site-packages\pyreadline\rlmain.py", line 422, in __init__
    BaseReadline.__init__(self)
  File "c:\python\lib\site-packages\pyreadline\rlmain.py", line 62, in __init__
    mode.init_editing_mode(None)
  File "c:\python\lib\site-packages\pyreadline\modes\emacs.py", line 633, in init_editing_mode
    self._bind_key('space',       self.self_insert)
  File "c:\python\lib\site-packages\pyreadline\modes\basemode.py", line 162, in _bind_key
    if not callable(func):
  File "c:\python\lib\site-packages\pyreadline\py3k_compat.py", line 8, in callable
    return isinstance(x, collections.Callable)
AttributeError: module 'collections' has no attribute 'Callable'
```

Unfortunately from that bug (and the repository in general), it would appear that `pyreadline` is no longer actively maintained (last update of any kind was in 2015), so for Hy to continue to support Windows on future Python versions will require some amount of workaround (a fork of `pyreadline`, dropping readline support on Windows, etc).

I'm not sure if there's a way to specify that Python 3.10+ and Windows are simply mutually incompatible, but that would be the "simplest" workaround if there's a clean way to specify that. :see_no_evil:

tianon:
Closes #2114

This is a bit hacky, but I'm not sure if there's a better way.  I'm confident on the `setup.py` change to avoid even installing `pyreadline` on Python 3.10+ (which in itself is enough to fix the problem since then the fallback happens properly), but I added more code to handle the `AttributeError` for cases where `pyreadline` gets installed separately and thus `import readline` still pulls it in (I'm not sure which cases actually invoke the `import pyreadline` fallback code -- maybe older versions of `pyreadline`?  older versions of Python's built-in `readline` module?).

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -9,10 +9,12 @@ Other Breaking Changes
   `(import [foo [bar]])` is now `(import foo [bar])`
   and `(import [foo :as baz])` is now `(import foo :as baz)`.
   To import all names from a module, use `(import foo *)`.
+* Dropped support for `pyreadline`
 
 Bug Fixes
 ------------------------------
 * Improved error messages for illegal uses of `finally` and `else`.
+* Fixed a crash on Windows due to Readline
 
 1.0a3 (released 2021-07-09)
 ==============================
--- a/hy/completer.py
+++ b/hy/completer.py
@@ -16,13 +16,14 @@ docomplete = True
 
 try:
     import readline
-except ImportError:
-    try:
-        import pyreadline.rlmain
-        import pyreadline.unicode_helper  # NOQA
-        import readline
-    except ImportError:
+except AttributeError as e:
+    # https://github.com/pyreadline/pyreadline/issues/65
+    if "module 'collections' has no attribute 'Callable'" in str(e):
         docomplete = False
+    else:
+        raise
+except ImportError:
+    docomplete = False
 
 if docomplete:
     if sys.platform == 'darwin' and 'libedit' in readline.__doc__:
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,6 @@ setup(
         'funcparserlib>=0.3.6',
         'colorama',
         'astor>=0.8 ; python_version < "3.9"',
-        'pyreadline>=2.1 ; os_name == "nt"',
     ],
     cmdclass=dict(install=Install),
     entry_points={


Kodiologist:
It looks like `pyreadline` is a dead library walking, so I would vote to just remove it entirely, and Windows users will have to use Cygwin if they want a decent CLI, as usual. So the branch `except ImportError` can be just `docomplete = False`. As for the other branch, can we do something more specific than checking for `AttributeError`? I feel like a lot of unforeseen bugs could trigger that.

tianon:
Yeah, makes perfect sense -- I've just force pushed another go at this that I hope is more in line with what you have in mind. :smile:

Kodiologist (approved):
Thanks. You can remove the comment referring to the Hy issue (the one referring to the pyreadline issue is useful, though, and should stay), and add a line to NEWS like "Fixed a crash on Windows due to Readline".

tianon:
Absolutely -- I also added an explicit note about `pyreadline`, but happy to adjust again if you'd prefer not to even mention that. :+1:

allison-casey (approved):

liyiliuxingyu:
Please see this link.
https://github.com/hylang/hy/issues/2114#issuecomment-950864067

PR #2118, opened 2021-07-13
Keep `let` from duplicating lambda lists

### LINKED ISSUE #2116: `let` duplicates the lambda list of function definitions
```
$ echo '(require hy.contrib.walk [let]) (let [] (defn f [a] 1))' | hy2py
...
def f(a):
    [a]
    return 1
```

`[a]` defines the parameters of `f` as it ought to, but is also then used as part of the function body. This is usually a no-op, but you can see a semantic fault when, for example, a default argument has side-effects:

```
(defn test-no-extra-eval-of-default-args []
  (setv l [])
  (defn f []
    (.append l 1))
  (let [a 1]
    (assert (= a 1))
    (defn g [[b (f)]]
      5))
  (assert (= (g) 5))
  (assert (= l [1])))
```

Kodiologist:
Fixes #2116.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -9,12 +9,11 @@ Other Breaking Changes
   `(import [foo [bar]])` is now `(import foo [bar])`
   and `(import [foo :as baz])` is now `(import foo :as baz)`.
   To import all names from a module, use `(import foo *)`.
-* Dropped support for `pyreadline`
 
 Bug Fixes
 ------------------------------
+* `let` should no longer re-evaluate default arguments.
 * Improved error messages for illegal uses of `finally` and `else`.
-* Fixed a crash on Windows due to Readline
 
 1.0a3 (released 2021-07-09)
 ==============================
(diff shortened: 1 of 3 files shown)


scauligi:
:+1: 
`(.tail self)` can be `(self.tail)` for consistency and `(hy.models.Symbol "defn")` can be shortened to just `'defn`.

-----

This is fine as-is, but maybe we can simplify the logic so it's a little clear what the intent is?
For example, something like
```hy
(setv head [(self.head)]
      tail (self.tail))
(if (= (get head 0) 'defn)
  (.append head (.pop tail 0)))
```
but maybe more functional/lispy for style points

or I guess at least a comment making it clear that we're dealing with the name that shows up in a `defn` form but not `fn`

Kodiologist:
Done.

scauligi (approved):

allison-casey (approved):

PR #2123, opened 2021-07-16
properly format bracketed strings in hy-repr

scauligi:
Also fixes a bug where FStrings models lost `brackets` information
after going through `as_model`.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -12,12 +12,15 @@ Other Breaking Changes
 * Functions that provide first-class Python operators, such as ``+``
   in constructs like ``(reduce + xs)``, are no longer brought
   into scope automatically. Say ``(import hy.pyops *)`` to get them.
+* The constructors of `String` and `FString` now check that the input
+  would be syntactically legal.
 
 Bug Fixes
 ------------------------------
 * `let` should no longer re-evaluate default arguments.
 * Improved error messages for illegal uses of `finally` and `else`.
 * `match` should no longer re-evaluate subect.
+* `hy-repr` now properly formats bracketed strings.
 
 New Features
 ------------------------------
--- a/hy/core/hy_repr.hy
+++ b/hy/core/hy_repr.hy
@@ -116,15 +116,17 @@
 (hy-repr-register [hy.models.Symbol hy.models.Keyword] str)
 (hy-repr-register [hy.models.String str hy.models.Bytes bytes] (fn [x]
   (setv r (.lstrip (_base-repr x) "ub"))
-  (+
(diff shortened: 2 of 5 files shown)


Kodiologist:
Don't forget some tests.

Kodiologist:
Looks good. I would write e.g. `x.brackets` instead of `(. x brackets)`, and you might also like to use an f-string or two, like `f"#[{x.brackets}[{x}]{x.brackets}]"`.

Kodiologist (approved):

scauligi:
done and done!
imagine, using f-strings in a PR dealing with f-strings... sigh

allison-casey (approved):

Kodiologist:
Oops, I realized a problem.

    => (hy.models.String "]foo]" :brackets "foo")
    '#[foo[]foo]]foo]
    => '#[foo[]foo]]foo]
    … hy.lex.exceptions.LexException

I think you need to change the constructors of some models to forbid syntactically illegal strings like this, similar to what I did in #2064.

allison-casey:
how do we feel about this fix for forbidding syntactically illegal bracket strings?

scauligi:
Need to make sure `brackets is not None` for those checks but otherwise that should work

allison-casey:
good catch, fixed

Kodiologist:
Good, but I think this change needs a test and a NEWS line, too.

allison-casey:
fixed bug in bracket detection in fcomponents, added tests, and news item. 

Assuming we're good with this. Mind if i rebase and clean this all up @scauligi?

scauligi:
dooooope yeah go for it!

allison-casey:
Done

Kodiologist:
I guess you use `isinstance(node, (String, str))` in place of just `isinstance(node, String)` because `node` may not yet have been fully boxed. But I think it would be better to call `as_model` first, in case any other objects get promoted to `String`s.

allison-casey:
That check was because adjacend `String` nodes are joined during construction and that returned a `str` and not a `String`. changed it to return `String`'s always

scauligi:
Alternatively you could implement `__add__` for HyObjects to wrap superclass `__add__`, which will also fix a similar thing happening for Lists; I think I did that in the reader PR, you can steal it from there

Kodiologist (approved):

allison-casey:
> Alternatively you could implement `__add__` for `Objects` to wrap superclass `__add__`

I like this, but i'd like to play around with it some to poke around the edge cases. I'll merge this and either open a PR for that specifically or just leave it as part of the reader PR

Issue #1994, opened 2021-03-15
`setx` is scoped inconsistently in comprehension forms


scauligi:
Illustrative example:
```
=> (lfor x (range 3) (setx last x))
[0, 1, 2]
=> last
2
```
But:
```
=> (lfor x (range 3) :do (print x) (setx also-last x))
0
1
2
[0, 1, 2]
=> also-last
[...] NameError: name 'also_last' is not defined
```

This is because the former is compiled to
```python
[(last := x) for x in range(3)]
last  # same scope as assignment
```
whereas the latter is
```python
def _hy_anon_var_1():
    for x in range(3):
        print(x)
        yield (also_last := x)
[].__class__(_hy_anon_var_1())
also_last  # different scope from assignment
```

Generally, if the body of `lfor` (or `dfor` or `gfor`) is not a simple expression, then hy compiles it to an anonymous function. That means anything with a `:do` clause, any body with a `(do)` form, and probably many macros. However, this changes the scope of the body expression in unexpected ways.

I don't have a proposed solution for this, just pointing out that this happens and is weird.

peaceamongworlds:
I justified this to myself by saying that list comprehensions in Python have a separate internal scope. Hence things that happen within the `:do` clauses might not be seen outside of the comprehension.

However, 
```hy
=> (lfor x (range 3) (setx last x))
[(last := x) for x in range(3)]

[0 1 2]
=> 
```
this is weird. It seems to me to be more of an issue with Python's lack of block scope and things leaking through in unexpected ways sometimes.

Kodiologist:
It was always my intention that the comprehension forms be regarded as introducing a new function-level scope. The fact that a real comprehension statement, with no new scope, can be produced by Hy in some cases wasn't supposed to affect the user, because there formerly was no way to set variables in a Python comprehension other than the loop variables, which are scoped to the comprehension just as if it were a function. `setx` has thrown a wrench in the works. I guess the thing to do is to force the function-compilation mode when `setx` is used, although this would introduce a perhaps surprising inconsistency from Python. It was quite intentional on Python's part to leak `:=` assignments in comprehensions, and we'd be preventing that.

scauligi:
This affects more than just `setx`, though admittedly all edge-case why-are-you-doing-this cases. E.g. `(lfor x (locals) expr-vs-stmt)` will have different behavior depending on whether it's an expression or function context.

allison-casey:
[...]

Issue #817, opened 2015-06-22
When doing a setv, magic (nonlocal) if the symbol is in a parent node's assignment place


paultag:
When doing a setv, magic `(nonlocal)` if the symbol is in a parent node's assignment place

I have no idea if this is doable. Like, at all.

Perhaps it's doable if we work out "Assignment" operators, and store names in parent scopes in the Hy Result object?

kirbyfan64:
Heh, I actually tried this once in Python 2 and 3. The main issues stemmed from two things:

1. Sometimes you actually *don't* want a variable to be nonlocal.
2. Being a dynamic language, Python makes it a pain to do analysis like this. It's definitely possible, but there are many, many corner cases.

paultag:
:+1: my vague emotions as well

Kodiologist:
While we may in the future implement a special operator that does something like this, I hope we can agree that ordinary Python assignment should continue to be available, whether it's named `setv` or something else.

Kodiologist:
Yeah, we might add a feature sort of like this, but trying to make the default semantics different from Python's is probably not wise.

PR #2125, opened 2021-07-17
yet another let implementation

scauligi:
Once again, a major breaking change that nobody asked for, even though I haven't even finished the last one.

Here's a (draft) implementation that adds tracking of scoping information, allowing us to implement `let` as a first-class result macro.

I did my best to have `let` bindings interact nicely with Python's normal scoping rules; that is, let bindings are as close to possible as standard assignment.

### Examples!

This example is fairly straightforward; `x` refers to the name outside the function.
```hy
(let [x 1]
  (defn foo []
    (print x)))
(foo)
```
```python
_hy_scope_1_x = 1
def foo():
    return print(_hy_scope_1_x)
foo()
```

If we assign to `x` inside the function, then per Python scoping rules, `x` is entirely local to the function despite any assignment outside:
```python
# showing "actual" assignment for this example
x = 1
def foo():
    print(x)  # UnboundLocalError
    x = 8
foo()
```
Therefore, this is how the corresponding Hy compiles:
```hy
(let [x 1]
  (defn foo []
    (print x)
    (setv x 8)))
(foo)
```
```python
_hy_scope_1_x = 1
def foo():
    print(x)  # UnboundLocalError, as "expected"
    x = 8
foo()
```
However, if we use a `nonlocal`/`global`, then all is well again!
```hy
(let [x 1]
  (defn foo []
    (global x)
    (print x)
    (setv x 8)))
(foo)
```
```python
_hy_scope_1_x = 1
def foo():
    global _hy_scope_1_x
    print(_hy_scope_1_x)
    _hy_scope_1_x = 8  # assigns to outer `x`!
foo()
```

I need to experiment further, but I think the scope tracking can also help us solve the problem from #1994 with `setx` and scope weirdness with generator forms.

### DIFF
--- a/hy/scoping.py
+++ b/hy/scoping.py
@@ -1,0 +1,369 @@
+"Scope and variable tracking for Hy/Python scopes."
+
+import ast
+import itertools
+from abc import ABC, abstractmethod
+
+import hy._compat
+from hy.errors import HyInternalError
+from hy.lex import mangle
+from hy.models import Expression, List, Symbol
+
+
+def is_function_scope(scope):
+    return isinstance(scope, ScopeFn) and scope.is_fn
+
+
+def is_inside_function_scope(scope):
+    "True if any enclosing scope (including this one) is a function scope."
+    cur = scope
+    while not isinstance(cur, ScopeGlobal):
+        if is_function_scope(cur):
+            return True
+        cur = cur.parent
+    return False
+
+
+def nearest_python_scope(scope):
+    "Return the closest enclosing scope that corresponds to a Python scope."
+    cur = scope
+    while isinstance(cur, (ScopeLet, ScopeGen)):
+        cur = cur.parent
+    return cur
+
+
+class NodeRef:
+    """
+    Wrapper for AST nodes that have symbol names, so that we can rename them if
+    necessary. Each `NodeRef` corresponds to one Python identifier. `ast` nodes
+    that accept multiple identifier names (`global`, `nonlocal`, etc) have
+    their specific identifier referenced by their `index` in the list of
+    `names` provided by the `ast` node.
+    """
+
+    ACCESSOR = {
+        ast.Name: "id",
+        ast.Global: "names",
+        ast.Nonlocal: "names",
+    }
+    if hy._compat.PY3_10:
+        ACCESSOR.update(
+            {
+                ast.MatchAs: "name",
+                ast.MatchStar: "name",
+                ast.MatchMapping: "rest",
+            }
+        )
+
+    def __init__(self, node, index=None):
+        self.node = node
+        self.index = index
+        self._accessor = NodeRef.ACCESSOR[type(self.node)]
+
+    @property
+    def name(self):
+        res = getattr(self.node, self._accessor)
+        if self.index is not None:
+            return res[self.index]
+        return res
+
+    @name.setter
+    def name(self, new_name):
+        "Used to rename `ast` identifiers"
+        if self.index is not None:
+            getattr(self.node, self._accessor)[self.index] = new_name
+        else:
+            setattr(self.node, self._accessor, new_name)
+
+    @staticmethod
+    def wrap(f):
+        "Decorator to convert AST node parameter to NodeRef."
+
+        def _wrapper(self, node, index=None):
+            if not isinstance(node, NodeRef):
+                node = NodeRef(node, index)
+            return f(self, node)
+
+        return _wrapper
+
+    def __repr__(self):
+        return (
+            f"NodeRef(name={self.name}, node={type(self.node).__name__},"
+            f" index={self.index})"
+        )
+
+
+class ScopeBase(ABC):
+    def __init__(self, compiler):
+        self.parent = None
+        self.compiler = compiler
+
+    def create(self, scope_type, *args):
+        "Create new scope from this one."
+        return scope_type(self.compiler, *args)
+
+    def __enter__(self):
+        self.parent = self.compiler.scope
+        self.compiler.scope = self
+        return self
+
+    def __exit__(self, *args):
+        self.compiler.scope = self.parent
+        self.parent = None
+        return False
+
+    # Scope interface
+    @abstractmethod
+    def access(self, node, index=None):
+        "Called when a symbol is accessed."
+        ...
+
+    @abstractmethod
+    def assign(self, node, index=None):
+        "Called when a symbol is assigned to."
+        ...
+
+    @abstractmethod
+    def define(self, name):
+        """
+        Called when a symbol is defined.
+        (e.g., function names, class names, imports)
+        """
+        ...
+
+    @abstractmethod
+    def define_nonlocal(self, node, root):
+        "Called when a symbol is declared nonlocal or global."
+        ...
+
+
+class ScopeGlobal(ScopeBase):
+    """Global scope."""
+
+    def __init__(self, compiler):
+        super().__init__(compiler)
+        self.defined = set()
+        "set: of all symbols created or assigned to in this scope"
+
+        self.nonlocal_vars = []
+        """
+        list: of all `nonlocal`'s defined in this scope.
+
+        Explicitly not a `set` so that we maintain the order they were defined in
+        """
+
+    def __exit__(self, *args):
+        nonlocal_vars = self.nonlocal_vars
+        self.nonlocal_vars = []
+        if not self.defined.issuperset(nonlocal_vars):
+            raise SyntaxError(f"no binding for nonlocal '{nonlocal_vars[0]}'")
+        return super().__exit__(*args)
+
+    @NodeRef.wrap
+    def access(self, node, index=0):
+        return node.node
+
+    @NodeRef.wrap
+    def assign(self, node):
+        self.define(node.name)
+        return node.node
+
+    def define(self, name):
+        self.defined.add(name)
+
+    def define_nonlocal(self, node, root):
+        if root == "nonlocal":
+            self.nonlocal_vars.extend(node.names)
+        else:
+            self.defined.update(node.names)
+
+
+class ScopeLet(ScopeBase):
+    """
+    Scope that supports let-binding by renaming bound symbols as they are
+    accessed/assigned. Defined symbols are never renamed.
+    """
+
+    def __init__(self, compiler):
+        super().__init__(compiler)
+        self.bindings = {}
+
+    def _rename_if_bound(self, node):
+        if node.name in self.bindings:
+            node.name = self.bindings[node.name]
+            return True
+        return False
+
+    @NodeRef.wrap
+    def access(self, node):
+        self._rename_if_bound(node) or self.parent.access(node)
+        return node.node
+
+    @NodeRef.wrap
+    def assign(self, node):
+        self._rename_if_bound(node) or self.parent.assign(node)
+        return node.node
+
+    def define(self, name):
+        self.bindings.pop(name, None)
+        self.parent.define(name)
+
+    def define_nonlocal(self, node, root):
+        # remove nonlocal defs of any let scopes in this Python scope
+        cur = self
+        while isinstance(cur, ScopeLet):
+            for name in node.names:
+                if root == "nonlocal":
+                    if name in cur.bindings and cur is not self:
+                        node.names.remove(name)
+                else:
+                    cur.bindings[name] = name
+            cur = cur.parent
+        cur.define_nonlocal(node, root)
+
+    def add(self, target, new_name=None):
+        """Add a new let-binding target, mapped to a new, unique name."""
+        if isinstance(target, (str, Symbol)):
+            if "." in target:
+                raise ValueError("binding target may not contain a dot")
+            name = mangle(target)
+            if new_name is None:
+                new_name = self.compiler.get_anon_var(f"_hy_let_{name}")
+            self.bindings[name] = new_name
+            if isinstance(target, Symbol):
+                return Symbol(new_name).replace(target)
+            return new_name
+        if new_name is not None:
+            raise ValueError("cannot specify name for compound targets")
+        if isinstance(target, List):
+            return List(map(self.add, target)).replace(target)
+        if (isinstance(target, Expression) and target and
+                target[0] in (Symbol(","), Symbol("unpack-iterable"))):
+            return Expression([target[0], *map(self.add, target[1:])]).replace(
+                target
+            )
+        raise ValueError(f"invalid binding target: {type(target)}")
+
+
+class ScopeFn(ScopeBase):
+    """Scope that corresponds to Python's own function or class scopes."""
+
+    def __init__(self, compiler, args=None):
+        super().__init__(compiler)
+        self.defined = set()
+        "set: of all vars defined in this scope"
+        self.seen = []
+        "list: of all vars accessedto in this scope"
+        self.nonlocal_vars = set()
+        "set: of all `nonlocal`'s defined in this scope"
+        self.is_fn = args is not None
+        """
+        bool: `True` if this scope is being used to track a python
+        function `False` for classes
+        """
+
+        if args:
+            for arg in itertools.chain(
+                args.args, args.posonlyargs, args.kwonlyargs, [args.vararg, args.kwarg]
+            ):
+                if arg:
+                    self.define(arg.arg)
+
+    def __exit__(self, *args):
+        for node in self.seen:
+            if node.name not in self.defined or node.name in self.nonlocal_vars:
+                # pass unbound/nonlocal names up to parent scope
+                self.parent.access(node)
+        return super().__exit__(*args)
+
+    @NodeRef.wrap
+    def access(self, node):
+        self.seen.append(node)
+        return node.node
+
+    @NodeRef.wrap
+    def assign(self, node):
+        self.access(node)
+        self.define(node.name)
+        return node.node
+
+    def define(self, name):
+        self.defined.add(name)
+
+    def define_nonlocal(self, node, root):
+        ((self.nonlocal_vars if root == "nonlocal" else self.defined)
+            .update(node.names))
+        for n in self.seen:
+            if n.name in node.names:
+                raise SyntaxError(
+                    f"name '{n.name}' is declared {root} after being used"
+                )
+        if root == "nonlocal":
+            # toss all nonlocal names up to parent scope
+            for i in range(len(node.names)):
+                self.parent.access(node, i)
+
+
+class ScopeGen(ScopeFn):
+    """
+    Scope that supports generator forms (`lfor`, `gfor`, ...). If this scope is
+    contained within a function scope or the global scope, then any variable
+    assignments are "leaked out" to the parent scope, mimicking Python's inline
+    generator semantics.
+
+    .. note::
+       see :pep:`572#why-not-use-a-sublocal-scope-and-prevent-namespace-pollution`
+       for more details on how and when variables are leaked into enclosing scopes.
+    """
+
+    def __init__(self, compiler):
+        super().__init__(compiler)
+        self.iterators = set()
+        self.assignments = []
+        self.nonlocals = set()
+        self.exposing_assignments = False
+
+    def __enter__(self):
+        super().__enter__()
+        enclosing = nearest_python_scope(self.parent)
+        if isinstance(enclosing, ScopeGlobal) or is_function_scope(enclosing):
+            # only "leak out" assignments if we're contained in a function
+            # (or global) scope
+            self.exposing_assignments = True
+        return self
+
+    def finalize(self):
+        """
+        Access and return all the identifiers created and potentially leaked
+        out by this generator. This should be called once and only once we've
+        processed all of the iterators in this generators so as not to leak out
+        any unwanted identifiers
+        """
+        res = set()
+        for node in self.assignments:
+            if node.name not in self.nonlocal_vars:
+                self.parent.access(node)
+                res.add(node.name)
+        return sorted(res)
+
+    @NodeRef.wrap
+    def assign(self, node):
+        self.access(node)
+        if node.name not in self.defined:
+            self.assignments.append(node)
+        return node.node
+
+    def iterator(self, target):
+        """
+        Declare an iteration variable name for this scope; as in Python, the
+        iteration variable(s) cannot be reassigned.
+        """
+        self.iterators.update(
+            name.id for name in ast.walk(target) if isinstance(name, ast.Name)
+        )
+        # remove potentially leakable identifiers that were actually iteration
+        # variables found in `target`
+        self.assignments = [
+            node for node in self.assignments if node.name not in self.iterators
+        ]
+        self.seen = [node for node in self.seen if node.name not in self.iterators]
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -23,11 +23,17 @@ Other Breaking Changes
 
   * Functions: `butlast`, `coll?`, `constantly`, `dec`, `destructure`, `distinct`, `drop-last`, `end-sequence`, `flatten`, `inc`, `macroexpand-all`, `parse-args`, `pformat`, `postwalk`, `pp`, `pprint`, `prewalk`, `readable?`, `recursive?`, `rest`, `saferepr`, `walk`
   * Classes: `PrettyPrinter`, `Sequence`
-  * Macros: `#%`, `#:`, `->`, `->>`, `ap-dotimes`, `ap-each`, `ap-each-while`, `ap-filter`, `ap-first`, `ap-if`, `ap-last`, `ap-map`, `ap-map-when`, `ap-reduce`, `ap-reject`, `as->`, `assoc`, `cfor`, `comment`, `defmacro!`, `defmacro/g!`, `defmain`, `defn+`, `defn/a+`, `defseq`, `dict=:`, `do-n`, `doto`, `fn+`, `fn/a+`, `ifp`, `let`, `let+`, `lif`, `list-n`, `loop`, `ncut`, `of`, `profile/calls`, `profile/cpu`, `seq`, `setv+`, `smacrolet`, `unless`, `with-gensyms`
+  * Macros: `#%`, `#:`, `->`, `->>`, `ap-dotimes`, `ap-each`, `ap-each-while`, `ap-filter`, `ap-first`, `ap-if`, `ap-last`, `ap-map`, `ap-map-when`, `ap-reduce`, `ap-reject`, `as->`, `assoc`, `cfor`, `comment`, `defmacro!`, `defmacro/g!`, `defmain`, `defn+`, `defn/a+`, `defseq`, `dict=:`, `do-n`, `doto`, `fn+`, `fn/a+`, `ifp`, `let+`, `lif`, `list-n`, `loop`, `ncut`, `of`, `profile/calls`, `profile/cpu`, `seq`, `setv+`, `smacrolet`, `unless`, `with-gensyms`
 
 * The constructors of `String` and `FString` now check that the input
   would be syntactically legal.
 * `hy.extra.reserved` has been renamed to `hy.reserved`.
+* Hy scoping rules more closely follow Python scoping in certain edge cases.
+* `let` is now a core macro. Semantics of `let` have changed in certain
+  edge cases: Any assignment or assignment-like operations (`with`, `match`,
+  etc.) will *assign* to the let-bound name, while any definintions or
+  definition-like operations (`defn`, `defclass`, `import`) will *shadow* the
+  let-bound name (and will also continue to be defined after the `let`-scope ends).
 
 Bug Fixes
 ------------------------------
@@ -40,6 +46,9 @@ Bug Fixes
 * Fixed a bug with self-requiring files on Windows.
 * The `repr` and `str` of string models now include `brackets` if
   necessary.
+* Complex comprehensions are now always treated as inline expressions for
+  variable scoping, rather than creating a new block scope. Specifically,
+  assignments within comprehensions are now always visible to the surrounding code.
 
 New Features
 ------------------------------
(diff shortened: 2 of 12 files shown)


scauligi:
Cool, I was able to "fix" the generator scoping issue #1994 by making Hy mimic Python's scoping rules in generators!
As an added bonus, `(nonlocal)` now gets translated to `global` if there's no enclosing function scope, so it's always ok to use.

Tests currently failing because of `setx` forms in the tests that I haven't compat-gated.
I also haven't investigated how any of this interacts with 3.10 match statements yet.

allison-casey:
this is very very cool scauligi! i'm back from vacation for a bit so i'll start looking through this and the new reader pr and see what i see

scauligi:
And now it interops with `match` expressions! And tested, too!
<sub>...I should probably get back to my dissertation...</sub>

Kodiologist:
Hey, I wish I procrastinated on my thesis as productively as you do.

scauligi:
Finally cleaned up and ready for review! Although I still need to add documentation for `let` and also should add documentation for the new scoping rules, now that we actually have defined scoping.

Brief summary of changes:
- `let` and `let*` provide parallel and sequential let-bindings
- `nonlocal` compiles to `global` when inside top- or first-level functions
- generators act as if they were inline even if they get compiled to anonymous functions (fixes #1994, as long as you're not doing any shenanigans with `(locals)` etc.)

Notes on `let` bindings:
- Any assignment or assignment-like operations (`with`, `match`, etc.) will _assign_ to the let-bound name
- Any definintions or definition-like operations (`defn`, `defclass`, `import`) will _shadow_ the let-bound name (and also will extend beyond the scope of the `let`)

Examples:
```hy
(let [x 3]
  (with [x (open "foo")])  ; assigns to let-bound `x`
    (print x)))  ; file object
(print x)  ; error: `x` is unbound

(let [re 4]
  (import re)
  (print re))  ; no longer let-bound since it was shadowed
(print re)  ; this still works
```
```python
_hy_scope_1_x = 3
_hy_anon_var_1 = None
with open('foo') as _hy_scope_1_x:  # assigns to let-bound name
    _hy_anon_var_1 = print(_hy_scope_1_x)
print(x)  # error: `x` is unbound

_hy_scope_2_re = 4
import re  # imports are not let-bound
print(re)  # from inside the let, but re was shadowed!
print(re)  # from outside the let, this still works
```

hy/scoping.py
```diff
@@ -0,0 +1,243 @@
...
+            self.node.names[self.index] = new_name
+        elif hy._compat.PY3_10:
+            if isinstance(self.node, (ast.MatchAs, ast.MatchStar)):
+                self.node.name = new_name
+            elif isinstance(self.node, ast.MatchMapping):
+                self.node.rest = new_name
+        self.node.id = new_name
+
+
+class Scope:
```
allison-casey, line 38: If this is intended as an abstract base clase, it should extend from `abc.ABC`

hy/scoping.py
```diff
@@ -0,0 +1,243 @@
...
+        if not isinstance(node, NodeWithName):
+            node = NodeWithName(node, index)
+        self._access(node)
+
+    def assign(self, node, index=None):
+        if not isinstance(node, NodeWithName):
+            node = NodeWithName(node, index)
+        self._assign(node)
+
+    def _access(self, node):
```
allison-casey, line 82: and should be decorated with `@abc.abstractmethod`

hy/scoping.py
```diff
@@ -0,0 +1,243 @@
...
+        pass
```
allison-casey, line 83: is it valid to access these `pass` methods on an instance of `Scope` or should this raise an error?

hy/scoping.py
```diff
@@ -0,0 +1,243 @@
...
+        if isinstance(nonlocal_node, ast.Nonlocal):
+            for i in range(len(nonlocal_node.names)):
+                self.parent.access(nonlocal_node, i)
+
+
+class ScopeStack:
+    """Scoping information that Hy adds on top of Python's scoping rules."""
+    def __init__(self):
+        self.last_id = 0
+        self.scope = Scope(self, self.last_id)
```
allison-casey, line 230: no other code instantiates `Scope` directly. would it be more appropriate to make it a true abstract base class inheriting from `abc.ABC` and adding a `ScopeRoot` class that is instantiated here instead?

hy/scoping.py
```diff
@@ -0,0 +1,243 @@
...
+        self.stack.scope = self.parent
+        return False
+
+    def is_global_scope(self):
+        return self.id == 0
+
+    def is_function_scope(self):
+        return isinstance(self, ScopeFn) and self.is_fn
+
+    def inside_function_scope(self):
```
allison-casey, line 58: predicate method doesn't following `is_` semantic used in other predicate methods in this class

hy/scoping.py
```diff
@@ -0,0 +1,243 @@
...
+                raise ValueError("binding target may not contain a dot")
+            name = mangle(target)
+            new_name = f"_hy_scope_{self.id}_{name}"
+            self.bindings[name] = new_name
+            return Symbol(new_name).replace(target)
+        elif isinstance(target, List):
+            return List(map(self.add, target)).replace(target)
+        elif isinstance(target, Expression) and target:
+            if isinstance(target[0], Symbol) and str(target[0]) in (",", "unpack-iterable"):
+                return Expression([target[0], *map(self.add, target[1:])]).replace(target)
```
allison-casey, line 128: this fails confusingly when attempting to use a star expression as the root symbol
```clojure
(let [#* a [1 2 3]] a)
```
`walk.let` fails like
```clojure
    (let [#* a [1 2 3]])
    ^------------------^
expanding macro let
  TypeError: bind targets must be symbol or iterable unpacking assignment
```

Kodiologist:
Sorry if I'm being dense, but: why? We already have a `let`, and it's less tightly coupled to the compiler than this version.

hy/scoping.py
```diff
@@ -0,0 +1,243 @@
...
+        self.node.id = new_name
+
+
+class Scope:
+    def __init__(self, stack, scope_id):
+        self.stack = stack
+        self.id = scope_id
+
+    def __enter__(self):
+        self.parent = self.stack.scope
```
allison-casey, line 44: parent isn't defined in `__init__` and will raise an unexpected exception in certain edge cases. i.e.

```clojure
=> (let [a 1 b 2] (nonlocal a) (global b))
  File "stdin-bf89d46423db3acec3100628e39c19de80650ffb", line 1, in <module>
    (let [a 1 b 2] (nonlocal a) (global b))
  File "<stdin>", line 1
    (let [a 1 b 2] (nonlocal a) (global b))
                   ^----------^
expanding macro nonlocal
  AttributeError: 'Scope' object has no attribute 'parent'
```
`self.parent` should be initialized in `__init__` to `None`

Kodiologist:
I'm not sure this is less brittle. It was formerly the case that all the code for `let` was in one place. Now we have a large amount of bookkeeping code scattered throughout the compiler, with a sizable support module, all to support this feature. `let` has always made me uncomfortable because of the herculean bug-fixing effort involved in what is ultimately a measly feature that fails to fix the mistakes of Python's scoping. I was hoping that by moving it to Hyrule, it could eventually be killed off more easily, or at least end up somewhere that core Hy developers didn't have to maintain it.

All this said, there's no doubt of the user demand for `let`, and if this version fixes a lot of bugs present in the old version, without introducing its own bugs, then that's justification enough. Make sure to add tests for those bugs, though. Relatedly, we want to remove the old `let` implementation and run its tests with the new one, right?

scauligi:
> less tightly coupled to the compiler

Not exactly; the contrib version of `let` is iterating directly over un-macro'd Hy AST after all, and essentially implements half of the compiler functionality on its own since it needs to expand all macros and special-case all special forms.

> Now we have a large amount of bookkeeping code scattered throughout the compiler, with a sizable support module, all to support this feature.

While the scope tracking code is necessary to enable in-house `let`, it also gives us many other benefits as well; I view it as an investment for a variety of other features rather than *just* for `let`:
- fix scoping issues when comprehensions are promoted to anonymous functions
- with a small amount of extra effort, we can get rid of `contrib.walk` entirely, since scope tracking takes care of its main use case: finding and renaming symbols within the AST
- like @allison-casey mentioned, we can use this for richer debug/error information
- eventually: better macro scope semantics? (i.e., no longer need to explicitly re-import/require/define helper modules and functions within quasiquotes, since we will know at compile time which names are in the local scope of the macro)
- eventually: related to the above, clojure-style namespacing features? (i.e., treat python modules as namespaces that have some non-shadowable way of being accessed)

> I'm not sure this is less brittle.

For partly the reasons above, I disagree! (But also I wrote this so I'm fairly biased...)
Instead of having the giant AST walker that is `contrib.walk` with all of its special casing for various forms, we now have a well-defined internal API that lets us clearly mark which forms access/assign/define symbols and which forms create new namespaces/scopes.

In fact, I argue it's more robust: If we add a new form, we only need to make sure any scoping calls within the new form's result macro are correct; the existing scope tracking system will automatically ensure it interacts properly with everything else.

Compare, for example, the scoping patch for match expressions (currently 5ff09957, "match expression handling", <10 small hunks inside one function) vs the code we needed to get `contrib.walk` to support match statements (0495662c, "Feature: python 10 'match' statement", ~50 loc reimplementing match pattern traversal).

> `let` has always made me uncomfortable because of the herculean bug-fixing effort involved

Same here, that's partly why I think this implementation will make things much easier for us!

> the mistakes of Python's scoping.

yeah you got me there


> Relatedly, we want to remove the old let implementation and run its tests with the new one, right?

Yup, tests I've added include all the tests (or at least the ones that make sense) from the `contrib.walk` test file. @allison-casey if you have examples of those other bugs/edge-cases you found, can you send them my way?

Kodiologist:
Related: #817.

Kodiologist:
It means that if the compiler saw `(setv x …)` when a variable named `x` already existed in a parent scope, it would add `(nonlocal x)` or `(global x)`. Which is similar to what you're doing here, if I'm not mistaken.

scauligi:
Ah, gotcha. Yeah I have an automatic nonlocal, but it only happens when we promote comprehensions to anonymous functions, so it shouldn't catch any users by surprise.

Kodiologist:
> I was able to "fix" the generator scoping issue #1994

I don't see a test or NEWS item for this.

> As an added bonus, `(nonlocal)` now gets translated to global if there's no enclosing function scope

Nor this, nor a mention of it in the documentation for `nonlocal`.

Your NEWS edit describes `let` as if it were a new feature, but users have already had `let` for a while. Instead you need to describe it as a breaking change: it's now available by default instead of being required from `contrib.walk`, and it has different semantics. It's not clear from the documentation nor from tests (since the heritage from the old `let` tests to the new is not made clear in the commit history) how the semantics are different. (Of course, the documentation for this `let` doesn't need to compare it to the old one. Just the NEWS entry does.) Also, the bugfix that you deleted mention of is also in the new implemention, if I understand correctly.

You mention `let*` in https://github.com/hylang/hy/pull/2125#issuecomment-886412625, but it's not in any of the commits.

> let's get asty

This commit message is unintelligible. The new code in this commit has a lot of redundancies that can be removed with loops.

> add hy/scoping.py

It's not clear from this commit message or from any docstrings or comments in the module what the module does. I see redundancy in the module, too. For example, the setter and getter definitions of `NodeWithName.name` are almost complete duplicates of each other, and so are `ScopeLet._access` and `ScopeLet._assign`.

```
# right now only ScopeFn overrides this;
#   should consider allowing nonlocals in other scopes
```

Ideas for future changes to the code should go in the issues list, not in the code.

Instead of having an abstract class `ScopeBase` which has a lot of stub methods that raise errors and then a concrete child class `ScopeGlobal` which also has stub methods, but they `pass` instead of raising errors, you can combine these into one class.

It seems roundabout to use hook methods like `_on_enter`, `_on_exit`, `_access`, and `_assign` when you can just override `__enter__` or `access` or whatever in subclasses, using `super` as appropriate.

It's hard to see what `self.is_global_scope()` adds beyond `isinstance(self, ScopeGlobal)`.

That's probably enough for now.

scauligi:
Eep, clearly I didn't clean this up as much as I should have. Thanks for taking the time to sift through this regardless.

> Your NEWS edit describes let as if it were a new feature, but users have already had let for a while.

Good point, I was treating the new `let` as a completely different feature, but from a user perspective it pretty much is just `let` being moved from contrib to core.

> Also, the bugfix that you deleted mention of is also in the new implemention, if I understand correctly.

Ech, this was again me treating `let` as a new feature instead of a "fix"/move of the old one and thus thinking that the news line of removing contrib's `let` superseded the bugfix. The new `let` doesn't have this bug, so I'll un-delete the bugfix line.

> `let*`

Originally I had `let` and `let*` which bind respectively in parallel and sequentially, as in lisps such as Racket or Scheme.
But, turns out, Clojure only has `let`, and it binds sequentially, so I changed the implementation to follow that. (Plus, that's the behavior that contrib's `let` had, so that'll stay consistent.)
Related, I'll also add a comment in this PR with the new summary of changes after I properly clean everything up so that that doesn't get confusing.

> NodeWithName redundancy

The NodeWithName setter needs to mutate its underlying AST node in-place, so I'm not quite sure the best way to remove this code duplication.

> ScopeBase vs ScopeGlobal

This came from https://github.com/hylang/hy/pull/2125#discussion_r676721303 and I think it's valuable to do; `ScopeBase` defines the interface, and `ScopeGlobal` makes it clear that the global scope itself does nothing special. If I remove the hook functions as suggested, then `ScopeGlobal` will probably just be `class ScopeGlobal(ScopeBase): pass`, but at least that will still keep it a distinct type for checking whether a given scope is the global one.

The remaining points I agree with, and I'll fix those up along with fixing the commits/docs/news.

Kodiologist:
> The NodeWithName setter needs to mutate its underlying AST node in-place, so I'm not quite sure the best way to remove this code duplication.

How about something like this (untested)?

```
def name_helper(self):
    return (
        'INDEX' if isinstance(self.node, (ast.Global, ast.Nonlocal)) else
        'name' if PY3_10 and isinstance(self.node, (ast.MatchAs, ast.MatchStar)) else
        'rest' if PY3_10 and isinstance(self.node, ast.MatchMapping) else
        'id')

@property
def name(self):
    return (self.node.names[self.index]
        if self.name_helper() == 'INDEX':
        else getattr(self, self.name_helper()))

@name.setter
def name(self, new_name)
   if self.name_helper() == 'INDEX':
      self.node.names[self.index] = new_name
   else:
      setattr(self, self.name_helper(), new_name)
```

It's clumsy because Python doesn't have a proper way to pass around lvalues, but at least it avoids duplicating a bunch of logic.

Kodiologist:

scauligi:
I'm baaaaaaaaack from the dead :tada: :tada: :confetti_ball: 
Here's a (hopefully) much cleaner PR!

Kodiologist:
Oh yes. This looks much more polished. Thanks for putting the time into it.

You've forbidden `global` and `nonlocal` in comprehension forms and in `let`. This is a regression; they seem to work fine on master. Can you fix it?

scauligi:
Now `nonlocal`/`global` work inside `let` and comprehensions (with accompanying tests!)

Kodiologist:
Very good.

"typo" isn't a great commit message because typos of different kinds of material (code vs. comments vs. error messages vs. documentation) vary in seriousness. If you still want to be telegraphic, you could say "typo in error message".

Do you have to use `if attr in node._attributes` in `Asty._replace_pos`, instead of `hasattr`, or `getattr` with a default value?

`scoping.py`
-----------------------------------

The exception handler in `NodeRef.__init__` doesn't work because `KeyError` would be raised in this case, not `IndexError`; try, for example, `NodeRef(1)`. (You'd think that `KeyError` would be a subtype of `IndexError`, right? But it isn't. Stupid Python. Instead, there's a different supertype for the both of them, `LookupError`.) But I don't think you need a `try` here anyway, because if I understand correctly, an error shouldn't happen here from a user's mistake, only a Hy implementer's mistake, so the default backtrace should be fine.

I notice that the docstring of `ScopeBase.create` says "Create new child scope", but it doesn't set `parent`. `parent` is set in `__enter__` and then not cleared in `__exit__`. Perhaps it would be more appropriate to set `parent` in `create` than `__enter__`.

It seems a little better with respect to separation of concerns to define `is_function_scope` as `return False` in `ScopeBase` and then override it with `return self.is_fn` in `ScopeFn`. Speaking of which, it might be clearer to rename `ScopeFn` to `ScopeFC`, since it's for both functions and classes.

Likewise (although, with the new definition, this is somewhat more verbose to the same end, so I would understand if you don't want to do it), you could call a method `….is_global()` instead of `isinstance(…, ScopeGlobal)`. Likewise `….is_let()` instead of `isinstance(…, ScopeLet)`.

The logic of `ScopeLet.add` can be simplified; e.g., the top-level `else` is not necessary, because the preceding `if` always ends in a return.

```
for name in ast.walk(target):
    if isinstance(name, ast.Name):
        self.iterators.add(name.id)
```

This passage could be written more directly as

```
self.iterators.update(name.id
    for name in ast.walk(target)
    if isinstance(name, ast.Name))
```

That's probably enough for now.

Kodiologist:
What's the deal with `smacrolet`, though? It seems not to be implemented.

Kodiologist:
I'm skeptical of the value of the detailed white-box test `test_compiler_nonlocal_global`. I think a simple native black-box test that showed that `nonlocal` could see a global variable would be better; something like

```
(setv my-global 7)

(defn test-nonlocal-as-global []
  (assert (= my-global 7))
  (defn f []
    (nonlocal my-global)
    (setv my-global 11))
  (f)
  (assert (= my-global 11)))

(test-nonlocal-as-global)
```

But in fact, this fails (`SyntaxError: no binding for nonlocal 'my_global' found`). Shouldn't it pass, or do I misunderstand the feature?

gilch:
I'm still not sure if I understand how the new let scope works. When I wrote the code-walking let, I considered direct variable assignments, but dict assignments worked better. One of the main uses of macros is to change when evaluation happens in a way that isn't possible for strict function evaluation. This could often mean expanding to hidden function definitions. For a simple example of such usage, in Hissp's `if-else` macro
```
(defmacro if-else (test then otherwise)
  "``if-else`` Basic ternary branching construct.

  Like Python's conditional expressions, the 'else' clause is required.
  "
  `((lambda (,'test : :* ,'then-else)
      ((operator..getitem ,'then-else (operator..not_ ,'test))))
    ,test
    (lambda : ,then)
    (lambda : ,otherwise)))
```
we expand to lambdas. Now suppose something like this is wrapped in a let. In cases like that, would the user have to realize that he is using a macro, know which parts expand to function definitions, and manually add a `nonlocal` form? The current let doesn't require that because the dict assignments cut though the scopes.

allison-casey:
I'll be back as well end of this month. In the middle of moving across the country :worried: I can help with this PR then if it hasn't already been merged

scauligi:
hokay lotta' stuff to catch up on.

> Lots of comments/suggestions/fixes from @Kodiologist 

Will do! I agree with pretty much all of these, thanks (again) for taking the time to review this rather large PR.

> smacrolet

`smacrolet` serves a different purpose from `let`. I can (re)implement it as a result macro if we want it, but even then I feel it's best left as a separate PR.

> `nonlocal` acting as `global`

My goal for this was to have `nonlocal` turn into `global` when the nearest definition of the variable is in fact a global definition. Turns out this will need a little bit of additional tracking in the scopes, so I'm just going to remove this feature from this PR, as it really is unrelated to the functionality of `let` other than piggy-backing on the scope information (I'll admit, I got overeager with this PR set). I'll submit it as a separate PR so we can discuss its implementation there.

> semantics of new `let`, assignments vs dict-assignments @gilch 

As much as possible, the new `let` semantics mirror the semantics of `setv` (just scoped to the syntactic form rather than the enclosing Python function/class).

If I'm understanding your example correctly, your `if-else` macro will still break for non-`let` variables -- it'll only properly handle reassignments to variables bound by Hyrule's implementation of `let`, since the use of `lambda`s to delay execution also creates implicit scopes, which is not the intention of this macro (and thus dict-assignment `let` is really just a hack).
An alternative method is to use a generator to delay evaluation instead, since (unlike lambdas) they do not create an enclosing semantic scope, and so will work even with non-dict assignments.

Here is an example:
```hy
(require hyrule [defmacro/g!
                 let :as mlet])

(defmacro if-else-lambda [test then otherwise]
  `((fn [test #* then-else]
      ((get then-else (not test))))
    ~test
    (fn [] ~then)
    (fn [] ~otherwise)))

(defmacro/g! if-else-gfor [test then otherwise]
  `((fn [test #* then-else]
      (get (list (get then-else (not test))) 0))
    ~test
    (gfor ~g!_ [0] ~then)
    (gfor ~g!_ [0] ~otherwise)))

(setv res (int (input "enter a number: ")))

(print)
(print "using if-else-lambda")
(setv x "unset")
(mlet [y "unset"]
  (let [z "unset"]
    (if-else-lambda (> res 5)
             (do (print "  (> 5 branch)")
                 (setv [x y z] (* ["> 5"] 3)))
             (do (print "  (<= 5 branch)")
                 (setv [x y z] (* ["<= 5"] 3))))
  (print "  setv:" x)
  (print "  mlet:" y)
  (print "   let:" z)))

(print)
(print "using if-else-gfor")
(setv x "unset")
(mlet [y "unset"]
  (let [z "unset"]
    (if-else-gfor (> res 5)
             (do (print "  (> 5 branch)")
                 (setv [x y z] (* ["> 5"] 3)))
             (do (print "  (<= 5 branch)")
                 (setv [x y z] (* ["<= 5"] 3))))
  (print "  setv:" x)
  (print "  mlet:" y)
  (print "   let:" z)))
```
```
enter a number: 4   
                    
using if-else-lambda
  (<= 5 branch)     
  setv: unset       
  mlet: <= 5        
   let: unset       
                    
using if-else-gfor  
  (<= 5 branch)     
  setv: <= 5        
  mlet: <= 5        
   let: <= 5        
```
**NOTE:** I did find a bug in testing this, in that generators weren't picking up let-bound names. You'll need the patch I just pushed for this to work properly for result-macro `let`, but my point still stands with its interaction with `setv`.

Kodiologist:
> `smacrolet` serves a different purpose from `let`. I can (re)implement it as a result macro if we want it, but even then I feel it's best left as a separate PR.

The current implementation of `smacrolet`, at least, reuses a lot of logic used by the current implementation of `let`, in the internal class `SymbolExpander`. So it looks like this PR would require us to keep `SymbolExpander` around and continue to maintain it, which defeats part of the purpose of the PR, right? Either that or we have to drop `smacrolet` entirely. 

> I'm just going to remove this feature [`nonlocal` acting as `global`] from this PR, as it really is unrelated to the functionality of `let` other than piggy-backing on the scope information (I'll admit, I got overeager with this PR set). I'll submit it as a separate PR so we can discuss its implementation there.

Okay, makes sense.

scauligi:
> SymbolExpander

That's a good point. I'd like to keep this PR from growing too much, so how about I migrate `smacrolet` as a separate PR, and then once `let` and `smacrolet` are integrated we can remove `hyrule/let.hy`?

Kodiologist:
Migrate it in what sense? Reimplement it with your new scoping tools?

scauligi:
Yeah, "migrate" as in reimplement using the scoping classes.

Kodiologist:
Okay. Yeah, you have a point; this PR is already big.

gilch:
Common Lisp:
```Lisp
* (let ((x 1))
    ((lambda ()
       (setq x 2)))
    x)

2
```
`let` declares a variable and its lexical scope. `setq` updates the variable without changing its scope.

I designed `let` so the equivalent Hy works the same way.
```Hy
(require [hy.contrib.walk [let]])

(let [x 1]
  ((fn []
     (setv x 2)))
  x) ; => 2
```
How can this work without a dict?

gilch:
I did consider making `let` readonly and single-assignment. You could maybe disallow assignments by detecting `setvs` and raising an error. Such a thing would be able to use direct gensym locals because it disallows the nonlocal reassignment edge case. Clojure's `let` is single assignment. (As is Hissp's basic let macro.) But the ability to create a mutable closure does have uses.

Kodiologist:
> How can this work without a dict?

The compiler adds a `nonlocal` statement. That's what this big new scoping module is for, to keep track of which variables are supposed to exist where so the compiler can do things like that.

Kodiologist:
Oh man, Sunjay, I'm a riot. I told you to replace `is_global_scope` with `isinstance` in August and then told you to basically do the opposite a month later. Evidently I can't make up my mind. Thinking about it, I now say: brevity is king, and being rigid about OO-style abstraction isn't particularly helpful when there is no expectation that subclasses will ever be defined outside this program, so I was probably right the first time.

scauligi:
> Common Lisp vs Hy

Hy isn't Common Lisp though; it's rooted in Python's existing semantics. As such, Hy's `fn` _is_ Python's `lambda` (or `def`) and Hy's `setv` _is_ Python's assignment statement. And Python already has clearly defined semantics for what assignment to an otherwise undeclared variable means inside of a function scope.
Thus to me, this Hy example
```hy
(let [x 1]
  ((fn []
     (setv x 2)))
  x)
```
**should** produce 1 and not 2, as that is directly analogous to Python's existing scope semantics:
```python
x = 1
def _anon():
    x = 2

_anon()
print(x)  # => 1
```
If you want to break out of a function scope, then use `nonlocal`, just as in Python:
```hy
(let [x 1]
  ((fn []
     (nonlocal x)
     (setv x 2)))
  x)  ; => 2
```
or instead use a generator, which doesn't create a new scope:
```hy
(let [x 1]
  (lfor _ (range 1)
     (setv x 2))
  x)  ; => 2
```

For general use cases, we could probably package up single-item generators in a macro for creating and using thunks.

allison-casey:
So what's left in this PR before we can initiate a final review?
just the moving of `smacrolet` out into it's own PR?

Kodiologist:
That and spinning out the `nonlocal`-to-`global` change and my comments from September.

scauligi:
Hey sorry I've been dragging my feet on this, I think I've been dealing with a bit of burnout after my dissertation. I'll try and get the changes in soon

Kodiologist:
I don't mind so long as we can wrap this up in time for the next release.

allison-casey:
@scauligi I can take a crack at hitting those last changes if reviewing my changes would be easier for you than coding yourself rn. Burnout's rough :heart:

scauligi:
That would be a boon, thanks! Feel free to hit me up though if I can answer questions about anything

allison-casey:
@scauligi sounds good! is there a way to talk to you outside of this PR, email or discord maybe (you can find my work email on my profile)? I'd rather not clutter this PR with these kinds of questions

allison-casey:
I've gone ahead and fixed the issues brought up as well as one or two @scauligi and I found @Kodiologist. Let me know if we missed anything.

Kodiologist:
Did you see [this comment](https://github.com/hylang/hy/pull/2125#issuecomment-923225015)? For example, you still have a commit message of "typo", there's still the `if attr in node._attributes`, and there's still the `except IndexError`.

allison-casey:
> commit message of "typo"

fixed

> if `attr in node._attributes`

changed to `hasattr`

> there's still the `except IndexError`

got rid of the `try/except`

> docstring of ScopeBase.create says "Create new child scope"

Removed `child` from docstring. `create` is just used to create new scopes with references to the currently instantiated compiler

> `parent` is set in `__enter__` and then not cleared in `__exit__`

cleared `parent` on exit

Also moved the helper functions out of the abstract super class and brought them into module scope. Children using `isinstance` of other children is fine so i left the is `ScopeGlobal` check in `ScopeGen` alone. If this seems like a good enough solution, i'll squash it into the scoping commit

This whole module could probably be simplified using mixins, but that's a job for another day.

Kodiologist:
> If this seems like a good enough solution, i'll squash it into the scoping commit

I am neutral as to whether these are functions or methods.

Kodiologist:
I'm still seeing a bunch of lingering issues, like the documentation saying that you can use `nonlocal` to affect a global name even thought that feature was removed. Tell you what. This PR is already a team effort, so let me know once you've decided what to do with that last commit and rebased onto master, and then I'll try to tie up all the loose ends.

allison-casey:
squashed the last commit and removed the ref to `nonlocal` acting as `global`. Feel free to go ahead and fix any issues I missed :+1:

Kodiologist:
Okay, I've read through and edited the PR. Here's a summary of changes, most of which are just things I asked for in September. If you think it's good, merge that sucker.

- Simplified the logic of `ScopeLet.add`.
- Eliminated use of a temporary variable in `ScopeFn.assign_local`.
- Edited "scoping for generators (pt 1)" so that it's same as the subsequent commit but without the `with` and indent, as Sunjay intended originally.
- Removed some `+=` statements in favor of just returning the same value later.
- Use return values from scope methods instead of temporary variables.
- Added a Hyrule commit hash to the message for the commit that copies a file from Hyrule.
- Changed a use of `let_scope` per my previous comment.
- Removed tests of exception messages when the exception message is already Python-standard.
- Replaced an `(assert True)` with a check of a return value.
- Undangled a parenthesis.
- Grouped some of the new documentation material into paragraphs.
- Minor documentation wording improvements.
- Changed the message about `let` and deallocation per my previous comment.
- Deleted the mention of `let` in the tutorial.
- Melded "bugfix: gen-exposed vars weren't reflecting let-binds" into the appropriate earlier commits.

Kodiologist:
Rats, there's a GitHub test failure on Python 3.10, which I think is unrelated to this PR. I'll look into it.

allison-casey:
oh its just our `python_requires` constraint being too strict. we should remove the `<=3.10`

Kodiologist:
Fixed and rebased.

allison-casey:
we all good to merge then?

allison-casey (approved):

Kodiologist:
Should be. Bombs away!

PR #2121, opened 2021-07-15
Make the shadow functions opt-in

Kodiologist:

Issue #1747, opened 2019-02-12
Slow startup when Hy is installed from a wheel


Kodiologist:
Testing the new release of 0.16.0, I see that startup is much slower when installing from the wheel than from the source distribution or directly from the repository. Likewise for older Hy releases. Even when I make sure the `__pycache__`s are included in the wheel and I can see they're installed. Either there's something wonky with my system, or wheel installation doesn't play nicely with premade byte-compiled files.

brandonwillard:
Maybe a bytecode version thing?  If you turn on debug, does it say it's recompiling?

Kodiologist:
Sorry, which debug mode do you mean? Should `HY_DEBUG` print such messages?

brandonwillard:
I was thinking of [`-d`](https://docs.python.org/3.6/using/cmdline.html#cmdoption-d).

Kodiologist:
I don't seem to get any messages about compilation from `python3 -d -m hy.core.language`.

PR #2126, opened 2021-07-17
Get Hy installing again

Kodiologist:
In #2121, I broke installation. `pip install -e` works, as we use in GitHub Actions for testing, but not without `-e`, because then `Install.__compile_hy_bytecode` runs, which tries to call `importlib.util.cache_from_source` with a nonexistent keyword argument `optimize` (a typo for `optimization`). The `cache_from_source` call was never previously executed because the glob `"hy/**.hy"` matched nothing (what was meant was `"hy/**/*.hy"`) until I added `hy.pyops`. If fixed, this code still wouldn't help, because `cache_from_source` just returns a path, without compiling anything. In short, I don't think #1865 ever really worked.

Rather than try to fix bytecode compilation, I've just removed the custom `Install` class, which makes `pip install` without `-e` work again. #1747 can be regarded as already covering the question of compiling and installing bytecode for Hy's standard library, so I haven't made any new issues.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,6 @@
 #!/usr/bin/env python
 
-import glob
-import importlib
-import inspect
-import os
-import sys
-
 from setuptools import find_packages, setup
-from setuptools.command.install import install
 import fastentrypoints   # Monkey-patches setuptools.
 
 from get_version import __version__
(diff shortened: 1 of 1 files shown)


scauligi:
I'll admit, I don't actually understand the setup tooling nor how to properly test it, which is why I haven't given a thumbs up or down on this. But at least nothing in Github CI is breaking?

PR #2127, opened 2021-07-19
 Remove copyright and license headers from code

Kodiologist:
In the spirit of continuing to shake off status-quo bias, here's another attempt at #1260.

I can't imagine a situation in which a court would allow a license violation because the individual files didn't have license headers, even though the repository has a license file in the usual place and the license is tagged in `setup.py`, GitHub, and PyPI. License headers are just a holdover from the uncertain early days of GNU. They're noise at the top of every file and they create diff noise every year when they get updated. Now that I'm working on Hyrule, I'd like to get rid of these first so we don't have to deal with them in both repositories.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -52,21 +52,6 @@ unsure of the proposed change, create an issue to discuss it before
 you write any code. This will allow others to give feedback on your
 idea, and it can avoid wasted work.
 
-File headers
-------------
-
-Every Python or Hy file in the source tree that is potentially
-copyrightable should have the following header (but with ``;;`` in
-place of ``#`` for Hy files)::
-
-      # Copyright [current year] the authors.
-      # This file is part of Hy, which is free software licensed under the Expat
-      # license. See the LICENSE.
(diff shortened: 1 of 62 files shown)


allison-casey:
cpython doesn't have any of this either. lets get rid of it

allison-casey (approved):

brandonwillard:
> @scauligi @brandonwillard When you press the thumbs-up button on the PR, do you mean to indicate that you approve it? Because I don't get email notifications for comment reactions, so if you do approve it, you should make a comment or a review.

Sometimes I'm not in a position to properly review explicit changes, but I'm able to quickly read a written summary of changes and show support for that.  This was one of those cases.

scauligi (approved):
So I was trying to look around online to see what the general consensus was and couldn't find any general consensus, but...
The Cpython code base doesn't seem to have license headers, at least not on the couple random files I looked at. So I'm cool with this 👍

scauligi:
ah didn't see that @allison-casey beat me to it orz

PR #2128, opened 2021-07-20
Feature/calls in dot forms

### LINKED ISSUE #1108: Consider extending Hy's `.` form to accept calls
Clojure's `..` can have method calls. This is especially useful when chaining accessors, [for example](http://clojure.org/reference/java_interop#_the_dot_special_form):

``` Clojure
 (.. System (getProperties) (get "os.name"))
```

expands to `.`, which can also have calls:

``` Clojure
(. (. System (getProperties)) (get "os.name"))
```

This doesn't appear to work in Hy:

``` Hy
=> (import os)
=> (. os (getcwd))

  (. os (getcwd))
        ^------^
HyTypeError: b'The attribute access DSL only accepts HySymbols and one-item lists, got HyExpression instead'
```

The error message seems to indicate that this should work, is this a bug? And why restrict it to one item? Clojure doesn't.

This makes chaining of accessors awkward (--spy):

``` Hy
=> (. ((. ((. os getcwd)) isalpha)) __class__ __name__)
os.getcwd().isalpha().__class__.__name__
'bool'
```

Although the `->` macro helps.

``` Hy
=> (-> os (.getcwd) (.isalpha) (. __class__) (. __name__))
os.getcwd().isalpha().__class__.__name__
'bool'
```

But Hy's `.` _can_ chain when not using calls

``` Hy
=> (. "" __class__ __name__)
''.__class__.__name__
'str'
```

This makes it more like Clojure's `..`, than Clojure's `.`, which can only take two arguments.

The proposed extended form would look like this: 

``` Hy
=> (. os (getcwd) (isalpha) __class__ __name__ [0])
os.getcwd().isalpha().__class__.__name__[0]
'b'
```

 The calls could also accept arguments, as Clojure.

allison-casey:
closes #1108 

This allows expressions in the attribute access macro `.` that compile to method calls, such that the following is now possible

```clojure
=> (. os (getcwd) (isalpha) __class__ __name__ [0])
os.getcwd().isalpha().__class__.__name__[0]
'b'
=> (. "ab hello" (strip "ab ") (upper))
"ab hello".strip("ab ").upper()
"HELLO"
```

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -18,6 +18,10 @@ Bug Fixes
 * `let` should no longer re-evaluate default arguments.
 * Improved error messages for illegal uses of `finally` and `else`.
 
+New Features
+------------------------------
+* the attribute access macro `.` now accepts method calls.
+
 1.0a3 (released 2021-07-09)
 ==============================
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -88,19 +88,19 @@ base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
 
    ::
 
-       (. foo bar baz [(+ 1 2)] frob)
+       (. foo (bar "qux")  baz [(+ 1 2)] frob)
 
    Compiles down to:
 
    .. code-block:: python
 
-       foo.bar.baz[1 + 2].frob
+       foo.bar("qux").baz[1 + 2].frob
 
    ``.`` compiles its first argument (in the example, *foo*) as the object on
    which to do the attribute dereference. It uses bare symbols as attributes
-   to access (in the example, *bar*, *baz*, *frob*), and compiles the contents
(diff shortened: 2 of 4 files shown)


Kodiologist:
Looks good, but why the extra newline in `test-attribute-access`?

allison-casey:
fixed

Kodiologist (approved):

scauligi:
Dope!
It looks like we're explicitly blocking `unpack-*` forms, but what about other specially-parsed forms like `quote` or `annotate`? (Should we do something about them explicitly?)

```
=> (. x ^y)
x.annotate(y)

=> (. x 'y)
x.quote(y)
```

Kodiologist:
While funny, that sort of behavior is probably not worth trying to prevent.

scauligi (approved):

PR #2129, opened 2021-07-21
Fix: filter internal importlib frames from stacktrace

### LINKED ISSUE #1693: Remove internal `importlib` frames from traceback
Python 3.x uses `_call_with_frames_removed` to remove internal frames from tracebacks.  I tried it in #1687, but couldn't get it to work; might be due to this [Python issue](https://bugs.python.org/issue23773).  
Regardless, it would be good to remove those frames somehow.

allison-casey:
closes #1693 

don't think this really merits a NEWS update. it doesn't affect users in any real way

### DIFF
--- a/hy/_compat.py
+++ b/hy/_compat.py
@@ -1,9 +1,11 @@
 import sys
+import platform
 
 PY3_7 = sys.version_info >= (3, 7)
 PY3_8 = sys.version_info >= (3, 8)
 PY3_9 = sys.version_info >= (3, 9)
 PY3_10 = sys.version_info >= (3, 10)
+PYPY = platform.python_implementation() == "PyPy"
 
 
 if not PY3_9:
--- a/hy/errors.py
+++ b/hy/errors.py
@@ -8,6 +8,7 @@ from functools import reduce
 from colorama import Fore
 from contextlib import contextmanager
 from hy import _initialize_env_var
+from hy._compat import PYPY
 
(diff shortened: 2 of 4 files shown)


Kodiologist:
Exception filtering is undocumented, so I agree that no NEWS line is required. But we should probably have a regression test, because it would be easy for a Python update to break this.

allison-casey:
added

allison-casey:
you were right, what an absolute pain in the ass. Pypy, < 3.8, and > 3.8 all handle it differently. added the test and fixed the inconsistencies across versions

Kodiologist:
Yeah, phew, thanks for wrangling that.

It looks like you could do the `_tb_hidden_modules` update a little less redundantly with something like

```
_tb_hidden_modules.update(
    f'<builtin>/frozen {x}' if PYPY else f'<frozen {x}>'
    for x in 'importlib._bootstrap', 'importlib._bootstrap_external')
```

I'm worried that the test is fragile because we're checking against the exact formatting of the backtrace, including even the line number of the test. Maybe just `assert "importlib" not in captured_w_filtering` and call it a day. I don't think you need `captured_wo_filtering` anymore, either.

allison-casey:
simplified the tb hidden modules update and removed the stacktrace string equality check

Kodiologist (approved):

scauligi (approved):
:metal:

PR #2131, opened 2021-07-25
remove just-in-time imports in top-level module

scauligi:
```python
import sys, hy

for module in sys.modules:
    if module.startswith('hy'):
        print(module)
```

```
hy.version
hy.lex.exceptions
hy.core
hy.errors
hy.models
hy.model_patterns
hy.lex
hy._compat
hy.macros
hy.compiler
hy.importer
hy.lex.lexer
hy.lex.parser
hy.core.result_macros
hy.core.macros
hy.core.language
hy
```

Everything except `hy.core.hy_repr` already gets imported (likely due to `hy.importer`), so the just-in-time machinery seems unnecessary.

### DIFF
--- a/hy/__init__.py
+++ b/hy/__init__.py
@@ -15,42 +15,9 @@ hy.importer._inject_builtins()
 
 from fractions import Fraction as _Fraction  # For fraction literals
 
-# Import some names on demand so that the dependent modules don't have
-# to be loaded if they're not needed.
+from hy.compiler import hy_eval as eval
+from hy.lex import read, read_str, mangle, unmangle
+from hy.models import as_model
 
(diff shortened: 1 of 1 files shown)


Kodiologist:
That's a bug. The whole point of the just-in-time imports is that you can `import hy` without a slowdown from loading Hy code. Probably some imports that were supposed to be in functions were moved to the top level.

scauligi:
Hmm, in that case I'll work this PR the other way and get these modules to not import until they're really needed

Kodiologist:
Sounds good. But you might want to wait until I have in the PR that will move a lot of things to Hyrule anyway.

PR #2135, opened 2021-07-27
add pre-commit hooks

### LINKED ISSUE #2134: Upgrade codebase to >= 3.6 python syntax
Hy still has a lot of code that was written to be compatible with python versions < 3.6 that make the code base noisier than it needs to be. Not using f-strings, classes extending from `object`, verbose `super(<type>, self)` calls that can all be removed now to simplify the code. 

This is a pretty easy one time job as a pre-commit with `pyupgrade` so we never have to deal with this again. You can find examples of pre-commits in other python repos like [flask](https://github.com/pallets/flask/blob/main/.pre-commit-config.yaml), [pandas](https://github.com/pandas-dev/pandas/blob/master/.pre-commit-config.yaml), [django](https://github.com/django/django/blob/main/.pre-commit-config.yaml), [scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/main/.pre-commit-config.yaml), etc.

allison-casey:
closes #2134 

pre-commits ensure consistency across the code base; ensuring we're not using legacy syntax, whitespace cleanup, trailing new lines, and generally making code review easier deal with by allowing us to focus on what's actually changed and not code quality minutiae no one wants to deal with anyways. 

the first pass can be ignored in git-blame through `.git-blame-ignore-revs` if you don't want to see it, but after that none of these things will ever reach the repo in the first place.

There's a one time setup by running `pre-commit install` after installing the dev dependencies.


Kodiologist:
Mostly this looks good. However, I still doubt it is useful or necessary to run `pyupgrade` on every commit. I also think removing the implied "r" from `open` is not helpful, and most of the uses of `.format` look better that way than as f-strings.

Kodiologist:
Although, it all runs locally, so none of these programs will do anything if somebody who submits a PR hasn't bothered to `pip install pre-commit`, right? That kind of newcomer is exactly the kind of person who we would fear will commit trailing whitespace and so on. So I think we should be running these in a GitHub Action instead, or at least, in addition.

Kodiologist:
I forgot to ask: why is the documentation excluded?

Kodiologist:
To top it off, `pyupgrade` is not very configurable, and I rustled the jimmies of a maintainer by asking for more options: https://github.com/asottile/pyupgrade/issues/516

allison-casey:
he's right though. highly configurable formatting tools are almost always a mistake and you likely won't find new ones being made (Google even made `yapf` and you'll find `black` in their newer pulbic repos). gofmt, rustfmt, js-beautify, black, and pyupgrade all exist to standardize code across repositories not to make each repository its own unique flavor of the tooling. This makes picking up the repo and code review easier to deal with since you always know what you're looking at. 

> Although, it all runs locally, so none of these programs will do anything if somebody who submits a PR 

It can be made to be automatically triggered as a github workflow by commenting something like `@github-actions pre-commit` in the PR comments if we want (this is how `pandas` does it)

> I forgot to ask: why is the documentation excluded?

i'll fix that 

>  I also think removing the implied "r" from open is not helpful

`'r'` hasn't been been required since python 2, why keep it? 

 > most of the uses of .format look better that way than as f-strings.
 
 I pretty generally disagree. pyupgrade only converts to f-strings if it makes the line shorter and contains only bare names or single function calls. Which ones do you not like?

Kodiologist:

PR #2149, opened 2021-08-14
Windows support

### LINKED ISSUE #2146: Get tests passing on Windows
GitHub Actions supports Windows, so we might as well bite this bullet. I'm currently working on this at https://github.com/Kodiologist/hy/tree/winders.

Kodiologist:
Closes #2146.

Most of these changes just generalize the tests to accomodate Windows, but there are two bugfixes.

I've verified that as it stands, this branch passes the test suite with all versions of Python on Windows, although the edits to the GitHub Action will only test one version of Python on Windows henceforth.


tests/importer/test_importer.py
```diff
@@ -84,29 +83,24 @@ def test_import_error_cleanup():
...
 
-        pyc_path = importlib.util.cache_from_source(f.name)
-
-        try:
-            os.remove(pyc_path)
-        except OSError:
-            pass
-
-        spec = importlib.util.spec_from_file_location('mymodule', f.name)
+    def ip(path):
```
allison-casey, line 92: This method name is meaningless. Please change it something more immediately clear.

tests/importer/test_importer.py
```diff
@@ -265,10 +259,10 @@ def test_shadowed_basename():
         assert os.path.isfile('tests/resources/importer/foo/some_mod.py')
 
         foo = importlib.import_module('foo')
-        assert foo.__file__.endswith('foo/__init__.hy')
+        assert os.path.basename(foo.__file__) == '__init__.hy'
         assert foo.ext == 'hy'
         some_mod = importlib.import_module('foo.some_mod')
-        assert some_mod.__file__.endswith('foo/some_mod.hy')
+        assert os.path.basename(some_mod.__file__) == 'some_mod.hy'
```
allison-casey, line 265: It would be more concise and idiomatic to use `pathlib` for these checks. `Path(some_mode.__file__).name == 'some_mod.hy'` etc

allison-casey (changes requested):

tests/importer/test_importer.py
Kodiologist: I'm always surprised when people want much more descriptive names from local subroutines than just-as-local temporary variables. I guess I still have some habits from programming in Haskell over a decade ago. But I'll change it, yes.

tests/importer/test_importer.py
Kodiologist: Aright. Might as well change all uses of `os.path` to `pathlib` in this file.

tests/importer/test_importer.py
allison-casey: Other people need to be able to work on and understand this code. It doesn't need to have the same the same explicitness as a top level public method `imppath` or the like would be fine

Kodiologist: But that would imply that people than me exist, and their needs even matter… are you sure about that?

allison-casey: a wild departure from your usual solipsism i know

Kodiologist:
Done.

allison-casey (approved):

PR #2159, opened 2021-08-30
Detecting cycles in `as_model` from `hy/models.py`

### LINKED ISSUE #2153: `hy.as-model` raises `RecursionError` for self-referential objects
```
=> (setv x [1 2 3])
=> (setv (get x 1) x)
=> x
[1 [...] 3]
=> (hy.as-model x)
…
RecursionError: maximum recursion depth exceeded while calling a Python object
```

Ultimately we don't want to try to support self-referential or otherwise infinite Hy programs, but we can do a little better here by checking for self-references instead of trying to recurse infinitely.

diminou:
Closes #2153.

In relation to `List`s and `Dictionaries`, this edit tracks parents of the elements to raise a `HyWrapperError` instead of relying Python being incapable of infinite recursion for recursive models.

This is done by passing an array of already seen parents for each recursive transformation in a List or a Dict.

Mysteriously, this breaks `tests/importer/test_importer.py::test_import_error_cleanup` as the imported module raises a `NameError` and not a `HyMacroExpansionError`.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -106,3 +106,4 @@
 * Peter Andreev <appa@gmx.co.uk>
 * Sunjay Cauligi <scauligi@eng.ucsd.edu>
 * David Tscheppen <david.tscheppen@gmail.com>
+* Dmitry Ivanov <dmitry.ivanov@divanov.eu>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -41,6 +41,7 @@ New Features
 ------------------------------
 * `hy.repr` now supports several more standard types.
 * the attribute access macro `.` now accepts method calls.
+* `as_model` checks for self-references in its argument.
 
 .. _Hyrule: https://github.com/hylang/hyrule
 
--- a/hy/models.py
+++ b/hy/models.py
@@ -109,7 +109,7 @@ class Object:
 
 
 _wrappers = {}
-
+_seen = set()
 
 def as_model(x):
     """Recurisvely promote an object ``x`` into its canonical model form.
@@ -142,6 +142,9 @@ def as_model(x):
        True  ; True because ``as-model`` has walked the expression and promoted
              ; the literal int ``2`` to its model for ``(hy.model.Integer 2)``
     """
+
+    if id(x) in _seen:
+        raise HyWrapperError("Self-referential structure detected in {!r}".format(x))
 
     new = _wrappers.get(type(x), lambda y: y)(x)
     if not isinstance(new, Object):
@@ -471,7 +474,15 @@ class List(Sequence):
 
 
 def recwrap(f):
-    return lambda l: f(as_model(x) for x in l)
+
+    def lambda_to_return(l):
+        _seen.add(id(l))
+        try:
+            return f(as_model(x) for x in l)
+        finally:
+            _seen.remove(id(l))
+
+    return lambda_to_return
 
 _wrappers[FComponent] = recwrap(FComponent)
 _wrappers[FString] = lambda fstr: FString(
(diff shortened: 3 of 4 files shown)


Kodiologist:
It looks like tests are failing, even with your change to `test_import_error_cleanup`.

You may want to look at how `hy/core/hy_repr.hy` deals wtih cycle detection, using a global variable, which may save you quite a few parameter additions.

Please indent by at least 2 spaces. 1-space indents are too small.

diminou:
Thanks for your swift comment and for the pointers!

On my laptop, `test_import_error_cleanup` produces a `NameError` instead of the `HyMacroExpansionError`, but seeing the results of the testing pipeline, this looks only valid for my setup. I will have to investigate this further.

I will also have to test on 3.10-dev locally to fix the build a bit later.

diminou:
I am at a loss at to why tests for 3.10-dev are failing.

On my local setup, I cannot reproduce this test failure. However, even on a clean `master` I have multiple other test failures locally:
```
FAILED tests/test_bin.py::test_bin_hy_error_parts_length - ValueError: substring not found
FAILED tests/test_bin.py::test_bin_hy_syntax_errors - AssertionError: assert 'SyntaxError: import * only allowed' in 'hy 1.0...
FAILED tests/test_bin.py::test_bin_hy_circular_macro_require - AssertionError: assert 1 == 0
FAILED tests/test_bin.py::test_bin_hy_macro_require - AssertionError: assert 1 == 0
FAILED tests/test_hy2py.py::test_hy2py_import - subprocess.CalledProcessError: Command '['hy2py', 'tests/resources/pydemo.hy...
```
for Python 3.10.0rc1, for example.

I would greatly appreciate any pointer or ideas.

Kodiologist:
I'll look into it. Python prelreases are moving targets, so it's common for them to break Hy without warning.

Kodiologist:
Okay, check it out. I think the problem wasn't with any specific version of Python so much as it was inconsistent when a test failure would occur, because duplication of `id` values depends on where data happens to land in RAM.

diminou:
>I think the problem wasn't with any specific version of Python so much as it was inconsistent when a test failure would occur, because duplication of id values depends on where data happens to land in RAM.

Your edits make perfect sense, thanks! I did not think of this scenario.

As for indentation - I believe `tests/native_tests/models.hy` has 2-space indentations and `models.py` has 4-space indentations.

Kodiologist:
> As for indentation - I believe tests/native_tests/models.hy has 2-space indentations and models.py has 4-space indentations.

Sorry, I was confused by GitHub's code view, which displays tabs incorrectly. Don't use tabs. Use spaces instead.

Let's finish this up:

- Don't add the trailing whitespace on `hy/models.py` line 145.
- `test_recursive_model_detection` is probably better written in Python and put in `tests/test_models.py`. Your new comments merely recapitulate the code and should be removed. But you should add a comment at the top with the URL `https://github.com/hylang/hy/issues/2153`.
- Squash your work into one commit.
- Rebase onto `master`.
- See `CONTRIBUTING.rst`. In particular, you need to add to `AUTHORS` and `NEWS`.

Kodiologist:
The `AUTHORS` addition should be in a separate commit. Please do read `CONTRIBUTING.rst`.

Lots of repetitive checks like those would be more concisely written as a loop, such as 

```
for x in [self_ref_list, self_ref_dict, mut_ref_list, mut_ref_dict]:
    with pytest.raises(HyWrapperError) as e:
        as_model(x)
    assert "Self-referential" in str(e)
```

Kodiologist (approved):
Well done. Thank you.

diminou:
Thank you for your patience!

PR #2163, opened 2021-09-12
Upgrade to funcparserlib 1.0.0a0

vlasovskikh:
Although I intended to make funcparserlib 1.0.0a0 backwards-compatible with 0.3.6, there is one incompatibility detected by the Hy tests, so I had to update `importlike(...)` to account for these changes.

`p1 + maybe(skip(p2))` no longer returns the result of `p1` if `p2` matches its tokens, so `skip(...)` it not distributive. The funcparserlib docs for 0.3.6 mentioned that you can use skip() **only** in `+`-expressions, but there is no enforcement for that. The current behaviour of `p1 + maybe(skip(p2))` is `(<the parse result of p1>, _Ignored(<the parse result of p2>))`.

### DIFF
--- a/hy/core/result_macros.py
+++ b/hy/core/result_macros.py
@@ -1397,7 +1397,7 @@ def compile_class_expression(compiler, expr, root, name, rest):
 def importlike(*name_types):
     name = some(lambda x: isinstance(x, name_types) and "." not in x)
     return [many(
-        SYM + maybe(sym("*")
+        SYM + maybe(sym("*") >> (lambda _: "*")
                     | (sym(":as") + name)
                     | brackets(many(
                         name + maybe(sym(":as") + name)))))]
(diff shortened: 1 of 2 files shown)


vlasovskikh:
See also vlasovskikh/funcparserlib/issues/65.

vlasovskikh:
There are some tests on GItHub failing for Python 3.10-dev as an internal error in pytest:

```
...
INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.10.0-rc.2/x64/lib/python3.10/site-packages/_pytest/_code/code.py", line 727, in get_source
64
INTERNALERROR>     lines.append(space_prefix + source.lines[line_index].strip())
65
INTERNALERROR> IndexError: list index out of range
```

I cannot reproduce them locally with:

> platform darwin -- Python 3.10.0rc1, pytest-6.2.5, py-1.10.0, pluggy-1.0.0

Kodiologist:
Thanks for taking this on. Could you reopen it with a request to pull from a new branch, rather than master? And don't forget an AUTHORS commit. See our CONTRIBUTING.rst. I'll look into the 3.10 issue in the meantime.

Kodiologist:
Turns out master is currently broken on Python 3.10 on GitHub Actions, even though it passed the last time we ran it, so you can ignore 3.10 failures for the time being. Either there's something wrong on GitHub's end or there was a backwards-incompatible configuration change. I'll look into it.

PR #2164, opened 2021-09-12
Upgrade to funcparserlib 1.0.0a0

vlasovskikh:
**Update:** I've created a new PR from a separate branch, as requested in #2163, and added an entry into AUTHORS. In the first PR I occasionally skipped these steps after my first quick look at CONTRIBUTING.rst.

Although I intended to make funcparserlib 1.0.0a0 backwards-compatible with 0.3.6, there is one incompatibility detected by the Hy tests, so I had to update `importlike(...)` to account for these changes.

`p1 + maybe(skip(p2))` no longer returns the result of `p1` if `p2` matches its tokens, so `skip(...)` it not distributive. The funcparserlib docs for 0.3.6 mentioned that you can use skip() **only** in `+`-expressions, but there is no enforcement for that. The current behaviour of `p1 + maybe(skip(p2))` is `(<the parse result of p1>, _Ignored(<the parse result of p2>))`.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -107,3 +107,4 @@
 * Sunjay Cauligi <scauligi@eng.ucsd.edu>
 * David Tscheppen <david.tscheppen@gmail.com>
 * Dmitry Ivanov <dmitry.ivanov@divanov.eu>
+* Andrey Vlasovskikh <andrey.vlasovskikh@gmail.com>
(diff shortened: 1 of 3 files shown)


Kodiologist:
Looks good, thanks.

Kodiologist (approved):

scauligi (approved):
Confirmed tests pass for 3.10rc2 after applying #2167!

vlasovskikh:
Cool! Should I rebase my PR onto master once #2167 gets merged?

Kodiologist:
No need. This is eligible to merge now, so it gets in first.

PR #2177, opened 2021-10-26
Remove old tool configuration

Kodiologist:


Kodiologist:
@allison-casey Still not convinced?

allison-casey:
No, but frankly I don't care enough it's not hard add back in if it becomes a problem down the road. Just make sure to update the "hacking on hy" page to drop the make commands it references

PR #2178, opened 2021-10-30
Fix a doc example for keyword-call changes

### LINKED ISSUE #2157: Docs not fully updated for `Keyword.__call__`
Trying to access a dict keyed by a keyword (as presented in some more complex examples from the documentation e.g. https://docs.hylang.org/en/alpha/api.html?highlight=person#fn) does not seem to work.

```hy
Hy 1.0a3+116.g0abaaaa7 using CPython(default) 3.8.10 on Linux
=> (:x {:x 1})
  File "stdin-5b93f6598dc2625adddf6d6eef831e622528a8f2", line 1, in <module>
    (:x {:x 1})
KeyError: 'x'
=> 
```
Using the `(get obj key)` form seems to work however

```
Hy 1.0a3+116.g0abaaaa7 using CPython(default) 3.8.10 on Linux
=> (get {:x 1} :x)
1
=> 
```

This seems to be a consequence of https://github.com/hylang/hy/blob/949e07b4abf639f911dfbda779d2eab2b2ecb62a/hy/models.py#L254 where `(:kw d)` is effectively `d["kw"]` after mangling+stringification.

Perhaps it would be reasonable to introduce a two-step lookup for keywords i.e. where keyword lookup is attempted first, and "mangled string lookup" is attempted if that does not work? Perhaps something like this would be a reasonable compromise:

```python
    def __call__(self, data, default=_sentinel):
        from hy.lex import mangle
        try:
            return data[self]
        except KeyError:
            try:
                return data[mangle(self.name)]
            except KeyError:
                if default is Keyword._sentinel:
                    raise
                return default
```

`{:x 1 "x" 2}` would then become ambiguous but in this case the user can utilize `(get {:x 1 "x" 2} :x)` or `(get {:x 1 "x" 2} "x")` to disambiguate.

Kodiologist:
Fixes #2157. (I searched for `{:` and the only other instance I found was in the style guide, where it doesn't make a difference.)

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -116,9 +116,9 @@ base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
    passed to another function for filtering output::
 
        => (setv people [(dict :name "Alice" :age 20)
-       ...             (dict :name "Bob" :age 25)
-       ...             (dict :name "Charlie" :age 50)
-       ...             (dict :name "Dave" :age 5)])
+       ...              (dict :name "Bob" :age 25)
+       ...              (dict :name "Charlie" :age 50)
+       ...              (dict :name "Dave" :age 5)])
 
        => (defn display-people [people filter]
        ...  (for [person people] (if (filter person) (print (:name person)))))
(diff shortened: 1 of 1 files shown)


Kodiologist:
Yeah, I noticed that, too. Done.

allison-casey (approved):

PR #2037, opened 2021-04-17
Docs: Migrate to github discussions

allison-casey:
We won't be able to close issue #1778 until we actually archive the channels, but lets update the docs and give it another month to let people filter over.

PR #2187, opened 2021-12-07
Removed IRC server in favor of GH Discussions

DrDub:
Just in case a new contributor is not aware of the migration to
libera.chat.

### DIFF
--- a/docs/hacking.rst
+++ b/docs/hacking.rst
@@ -11,7 +11,7 @@ Join our Hyve!
 
 Please come hack on Hy!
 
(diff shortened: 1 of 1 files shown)


Kodiologist:
Actually, we abandoned IRC entirely in lieu of GitHub Discussions (#2037), so you can just delete this paragraph, or replace it with a link to the discussions page.

DrDub:
Ah, missed that. What about now?

Kodiologist:
Looks good. Squash everything into one commit (with an updated commit message) and I'll merge.

Kodiologist:
Gracias.

Issue #1738, opened 2019-01-31
Make a fancy landing page for Hy


Immortalin:

Kodiologist:
Did you forget to add a description of your issue?

Kodiologist:
It sounds like you're under the impression that we don't have a website, and intended this issue to be a request for one. We do have one. It's http://hylang.org .

Immortalin:
Yes but a docs page isn't quite the same caliber as e.g. Nim or Rust's landing pages. Being flashy maybe be gimmicky but it helps conversion rates.

Immortalin:
Even Clojure has a dedicated landing page.

Kodiologist:
Okay, a hard-selling website is a valid feature request. I'm not personally interested in it (I think the kind of rhetoric currently shown on http://www.rust-lang.org is asinine, and encourages the naive chauvinism that some Rust programmers espouse), but perhaps other people feel differently.

kirbyfan64:
Hmm, if no one else is interested I might try tackling this.

yuhan0:
I think a better comparison would be with [Coconut](http://coconut-lang.org/) or [Dogelang](http://pyos.github.io/dg/) - both also being "compile-to-Python" languages with nice landing pages. 

Having side-by-side syntax examples up front showcasing some of the killer features would be great. 
(Threading macros? Multiline lambdas? Keyworded dictionaries without need for commas?)

Even better a live interpreter like on the Coconut page, showing the direct transpilation between Hy -> Python. The current one on https://try-hy.appspot.com is awesome but not very functional :P 

Here's one for Clojurescript -> Javascript: http://app.klipse.tech/

Kodiologist:
Fair enough. I guess the chief features distinguishing Hy from Python are:

- Free-form, homoiconic syntax
- Macros and arbitrary compile-time computation
- Erasure of the distinction between statements and expressions (mostly)

yuhan0:
Yeah, it depends who the target audience is - for anyone who's not already familiar with Lisps the advantages of "homoiconicity" is notoriously hard to explain.  
The ones I listed are more like immediate syntactical conveniences off the top of my head that someone coming from Python would be able to understand immediately?

asemic-horizon:
Can we publicize issues in bounty hunting websites? There are some that tightly integrate to Github too.

[...]

PR #2190, opened 2021-12-10
Add: project_urls to package info

### LINKED ISSUE #2189: Add `project_urls` to `setup.py`
This would allow us to provide links to our GitHub repository etc. in a sidebar on PyPI.

allison-casey:
closes #2189

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,5 +66,9 @@ setup(
         "Topic :: Software Development :: Code Generators",
         "Topic :: Software Development :: Compilers",
         "Topic :: Software Development :: Libraries",
-    ]
+    ],
+    project_urls={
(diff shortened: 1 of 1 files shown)


Kodiologist:
PyPI has one page for all versions of the package, so the links have to be version-neutral, to the extent that's possible. How about we just make the documentation link `https://docs.hylang.org`, thus passing the buck to our Read the Docs configuration to decide which version to default to?

Kodiologist:
Given, it will currently go to the same place as the "Homepage" link (which is set by the `url` argument to `setuptools.setup`), but maybe that will change someday (#1738).

allison-casey:
fixed

PR #2191, opened 2021-12-18
Fix `python_requires` in `setup.py`

Kodiologist:


allison-casey (commented):
do we even need the upper bound? Most python libraries don't bother with it

Kodiologist:
I think we should have it, yes, because I have yet to see a major release of Python that didn't break Hy. That is to say, I guarantee that 3.11 won't work until we make some changes to support it. The reason is that Hy is a lot more sensitive to Python internals than most Python libraries are.

Oddly, `<= 3.10.*` doesn't work either, so I went with `< 3.11`. Version specifiers seem to be hairy, especially since I couldn't find a library to let me quickly check a version against a specifier.

allison-casey:
ahh didn't know there was a history of new python versions breaking hy.

allison-casey (approved):

Kodiologist:
Yep. Hence I always construe support for a new major version of Python as a feature in NEWS.

PR #2196, opened 2022-01-03
Release Hy 1.0a4

Kodiologist:
I've decided that the right order of operations is to release Hy first and Hyrule second, because Hyrule depends on Hy and not the reverse. But the goal is to be practically simultaneous, so I wouldn't merge this before Hyrule was ready, too.


Kodiologist:
@hylang/core Remember, I need two approvals on this (and on hylang/hyrule#19) to merge.

allison-casey (approved):

brandonwillard (approved):

Kodiologist:
Thanks, guys. I'll make this happen once I have enough approvals for Hyrule, too.

PR #2197, opened 2022-01-08
Add "import os" in setup.py

Arkelis:
When adding `hy` as a `git` depedency with Poetry, it tries to read `setup.py` and crashes because `os` is not imported

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+import os
+
 from setuptools import find_packages, setup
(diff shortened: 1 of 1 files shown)


Kodiologist:
Now I wonder why `setup.py` works in the first place. Doesn't `pip install` execute it? So shouldn't it throw `NameError` when it hits `os`? Man, Python package management is mysterious.

Anyway, this is a very small change and seems fine to me. Thanks for catching it.

PR #2206, opened 2022-01-16
Typo in docs: CherryPy link

Arkelis:
Just found out by accident that CherryPy website changed, I replaced it with the Github repo, I guess it should be more "stable"


Kodiologist:
Why not link to https://cherrypy.dev instead?

Arkelis:
I think Github repo url is less likely to change on the long run, but if you prefer I'll go for cherrypy.dev

Kodiologist:
A domain name is controlled by the organization and can be retargeted at will, whereas a GitHub URL is only valid for as long as the organization uses GitHub. cherrypy.org was, I believe, a pathological case where the guy who registered the domain abruptly ceased all contact with other CherryPy developers.

Arkelis:
Changed to cherrypy.dev

PR #2220, opened 2022-01-28
Docs: Add stable branch warning header for active alpha cycle

### LINKED ISSUE #2219: Add header notice to "stable" line documentation to point users to the alpha cycle documentation
I was reading documentation and noticed that hy.contrib.walk is mentioned there:
https://docs.hylang.org/en/stable/contrib/walk.html

however it appears that hy.contrib.walk file is no longer on the master branch. 
https://github.com/hylang/hy/blob/6ba90fd3f853b2ddc391aa3358f9386c41d831c4/hy/contrib/walk.hy

is it a bug in documentation or I'm missing something?

allison-casey:
closes #2219 

Adds `has_active_alpha` bool we can toggle off when we're not in an alpha cycle and don't need the warning

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -1,0 +1,15 @@
+{% extends "!layout.html" %}
+
+{% block extrabody %}
+{% if has_active_alpha %}
+<div class="wy-nav-content-wrap">
+    <div id="dev-warning" class="wy-nav-content" style="overflow-wrap: breakword; padding: 1em 1em 0.1em 1em; background: #ffe761;">
+        <p>
+            Hy is currently in an active alpha cycle. Make sure you're looking
+            at the correct version of the documentation for your install by
+            using the version selector in the bottom-left corner of this page.
+        </p>
+    </div>
+</div>
+{% endif %}
+{% endblock %}
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -56,7 +56,9 @@ html_use_smartypants = False
 html_show_sphinx = False
 
 html_context = dict(
-    hy_descriptive_version = hy_descriptive_version)
+    hy_descriptive_version = hy_descriptive_version,
+    has_active_alpha = True,
+)
 
 highlight_language = 'clojure'
 


Kodiologist:
Looks helpful, but is this enough? We now have three different branches of the documentation running around (`stable`, `alpha`, `master`) and it seems like a user could also make a mistake like looking at the alpha manual when they're running 0.20.0. What do you think?

Sidebar: I'm always surprised that Git allows slashes in branch names, since slashes are also used to qualify branch names with remotes. Seems to work fine, though.

allison-casey:
Let's enable it for all branches while we're in an alpha/release candidate cycles then. 

It's very helpful to namespace branch types and I use it extensively

Kodiologist:
Okay. Aim for brevity in the warnings, since they'll then appear at the top of every single documentation page.

allison-casey:
> Hy is currently in an active alpha cycle. Make sure you are looking
            at the correct version of the documentation for your install by
            using the version selector in the bottom left corner of this page.

Seems good enough for me. what do you think?
Let me also remove some of the vertical padding

allison-casey:
image

Kodiologist:
I like it. To be extremely nitpicky, I'd say "you're" instead of "you are", and hyphenate "bottom-left".

I know that there's no version selector when you're building the documentation locally, and probably there's a different mobile stylesheet for Read the Docs that doesn't put it in the bottom left, but eh, I can't quite bring myself to care, because there are bigger web-design issues with the documentation.

Kodiologist (approved):
Merge when ready.

allison-casey:
Done. 

If you're building locally, you probably know that you're building documentation for your local instance only anyways.

Kodiologist:
Don't forget to push a new commit to `alpha`, too. To change the stable version, I think you have to make a new branch and then change what the "stable" docs version points to on Read the Docs.

allison-casey:
Done. the docs now point to the "stable" branch and the header is activated on all versions

Kodiologist:
@allison-casey There's a hitch: the stable header now says "Hy version 0.20.0+1.g0e4d576c (unstable)" and similarly for the alpha. I think you need to fake `hy_descriptive_version`. How you do it doesn't matter too much, since the commits you do it in will never get merged into `master`.

PR #2223, opened 2022-02-04
Miscellaneous test and command-line cleanup

Kodiologist:
The most notable thing here is a small bugfix relating to overriding the `builtins` module. The removed Easter eggs and the delay of `colorama.init()` are theoretically user-visible, but probably too minor to mention in NEWS.


allison-casey:
Is there an issue associated with the builtins monkeypatching?

allison-casey (approved):

Kodiologist:
Nope, I don't think anybody noticed the bug behaviorally. I just noticed it from reading the code.

Kodiologist:
And `test_bin_hy_builtins` was construing it as a feature—for shame!

PR #2228, opened 2022-02-07
Fix:let loop clauses [2224]

### LINKED ISSUE #2224: `let` doesn't detect assignments in loop clauses of comprehension forms
```
=> (let [x 10] (lfor  x (range 5)  :if (> x 1)  x))
[10 10 10 10 10]
```

`hy --spy` shows this compiles to:

```
_hy_let_x_1 = 10
[_hy_let_x_1 for x in range(5) if _hy_let_x_1 > 1]
```

`let` is supposed to allow some kinds of assignments to shadow its own variables, while allowing other kinds to modify its variables, but its treatment of the loop variable seems inconsistent here.

allison-casey:
fixes #2224

This makes `let` rebinding better respect the distinction in Python's scoping rules regarding for loops vs. comprehensions by not renaming comprehension iteration targets in generators (as generators introduce their own scopes), ***but*** renaming in `for` loops as `for` loops are not their own scopes and their iteration targets leak into the parent scope. 

```clojure
(let [x 10]
  (lfor x (range 5) :if (> x 1) x) ; => [2 3 4]
  x) ; => 10

(let [x 10]
  (dfor x (range 5) :if (> x 1) [x x]) ; => {2 2  3 3  4 4}
  x) ; => 10

(let [x 10
      l []]
  (for [x (range 5) :if (> x 1)]
    (l.append x))
  [l x]) ; => [[2 3 4] 4]
```

```python
_hy_let_x_50 = 10
[x for x in range(5) if x > 1]
_hy_let_x_50
# >>> 10

_hy_let_x_51 = 10
{x: x for x in range(5) if x > 1}
_hy_let_x_51
# >>> 10

_hy_let_x_52 = 10
_hy_let_l_53 = []
for _hy_let_x_52 in range(5):
    if _hy_let_x_52 > 1:
        _hy_let_l_53.append(_hy_let_x_52)
[_hy_let_l_53, _hy_let_x_52]
# >>> [[2 3 4] 4]
```

also split apart the comprehension/for result macro because it had become a 237 line monstrosity that was essentially impossible to understand in any meaningful way

### DIFF
--- a/hy/core/result_macros.py
+++ b/hy/core/result_macros.py
@@ -698,243 +698,292 @@ loopers = many(
 )
 
 
-@pattern_macro(
-    ["for"], [brackets(loopers), many(notpexpr("else")) + maybe(dolike("else"))]
-)
+def _compile_tags(compiler, tags, scope=None):
+    """
+    Compiles list of generator `Tag`s from the parsed `tags` expression list.
+    When given a `ScopGen` `scope`, adds comprehension targets to the scope's
+    seen iterators.
+    """
+    parsed = []
+    for tag in tags:
+        if tag.tag in ("if", "do"):
+            tag_value = compiler.compile(tag.value)
+        else:
+            tag_value = [
+                compiler._storeize(tag.value[0], compiler.compile(tag.value[0])),
+                compiler.compile(tag.value[1]),
+            ]
+            if scope:
+                scope.iterator(tag_value[0])
+        parsed.append(Tag(tag.tag, tag_value))
+    return parsed
+
+
+def _compile_nested_for(parts, body, else_clause=None):
+    """
+    Compiles a nested for loop from generator clauses `parts` with the
+    inner body of result `body`. Optionally, add an `else` clause
+    to the outer for loop with `else_clause`
+    """
+
+    def f(parts, else_clause=None):
+        if not parts:
+            return body
+        (tagname, v), parts = parts[0], parts[1:]
+        if tagname in ("for", "afor"):
+            node = asty.AsyncFor if tagname == "afor" else asty.For
+            return v[1] + node(
+                v[1],
+                target=v[0],
+                iter=v[1].force_expr,
+                body=f(parts).stmts,
+                orelse=else_clause and else_clause.stmts or [],
+            )
+        elif tagname == "setv":
+            return (
+                v[1]
+                + asty.Assign(v[1], targets=[v[0]], value=v[1].force_expr)
+                + f(parts, else_clause)
+            )
+        elif tagname == "if":
+            return v + asty.If(
+                v, test=v.force_expr, body=f(parts, else_clause).stmts, orelse=[]
+            )
+        elif tagname == "do":
+            return v + v.expr_as_stmt() + f(parts, else_clause)
+        else:
+            raise ValueError(
+                "Found unsupported tag '{tagname}' while compiling generator tags"
+            )
+
+    return f(parts, else_clause)
+
+
+def _is_real_python_generator(exprs, parts):
+    """
+    Can the comprehension defined by `parts` and `exprs` be expressed as a real
+    python comprehension
+    """
+    return not (
+        any(expr.stmts for expr in exprs if expr)
+        or any(
+            p.tag == "do"
+            or (p.value[1].stmts if p.tag in ("for", "afor", "setv") else p.value.stmts)
+            for p in parts
+        )
+    )
+
+
+def _generators_from_parts(tags):
+    """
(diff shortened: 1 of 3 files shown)


Kodiologist:
Fundamentally, the motivation for this choice makes sense, but: are you sure you don't want to let the user use `let` for name replacement in comprehensions, which is useful for macros? I don't think there's any other way to use the current scope-tracking machinery to do all that work for you. For example, the anaphoric macros will currently replace `it` even if it wouldn't become a variable name. If I did (or rather, could do) the replacement with `let`, that wouldn't happen. Nesting would also work.

allison-casey:
I'm sure. Python comprehension forms are essentially sugar for `for` + `list.append` constructs. The only difference being that they introduce their own scope in which their vars are assigned. If we translate that into Hy it would be identical to 
```clojure
(let [x 10]
  ;; Our unwound list comprehension
  (let [out []
        x None]
    (for [x (range 5)]
      (when (> x 1)
        (out.append x)))
    out) ; => [2 3 4]
  x) ; => 10
```
Which is the correct behavior according to python scoping.

It's also the expected behavior in 
Common Lisp:
```lisp
(let ((x 10))
  (loop for x from 0 to 4 if (> x 1) collect x) ; => '(2 3 4)
  x) ; => 10
```
Clojure
```clojure
(let [x 10] 
  (for [x (range 5) :when (> x 1)] x) ; => '(2 3 4)
  x) ; => 10
```
Racket:
```racket
(let ([x 10])
  (for/list ([x (range 5)] #:when (> x 1)) x) ; => '(2 3 4)
  x) ; => 10
```
and Fennel:
```lua
(let [x 10] 
     (icollect [_ x (ipairs [1 2 3 4])] 
       (when (> x 1)
         x)) ; => [2 3 4]
     x) ;=> 10
```
I think your use case would be better accomplished with a macro á la `smacrolet`which itself does some custom stuff with scoping you could take advantage of.

Kodiologist:
> I think your use case would be better accomplished with a macro á la `smacrolet` which itself does some custom stuff with scoping you could take advantage of.

Perhaps I'm missing something, but it looks like your changes are to how the comprehension macros manage their scopes rather than to `let`, so I don't think a different macro from `let` can change this behavior. Instead, you'd need to replace the comprehension macros themselves. I guess I'd rewrite the anaphoric macros to use `for` instead of `gfor` etc.

Kodiologist:
Anyway, thanks for working on this. I think I can do a real review of it tomorrow.

> also split apart the comprehension/for result macro because it had become a 237 line monstrosity that was essentially impossible to understand in any meaningful way

I think Black butchered some of the more complicated expressions. `# fmt: off` is handy.

Kodiologist:
The tests and functional changes look good. I wonder why `test-generator-scope` is in `native_tests/let.hy` in the first place. Oh well. Something to clean up some other time, perhaps.

I'm not sure this new family of functions is an improvement over `compile_comprehension`. The old one is big and complicated, but it's linear, and this version adds redundancies such as the similar-but-not-identical `compile_sequential_comprehension` and `compile_dict_comprehension`. Besides, the potential value for intelligibility of spinning off logic into separate functions is not realized when the functions, like `_compile_parts` and `_is_real_python_generator`, are only fully understandable in the context in which they're called—that is, when you can't actually achieve separation of concerns. It's annoying that this much logic is required to implement these macros, but ultimately I can't think of a way around that.

An easier way to simplify `compile_comprehension`, if you want to do that in this PR, is to chip away at existing redundancy. For example, it looks like

```
if tagname in ("for", "afor"):
    generators.append(
        ast.comprehension(
            target=v[0],
            iter=v[1].expr,
            ifs=[],
            is_async=int(tagname == "afor"),
        )
    )
elif tagname == "setv":
    generators.append(
        ast.comprehension(
            target=v[0],
            iter=asty.Tuple(v[1], elts=[v[1].expr], ctx=ast.Load()),
            ifs=[],
            is_async=0,
        )
    )
```

could be written

```
if tagname in ("for", "afor", "setv"):
    generators.append(
        ast.comprehension(
            target=v[0],
            iter=(asty.Tuple(v[1], elts=[v[1].expr], ctx=ast.Load())
               if tagnme == "setv"
               else v[1].expr),
            ifs=[],
            is_async=int(tagname == "afor"),
        )
    )
```

Kodiologist:
@allison-casey I'd like to get this in, and I know you're busy with the reader PR. Shall I just reduce this to the bugfix and the test, and you can do refactoring in a later PR if you want to?

Kodiologist:
It is done.

PR #2230, opened 2022-02-09
Fix: crash when using keywords as match value

allison-casey:
Something i noticed fiddling around today. 

Hy would crash when trying to pattern match against keywords

```clojure
(match :hello
       :hello 1
       :world 2)
=> hy.errors.HySyntaxError: parse error for pattern macro 'match': got unexpected token: :hello, expected: end of file
```

Turns out I forgot that Keywords are compiled into a class instantiation calls not values and need to be handled as such.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -9,8 +9,13 @@ Bug Fixes
   CPython bug regarding the standard module `curses` (`bpo-2675`_).
 * Elements of `builtins` such as `help` are no longer overridden until
   the Hy REPL actually starts.
+* Crash when using Keywords as `match` value.
 
 .. _bpo-2675: https://bugs.python.org/issue2675#msg265564
+
+New Features
+------------------------------
+* new function `hy.model_patterns.parse_if`
 
 1.0a4 (released 2022-01-09)
 ==============================
--- a/hy/core/result_macros.py
+++ b/hy/core/result_macros.py
@@ -32,6 +32,7 @@ from hy.model_patterns import (
     dolike,
     keepsym,
     notpexpr,
+    parse_if,
     pexpr,
     sym,
     tag,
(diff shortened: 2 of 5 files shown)


Kodiologist:
Looks like we have some test failures on GitHub Actions. Do they work locally for you?

allison-casey:
My bad i forgot to test nesting keywords inside match class patterns. I'll get on that

allison-casey:
So this works, but it feels wrong? Like i feel like there should be a better way to define a parser that is an `and` relationship with a predicate. Like `some` but with a custom parser. I looked around funcparserlib but there didn't seem to be anything if anyone has any ideas.

allie-jo:
There was, in fact, a better way. 
```python
def parse_if(pred, parser):
    """
    Return a parser that parses a token with `parser` if it satisfies the
    predicate `pred`.
    """

    @Parser
    def _parse_if(tokens, s):
        some(pred).run(tokens, s)
        return parser.run(tokens, s)

    return _parse_if
```

Kodiologist:
I think it may also be possible by just removing the `KEYWORD` alternative in `_pattern` and setting `_kpattern = KEYWORD | _pattern`, then using `_pattern` or `_kpattern` when appropriate. But, `parse_if` might be a useful addition to `hy.model-patterns` in any case.

allie-jo:
I don't believe we can do `_kpattern` in this case since the whole parser is recursive and every other branch of the recursion tree can take keywords. I've moved `parse_if` to `model_patterns` in any case.

Kodiologist (approved):
Don't forget a NEWS item for `parse_if`, since `hy.model_patterns` is also for public use.

allison-casey:
Done

Kodiologist:
You don't have to tag PRs with GitHub labels, by the way. The reason I use those is to categorize outstanding issues.

PR #2209, opened 2022-01-17
Set up autoformatting of Python source

lafrenierejm:
Autoformat all Python source code via isort and black. Also add a GitHub Actions check so PRs that do not comply with the autoformatting rules will show a CI failure.

Closes #1704.

PR #2232, opened 2022-02-14
Add `_hy_export_macros` and `export`

### LINKED ISSUE #1934: No built-in method to create `__all__` using unmangled hy names
as far as i can tell, Hy doesn't have a way to define a module's `__all__` attribute using unmangled names. As in a method defined as `hello?` has to manually be set as
`(setv --all-- ["is_hello"])`
Which is very unintuitive, especially for the more complicated mangled names.
This is fixed pretty easily with a macro such as 
```
(defmacro defall [&rest symbols]
  `(setv __all__ ~(lfor sym symbols (mangle (name sym)))))
```
So the same `--all--` can be defined as
`(defall hello?)`
This seems like something Hy should be able to do by default as part of the core library.

### LINKED ISSUE #2142: `(require foo *)` unconditionally gets all macros
`(import foo *)`, [per Python](https://docs.python.org/3/reference/simple_stmts.html#the-import-statement), uses the variable `__all__` to decide what to import, or if `__all__` is not defined, it imports every name not beginning with an underscore. `(require foo *)`, by contrast, always gets all macros, although it's supposed to be analogous. So, what `(require foo *)` should do is:

1. Check a module variable, say `_hy_all_macros`, for a list of public names.
    - A name present in `_hy_all_macros` with no corresponding macro defined is an error at the time of `(require foo *)`.
2. If there is no such variable, require all macros whose names don't begin with an underscore.

Kodiologist:
- Closes #1934 
- Closes #2142

I thought `eval-and-compile` would be necessary for setting `_hy_export_macros`, but of course, by the time you're requiring a module, the whole thing has been executed.

### DIFF
--- a/hy/macros.py
+++ b/hy/macros.py
@@ -164,8 +164,7 @@ def require(source_module, target_module, assignments, prefix=""):
         package = None
         if source_module.startswith("."):
             source_dirs = source_module.split(".")
-            target_dirs = (getattr(target_module, "__name__", target_module)
-                .split("."))
+            target_dirs = getattr(target_module, "__name__", target_module).split(".")
             while len(source_dirs) > 1 and source_dirs[0] == "" and target_dirs:
                 source_dirs.pop(0)
                 target_dirs.pop()
@@ -176,8 +175,11 @@ def require(source_module, target_module, assignments, prefix=""):
             raise HyRequireError(e.args[0]).with_traceback(None)
 
     source_macros = source_module.__dict__.setdefault("__macros__", {})
-    source_exports = getattr(source_module, "_hy_export_macros",
-        [k for k in source_macros.keys() if not k.startswith('_')])
+    source_exports = getattr(
+        source_module,
+        "_hy_export_macros",
+        [k for k in source_macros.keys() if not k.startswith("_")],
+    )
 
     if not source_module.__macros__:
         if assignments in ("ALL", "EXPORTS"):
@@ -203,10 +205,15 @@ def require(source_module, target_module, assignments, prefix=""):
     if prefix:
         prefix += "."
 
-    for name, alias in (assignments if assignments not in ("ALL", "EXPORTS") else (
+    for name, alias in (
+        assignments
+        if assignments not in ("ALL", "EXPORTS")
+        else (
             (k, k)
             for k in source_macros.keys()
-            if assignments == "ALL" or k in source_exports)):
+            if assignments == "ALL" or k in source_exports
+        )
+    ):
         _name = mangle(name)
         alias = mangle(
             "#" + prefix + unmangle(alias)[1:]
--- a/hy/macros.py
+++ b/hy/macros.py
@@ -129,7 +129,8 @@ def require(source_module, target_module, assignments, prefix=""):
         target_module (Optional[Union[str, ModuleType]]): The module into which the
             macros will be loaded.  If `None`, then the caller's namespace.
             The latter is useful during evaluation of generated AST/bytecode.
-        assignments (Union[str, typing.Sequence[str]]): The string "ALL" or a list of macro name and alias pairs.
+        assignments (Union[str, typing.Sequence[str]]): The string "ALL", the string
+            "EXPORTS", or a list of macro name and alias pairs.
         prefix (str): If nonempty, its value is prepended to the name of each imported macro.
             This allows one to emulate namespaced macros, like "mymacromodule.mymacro",
             which looks like an attribute of a module. Defaults to ""
@@ -175,9 +176,11 @@ def require(source_module, target_module, assignments, prefix=""):
             raise HyRequireError(e.args[0]).with_traceback(None)
 
     source_macros = source_module.__dict__.setdefault("__macros__", {})
+    source_exports = getattr(source_module, "_hy_export_macros",
+        [k for k in source_macros.keys() if not k.startswith('_')])
 
     if not source_module.__macros__:
-        if assignments == "ALL":
+        if assignments in ("ALL", "EXPORTS"):
             return False
         for name, alias in assignments:
             try:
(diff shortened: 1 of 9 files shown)


hy/macros.py
```diff
@@ -1,3 +1,5 @@
+# fmt: off
```
allison-casey, line 1: Delete this

allison-casey (changes requested):

hy/macros.py
Kodiologist: Are you going to object to every single one of my PRs if I try to fix spacing in it, no matter how unreasonable Black's choices are?

allison-casey: Black's changes are not unreasonable. Your code introduced 2 additional [pep8 errors](https://www.python.org/dev/peps/pep-0008/#indentation) that black would have otherwise caught and was the whole point of #2209. If you'd like an approval for this PR you can remove this and I'll happily continue the review as per usual.

Kodiologist: Please answer the question.

allison-casey: I'll object to any PR that attempts to circumvent #2209 by turning off formatting, yours or otherwise.

Kodiologist: I'm having flashbacks to 2018 when Matthew got fixated on the indentation of my Hy code. Strong words were said and he eventually decided to make his own Lisp-on-a-Python instead (with blackjack and hookers). It wasn't a high point for the project.

Now, I want to emphasize that:

1. I'm not trying to move away from the Black style for a large majority of the Python code. I'd estimate that less than 1% of Hy's Python code, counting by lines, needs to be changed. A lot of Black's changes were pointless but harmless, like changing all the single quotes to double quotes; all I'm trying to revert now is the changes that are actually bad.
2. I'm not asking you to do the work of finding all these mistakes and cleaning them up. I'm only asking you to not forbid me from doing it.
3. No matter what formatting decisions I make, you can always run Black locally with a single command and see the code in the official Black-approved style. I'm not taking it away from you.

I don't think this is a lot to ask, and I don't know why you're being fanatical about this of all issues, especially after over a year of close work with you on Hy during which you've shown good judgment on many occasions.

Can't we broker some kind of compromise? I would happily prioritize work on some difficult Hy feature or bug that's important to you if it would help sway you.

Kodiologist: @allison-casey Let's wrap this up.

allison-casey: It is not a fanatical position. It's a common industry practice with proven benefits. Across Pandas, Scikit-learn, and Django, and their 1,287,351 lines of code, `fmt: off` shows up *zero* times. Black's choices are well founded and it's changes ensure consistent pep8 compliant code in all cases. 

@brandonwillard as the other approver of #2209, what are you feelings on kodi using `fmt: off` to adjust portions of the codebase away from the black standard?

Kodiologist: Just because a lot of people do something doesn't mean it's beneficial, let alone that any supposed benefits have been proven. Justifying Black's choices in terms of PEP 8 compliance only passes the buck to PEP 8, which would be no less unreasonable to rigidly adhere to.

I'd be happy to show you instances in which Black's choices aren't well founded, but there's no point in me doing that if you've already made up your mind that any deviation from Black's style, however slight, rare, or justified, is unacceptable, no matter how easy it is for you to run Black locally.

brandonwillard: > @brandonwillard as the other approver of #2209, what are you feelings on kodi using `fmt: off` to adjust portions of the codebase away from the black standard?

Unless something is functionally broken, which it doesn't appear to be in this case, disabling `black` is unwarranted and completely defeats its purpose.

Kodiologist: If it's not warranted when it makes the code more readable, that seems to imply a belief that consistency with Black's formatting is more important than readability. So then what motivates an interest in consistency? Readability is literally the only thing at stake with respect to formatting, yes?

allison-casey: This is one of the snippets you disabled black in order to write:
```diff
# your's
-    source_exports = getattr(source_module, "_hy_export_macros",
-        [k for k in source_macros.keys() if not k.startswith('_')])
# black's
+    source_exports = getattr(
+        source_module,
+        "_hy_export_macros",
+        [k for k in source_macros.keys() if not k.startswith("_")],
+    )
```

pep8 says it's forbidden:
> image

Google says it's forbidden:
> image

Google dislikes it so much they wrote about it as the **Rectangle Rule**
> image

Which underpins their wrapping rules for Python, Java, and Javascript and you'll see enforced in their repos with `yapf/black`, `google-java-format`, and facebook's `prettier` respectively

Clojure says arguments should be vertically aligned and CIDER and Emac's indent region will fight me if i attempt to push an arg before it's indent block
> image

Hy *itself* says it's "BAD. And evil."
> image

Black's choices follow language agnostic standard practice for writing readable code. They may not be your preferred style, but they're certainly not "bad" or "unreasonable". Brandon and I are in agreement then that nothing is obviously broken, and so the use of `fmt: off` is unwarranted. If you'd like to continue with this PR, I would be happy to do a full review once this line has been removed.

Kodiologist: I think Black's choice in that case is fine, actually. It's not what I wrote, but it's perfectly reasonable.

I don't think style guides are to your point so long as your position is that we have to do what Black says no matter what. The only "style guide" whose authority you recognize in that case is the one implicitly defined by Black. Otherwise, it seems we're in agreement that we don't care for style guides.

Hy's style guide is not Hy itself (thank goodness). It's something I've had on my mental chopping block for a long time, although I've yet to do the big documentation overhaul that would involve its removal.

> If you'd like to continue with this PR, I would be happy to do a full review once this line has been removed.

But won't you then just oppose the PR because it doesn't pass the Black check, as well as all future PRs with any spacing fixes, no matter what? So then it looks like your bigger issue still needs to be resolved. You're more than welcome to review this PR in the meantime, of course.

allison-casey: format the file with black, it will pass CI, and I will approve the PR assuming there's nothing else that needs addressing. you can format the file before each commit or at the very end, CI does not care.

Kodiologist: But you're still going to object to every single PR with spacing fixes no matter what. I want to negotiate, but in order for that to happen, you need to be willing to negotiate.

allison-casey: black handles everything. there are no spaces to fix.

PR #2235, opened 2022-02-20
Add type annotations to `hy` package

lafrenierejm:
This is still a work in progress.

### DIFF
--- a/hy/macros.py
+++ b/hy/macros.py
@@ -8,8 +8,8 @@ import os
 import pkgutil
 import sys
 import traceback
+import typing as T
 from ast import AST
-from typing import TYPE_CHECKING, cast
 
 from funcparserlib.parser import NoParseError
 
@@ -25,16 +25,16 @@ from hy.lex import mangle, unmangle
 from hy.model_patterns import whole
 from hy.models import Expression, Symbol, as_model, is_unpack, replace_hy_obj
 
-if TYPE_CHECKING:
+if T.TYPE_CHECKING:
     from types import ModuleType
-    from typing import Callable, List, Optional
-    from typing import Sequence as SequenceT
-    from typing import Tuple, TypeVar, Union
+
+    from funcparserlib.parser import Parser
 
     from hy.compiler import HyASTCompiler, Result
     from hy.models import Object
 
-    F = TypeVar("F", bound=Callable)
+    MacroNames = T.Union[str, list[str]]
+    F = T.TypeVar("F", bound=T.Callable)
 
 EXTRA_MACROS = ["hy.core.result_macros", "hy.core.macros"]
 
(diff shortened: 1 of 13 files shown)


Kodiologist:
Don't forget you can use parentheses to split one `import` over multiple lines, if that's why you're importing from the same module more than once.

It looks like GitHub doesn't want to run the tests if a PR is in draft mode, so I'll hit "Ready for review" although I understand you're still working on it.

Kodiologist:
Well, that didn't work. The action to run the tests still isn't queued. Maybe it doesn't like that there are merge conflicts? But the tests should be run on the actual commit at the tip of the PR branch, not on some merged commit, so that shouldn't matter.

Oh, well. Worse comes to worst, we can run the tests locally.

hy/macros.py
```diff
@@ -21,6 +25,17 @@
 from hy.model_patterns import whole
 from hy.models import Expression, Symbol, as_model, is_unpack, replace_hy_obj
 
+if TYPE_CHECKING:
+    from types import ModuleType
+    from typing import Callable, List, Optional
+    from typing import Sequence as SequenceT
```
lafrenierejm, line 31: @Kodiologist I'm assuming these sort of patterns are what you're referring to in your comment about mult-iline import statements. (Please correct me if I'm wrong.)

This ended up being a multi-line import because I'm aliasing `Sequence` within the imports and just letting `isort` take over from there. The reason I aliased `Sequence` was because at some point while I was working on this I was using `hy.models.Sequence` instead of `typing.Sequence`. It looks like that's no longer an issue in this specific file, I just hadn't noticed and removed the alias before creating this draft PR.

Kodiologist: Yeah, that's what I meant.

lafrenierejm: Thanks for confirming.

On this topic: At some point it might be would be worth having a conversation about how Hy's maintainers would like to handle `typing` imports whose names conflict with custom classes defined in the `hy` package. As I've been working on this draft there have been a couple times that I've become confused about which class a name was referring to, though others' experiences might be different.

Kodiologist: `typing` is a standard module, so we might want to change our names, but it's pretty easy to just say `import typing as T` and then `T.Sequence` or whatever, too. Shouldn't be a big issue. My gut reaction would be to pick the latter because e.g. renaming `hy.models.List` would be a breaking change, and while we **do** make breaking changes like it's going out of style, we might as well avoid it for such a small benefit.

allison-casey: [pep585](https://www.python.org/dev/peps/pep-0585/) added type hinting of generic typess in 3.7, which is the earliest python we support, which we can take advantage of for most things. Using the `T` alias for everything else like Kodi mentioned is likely the best route

allison-casey: There's also the official python [`typing_extensions`](https://github.com/python/typing/blob/master/typing_extensions/README.rst) module for backports of certain types if we need them

lafrenierejm: > [pep585](https://www.python.org/dev/peps/pep-0585/) added type hinting of generic typess in 3.7

Oh, that's great! For some reason I had thought PEP 585 was limited to 3.8; the fact that we can use it will make things nicer.

> it's pretty easy to just say import typing as T and then T.Sequence or whatever, too. Shouldn't be a big issue.

Agreed. Especially since we can make use of PEP 585, I do not think it is worth breaking backwards compatibility for this.

Kodiologist:
I'm closing this pull request due to lack of activity. If you resume work on it, we can reopen it once you want us to look at it again.

Issue #2210, opened 2022-01-21
Distinguishing `if`, `when`, and `cond`


Kodiologist:
Let's talk about how exactly these three similar constructs should differ.

`cond` requires brackets around each condition-result pair. Previously the idea of dropping those brackets, and thus requiring `do` when you want a multi-form result, has been floated. I've warmed up to that idea, particularly since we've been dropping brackets from other stuff (like `import`) and `if` can't be used like `cond` anymore. I feel like in most cases, I won't need the `do`.

`if` and `when` are identical when you use exactly two arguments. What if `if` was required to have exactly three arguments? Then, when you see `if`, you know that an else-clause is coming. The expectation is that you'll use `when` when you're happy with the implicit `None` in the else-case, although you can always write `(if condition then-clause None)` if you want to be explicit.

allison-casey:
There is precedent for this as Racket does enforce this distinction. Let me try it locally and see how it feels.

allison-casey:
I have a test passing version of a strict if on my [strict-if](https://github.com/allison-casey/hy/tree/feature/strict-if) branch if you want to take a look

Kodiologist:
It doesn't look like it had any surprising consequences. That's good.

allison-casey:
most the changes were for side effecting python constructs like `.append`, `break`, `continue` etc. There were a few cases where we were using the implicit `None` as a return value (mostly in `hy_repr`), but those should have either been explicitly defined as they are expected values of slices or done via `when` or as a short circuiting `and`.

Kodiologist:
I recall using `if` in the `hy.repr` code in a lot of cases where I wanted to use `when`, because the way things were originally set up years ago, `hy.repr` had access to special forms but not any core macros.

allison-casey:
I'm leaning in favor of this change at the moment. Using `if`'s return value with the implicit `None` is an anti pattern imo and should be discouraged. `when` does a much better job of indicating that you should be aware of the `None`  and generally leads to more easily understood code. Lets let this sit for a bit though to let the initial feeling sink in and see if anyone else wants to comment though. 

Also, you said you're warming up to bracketless `cond`. are you saying you'd be open to a PR to change that at some point? It does seem like most of our constructs are moving away from those kinds of brackets

Kodiologist:
[...]

PR #2237, opened 2022-02-24
Unbracketed `cond`

Kodiologist:
Partly fulfills #2210.

### DIFF
--- a/tests/native_tests/language.hy
+++ b/tests/native_tests/language.hy
@@ -340,18 +340,18 @@
 
 (defn test-cond []
   (cond
-   [(= 1 2) (assert (is True False))]
-   [(is None None) (setv x True) (assert x)])
+    (= 1 2) (assert (is True False))
+    (is None None) (do (setv x True) (assert x)))
   (assert (is (cond) None))
 
   (assert (= (cond
-    [False]
-    [[]]
-    [8]) 8))
+    False 1
+    [] 2
+    True 8) 8))
 
   ;make sure test is only evaluated once
   (setv x 0)
-  (cond [(do (+= x 1) True)])
+  (assert (= (cond (do (+= x 1) True) 2) 2))
   (assert (= x 1)))
 
 
--- a/tests/native_tests/language.hy
+++ b/tests/native_tests/language.hy
@@ -342,12 +342,12 @@
   (cond
    [(= 1 2) (assert (is True False))]
    [(is None None) (setv x True) (assert x)])
-  (assert (= (cond) None))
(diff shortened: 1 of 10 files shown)


tests/native_tests/language.hy
```diff
@@ -340,19 +340,29 @@
...
+    (is None None) (do (setv x True) (assert x)))
   (assert (= (cond) None))
 
   (assert (= (cond
-    [False]
-    [[]]
-    [8])) 8)
+    False 1
+    [] 2
+    True 8)) 8)
```
allison-casey, line 350: so it turns out this paren is one short and we need to slurp that 8 up to actually check equality

tests/native_tests/language.hy
```diff
@@ -340,19 +340,29 @@
 
 (defn test-cond []
   (cond
-   [(= 1 2) (assert (is True False))]
-   [(is None None) (setv x True) (assert x)])
+    (= 1 2) (assert (is True False))
+    (is None None) (do (setv x True) (assert x)))
   (assert (= (cond) None))
```
allison-casey, line 345: also can we get this `=` changed to `is` while we're here?

allison-casey (commented):
Fantastic! this is pretty much what I had in my branch. Everything looks good! just noticed two things we might as well fix in the cond test unrelated to this PR we might as well pick up while we're here. Also, I have a branch with the strict if form that would finish the conditional forms refactor we've been talking about if you want to yoink that and finish it up while i work on those sphinx hydomain issues you brought up

Kodiologist:
> this is pretty much what I had in my branch.

Oops, I did this because I thought it was the part you hadn't already done.

> Also, I have a branch with the strict if form that would finish the conditional forms refactor we've been talking about if you want to yoink that

Okay. Let's do it as another PR.

tests/native_tests/language.hy
Kodiologist: Good catch. Fixed.

tests/native_tests/language.hy
Kodiologist: You mean changing `(= 1 2)` to `(is 1 2)`? Aren't the semantics of `is` with numbers a bit dicey?

allison-casey: Oh no the `(= (cond) None)` one

Kodiologist: Oh, right. My eyeballs just skipped over that line because I hadn't edited it. Done.

allison-casey (approved):

PR #2240, opened 2022-02-27
Make `if` require three arguments

### LINKED ISSUE #2210: Distinguishing `if`, `when`, and `cond`

Kodiologist:
Closes #2210.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -5,6 +5,9 @@ Unreleased
 
 Breaking Changes
 ------------------------------
+* `if` now requires all three arguments. For cases with less than
+  three arguments (i.e., with no else-clause), `when` is a drop-in
+  replacement.
 * `cond` has a new unbracketed syntax::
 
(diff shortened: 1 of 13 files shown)


allison-casey (approved):
:tada:

PR #2241, opened 2022-02-27
Add an xfail test for newline styles in string literals

### LINKED ISSUE #2239: In code with DOS-style newlines, literal newlines are treated differently in double-quoted strings and bracket strings
This bug is most relevant to Windows users, but can be observed on Linux. For example, make a file with the contents

```
(print (repr r"hello
goodbye"))
(print (repr #[[hello
goodbye]]))
(print (repr #[[
hello
goodbye]]))
```

where the newlines are all 0x0D 0x0A (`unix2dos` is handy). The output is then

```
'hello\ngoodbye'
'hello\r\ngoodbye'
'\r\nhello\r\ngoodbye'
```

So in the case of double-quoted strings, Hy reads in `"\n"` for the newlines regardless of the newline style in the source code, as Python does for triple-quoted strings, but bracket strings keep whatever newline style was used in the source. (Furthermore, Hy doesn't trim a leading newline, as it's supposed to for bracket strings, if that newline is 0x0D 0x0A.) My feeling is that Python made the wrong choice here with this bit of magic, particularly for raw strings, but we should probably aim for consistency with it.

Tests for any fix to this bug should probably add the step `run: git config --global core.autocrlf false` to `.github/workflows/tests.yml`. By default, the GitHub Actions runner for Windows translates all newlines at checkout-time, which can complicate things.

Kodiologist:
Tests #2239.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,6 +27,7 @@ jobs:
       PYTHONPATH: .
 
     steps:
+    - run: git config --global core.autocrlf false
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
       with:
--- a/tests/compilers/test_ast.py
+++ b/tests/compilers/test_ast.py
@@ -1,3 +1,5 @@
+# fmt: off
+
 import ast
 
 import pytest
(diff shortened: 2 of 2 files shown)


allison-casey:
Is this worth doing here? https://github.com/allison-casey/hy/tree/custom-parser may have solved this already and even if it hasn't it will completely remove the lexer that this PR changes.

Kodiologist:
I try not to make any assumptions about when or if that will land, but if it does and wipes out the lexer change, at least we'll still have the test?

allison-casey:
I was actually gonna open a PR for it as soon as i can clean up the commits. Maybe we can just xfail the test and discuss it again once I do in the next day or two?

Kodiologist:
Oh, if it's coming that soon, sure. I'll rebase out the actual fixes and xfail the test.

Kodiologist:
Done.

PR #2242, opened 2022-03-02
Custom Parser and Reader Macros

### LINKED ISSUE #722: Fully functional reader macros

allison-casey:
closes #722 

Open question if hy should support such a permissive reader macros. This PR has it so you can define a reader for any single char or any sym prefixed by a `#`.

In my opinion reader macros prefised by `#` are more idiomatic and restricting user defined macros to only ones prefixed by `#` would make Hy more predictable and tooling easier down the road. I'd be in favor of that kind of restriction, but i'll leave that as an open question for discussion. 

There's also the behavior of reader macros on require. Right now, readers  macros need to be required *per module* and then enabled *per module*. This is how Racket works and to me seems like the right thing to do for a macro that *drastically* change how code looks and is written. Again another question for discussion though. 

@scauligi as this was your baby, do you have any thoughts?


Kodiologist:
Ho nelly! It's happening! Thanks for all the work you two put into this.

I agree that we don't need single-character reader macros in addition to `#`-prefixed reader macros. However, I'm not sure `#` is the best choice of prefix because we already use that for tag macros. (This is not a coincidence. Tag macros were originally called "reader macros", although they've always had the semantics of regular macros, hence the rename.) Perhaps we should use a different character. Alternatively, we could get rid of tag macros as a concept, so to get the effect of what is currently a tag macro, the programmer would write a reader macro that immediately falls back on the standard reader to do the reading part.

> There's also the behavior of reader macros on require. Right now, readers macros need to be required *per module* and then enabled *per module*.

I think that's redundant because the syntax for defining or requiring reader macros already makes it clear that they're reader macros. When you see `(require mymodule :readers ["!"])` or `(defreader "!" …)`, you know that `!` is going to be used as a reader macro, so why have the extra step?

> Unlike requiring regular macros, reader macros must be specified as strings, cannot be renamed with ``:as``, and are not made available under their absolute paths to their source module

Is that for implementation reasons or theoretical reasons?

I don't blame you for wanting to rename `parse_one_thing`, which is probably only named that because of [an annoying song I had stuck in my head at the time](https://www.youtube.com/watch?v=iMfoM9Y8viE), but maybe "parse_one_form" makes more sense than "parse_one_node".

allison-casey:
I wonder if we can make tag macros just a syntactic sugar for a reader macro that parses one thing and then passes that thing to the underlying macro function? Reader macros are a bit unwieldy by their nature and a sugared form could continue to be useful if you don't need to take advantage of the rest of the reader's functionality. 

> I think that's redundant because the syntax for defining or requiring reader macros already makes it clear that they're reader macros

That's valid. Let me test enabling readers at require to and see how it feels. 

> Is that for implementation reasons or theoretical reasons?

This is currently an implementation thing, but it think it could be done to make `:as` and absolute paths work. Though, those changes would be non breaking and i think can wait for a future PR so as not to make this take any longer than it has. 

> but maybe "parse_one_form" makes more sense than "parse_one_node".

Makes sense enough. Let me double check though that there isn't some other reader related stuff that uses `node` as the idiom

Kodiologist:
> I wonder if we can make tag macros just a syntactic sugar for a reader macro that parses one thing and then passes that thing to the underlying macro function?

Probably. One can imagine a macro-writing macro in Hyrule like `(deftag my-tag-macro [arg] (f arg))` that defines `my-tag-macro` as a reader macro that calls the reader, binds the result to `arg`, and then evaluates `(f arg)` like a normal macro.

> This is currently an implementation thing, but it think it could be done to make `:as` and absolute paths work. Though, those changes would be non breaking and i think can wait for a future PR so as not to make this take any longer than it has.

Okay.

allison-casey:
`deftag` as a sugared reader macro works nicely. Just need to go through and replace old `defmacro` tag macros with reader macros and yoink that functionality out of `defmacro`. Otherwise, i've restricted user readers to `#` prefixed and enabled them by default in the module they're declared and on `require`.

Kodiologist:
Neat.

allison-casey:
* `deftag` is a sugar for simple read time macros
* `defmacro` no longer supports tag macro semantics
* `defreader` macros are always prefixed with a `#`
* reader macros are enabled immediately after they're defined and on `require`

did i miss anything?

Kodiologist:
That sounds like everything we've discussed. I hope to take another look at the code soon.

hy/lex/hy_reader.py
```diff
@@ -0,0 +1,456 @@
...
+        """General handler for reader macros (and tag macros).
+
+        Reads a full identifier after the `#` and calls the corresponding handler
+        (this allows, e.g., `#reads-multiple-forms foo bar baz`).
+
+        Failing that, reads a single character after the `#` and immediately
+        calls the corresponding handler (this allows, e.g., `#*args` to parse
+        as `#*` followed by `args`).
+
+        Failing that, parses the text as an old-style tag macro (e.g., `#ident
```
scauligi, line 271: Minor nit, this part of the docstring is no longer true

scauligi (commented):
This is shaping up super well, once again many thanks for taking this on allie!

Here are some things I came across while taking this for a test run:
- What is the expected interface for `deftag`/`defreader`? Ie for `(defreader THING ...)` is THING supposed to be a (well-formed) symbol, a string, or any sequence of characters?
  - If we want to support the latter case (eg `(defreader : ...)` instead of `(defreader ":" ...)` then `defreader` itself will likely need to be a reader macro
- Along those lines, bad error message for using the wrong type for the reader name, eg `(defreader : '5)` produces `TypeError: can only concatenate str (not "Keyword") to str` (similar for other types, like Integer if you try to define `(defreader 4 '5)` instead of `(defreader "4" '5)`)

- `(defreader name [expr] ...)` looks like it's specifying args, but this actually just a list as the first body expression. This could just be a #docmebro? or alternatively we can explicitly require an empty args list (eg `(defreader name [] ...)`) to make this explicit. I'm leaning towards the empty args list to make it consistent with all the other `def...` forms.

hy/lex/hy_reader.py
allison-casey: fixed

Kodiologist:
It looks there is no equivalent of the old `hy.read`, and the new function of that name actually works like `hy.read-str`. Is that a mistake?

```python
assert can_compile(r'f"hello {"n"} world"')
```

Are you sure this is correct? It seems counterintuitive to me, and it's not consistent with Python.

I think there's a mistake in `test-tag-macros`. You're testing `(deftag [a b] '1)` and `(deftag [] '1)` when you mean `(deftag foo [a b] '1)` and `(deftag foo [] '1)`.

Some tests seem to have been lost, such as `test_bad_tag_macros` for `https://github.com/hylang/hy/issues/1965` (or if you did make a new version of that test, it's missing the issue link).

scauligi:
> `assert can_compile(r'f"hello {"n"} world"')`
> Are you sure this is correct? It seems counterintuitive to me, and it's not consistent with Python.

My vote is to keep this; it's a lot easier to use format strings (and feels lispier) when curly braces work like `unquote` rather than being "just" a part of the string.

> Putting in [] mistakenly is harmless because it's a no-op

Fair point!

Kodiologist:
Oh right, I forgot to mention: `deftag` and `@#` can be moved to Hyrule, since they just use the built-in reader-macro features now.

allison-casey:
- user defined reader macros only allow symbols now
- require `:readers` uses bare symbols and not `#` prefixed strings
- added back the missing `test_bad_tag_macros` test

this does mean that `#:` slice tag macro is no longer possible (at least from the user perspective). there's nothing stopping us from adding it as a built in reader. Which i would be for anyways given how extensive slicing is in python and its libraries. 

@scauligi can you speak to the `read` vs `read-str` point kodi brought up? 

I think that's everything?

Kodiologist:
If you're going ahead with `f"hello {"n"} world"`, don't forget a NEWS line for it, since it's a breaking change.

> this does mean that `#:` slice tag macro is no longer possible (at least from the user perspective).

No reason we couldn't just call it something else, right? We should have plenty of decent options without having to give it special accommodation in the reader. Perhaps `#s`, for "slice".

> I'll open a followup PR with moving `#@` to hyrule after this so tests will pass in hyrule

It looks like Hyrule is already broken by this PR. Not your fault; that's what usually happens with breaking changes to Hy. My recommendation is, remove `deftag` from and delete `#@` in this PR, and then we can add them to Hyrule in the same PR that unbreaks it.

scauligi:
> It looks there is no equivalent of the old hy.read, and the new function of that name actually works like hy.read-str. Is that a mistake?

That was intentional; the API is different since the new reader can directly take a stream where (iirc) the old reader needed the entire input at once as a string; so there's no need to have a separate `read` and `read-str` anymore.

> this does mean that #: slice tag macro is no longer possible

Hmm, I am leaning slightly towards adding `#:` as an additional baked-in reader; it should actually be very easy in the new reader. It's also one of the few non-user-definable constructs because of the `:` but also a very good mnemonic for what it does.

Kodiologist:
> the API is different since the new reader can directly take a stream

If that was intended, it doesn't seem to work:

```
=> (with [o (open "/tmp/test.hy")] (hy.read o))
  File "stdin-1219427163456d350c0c7b476cc55454c5eff66a", line 1, in <module>
    (with [o (open "/tmp/test.hy")] (hy.read o))
TypeError: initial_value must be str or None, not _io.TextIOWrapper
```

scauligi:
Ah, looks like I mixed myself up; internally it uses a stream but the reader interface indeed still expects the whole string up front.

I expect changing the function names shouldn't affect users too much, but do you want to keep the original names?

Kodiologist:
What are you proposing to change them to? As the PR stands, one of them is gone and the other was given its name.

scauligi:
Ah as in keep the original API (bring back `read-str`) or keep the PR as it currently is (no `read-str`).

Kodiologist:
I think we should bring back the original API, then. It seems a waste to implement a reader that is internally stream-based and then require the user to read an entire file into a string before using it, doesn't it?

allison-casey:
we will always have to read the entire file in as a string to enable source tracking in the same way as python. The stream basis of the reader is mostly to enable incremental compilation. I pushed a commit as a little test that changes the parser to take a stream only and then allowing the existing lex methods to take either a `TextIOBase` stream or a `str` that gets converted to a stream as a convenience which works pretty well. 

I've also deleted the `#@` tag macro and `deftag`.
It's worth noting that it's still possible for us to define a `#:` reader macro with some shenanigans via:
```clojure
(eval-when-compile
  (hy.eval
    `(deftag ~(hy.models.Symbol ":" :from-parser True) [expr]
       '1)))
```
:woman_shrugging:

Kodiologist:
> we will always have to read the entire file in as a string to enable source tracking in the same way as python. The stream basis of the reader is mostly to enable incremental compilation.

I see. I'll probably understand the details better once we've hammered out the user-facing stuff and I look deeper into the implementation.

>  I pushed a commit as a little test that changes the parser to take a stream only and then allowing the existing lex methods to take either a `TextIOBase` stream or a `str` that gets converted to a stream as a convenience which works pretty well.

Okay. What do you want to do about `read` vs. `read_str`? Add back `read_str` and make `read` stream-only? Keep `read_str` out in favor of letting `read` do the jobs of both of the old functions?

allison-casey:
I think we leave out `read-str` in favor of `read` taking a string or string stream. This way all the lex functions have the same interface of

`stream: TextIOBase | str, filename: None | str`

which is nice. what do you think?

Kodiologist:
The Python standard library has a few cases where different functions are provided for string and file input, like `json.reads` vs. `json.read`, but that should be overkill, since anything that's like a file object but that you want to treat as a string could always be converted with `str`.

allison-casey:
Will do!

I thought about the read/reads distinction, but it doesn't really work for module since that's just plural 🤷‍♀️

Anyways either would be converted to stream under the hood regardless. I'll try and get the rebase in tonight

allison-casey:
this is a spicy rebase :worried:

Kodiologist:
Sorry about that. I imagined it would be a little laborious (because e.g. the mangling code, which you're moving, changed retroactively) but not conceptually difficult. Let me know if you want me to try. I like to think I'm good at resolving merge and rebase conflicts.

allison-casey:
Oh i should be okay. Just need to hunt down some extra stuff that got silently eaten by the rebase. I appreciate the offer though!

allison-casey:
Alright, I think this is everything! Let me know if I missed something in the rebase @Kodiologist & @scauligi

allison-casey:
sounds good. 

although, `core tag macros enabled by default` wasn't supposed to be a no-op. it enabled `hy.core.macros` to define tag macros like it does for regular macros. Maybe it got eaten by my rebase? I'd still like to have the functionality even though we aren't using it currently. Let me see if i can salvage it from my local copy or otherwise add it back in correctly.

Yeah, there's definitely some stuff in here I'd like to refactor. I'll make a tech debt card to get back to it later.

Kodiologist:
But we have no core tag macros, right? So it would just be dead code unless (until?) we added some.

Kodiologist:
Especially since tag macros themselves are no longer part of core. Maybe you were thinking of reader macros. But we don't have any core reader macros, either, I think.

Kodiologist:
Unfortunately, we can't actually build the documentation, because `sphinxcontrib-hydomain` uses internal Hy bits that have changed. Presumably you'll worry about that once this is merged.

Summary of new changes:

- Added back a test of `r` in f-strings.
- Added back `test-read` as `test-read-file-object`.
- Made the new `read` raise `EOFError` for EOF, like the old one did, instead of returning `None`.
- Changed `def read_many` to `def tokenize` in `test_lex.py` to save a huge number of diff lines.
- More miscellaneous code simplification.
- Removed obsolete documentation for `enable-readers`.
- Rewrote the tutorial addition to be shorter and use an example that can't be implemented as an ordinary macro.
- Moved non-tutorial documentation to `api.rst`.
- Small rewording to `require` documentation for correctness.

Kodiologist:
Okay, this is pretty much done now. I should take another glance at it another day, just to check I didn't do anything stupid during all that rebasing, but otherwise I think it's good. Let me know any suggestions, complaints, etc. Teamwork makes the dream work.

Summary of changes:

- Changed the words "tag macro" to "reader macro" in an error message.
- Removed documentation of tag macros.
- Changed the argument order in `=` forms in tests to put the constant second instead of first.
- Added back a reader-macro version of the removed `test-tag-macro` in `native-tests.mangling`. The new test failed, so I fixed the bug.
- Reverted a large unrelated whitespace change in "Clean: Remove #@ tag macro".
- Removed the dependency on `rply`.
- Added some macro and reader-macro tests for `hy2py`.
- Filled out NEWS updates.

Kodiologist:
Yep, looks good.

As you can see in NEWS, I came down on the `deftag` question to thinking that there's no need to provide such a macro, or tag macros as a concept, because it's easy, and not terribly useful, to write a reader macro that just consumes the next form.

Kodiologist:
If there are no objections, let's merge this.

allison-casey:
I'll be back from my walk in a bit and i can give this a quick once over

allison-casey:
think everything looks good. only thing i'm noticing now is that i didn't let `:readers` require all via `(require some.module :readers *)` which was a mistake on my part. I can try and add that real quick since we kinda need that for `(require hyrule :macros * :readers *)`

Kodiologist:
I see, that's a good point.

Kodiologist:
Actually, since this PR is already so big, and we probably have more such things we need to do, would you like to do it as a follow-up PR?

allison-casey:
works for me

allison-casey:
i'll rebase the tuple syntax pr and remove the whitespace stuff once we merge

Kodiologist:
Very good. I shall rebase and push the merge button.

PR #2247, opened 2022-03-14
Fix scoping in comprehensions in `let`

### LINKED ISSUE #2224: `let` doesn't detect assignments in loop clauses of comprehension forms

Kodiologist:
- Fixes #2224.
- Replaces #2228.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -20,6 +20,7 @@ Bug Fixes
 * Elements of `builtins` such as `help` are no longer overridden until
   the Hy REPL actually starts.
 * Crash when using Keywords as `match` value.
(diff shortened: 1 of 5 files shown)


scauligi (approved):
lgtm :+1:

PR #2248, opened 2022-03-15
Fix completion not unmangling names

scauligi:
This was really getting under my skin.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -21,6 +21,7 @@ Bug Fixes
   the Hy REPL actually starts.
 * Crash when using Keywords as `match` value.
 * Fixed a scoping bug in comprehensions in `let` bodies.
+* Tab completion in the Hy REPL now properly unmangles symbol names.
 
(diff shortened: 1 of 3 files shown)


Kodiologist:
Because `unmangle` can't handle everything, this seem to prevent the completer from being able to complete some symbols, such as the malformed mangled symbol `hyx_XaX`.

scauligi:
Good catch, should be handled now. Are there any other edge cases that we might run into?

Kodiologist:
Not that I can think of.

Catching `Exception` seems dangerous; can you narrow which types you catch?

`namespaces` and `test_cases` are only used once each, so they can be inlined.

I think that's about it.

Kodiologist:
@scauligi Are you still working on this, or shall I look at it again?

scauligi:
oh shoot, forgot to post here
yeah it should be good for review

Kodiologist:
Looks good to me. Rebase it and merge.

Kodiologist (approved):

PR #2250, opened 2022-03-20
Refinement of command-line options

### LINKED ISSUE #1445: Mangle module names passed to `hy -m`
$ echo '(print "hello")' > hi_ho.hy
    $ hy -m hi_ho   
    hello
    $ hy -m hi-ho
    hy: module 'hi-ho' not found.

Kodiologist:
Closes #1445.

### DIFF
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -498,54 +498,25 @@ def cmdline_handler(scriptname, argv):
     # options, such as `-c`. So, we can't use `argparse`.
 
     defs = [
-        dict(
-            name=["-h", "--help"], action="help", help="show this help message and exit"
-        ),
-        dict(
-            name=["-c"],
-            dest="command",
-            terminate=True,
-            help="program passed in as a string",
-        ),
-        dict(
-            name=["-m"],
-            dest="mod",
-            terminate=True,
-            help="module to run, passed in as a string",
-        ),
-        dict(
-            name=["-E"],
-            action="store_true",
-            help="ignore PYTHON* environment variables",
-        ),
-        dict(
-            name=["-B"],
-            action="store_true",
-            help="don't write .py[co] files on import; also PYTHONDONTWRITEBYTECODE=x",
-        ),
-        dict(
-            name=["-i"],
-            dest="icommand",
-            terminate=True,
-            help="program passed in as a string, then stay in REPL",
-        ),
-        dict(
-            name=["--spy"],
-            action="store_true",
-            help="print equivalent Python code before executing",
-        ),
-        dict(
-            name=["--repl-output-fn"],
-            dest="repl_output_fn",
+        dict(name=["-h", "--help"], action="help",
+            help="show this help message and exit"),
+        dict(name=["-c"], dest="command", terminate=True,
+            help="program passed in as a string"),
+        dict(name=["-m"], dest="mod", terminate=True,
+            help="module to run, passed in as a string"),
+        dict(name=["-E"], action='store_true',
+            help="ignore PYTHON* environment variables"),
+        dict(name=["-B"], action='store_true',
+            help="don't write .py[co] files on import; also PYTHONDONTWRITEBYTECODE=x"),
+        dict(name=["-i"], dest="icommand", terminate=True,
+            help="program passed in as a string, then stay in REPL"),
+        dict(name=["--spy"], action="store_true",
+            help="print equivalent Python code before executing"),
+        dict(name=["--repl-output-fn"], dest="repl_output_fn",
             help="function for printing REPL output "
-            "(e.g., hy.contrib.hy-repr.hy-repr)",
-        ),
-        dict(
-            name=["-v", "--version"],
-            action="version",
-            help="show program's version number and exit",
-        ),
-    ]
+                "(e.g., hy.contrib.hy-repr.hy-repr)"),
+        dict(name=["-v", "--version"], action="version",
+            help="show program's version number and exit")]
 
     # Get the path of the Hy cmdline executable and swap it with
     # `sys.executable` (saving the original, just in case).
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -640,7 +640,7 @@ def cmdline_handler(scriptname, argv):
     if "mod" in options:
         set_path("")
         sys.argv = [program] + argv
-        runpy.run_module(options["mod"], run_name="__main__", alter_sys=True)
+        runpy.run_module(hy.mangle(options["mod"]), run_name="__main__", alter_sys=True)
         return 0
 
     if "icommand" in options:
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -498,23 +498,23 @@ def cmdline_handler(scriptname, argv):
     # options, such as `-c`. So, we can't use `argparse`.
 
     defs = [
-        dict(name=["-h", "--help"], action="help",
-            help="show this help message and exit"),
-        dict(name=["-c"], dest="command", terminate=True,
-            help="program passed in as a string"),
-        dict(name=["-m"], dest="mod", terminate=True,
-            help="module to run, passed in as a string"),
-        dict(name=["-E"], action='store_true',
-            help="ignore PYTHON* environment variables"),
         dict(name=["-B"], action='store_true',
             help="don't write .py[co] files on import; also PYTHONDONTWRITEBYTECODE=x"),
+        dict(name=["-c"], dest="command", terminate=True,
+            help="program passed in as a string"),
+        dict(name=["-E"], action='store_true',
+            help="ignore PYTHON* environment variables"),
+        dict(name=["-h", "--help"], action="help",
+            help="show this help message and exit"),
         dict(name=["-i"], dest="icommand", terminate=True,
             help="program passed in as a string, then stay in REPL"),
-        dict(name=["--spy"], action="store_true",
-            help="print equivalent Python code before executing"),
+        dict(name=["-m"], dest="mod", terminate=True,
+            help="module to run, passed in as a string"),
         dict(name=["--repl-output-fn"], dest="repl_output_fn",
             help="function for printing REPL output "
                 "(e.g., hy.contrib.hy-repr.hy-repr)"),
+        dict(name=["--spy"], action="store_true",
+            help="print equivalent Python code before executing"),
         dict(name=["-v", "--version"], action="version",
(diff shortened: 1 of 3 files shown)


tests/test_bin.py
```diff
@@ -623,3 +631,21 @@ def test_hystartup():
...
+
+
+def test_output_buffering(tmp_path):
+    tf = tmp_path / "file.txt"
+
+    pf = tmp_path / "program.hy"
+    pf.write_text(f'''
+        (print "line 1")
+        (import  sys  pathlib [Path])
+        (print :file sys.stderr (.strip (.read-text (Path #[=[{tf}]=]))))
```
allison-casey, line 643: placing the text after the keyword argument shouldn't be legal. it should crash with `positional argument follows keyword argument` as it does in python. Unrelated to this PR, but we should open an issue for this.

hy/cmdline.py
```diff
@@ -1,10 +1,13 @@
+# fmt: off
```
allison-casey, line 1: delete this

allison-casey (changes requested):

hy/cmdline.py
Kodiologist: I thought you were over this, or at least had decided that deviation from Black was acceptable in some circumstances, since you didn't object to some other PRs with `fmt: off`. What's going on?

tests/test_bin.py
Kodiologist: I've always seen this as a feature, and went out of my way to make it work for method calls (#1167). A lot of times we can't effectively get around Python limitations, but it's nice when we can.

Kodiologist:
@allison-casey Let's wrap this up. This has gone on for nearly two months. If you think Black formatting need not be adhered to unconditionally, then explain to me where you're drawing the line and why. I'm asking for pretty occasional exceptions, so maybe we can work something out.

You've said that `fmt: off` "defeats the whole purpose of formatting to a consistent style". So if you think it'd be better to remove Black and undo all the changes it made than to ever use `fmt: off`, I'm certainly game for that.

If, on the other hand, you want to keep Black, and you still don't intend to budge an inch, now or in the future, and the only reason you didn't also stop those other PRs is because you didn't notice the spacing-fix commits in them, then I'm just going to overrule these objections. Stopping every PR with the slightest deviation from a machine-enforced whitespace style, heedless of any consequences for readability, while rejecting any possibility of compromise, all to save yourself the inconvenience of typing `black` (or rather: `perl -pi -e 's/^# fmt: off\n//m' ./**/*.py; black`), is not a responsible use of your privileges as a core developer.

I want to be clear, if there's any doubt: you're a very smart, capable, and hardworking member of Hy's development team, and I greatly value everything you've done for Hy over the past year and a half. I'm not trying to punish you. It's just that, on this oddly specific matter, you've been disruptive, in a manner that is uncharacteristic for you. You're due a lot of patience and deference, but not an infinite amount.

allison-casey:
If you'd like to overrule @brandonwillard and I's approval of #2209, as well as @gilch and the community members that voiced support for it, all to save yourself the inconvenience of following basic industry best practices (including those defined by Python itself) then go ahead. But know that I'm out if you do

Kodiologist:
Undoing #2209 is definitely overkill so long as there's no change in people's opinions. All I'm going to do is merge PRs that you're stopping because of whitespace.

> But know that I'm out if you do

Do you mean you're going to quit the whole project over whitespace? Seriously? Well, I'm gonna miss you a lot, but I can't make you stay if you don't want to.

paultag:
I've been out of pocket due to personal life stuff, but we really need to resolve this and get everyone on the same page - I don't want to put my thumb on the scale, but if we're at an actual impasse that's going to cause people to stop working on the project, I can play that role.

I'd much rather the active committers come to consensus here, though! I don't know how we want to deal with it. I would like to be helpful, but I'd like to not over-involve myself in routine matters here.

My initial thought is to have @allison-casey and @Kodiologist iron this out between the two of you in some way that's most productive, and set the standard on what  commits going forward look like. If that's not going well / stuck at an unresolvable deadlock, I can weigh in.

Kodiologist:
Yes, to be clear, if there's some sort of compromise we can come to, rather than me applying brute force and Allie quitting the project, that would be a much better outcome.

lafrenierejm:
I had raised this [before](https://github.com/hylang/hy/pull/2221#discussion_r810558135), but only ever saw [a response from @Kodiologist](https://github.com/hylang/hy/pull/2221#discussion_r810558496). As a possible point of compromise, maybe it's worth a larger discussion?

> What are both of your opinions on delimiting individual blocks with fmt: {off,on}? Obviously that's more verbose than either a file-wide fmt: off or giving up manual formatting entirely, but maybe would it be an acceptable compromise for now? See https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html?highlight=skip#code-style for the relevant docs.

Kodiologist:
Yeah, still seems better than nothing to me.

allison-casey:
There's very little point to having black if it's going to be incrementally chipped away by chunks of `fmt: off/on` that i'd inevitably undo the next time i touched that code. Those chunks would have to be manually fixed regardless if flake8 or other code quality checks were ever introduced anyways as they already fail as they are. Either we use it or we don't.

Kodiologist:
So then let's not have Black. But you're also against that. So, if there's anything you would accept short of 100% fealty to the Dark One, I can't figure out what it is.

lafrenierejm:
Short preface:
1. I acknowledge that I have very little investment in Hy relative to most here.
1. I _do_ find value in using Black.
My intent is not to invalidate anyone's feelings.

> There's very little point to having black if it's going to be incrementally chipped away by chunks of `fmt: off/on`

IMO this feels like a straw man. I don't think it's necessarily the case that all benefits of Black would be lost by allowing blocks of hand-formatted code. My impression is that "even" Kodi (using @Kodiologist as an example because so far they have expressed the most discomfort with Black) is not 100% opposed to Black's formatting. I agree that the benefits of using an auto-formatter would be reduced by allowing hand-formatted blocks, but IMO auto-formatting should ultimately exist to serve current and future contributors rather than as an end in itself. That said, would it at least be worth experimenting with allowing per-block formatting before anyone here decides they need to step away from the project?

Kodiologist:
> My impression is that "even" Kodi … is not 100% opposed to Black's formatting.

For sure. I don't think autoformatting is a good idea, outside of desperate cases like deobfuscation, but a lot of Black's changes to this codebase were either innocuous or improvements to what was there before. The trick is the small minority of cases where its choices are not so great. And I don't think those cases arise because the Black style is a bad one overall, but because Black is a computer program and necessarily can't capture a human's understanding of code.

scauligi:
> delimiting individual blocks with `fmt: {off,on}`?

My worry with this is with logistics and the inevitable gray areas of both review and later editing.

Suppose, e.g., that a poorly stylistic developer submits a PR to Hy that includes a *clearly* badly formatted block, but enclosed with `fmt: off`.
I think consensus would definitely be to request the removal of the formatting comments and reformat.

But now suppose a more stylistically aware developer submits a PR with a maybe-or-maybe-not well-formatted block, enclosed with `fmt: off`. Do we reject? Accept with a warning? Where do we draw the line? Is it just up to whichever core developers respond to the PR? What if there's disagreement among them?

Suppose there's a code block already in the codebase delimited by `fmt: off`, and I happen to edit some part of it in a PR. If I remove the `fmt: off` comments because I think black's formatting looks better, will my PR be rejected until I restore the original formatting? Should I match the existing formatting if I add to the same code block even if I disagree with its style? What if I add a different code block elsewhere in the file, but it has similar or analagous code to the custom-styled one? Should I match the custom style to keep it consistent?

If we're going to allow `fmt: off` blocks, then I feel we would need a very strict set of guidelines/rules for these kinds of cases to avoid this sort of spiraling discussion again.

---

My personal preference is black everywhere, as I find it quite reasonable. E.g., I much prefer the black-formatted version of the argparse commands in this PR, because it's a lot easier for me to see each keyword argument.

I'm very cautiously open to allowing `fmt: off` as subjectively desired for *data* (array/dict constants, test data, etc.) but I do prefer auto formatting for all code.

Kodiologist:
I think we can handle it the same way we handled it in the pre-Black days. If you're reviewing a PR and you think the whitespace is unreasonable, then negotiate with the PR author. If you're writing a PR, do whatever makes sense to you and see if anybody objects. Somehow, as a community, we managed to be level-headed and proportionate about whitespace for a little under a decade. We can do it again. The trick to it isn't "a very strict set of guidelines/rules" but an interest in getting things done.

> Suppose, e.g., that a poorly stylistic developer submits a PR to Hy that includes a *clearly* badly formatted block

For what it's worth, I don't think I've ever seen this happen. Probably because the kind of programmer who submits PRs to a Lisp implementation is the not the kind of programmer who's desperately banging out a guess-the-number program for Computer Science 101.

allison-casey:
django, scikit-learn, and pandas all use black. Between their _million_ lines of code between them, `fmt: off` shows up exactly _zero_ times. Hy is not special. There's nothing we've done that is different or new from what the Python community has broadly accepted black capable of doing. 

I'm with scauligi. I much prefer the black formatted versions of this PR as well as all the PRs kodi has turned black off in.

Kodiologist:
That would set a precedent that if a core developer is unhappy with something, they can refuse all attempts to reason with them and threaten to resign. What if I threatened to resign? I'm not going to pretend I want to resign, because I don't, least of all over not getting my way with respect to whitespace, but I'm trying to bend over backwards to accommodate what I thought from the beginning was a really bad idea, and I'm getting zero consideration.

I've said it before and I'll say it again: if you can't stand seeing any deviation from the Black style, then just run `black` locally. I get to see code as it was intended to be formatted by a human, you get to see the Umbral Entity's standard, and everyone's happy. If this seems like too much of an inconvenience to help me read the codebase I've been formatting with tender loving care for the last 6 years, then… I don't know what to say beyond that. I will simply be sad. And continue to be sad as Black messes up my formatting every day for the subsequent 6 years.

paultag:
@Kodiologist the precedent I'm committing to here is that there was a deadlock in our normal process. I put on the BDFL hat and came in over the top with an edict in order to break the stalemate according to my own judgement. 

I don't like doing it - and it's perhaps a worse precedent, but I don't see this thread resolving itself

Kodiologist:
At least we can put this behind us. I'll add some blackening commits and merge these hoary PRs later today.

PR #2257, opened 2022-04-11
Unified Tuple Syntax

allison-casey:
There was some discussion in the new reader PRs about giving tuples their own syntax of `#(1 2 3)` so I thought I'd give it a whirl and see how it feels now that the reader PR is just about done. 

It's actually quite nice and gives us a number of nice benefits

* It makes all datastructures distinct from expressions
```clojure
[1 2 3]
{"a" 1  "b" 2}
#{1 2 3}
#(1 2 3) 

(, 1 2 3)  ; is an expression that actually represents a datastructure
```

* It lets use make `,` act as whitespace. which makes delimiting single line paired constructs easier to parse if that's your thing
```clojure
(import hy [read-module], typing [List])
{"a" 1, "b" 2}
```

* It lets use simplify logic around identifying tuples 
```clojure
(defmacro something [x]
  ;; you don't need to check for an expression with a comma 
  ;; as the first symbol anymore!
  (if (isinstance x hy.models.Tuple)
    "I'm a tuple!"
    "I'm not a tuple :("))
```
This simplified some stuff in the compiler as well. 
```diff
# in  `match`
-    elif (
-        isinstance(value, List)
-        or isinstance(value, Expression)
-        and value[0] == Symbol(",")
-    ):
-        patterns = value[0] if isinstance(value, List) else value[1:][0]
+    elif isinstance(value, (Tuple, List)):

# in `scoping`
-        if isinstance(target, List):
-            return List(map(self.add, target)).replace(target)
+        if isinstance(target, (List, Tuple)):
+            return target.__class__(map(self.add, target)).replace(target)
         if (
             isinstance(target, Expression)
             and target
-            and target[0] in (Symbol(","), Symbol("unpack-iterable"))
+            and target[0] == Symbol("unpack-iterable")
         ):
```

* It's a character shorter
```clojure
(, 1 2 3)
#(1 2 3)
```

It does mean you can't use commas in numerical values anymore. i.e. `65,432` is not longer valid. Which i think is fine given that it's already the python idiom to use `65_432`

I opened this as a draft so people can try it out and see how it feels. I personally quite like it, but i'd love to hear what other people think.


Kodiologist:
I like the sound of it except for the part where a comma is allowed as whitespace. That seems weird (in the context of Lisp) and not too useful, since you can always add more space characters to get more visual separation. I agree that the new syntax makes the `,` macro redundant and it can go. And we could live without commas in numeric literals, although I like them, because underscores serve the same function. But I'd rather `,` stay as an identifier character for the user's use than a whitespace character.

allison-casey:
Fair. A lot of lisps use `,` for unquoteing. I use clojure a lot and wanted to see what whitespace comma felt like in Hy. I can remove it of we go with this new tuple syntax.

Kodiologist:
I was thinking the other day that I'm glad we went with Clojure-style `~` for unquote instead of the traditional `,`. It's more visible.

allison-casey:
oh definitely. i've always found comma unquoting confusing looking

allison-casey:
rebased and removed comma whitespace. Should be ready for review @Kodiologist

Kodiologist:
This looks good. Thanks. I'm not a huge fan of the whitespace change in `tests/native_tests/comprehensions.hy`, but if you want to do it, that's okay; just do it in its own commit.

Next up, documentation changes. For NEWS, don't forget to mention that tuples now wrap to `Tuple` instead of `List`. In the past, if I recall correctly, I've done things like substitute a tuple in as the lambda list in a `defn`, relying on the promotion to `List`.

allison-casey:
undid the comprehensions stuff and updated news

Kodiologist:
Cool, I think all that's needed is documentation updates, then.

allison-casey:
docs updated

Kodiologist (approved):

scauligi (approved):
heck yeah

Issue #2096, opened 2021-06-13
Dotted symbols should probably be parsed as expressions


Kodiologist:
Currently, `foo.bar` parses as `(Symbol "foo.bar")`, and the compiler is responsible for interpreting the dot. Macros often have to specifically consider dotted identifiers, too. Things would be simpler if `foo.bar` parsed the same way as `(. foo bar)`. Fully excising dots from identifiers (other than float literals) may require using a new model type for dots, in constructs like `(. foo bar)` as well as `(.foo bar)`.

allison-casey:
i think this is worth exploring for 1.0 since it would be a breaking change for macros, but the upside seem pretty nice. Probably should be one of the last things we do though after the less parser based changes are done

scauligi:
This could be implemented in the [reader](https://github.com/hylang/hy/issues/722#issuecomment-862040753)! At read time we can read identifiers like `foo.bar` as `(. foo bar)` and expressions like `(.bar foo arg)` as `((. foo bar) arg)`.

allison-casey:
so poking around at this i don't think we'll be able to excise dots from the beginning of symbols. i.e. `(.bar foo arg)` because the conversion changes the order of the arguments in a way i don't think threading macros (or any argument modifying macro) can understand (`foo` moving in front of `bar` in this case)

How would the following be parsed?
```clojure
(-> foo
  .bar)
```

I don't see the same issue for dots in the middle of symbols, but i can't see a way out of this one.

Kodiologist:
Something like `(-> foo .bar)`, or just plain `(do .bar)`, presumably would be illegal at the reader level, since there's no attribute to get. A leading period would only be allowed right after `(`. Reader macros would still be an option if you wanted a macro that could do something else with leading periods in it, I guess. But it might be better to just have the macro treat some other character, like `,`, as a sort of escaped period.

Another option would be to choose a built-in default attribute for these cases, such as `X`, so `(do .bar)` would be understood as `(do (. bar X))`. I'm not sure if that would help at all.

scauligi:
Alternatively, we could special-case a bar leading-dot form as a "method getter", kind of like how we special-case bare pyops; e.g., `.bar` on its own would be equivalent to something like
```hy
(fn [this #*args] (.bar this #*args))
```

Kodiologist:
[...]

PR #2263, opened 2022-04-22
Add `...` as special syntax for Ellipsis

### LINKED ISSUE #2143: Allow `...` as shorthand for `Ellipsis` as in Python

Kodiologist:
Closes #2143.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -69,6 +69,7 @@ New Features
 * You can now set the variable `_hy_export_macros` to control what macros are
   collected by `(require module *)`
 * New macro `export`
+* `...` is now understood to refer to `Ellipsis`, as in Python.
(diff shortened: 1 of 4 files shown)


allison-casey:
Neat! Will review this when I get home tonight

allison-casey:
so everything seems good. I am curious how this will affect #2096 will we just have to special case three `.`s in the reader or something?

scauligi:
I was going to raise that point as well; it might make more sense to implement this directly in the reader?

Kodiologist:
Yes, I think the implementation will have to change to support that, and `...` will probably have to be special-cased in the reader. Fortunately, the current implementation is very simple, so very little will be lost.

allison-casey (approved):

scauligi (approved):

Issue #2150, opened 2021-08-14
Inconsistent number of executions of a file that requires itself 


Kodiologist:
Consider the following code similar to Hy's `tests/resources/bin/circular_macro_require.hy`, placed in `mydir/t1.hy`:

```
(defmacro bar [expr]
  `(print (.upper ~expr)))

(defmacro foo [expr]
  `(do (require mydir.t1 [bar])
       (bar ~expr)))

(foo "wowie")
```

If executed from outside the directory with `PYTHONPATH=. hy mydir/t1.hy`, it prints "WOWIE" once. If, on the other hand, you run the same file but with its parent directory set to your current working directory, and with the `require` form changed appropriately, you get two copies of "WOWIE".

This issue is probably related to how the `require` macro only adds a call to `hy.macros.require` in the generated Python AST if the compile-time call to `hy.macros.require` returns true.

Kodiologist:
I'm no longer entirely convinced this is a Hy bug, because [Python itself can execute the same Python file more than once in some situations](https://stackoverflow.com/questions/74677566/why-does-a-self-importing-module-execute-once-in-some-cases-and-twice-in-others) if that file is used for `__main__`, as in this case. I'm closing this issue till I get some indication we're doing the wrong thing, instead of just coping with Python's weirdness as best we can.

PR #2264, opened 2022-04-24
delmacro and macro redefinition warning

### LINKED ISSUE #1689: Macro name shadowing/collisions and removal

allison-casey:
closes #1689 

adds a `RuntimeWarning` when a user attempts to redefine a core macro and a new core macro `delmacro` a macro equivalent of `del` that deletes macros in the calling namespace. 

So the case of redefining `require` would look like:
```clojure
=> (require tests.resources.tlib [qplah])
hy.macros.require('tests.resources.tlib', None, assignments=[['qplah', 'qplah']], prefix='')
None
=> (defmacro require [] 1)
<string>:1: RuntimeWarning: require already refers to: `require` in module: `builtins`, being replaced by: `__console__.require`
  (defmacro require [] 1)
hy.macros.macro('require')(lambda hyx_XampersandXcompiler: 1)
None
=> (require)
1
1
=> (delmacro require)
None
=> (require tests.resources.tlib [parald])
hy.macros.require('tests.resources.tlib', None, assignments=[['parald', 'parald']], prefix='')
None
```

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -72,6 +72,7 @@ New Features
 * new function `hy.model_patterns.parse_if`
 * Added a command-line option `-u` (or `--unbuffered`) per CPython.
 * new function `hy.model_patterns.in_tuple`
+* new core macro `delmacro` to delete user defined or require'd macros.
 
 .. _bpo-2675: https://bugs.python.org/issue2675#msg265564
 .. _bpo-46520: https://bugs.python.org/issue46520
--- a/hy/core/macros.hy
+++ b/hy/core/macros.hy
@@ -161,6 +161,33 @@
         (lfor  x a  (hy.models.String x))
         (raise (TypeError "arguments must be keywords or lists of symbols"))))))))
 
+(defmacro delmacro
+  [#* names]
+  #[[Delete a macro(s) from the current module
+  ::
+
+     => (require a-module [some-macro])
+     => (some-macro)
+     1
+
+     => (delmacro some-macro)
+     => (some-macro)
+     Traceback (most recent call last):
+       File "<string>", line 1, in <module>
+         (some-macro)
+     NameError: name 'some_macro' is not defined
+
+     => (delmacro some-macro)
+     Traceback (most recent call last):
+       File "<string>", line 1, in <module>
+         (delmacro some-macro)
+     NameError: macro 'some-macro' is not defined
+  ]]
+  (let [sym (hy.gensym)]
+    `(eval-and-compile
+       (for [~sym ~(lfor name names (hy.mangle name))]
+         (when (in ~sym __macros__) (del (get __macros__ ~sym)))))))
+
 
 ;; Placeholder macros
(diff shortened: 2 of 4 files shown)


allison-casey:
oh looks like a fun situation, where using `eval-and-compile` makes tests pass, but breaks the repl and `eval-when-compile` makes repl work but breaks tests lol. will fix

allison-casey:
so pretty sure this is an issue related to #2150 where different hy files can be evaluated a different number of times. so we might be attempting to delete a macro that was already deleted previously in the case of `eval-and-compile` or not deleting a macro that was defined more than once in the case of `eval-when-compile`. Easy solution would be to not `NameError` on a `delmacro` call that doesn't exist, but that would make `delmacro `act differently than python `del`. I have no problem with that personally. we can always start `NameError`ing later if we fix #2150. What do y'all think?

Kodiologist:
As a general rule, I think it's fine for code that's supposed to delete something to silently do nothing if the thing doesn't exist. When it raises an error, that usually just makes programmers need to catch and ignore it. The inconsistency with `del` isn't too important considering how complex the question of whether a macro exists is.

allison-casey:
For what it's worth i don't think clojure attempts to solve the whole problem. Just warn and assume the user knows what their doing. Better macro introspection would definitely help and should probably be on the post 1.0 roadmap for sure

Fair enough. I'll change that

Kodiologist:
But `record` is still there. I mean that you don't need to bind the context manager to a name at all, because you won't use it.

allison-casey:
woops my b, missed that

Kodiologist:
Ah, shoot, I didn't rebase because I assumed you already had. My mistake.

PR #2267, opened 2022-04-26
Allow `(require ... :readers *)`

### LINKED ISSUE #2265: Support `(require module :readers *)`
Following up from https://github.com/hylang/hy/issues/2242#issuecomment-1103133136 just so we don't lose track.

I can take a swing at this pretty soon, doesn't seem like it should be too difficult.

scauligi:
Closes #2265.

Breaks latest hyrule since the API for `hy.macros.require_reader` is slightly different, but it's a very easy fix.
I can submit a PR once this one goes through.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1104,6 +1104,7 @@ base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
    Reader macros are required using ``:readers [...]``.
    The ``:macros`` kwarg can be optionally added for readability::
 
+       => (require mymodule :readers *)
        => (require mymodule :readers [!])
        => (require mymodule [foo] :readers [!])
        => (require mymodule :readers [!] [foo])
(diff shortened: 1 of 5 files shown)


Kodiologist:
Are you sure it makes sense to unmangle in the error message? The mangled form is what shows up in `import` errors, for example.

The test uses `#[f[ … ]f]`, but has no replacement fields. Did you mean `#[=[ … ]=]` or so?

I'd recommend using `setv` instead of `let` when you can, at least in Hy itself, because history shows that `let`'s semantics are liable to change.

scauligi:
> unmangle in the error message?

To me it's more ergonomic to display the unmangled names in errors (and I would propose that we unmangle the names in the other import/require errors as well).

> f-string

Yup, simple copy-paste error. Fixed.

> `setv` instead of `let`

I hope `let`'s common-case semantics will be fairly stable from here out, but you do have a point. Changed.

Kodiologist:
> I would propose that we unmangle the names in the other import/require errors as well

I don't think we can, specifically for `import`, because such errors are raised at run time. Likewise for any other Python-level errors that involve variable names. So I think it's easier to understand if we always use mangled names.

scauligi:
Ahh, Python. Curses.

Kodiologist:
Sometimes it really be like that.

Kodiologist:
After that, we need a line in the documentation of `require` about the syntax `:readers *`. Then I think this will be good to go.

scauligi:
Done!

Kodiologist:
I still see `unmangle(name)` in `macros.py` line 181. Did you forget to push?

scauligi:
woops, I reverted the one for `:readers` but forgot that I had also changed normal `require`. Fixed now.

Kodiologist:
Thanks.

PR #2268, opened 2022-05-01
Load reader macros from $HYSTARTUP

scauligi:
Allows reader macros to be defined/required in the $HYSTARTUP file and be available in the REPL.


Kodiologist:
Looks good, except you can inline `names = …` into the following line, and there's no need for a NEWS entry because reader macros themselves are new.

scauligi:
(done)

Kodiologist (approved):

PR #2270, opened 2022-05-06
Replace `with-decorator` with decorator lists in `defn` and `defclass`

### LINKED ISSUE #952: Decorators are ugly

### LINKED ISSUE #1624: we should have better documentation and error message for `with-decorator`
### Scenario 1

```clojure
hy 0.14.0 using CPython(default) 3.6.5 on Linux
=> (with-decorator identity (defn a [] 42))
=> (with-decorator identity (fn [] 42))

  (with-decorator identity (fn [] 42))
  ^----------------------------------^
HyTypeError: Decorated a non-function
```

It works on anonymous functions before (for example, 0.12.1), I believe decorating `(fn ...)` is useful sometimes. In python, there is no 'auto return', no multi-line lambda and `def` is not an expression, so we must use a named function, and the name is usually of no importance beside being used to be returned. To me, it looks good enough to have an `fn` as the last form.

For Example:
```clojure
(defn my-decorator [f]
  (with-decorator (functools.wraps f)
      ...)))
```

I think we should support `fn` (as before) here, if not, we should mention that in the documentation, and using a better error message, `fn` is obviously a function.

### Scenario 2
```clojure
=> (with-decorator identity (defclass C))
=> (defclass C)
=> (with-decorator identity C)

  (with-decorator identity C)
  ^-------------------------^
HyTypeError: Decorated a non-function
```

Here, feeding a type that's already been defined (not on the spot) raises an exception, which is reasonable BTW, but we should have more descriptive error message, `with-decorator` supports classes, too, and the wording implies only functions can be used here.

### LINKED ISSUE #2271: Decorating anonymous functions is inconsistent
Attempting to decorate an `fn` that can compile down to a `lambda` fails, but it won't fail if the `fn` contains statements that force compilation to an anonymous `def`

```clojure
=> (with-decorator cache (fn [x] x))
  File "stdin-ecb69bbf92e18e5e87b4bc8c6fd3222c5969d4a8", line 1, in <module>
    (with-decorator cache (fn [x] x))
  File "<stdin>", line 1
    (with-decorator cache (fn [x] x))
                          ^
hy.errors.HySyntaxError: Decorated a non-function
```

```clojure
=> (with-decorator cache (fn [x] (print 1) x))
@cache
def _hy_anon_var_2(x):
    print(1)
    return x
_hy_anon_var_2
```

I think it makes the most sense to force the anon `def` when an `fn` uses a decorator like it already does when it contains statements or type annotations.

Kodiologist:
- Closes #952
- Closes #1624
- Closes #2271

I haven't updated the documentation yet.


Kodiologist:
A possible extension would to make the brackets optional for the common case of only one decorator. However, `(defn foo bar [] …)` arguably looks like you're defining a function named `foo`, anyway.

allison-casey:
Neat!! I'll take a look after work tonight. I do think we should keep the explicit brackets for single decorators. the ambiguity of which is the fn name doesn't seem worth the bracket saving two characters on the brackets. 

What is the evaluation order of multiple decorators using the new syntax? Python is right to left/bottom up. Is this the same way?

Kodiologist:
It should be the same as Python's. I'll tweak a test to make it sensitive to the order in which decorators are called.

Kodiologist:
And yes, the decorator expressions themselves are evaluated outer-first whereas the functions obtained by this evaluation are called inner-first, in both Python and this Hy implementation. Thus both of these  programs print 1 through 4 in order:

```
def d(n):
   return lambda f: [print(n), f][1]

@[print(1), d(4)][1]
@[print(2), d(3)][1]
def f():
    pass
```

```
(defn d [n]
  (fn [f] (print n) f))

(defn
  [(do (print 1) (d 4)) (do (print 2) (d 3))]
  g [])
```

allison-casey (approved):
Looks good! just needs documentation updated

Kodiologist:
Done.

PR #2273, opened 2022-05-08
Fix: subscript type annotations

allison-casey:
hy reader wasn't properly parsing expressions after `^`

### DIFF
--- a/hy/lex/hy_reader.py
+++ b/hy/lex/hy_reader.py
@@ -315,13 +315,14 @@ class HyReader(Reader):
 
     @reader_for("^")
     def annotate_or_xor(self, _):
-        suffix = self.read_ident(just_peeking=True)
-        if suffix == "":
+        next_char = self.peekc()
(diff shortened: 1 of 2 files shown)


Kodiologist:
The test results suggest a conflict with the exclusive-or operator.

allison-casey:
yeah just noticed that slipped past since i forgot to remove `__pycache__` when running tests

Kodiologist:
My usual local testing command is

    git clean -fx '*/__pycache__/*' && pytest --ignore=tests/test_bin.py

allison-casey:
there we go. all better

Kodiologist (approved):
This seems to have been a regression since 1.0a4, so no NEWS line is required.

allison-casey:
yeah, we just didn't have a test for subscript types. Can we actually get this merged before the 3 day wait since it's a small fix and a pretty impactful regression

Kodiologist:
Yeah, that's legit.

PR #2281, opened 2022-05-15
#^ annotations

### LINKED ISSUE #2278: Parsing `^` is weird
A wart in the syntax is that `^` marks an annotation unless it's followed by whitespace or `=`. The `^` and `^=` macros are rarely used, so I think it's safe to rename them. Then `^` will only be for annotations. Another option is to use some other syntax for annotation, like `#^`, and let `^` be a regular character.

allison-casey:
closes #2278

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -46,6 +46,8 @@ Breaking Changes
   `hy.models.Tuple`
 * Tuples now wrap to `hy.models.Tuple` during model promotion
 * Fraction literals have been removed
+* Annotations now use `#^` instead of `^`
+* ``annotate`` takes the target first and the type second. i.e. ``(annotate x int)``
 
 Bug Fixes
 ------------------------------
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -8,9 +8,9 @@ Core Macros
 The following macros are auto imported into all Hy modules as their
 base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
 
-.. hy:data:: ^
+.. hy:data:: #^
 
-   The ``^`` symbol is used to denote annotations in three different contexts:
+   The ``#^`` symbol is used to denote annotations in three different contexts:
 
    - Standalone variable annotations.
    - Variable annotations in a setv call.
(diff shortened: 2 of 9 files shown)


allison-casey:
while we're here, is there any way to remove the parens around `#^` in `defclass` when writing dataclasses? it feels like a weird exception and would look cleaner without it

```clojure
(defclass [dataclass] Point []
  #^int x
  #^int y)
```

Kodiologist:
Why are the parens necessary in the first place? Reader macros usually don't need them.

allison-casey:
Seems to be that the annotate form only takes the type but not the value to annotate and then `compile_assign` moves things around. So the parents are there to group the annotate with the form being annotated. It's weird and the more i look at it the less i like how we handle annotate under the hood

allison-casey:
from messing around with it, it seems like normalizing all `annotate` exprs to take the type and the target lets us drop the parens. We just need to fix where `OPTIONAL_ANNOTATION` is used to recognize that. i.e.
`setv` currently works like:
```clojure
(setv #^int x 1)
;; currently reads as
(setv (annotate int) x 1)
;; when it probably should read as
(setv (annotate int x) 1)
```

allison-casey:
So this last commit lets us drop the parens in `defclass`. Instead of having to parse an `(annotate FORM)` and it's relationship to the subsequent form, `annotate` now just takes the type and the target together as `(annotate TYPE FORM)`. this feels much cleaner and more consistent imo. no more contextual understanding needed to parse annotations. I haven't updated documentation or news yet so people can take a look and see how it feels.

Kodiologist:
While we're changing it, should we use the Python order, too, with the variable name first?

scauligi:
For the `(annotate)` form itself, `(annotate sym type)` feels more consistent with the rest of Hy to me.
For the macro, I think we should keep the `#^type sym` order (unless that wasn't in question).

allison-casey:
I'm in favor of keeping the `#^type sym` order for the macro, but i'd be for swapping the order in the `annotate` form.

Kodiologist:
I never use annotations and I'm not invested in this, so do whatever you think is best.

allison-casey:
done

allison-casey:
noted in docs

That was my first though, but not without forcing a space between the `^` unfortunately. Since reader macro dispatch reads the whole sym and I'd rather not break the type off from the `#^` (it gets pretty hard to parse when annotating a bunch of things imo) or add extra character to annotations which are already a character longer than they are in python.

allison-casey:
oh, yeah in that case that is a limitation. i'm honestly okay with that personally (given we make it clear in the docs obvi). annotations are going to get much much more use than reader macros in general and then reader macros that specifically start with a `^`

allison-casey:
in that case, should we add a check to `defreader` to forbid names starting with `^`?

Kodiologist:
Probably, yeah.

allison-casey:
done

Kodiologist:
Why add new documentation for `defreader` in `api.rst`? `defreader` already has a docstring. Is it not getting rendered for some reason?

allison-casey:
Lol woops. It was late and i forgot that was autodoc'd my b

allison-casey:
fixed

Kodiologist (approved):

PR #2282, opened 2022-05-15
remove fraction literals

### LINKED ISSUE #2277: Remove fraction literals
They don't have their own Python AST node, so there's no real need to give them special support in the core language. Hyrule could provide some kind of syntactic sugar for them, but the need for fractions is pretty rare anyway.

allison-casey:
closes #2277


Kodiologist:
Instead of removing its `hy.repr` entirely, how about changing it to the style `(Fraction 1 2)`? I've tried to cover a lot of types in Python's standard library.

allison-casey:
fair enough, done

Kodiologist (approved):

PR #2283, opened 2022-05-15
narrow defmacro lambda list parser

### LINKED ISSUE #1730: Macros do not like unpack-mapping
```hylang
(defmacro id [x] x)
(print #** {})  ; no problem
(print (id #** {}))  ; error

;   File "<input>", line 1, column 12
; 
;   (print (id #** {}))
;              ^----^
; HyTypeError: The special form 'unpack-mapping' is not allowed here
```

I tried to test this on devel, but devel wouldn't compile :neutral_face:. Perhaps I'm doing something wrong; not sure.

allison-casey:
closes #1730 

`defmacro` no longer allows the position only delimiter `/` (values are always passed positionally), the keyword only delimiter `*`, kwargs  `#** kwargs`, or arguments after an `#*` args as part of it's parser.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -60,6 +60,7 @@ Bug Fixes
 * Module names supplied to `hy -m` are now mangled.
 * Literal newlines (of all three styles) are now recognized properly
   in string and bytes literals.
+* ``defmacro`` no longer allows arguments after ``#* args``
 
 New Features
 ------------------------------
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -341,6 +341,20 @@ base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
         => (infix (1 + 1))
         2
 
+   .. note:: because all values are passed to macros unevaluated, ``defmacro``
+             cannot use keyword arguments, or kwargs. All arguments are passed
+             in positionally. Parameters can still be given default values
+             however::
+
+                => (defmacro a-macro [a [b 1]]
+                ...  `[~a ~b])
+                => (a-macro 2)
+                [2 1]
+                => (a-macro 2 3)
+                [2 3]
+                => (a-macro :b 3)
+                [:b 3]
+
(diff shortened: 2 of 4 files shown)


Kodiologist:
I could go either way on `/`, since it's merely redundant, but the rest makes sense.

Kodiologist:
How about testing that a compile-time error is produced in these cases?

allison-casey:
so i'm realizing that doing this at the parser level makes our error messages less specific. 
```clojure
=> (defmacro a-macro [* a b])
  File "stdin-264fe9f33e62fddb584f6b604556d52ff12ea9b8", line 1
    (defmacro a-macro [* a b])
              ^
hy.errors.HySyntaxError: parse error for pattern macro 'defmacro': got unexpected token: hy.models.Symbol('*'), expected: end of input: hy.models.Symbol('a-macro'), expected: end of input
```
vs what it is currently
```clojure
=> (defmacro a-macro [* a b])
  File "stdin-264fe9f33e62fddb584f6b604556d52ff12ea9b8", line 1, in <module>
    (defmacro a-macro [* a b])
  File "<stdin>", line 1
    (defmacro a-macro [* a b])
                       ^
hy.errors.HySyntaxError: macros cannot use '*'
```
maybe we should use the same lambda-list and just add another check like we do with `#**` and `*`?

Kodiologist:
I know, it's true, but my philosophy is to use model patterns, which can implicitly check for a million different problems but produce imprecise error messages, rather than to write explicit checks, which can produce hand-tailored error messages but involve a lot more code.

allison-casey:
done

Kodiologist:
I believe you can write `(with [_ (pytest.raises HySyntaxError)] …)` as just `(with [(pytest.raises HySyntaxError)] …)`.

allison-casey:
done

Kodiologist (approved):

PR #2286, opened 2022-05-22
Fix: bad var name in reader require exception

allison-casey:
small bug in reader exception raising


Kodiologist:
I think it's better not to unmangle, for consistency with other error messages.

Kodiologist (approved):

allison-casey:
it's a small fix, just gonna merge this now

PR #2294, opened 2022-05-26
Remove `Keyword.__slots__`

### LINKED ISSUE #1928: Consider changing the Hy model objects' design
Since `HyKeyword` is inheriting from a non-`__slots__` object, I believe it will have a `__dict__` attribute anyway, in which case we would be missing out on some of the space-saving advantages of using `__slots__`.  To fix that 
 we can make `HyObject` use `__slots__`; however, this will clash with all the `HyObject` + primitive subclasses.  

If the Hy model objects held references to the actual primitives, instead of actually _being_ those primitives, this problem would be easily solved.

Otherwise, we can probably just use the `abc` and/or `collections.abc` interfaces.

_Originally posted by @brandonwillard in https://github.com/hylang/hy/issues/1924#issuecomment-731694498_

Kodiologist:
Closes #1928.

### DIFF
--- a/hy/models.py
+++ b/hy/models.py
@@ -248,7 +248,6 @@ class Keyword(Object):
       mangling is performed.
     """
 
-    __slots__ = ["name"]
     __match_args__ = ("name",)
 
     def __init__(self, value, from_parser=False):
@@ -310,19 +309,6 @@ class Keyword(Object):
             if default is Keyword._sentinel:
                 raise
             return default
-
-    # __getstate__ and __setstate__ are required for Pickle protocol
-    # 0, because we have __slots__.
-    def __getstate__(self):
-        return {
-            k: getattr(self, k)
-            for k in self.properties + self.__slots__
-            if hasattr(self, k)
-        }
-
-    def __setstate__(self, state):
-        for k, v in state.items():
-            setattr(self, k, v)
 
 
 def strip_digit_separators(number):


brandonwillard:
@Kodiologist, are you just merging your own PRs now?

Kodiologist:
I don't understand the question.

brandonwillard:
Who else reviewed and approved this?

Kodiologist:
Nobody. Have you forgotten [the merge policy](https://docs.hylang.org/en/master/hacking.html#the-pr-itself)?

brandonwillard:
Did you even ask for a review from anyone else?  How about the people involved in the issue it was supposed to close?

Kodiologist:
No, in fact, I've never asked for a review except when it's absolutely required for a PR to be merged.

brandonwillard:
Apparently, reviews aren't required at all, especially when one doesn't even feel the need to request them.

Kodiologist:
Right, unless there's a veto or it's for a new release.

Kodiologist:
Listen, I'm sorry if you felt snubbed by the fact that I didn't ask you for a review. Please be assured it's nothing to do with your personally and you're still allowed to review whatever PRs you want. An important reason not to expect reviews is that, very often, people aren't free to do them. The reason I made this policy in the first place (#1508) is that Hy never had have enough active reviewers to make the old two-reviews-for-everything policy practical. Then I lowered the interval from 2 weeks to 3 days on Allie's urging (#1964) because we were both tired of waiting around 2 weeks for PRs that were obviously never going to get more comments on them.

PR #2305, opened 2022-06-09
Add missing word in tutorial

sunarch:
Hi,
I know that the contrubution guidelines specify to base PRs on `master`, but I thought that this change in the `stable` docs could still make sense until the the new stable version is merged onto it.


Kodiologist:
I'd rather not have to decide which changes are important enough or small enough to backport, and just leave 0.20.0 as it is. We're already pretty thin on manpower for the development of master.

sunarch:
You're right, thank you for the reply.

PR #2307, opened 2022-06-10
Rename `hy.lex` to `hy.reader`

Kodiologist:


scauligi (approved):
I'm fine with the format check failure since this is just a renaming change.

Kodiologist:
Yes, that's probably because a line length was exceeded or something.

Kodiologist:
No, it was because changing the module name led to sequences of import statements no longer being in alphabetical order. Legend has it that there's a reason to care whether your imports are in alphabetical order, but scholars have continued to search without success.

PR #2308, opened 2022-06-13
Document more gotchas

### LINKED ISSUE #1324: .pyc files don't get recreated when they use a macro from a file that was changed

### LINKED ISSUE #1342: Arguments aren't always evaluated left to right

Kodiologist:
- Closes #1324
- Closes #1342

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -24,6 +24,7 @@ the command ``hy``, or run a Hy program with ``hy myprogram.hy``.
    whyhy
    tutorial
    syntax
+   semantics
    macros
    repl
    env_var
--- a/docs/semantics.rst
+++ b/docs/semantics.rst
@@ -1,0 +1,71 @@
+==============
+Semantics
+==============
+
+This chapter describes features of Hy semantics that differ from Python's and
+aren't better categorized elsewhere, such as in the chapter :doc:`macros`.
+
+Implicit names
+--------------
+
+Every compilation unit (basically, module) implicitly begins with ``(import
+hy)``. You can see it in the output of ``hy2py``. The purpose of this is to
+ensure Hy can retrieve any names it needs to compile your code. For example,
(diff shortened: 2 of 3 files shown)


Kodiologist:
I wouldn't recommend that. It doesn't save the user from having to understand the underlying phenomenon, so long as they're still using bytecode at all.

scauligi:
> so long as they're still using bytecode at all.

You mean if they're running `.pyc` files directly instead of just implicitly via the pycache? Because in the typical pycache case I have some code in the importer that can check the (tree of) macro dependencies and recompile (that is, consider the cached `.pyc` invalid) if any of the dependencies are themselves stale.

To me this feels seems to give a more intuitive experience, e.g. if you edit a file that exports macros then try to re-run a dependee script.

Kodiologist:
> > so long as they're still using bytecode at all.
>
> You mean if they're running `.pyc` files directly instead of just implicitly via the pycache?

Just implicitly.

I get that dependency-tracking can work in many common cases, but I don't think it's fully general; thus, "it doesn't save the user from having to understand the underlying phenomenon, so long as they're still using bytecode at all".

Kodiologist:
@scauligi Any objections to merging this?

scauligi:
Ah nope, no objections

PR #2309, opened 2022-06-17
Add "wheel" to `setup_requires`

scauligi:
To let https://github.com/hylang/hyrule/pull/46 work.


Kodiologist:
This looks good, but I would put the first diff in its own commit, since it appears to be unrelated, and I would explain in the commit message why we're adding `wheel`.

PR #2314, opened 2022-07-02
Documentation improvements

### LINKED ISSUE #2313: Error in `lfor` docs, using `if` instead of `when`.
The lfor docs contain the following code sample using `if` instead of `when`:

> `:if CONDITION`, which is equivalent to `:do (if (not CONDITION) (continue))`.

image

Kodiologist:
Fixes #2313.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -111,7 +111,7 @@ base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
        ...              (dict :name "Dave" :age 5)])
 
        => (defn display-people [people filter]
-       ...  (for [person people] (if (filter person) (print (:name person)))))
+       ...  (for [person people] (when (filter person) (print (:name person)))))
 
        => (display-people people (fn [person] (< (:age person) 25)))
        Alice
@@ -714,7 +714,7 @@ base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
      iteration clause before the ``:do``.
    - ``:setv LVALUE RVALUE``, which is equivalent to ``:do (setv LVALUE
      RVALUE)``.
-   - ``:if CONDITION``, which is equivalent to ``:do (if (not CONDITION)
+   - ``:if CONDITION``, which is equivalent to ``:do (when (not CONDITION)
      (continue))``.
 
    For ``lfor``, ``sfor``, ``gfor``, and ``dfor``,  variables defined by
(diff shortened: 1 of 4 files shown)


docs/api.rst
```diff
@@ -96,64 +96,34 @@ base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
...
+   to an assignment of a :ref:`lambda expression <py:lambda>`). It requires two
+   arguments: a name (given as a symbol; see :hy:func:`fn` for anonymous
+   functions) and a "lambda list", or list of parameters (also given as
+   symbols). Any further arguments constitute the body of the function::
 
        (defn name [params] bodyform1 bodyform2...)
 
-   If there at least two body forms, and the first of them is a string literal,
-   this string becomes the :term:`py:docstring` of the function.
+   An empty body is implicitly ``(return None)``. If there at least two body
```
scauligi, line 122: Nitpicking here, not sure if it actually matters, but an empty body actually compiles to `pass`.

docs/api.rst
```diff
@@ -96,64 +96,34 @@ base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
...
+   forms, and the first of them is a string literal, this string becomes the
+   :term:`py:docstring` of the function. The final body form is implicitly
+   returned; thus, ``(defn f [] 5)`` is equivalent to ``(defn f [] (return
+   5))``. By contrast, the ``defn`` form itself always returns ``None``.
```
scauligi, line 126: What does "the `defn` form itself" mean? I'm not quite sure what you're trying to get at here.

docs/api.rst
Kodiologist: Right, which is equivalent to Hy's `(return None)`.

docs/api.rst
Kodiologist: `(defn f [] 3)` returns `None`, whereas `(f)` returns `3`.

scauligi: Hmm... I feel like explaining that distinction only serves to confuse more than elucidate, personally.

Kodiologist: It seems that the return value of `defn` has to be documented somewhere, particularly since it's a chief difference from `fn`. If you don't think this is the right way to document it, what do you think is the right way?

scauligi: We don't have any documentation for `setv` saying it evaluates to `None`, only that `setx` is the "expression" form of `setv`; I think we can document `defn`/`fn` similarly.

If you feel strongly that we should mention it though, I think it'll be less confusing up near the top, a la something like the following:
```
   ``defn`` compiles to a :ref:`function definition <py:function>` (or possibly
   to an assignment of a :ref:`lambda expression <py:lambda>`). Unlike ``fn`` however,
   the ``defn`` form itself evaluates to ``None``. It requires two arguments...
```

Kodiologist: > We don't have any documentation for `setv` saying it evaluates to `None`

We absolutely should. `setv` is among the many core macros whose documentation needs a rewrite.

> If you feel strongly that we should mention it though, I think it'll be less confusing up near the top

Okay, how's this?

scauligi (approved):

PR #2333, opened 2022-08-26
Support Python 3.11 and PyPy 3.9

Kodiologist:


hy/macros.py
```diff
@@ -439,7 +440,21 @@ def macroexpand_1(tree, module, compiler=None):
...
+            co_name=new_name,
+            **(
+                {
+                    "co_qualname": re.sub(
+                        r"\.[^.+]\Z", "." + new_name, f.__code__.co_qualname
+                    )
+                    if "." in f.__code__.co_qualname
+                    else new_name
+                }
+                if PY3_11
```
tianon, line 454: With PyPy's 3.11 (https://pypy.org/posts/2025/02/pypy-v7318-release.html), it looks like this might need to be updated to exclude PyPy :eyes:

I don't think it's implemented `co_qualname` (but I'll very actively admit I'm way out past my skis in pretending to understand the backtrace I'm seeing :sweat_smile:).

Kodiologist: Thanks for the heads up. I guess I'll wait till GitHub Actions is using the new PyPy to change anything.

tianon: You responded faster than I could even get to it, but I upgraded my comment to an issue so it doesn't get lost; hopefully that's OK! :smile: :heart:

https://github.com/hylang/hy/issues/2631

PR #2354, opened 2022-11-06
Implement import from ZIP archives

### LINKED ISSUE #659: Import from zip files
It would be handy to be able to import .hy files from .zip archives (for those of us who like to deploy stuff in single file bundles).

Right now I'm working around this by doing:

```
BUNDLE=sushy.zip

clean:
    rm *.zip

bundle:
    hyc sushy/*.hy
    zip -r9 $(BUNDLE) . -i *.py *.pyc

test-bundle:
    PYTHONPATH=$(BUNDLE) python -m sushy
```

Kodiologist:
- Closes #659.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,5 +1,15 @@
 .. default-role:: code
 
+Unreleased
+=============================
+
+New Features
+------------------------------
+* On Pythons ≥ 3.7, Hy modules can now be imported from ZIP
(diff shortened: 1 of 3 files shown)


Kodiologist:
It looks like Python 3.7 implemented `zipimport` in C. That would make monkey-patching it pretty difficult. 3.7 has less than a year until end-of-life, anyway, so I think I'll just say that this is only supported on 3.8 and up.

PR #2370, opened 2022-12-03
Add reader macro to escape reserved macros

Techcable:
Some names like `export`, `quasiquote`, `quote`, etc are reserved in Hy, but useable in Python.

This allows accessing these names via a `#'export` syntax.

Not exactly sure what an approriate syntax is. I just used `#'foo` as a temporary placeholder.

### DIFF
--- a/hy/reader/hy_reader.py
+++ b/hy/reader/hy_reader.py
@@ -366,6 +366,15 @@ class HyReader(Reader):
             self.parse_one_form(),
         )
 
+    @reader_for("#'")
+    def escaped_identifier(self, _):
+        """Parses an escaped identifier"""
+        form = self.parse_one_form()
+        if isinstance(form, Symbol):
+            return mkexpr("py", str(form))
+        else:
+            raise LexException.from_reader("expected a symbol", self)
+
     ###
     # Strings
     # (these are more complicated because f-strings


Kodiologist:
I think you may be confused about some of the details of macro-expansion. As an example, why don't you show me the specific problem you encountered that motivated this PR?

Techcable:
I want to call `export(3, 7)` in Hy, but `export` is a reserved name

Kodiologist:
Are you aware that the following works?

```
=> (defn export [x y] (+ x y))
=> ((do export) 3 7)
10
```

Macros are only expanded at the head of an expression.

Techcable:
> Are you aware that the following works?
> 
> ```
> => (defn export [x y] (+ x y))
> => ((do export) 3 7)
> 10
> ```
> 
> Macros are only expanded at the head of an expression.

No, I was not aware. I would say this is preferable to a special syntax.

Should this specific use case be added to the documentation or some sort of FAQ?

Kodiologist:
> Should this specific use case be added to the documentation or some sort of FAQ?

Probably there's a helpful place one could mention this in the manual, yes. Perhaps `macros.rst`. There isn't much there so far, but I've always thought of expanding it.

PR #2373, opened 2022-12-13
Update CherryPy example to hy0.25.

fgui:
The example was not a valid.
#@ is not supported in 0.25 use [] as first parameter in defn.


Kodiologist:
Good catch, thanks. You have to open the PR against `master`, though, not `stable`.

fgui:
Ok, sorry. I opened the PR against `master`.

Kodiologist:
Closed in place of #2374.

PR #2375, opened 2022-12-14
More documentation improvements

Kodiologist:


Kodiologist:
I'm not sure how I mistakenly pushed this to hylang instead of my own fork, but I did.

PR #2389, opened 2023-01-18
Add folder output for hy2py

### LINKED ISSUE #2079: Recursive `hy2py`
We could give `hy2py` a recursive mode that goes through a source tree and saves a matching Python file for each Hy file, if an up-to-date one doesn't already exist. Optionally, it could delete all such Python files and bytecode first, in order to ensure all macro expansions are fully up to date.

I'm not sure how often this would be useful in practice, but it seems like an obvious feature to have.

danieltanfh95:
- Closes #2079

hy2py currently only support single files.

This PR adds two functionality:
1. support for `output` parameter to specify where output should go for file or folder.
2. support for recursive transformation of hy files in folders. Outputs to original folder if output is not specified, follows original folder structure.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -109,3 +109,4 @@
 * Dmitry Ivanov <dmitry.ivanov@divanov.eu>
 * Andrey Vlasovskikh <andrey.vlasovskikh@gmail.com>
 * Joseph LaFreniere <joseph@lafreniere.xyz>
+* Daniel Tan <danieltanfh95@gmail.com>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -38,6 +38,7 @@ New Features
   remaining special rule that `...` compiles to `Ellipsis`)
 * On Pythons ≥ 3.7, Hy modules can now be imported from ZIP
   archives in the same way as Python modules, via `zipimport`_.
+* `hy2py` now supports directory input, and will recursively convert hy source code into python source code.
 
 .. _zipimport: https://docs.python.org/3.11/library/zipimport.html
 
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -43,7 +43,7 @@ for a complete list of options and :py:ref:`Python's documentation
 hy2py
 -----
 
-``hy2py`` is a program to convert Hy source code into Python source code. Use ``hy2py --help`` for usage instructions. It can take its input from standard input or from a filename provided as a command-line argument. The result is written to standard output.
+``hy2py`` is a program to convert Hy source code into Python source code. Use ``hy2py --help`` for usage instructions. It can take its input from standard input, from a filename, or folder name provided as a command-line argument. If it is a folder, the output parameter (--output/-o) must be provided. When the output parameter is provided, the output will be written into the folder or file, otherwise the result is written to standard output.
 
     .. warning::
        ``hy2py`` can execute arbitrary code. Don't give it untrusted input.
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -7,6 +7,7 @@ import platform
 import py_compile
 import runpy
 import sys
+from contextlib import nullcontext
 from pathlib import Path
 
 import hy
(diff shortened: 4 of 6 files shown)


Kodiologist:
Is this intended to close #2079?

danieltanfh95:
> Is this intended to close #2079?

No, I was not aware this issue existed. This was more for my own use when using hy in production, very annoying to deal with multiple file dependencies.

Kodiologist:
Okay. I think we can close it with this approach if we're thoughtful about it.

In terms of the interface, I don't think it makes sense to offer the ability to put the generated Python files in the same directories as the Hy files, and with the same names (other than file extensions). That creates a situation ripe for surprises about which file gets imported when. So, `--output` should be required for a directory input.

I also think that directory inputs should be disallowed unless an `-r/--recursive` option is set. This would be consistent with the general principle that Unix command-line tools don't recurse into directories without such an option. If the input is a plain file, the option will have no effect, which is fine.

We also need a test. I think a concise way to do this is to add a function to `test_bin.py` that uses the Pytest fixture `tmp_path` to create a tree with two or three one-line Hy files, calls `hy2py` on it, and checks the output of executing the result in Python.

Regarding non-code changes:
- The prose in `cli.rst` needs to be reworded a bit now that `hy2py` can take a directory.
- Update `NEWS.rst`.
- Add yourself to `AUTHORS` in a separate commit. (Have you read `CONTRIBUTING.rst`? You should.)

Kodiologist:
Eh, on second thought, scratch the bit about `--recursive`. It's overkill because the files you're recursing over are only being executed, anyway, not modified.

Kodiologist:
It looks like tests are failing on Pythons ≥ 3.9.

Kodiologist:
This is really a test of `hy2py`'s command-line interface and not the substance of what it does. That's why I suggested putting it in `test_bin.py`. If you do that, it will automatically be skipped on Pyodide.

danieltanfh95:
@Kodiologist seems like there's `/lib/python3.10/subprocess.py:1775: OSError` for pyodide (might be due to subprocess not being implemented), is there a flag i can use to remove it from pyodide tests temporarily?

danieltanfh95:
> 

got it, lemme do that

Kodiologist:
Thanks for doing this. Here are some things I'd like you to address before we merge this:

- Let's not mention `hy2py3` in NEWS. It's a mistake that `hy3` etc. still exist, years after Hy dropped support for Python 2, so I'll remove them instead in a separate PR.
- It looks like there's duplicated logic between the recursive and non-recursive cases of `hy2py` that you haven't put in `hy2py_worker`, like reading the source file and opening the output file
- Use the word "directory" rather than "folder" in command-line help text.
- I think the comment `# output directly into the same folder` is no longer applicable.
- "test failure if output parameter is not provided when directory input is given" — You're not actually testing for the expected exception, just suppressing the exception if it arises. You want `pytest.raises`, not `contextlib.suppress`. However, it's better to use `run_cmd` (defined at the top of the file) and check for results indicating an error, as shown in the other tests in this file, rather than using `subprocess.check_call` and then looking for an exception.
- Instead of testing the number of files in the output directory, test the set of filepaths for equality against what it ought to be.
- Instead of testing the contents of the generated Python files, run one (with Python via `run_cmd`, or by importing as in `test_hy2py.py`) and check the result. The exact Python code that `hy2py` produces is just an implementation detail; what we want to test is the effect of the code.

danieltanfh95:
> It looks like there's duplicated logic between the recursive and non-recursive cases of hy2py that you haven't put in hy2py_worker, like reading the source file and opening the output file

I chose not to do this in hy2py_worker so I can still use `with`, since `output_file` may be `None`

Kodiologist:
You can still use `with`. I've pushed a demonstration. Please continue pulling out duplicated logic from here. For `source`, you don't even need the `with` because you're reading it all in at once; `Path(…).read_text()` suffices.

danieltanfh95:
@Kodiologist need your help for the tests on win-3.11

Kodiologist:
Try making this change:

> it's better to use `run_cmd` (defined at the top of the file) and check for results indicating an error, as shown in the other tests in this file, rather than using `subprocess.check_call` and then looking for an exception.

If it still doesn't work, I'll see what I can do.

danieltanfh95:
> it's better to use `run_cmd` (defined at the top of the file) and check for results indicating an error, as shown in the other tests in this file, rather than using `subprocess.check_call` and then looking for an exception.

I'm using check_call because run_cmd assert is False, due to exit code being 1 and not 0

Kodiologist:
Look at the `assert` line. What is `p.wait()` being compared to?

danieltanfh95:
> Look at the `assert` line. What is `p.wait()` being compared to?

I see, we can pass in an expected error code. Okay, will change the code to fix this

danieltanfh95:
@Kodiologist I dont have a windows machine so I can't debug these https://github.com/hylang/hy/actions/runs/4020969124/jobs/6909418144 , let me know what I can do to fix these

Kodiologist:
I rarely bother to run Hy on Windows, either; I typically just take a good guess and leave it to GitHub Actions to check my work. Anyway, thanks for your continued work. I'll take a stab at it tomorrow.

Kodiologist:
Okay, fixed. Time to clean up the commits and we should be done. You can probably squash everything together (with a cleaned-up commit message) except for your AUTHORS commit. Just be sure to remove unrelated whitespace changes. I also don't think the changes to `tests/test_hy2py.py` to use `Path`s instead of strings are useful. By contrast, I think the removal of unused imports in that file is useful, but they should also be in their own commit, since they're unrelated to the main changes here.

danieltanfh95:
@Kodiologist is this merge-able?

Kodiologist:
I'd still like to see more commit cleanup. You have a bunch of merge commits and these giant commit messages that are the concatenation of the long and uninteresting history of your final changes. So, rebase onto the current `master` and tend to your commit messages.

Kodiologist:
I can do it for you if you're having Git issues.

Kodiologist:
By the way, I'm hoping to do a new release by February 10th, so you gotta get this in by then if you want it to be included.

danieltanfh95:
> I can do it for you if you're having Git issues.

Yes please, would be great if you could help me with this

Kodiologist:
No problem, my man.

Kodiologist:
All done. Thanks again, and welcome to the Hyve.

PR #2400, opened 2023-01-31
Hy support multiple icommand

danieltanfh95:
This would allow the `hy` command to load multiple files into the repl (helpful for preloading a bunch of util functions or a folder of hy files)

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -21,6 +21,7 @@ Breaking Changes
 * New macro `do-mac`.
 * New macro `pragma` (although it doesn't do anything useful yet).
 * `hy.cmdline.HyREPL` is now `hy.REPL`.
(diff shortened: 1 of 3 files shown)


danieltanfh95:
> The code in fy.hy is executed in the same scope as fx.hy entirely implicitly. 

This is true. The initial motivation was to provide something akin to clojure user profiles, preloading different sets of tools/environments per REPL session. Maybe there's a better way to do this? 

My use case is loading a bunch of debugging tools/macros for development each time I reload a REPL, which should not be loaded when in production. I guess the same can be done if we have some sort of packing step like what javascript does.

Kodiologist:
`HYSTARTUP` exists pretty much for that purpose. You should be able to just say `HYSTARTUP=$some_path hy` to get the tools you want for each REPL session.

PR #2425, opened 2023-03-28
Add `hy.M` import sugar

### LINKED ISSUE #2416: `hy.M` sugar for import on demand
It's annoying to have to import to a gensym when you want to use a function from another module in a macro expansion. Suppose you could say `(hy.M.math.sqrt 2)` to call `math.sqrt` without actually binding `math` or `math.sqrt` to some name in the local scope. In addition to making macros neater, this could be convenient as a general shorthand for using something from a module, particularly a module with a long name, without having to add a separate `import` form.

To get a module whose name itself has dots, one could use `/` instead of the dots, so `(hy.M.foo/bar/baz.bing)` would work like `(import foo.bar.baz) (foo.bar.baz.bing)`.

Here's a simple implementation:

```python
import hy

class M:
    def __call__(self, module_name):
        import importlib
        return importlib.import_module(module_name)
    def __getattr__(self, s):
        return self('.'.join(hy.unmangle(s).split('/')))

hy.M = M()
hy.eval(hy.read_many('''
(print hy.M.sklearn/neighbors.KNeighborsRegressor)
'''))
```

The nice thing about this feature is that it can be added without changing Hy's syntax or touching the set of core macros.

`__call__` is provided so you can also say `(hy.M "foo.bar.baz")` to get the module `foo.bar.baz`, which may be more convenient when the module name isn't known until runtime.

This sort of feature could probably be provided for requiring macros, too.

Kodiologist:
- Closes #2416 

I intend no prejudice against a reader macro as proposed by @scauligi in the linked issue, but it would need to expand to something, and I propose this as what it would expand to if we had it.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,10 +39,11 @@ jobs:
       shell: bash
       run: |
         if [[ ${{ matrix.python }} = pyodide ]] ; then
-            npm install pyodide
+            npm install pyodide@0.22.1
+              # 0.23.0 has a regression: https://github.com/pyodide/pyodide/issues/3730
             pip install 'pip >= 22.3.1'
               # Older pips may fail to install `pyodide-build`.
-            pip install pyodide-build
+            pip install 'pyodide-build == 0.22.1'
             pyodide venv .venv-pyodide
             source .venv-pyodide/bin/activate
         fi
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -16,6 +16,7 @@ New Features
   case they're no-ops.
 * The `py` macro now implicitly parenthesizes the input code, so Python's
   indentation restrictions don't apply.
+* New built-in object `hy.M` for easy imports in macros.
 
 0.26.0 (released 2023-02-08)
 =============================
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1312,6 +1312,8 @@ the following methods
 
 .. hy:autofunction:: hy.as-model
 
+.. hy:autoclass:: hy.M
+
 .. _reader-macros:
 
 Reader Macros
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -70,6 +70,10 @@ intersphinx_mapping = dict(
     py3_10=("https://docs.python.org/3.10/", None),
     hyrule=("https://hyrule.readthedocs.io/en/master/", None),
 )
+
+import hy
+hy.M = type(hy.M)  # A trick to enable `hy:autoclass:: hy.M`
+
 # ** Generate Cheatsheet
 import json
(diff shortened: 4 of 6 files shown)


scauligi:
I'm still not convinced what the value of `(hy.M module)` is, since it's exactly equivalent to `hy.M.module`?

scauligi (changes requested):
Submodules with leading underscores get borked during the mangling dance:

```
=> hy.M.slyde/skip/_skip
  File "stdin-6f407d820ddb70c45c408b6dc6756be7a9df566c", line 1, in <module>
    hy.M.slyde/skip/_skip
     ~~~~~~~~~~~~~~^~~~~~
  File "/Users/scauligi/workspace/hy/hy-git/hy/__init__.py", line 27, in __getattr__
    return self(hy.mangle('.'.join(hy.unmangle(s).split('/'))))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/scauligi/workspace/hy/hy-git/hy/__init__.py", line 25, in __call__
    return importlib.import_module(module_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.2_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/importlib/__init__.py", line 126, in import_module
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'slyde.skip.hyx_XhyphenHminusXskip'
```

Kodiologist:
Good catch, thanks. That should be fixed and tested now.

Kodiologist:
Hmm, Pyodide is failing, but I bet that's a Pyodide bug. As much as I like having Hy on Pyodide, running it on pytest really hasn't been productive so far because it's resulted in zero changes to Hy itself to accomodate Pyodide, just changes to the tests.

Kodiologist:
I spent a lot of time figuring out the need for reloading `foo` in the test, hence https://github.com/pytest-dev/pytest/issues/10854.

Kodiologist:
It turns out the Pyodide problem can be reproduced on the current master (8ee37b01a91fca679c430951fe5fc1a459fd2aa2), so it's from a server-side change. Ignore Pyodide failures till I can get it working again.

Kodiologist:
Okay, fixed.

Kodiologist:
@scauligi Do you approve?

scauligi (approved):
Oof, hacky but I'll take it. lgtm!

Issue #2306, opened 2022-06-10
`#_`, `#*`, and `#**` should probably be reader macros


Kodiologist:
Right now they're hard-coded into `HyReader`.

Kodiologist:
In fact, as things stand, the way something like `#_foo` or `#*foo` is parsed depends on whether a reader macro named `_foo` or `#*foo` happens to exist:

```
=> (setv foo [1 2 3])
=> (print #*foo)
1 2 3
=> (defreader *foo 5)
=> (print #*foo)
5
```

scauligi:
Hmm, I know one of the iterations of the reader macro dispatcher would prefer single-character readers over identifiers starting with that character, precisely so you could dispatch things like `#*args` or `#%val`. I like having that as a feature, but it might be a little unintuitive if you're not expecting it?

Kodiologist:
I think that would get confusing once you have something like one reader macro named `a` and another named `all`. Is `#all-boots` parsed as `#a ll-boots` or `#all -boots`?

scauligi:
This seems like the most relevant place to discuss single-character tag dispatch before I get back to #2426 

So an interesting (and likely unintentional) quirk of tag dispatch is that a user can never actually define one of these single-character readers:

```hy
=> (defreader w '2)
=> (setv eirdness 3)
=> (print #w eirdness)
2 3
=> (print #weirdness)
hy.reader.exceptions.LexException: reader macro '#weirdness' is not defined
```

(You can also test this with the `#s` and `#%` macros from hyrule)

---

The reason it's like this is because of mangling quirks:

```hy
=> (eval-when-compile (print (-> hy.&reader.reader-table (.keys) list)))
[
    # ...other keys...
    '#_',
    '#*',
    'hyx_Xnumber_signXXpercent_signX',
    'hyx_Xnumber_signXs',
    'hyx_Xnumber_signXw',
    'hyx_Xnumber_signX_uhoh'
]
```

In `HyReader.tag_dispatch()`, we first check for the *mangled* full symbol, and failing that we check for literal "#" plus one *unmangled* character. So to get your own single-character reader, you would have to manually insert your reader function into the table, since `defreader` always mangles the tag name.

---

I like the aesthetics of `#*args` over `#* args`, but while typing out this comment I realized that (unless we start special-casing things) this would allow expressions like `#_uhoh` to actually produce something if eg `(defreader _uhoh "badbadbad")` has been defined, which seems like a big no-no.

[...]

PR #2426, opened 2023-04-01
Add `#/` import reader macro

scauligi:
Reader macro `#/` for accessing `hy.M` import helper.
Examples:
```hy
(#/re.match r"re(gex)?" line)
(= (type #/collections/abc #/types.ModuleType))
(print #/os/path.curdir)
```

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -16,6 +16,7 @@ New Features
   case they're no-ops.
 * The `py` macro now implicitly parenthesizes the input code, so Python's
   indentation restrictions don't apply.
+* New built-in object `hy.M` for easy imports in macros.
 
 0.26.0 (released 2023-02-08)
 =============================
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1312,6 +1312,8 @@ the following methods
 
 .. hy:autofunction:: hy.as-model
 
+.. hy:autoclass:: hy.M
+
 .. _reader-macros:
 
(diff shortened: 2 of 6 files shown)


scauligi:
Ah should have submitted this as a draft for now, it depends on https://github.com/hylang/hy/pull/2425

scauligi (commented):
> Pull request authors can't request changes on their own pull request


Well that's just silly.

---

Anyways @scauligi, speaking as a maintainer, you need to wait until #2425 lands.
Once it does, rebase your commits and add a line to `NEWS.rst`. After that we'll take a look at your PR.

Kodiologist:
Your comment made me giggle.

I like this syntax, but do you think this should be in Hyrule rather than Hy? Since it's a reader macro, it's easy to use in a quasiquoted construct without having to import or require or whatever in the expansion, so being in Hyrule shouldn't keep it from being convenient.

Kodiologist:
And that would allow users to use the rather desirable reader-macro name `/` for something else if they desire.

Kodiologist:
Also, before I forget: it's probably better to write examples with a space after the reader macro name, e.g. `#/ collections/abc`. `/` is an identifier character, so in order to get more consistent behavior in parsing reader-macro calls, I'd like to eventually get `#/collections/abc` to unconditionally attempt to call a reader macro named `/collections/abc`. See our conversation in #2306.

PR #2436, opened 2023-05-05
In the REPL, print a delimiter after `spy` output

### LINKED ISSUE #1435: Delimit python "--spy" output.

Kodiologist:
- Closes #1435

I arbitrarily chose `'-' * 30` as the (default) delimiter. I don't have a strong preference, so if you'd like to bikeshed about it, now's the time.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -23,6 +23,8 @@ New Features
 * The `py` macro now implicitly parenthesizes the input code, so Python's
   indentation restrictions don't apply.
 * New built-in object `hy.M` for easy imports in macros.
+* `hy --spy` now prints a delimiter between the Python equivalent of
+  your code and the result of evaluating the code for easier reading.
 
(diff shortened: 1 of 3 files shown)


Kodiologist:
I opened #2437 to deal with the Windows test failures.

PR #2442, opened 2023-05-22
Fix tests for MacOS (and also for running locally)

scauligi:
im on a mac now and the tests wernt working so i fixed them

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,10 +12,13 @@ jobs:
         os: [ubuntu-latest]
         python: [3.8, 3.9, '3.10', 3.11, 3.12-dev, pypy-3.9, pyodide]
         include:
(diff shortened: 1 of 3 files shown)


Kodiologist:
Thanks.

If I understand correctly, the `PYTHONPATH` assignment is still not related to Mac OS, but allows `test_circular_macro_require` to pass on platforms other than GitHub Actions. Could you put that change in its own commit? Also, for that comment URL, I'd use https://docs.python.org/3.11/library/readline.html , since the documentation can change between 3.x releases of Python.

Otherwise, looks good.

Kodiologist (approved):

scauligi:
No new features, just bug fixes for tests, so I'm going ahead and merging this under the "trivial changes" clause.

PR #2444, opened 2023-05-25
Reader macros always take a full identifier

scauligi:
Per discussion in #2306, we no longer have tagged reader macros that immediately
dispatch after a single character. E.g., `#*args` now only refers to a reader macro defined using `(defreader *args ...)`,
whereas unpacking a symbol is now consistently written as `#* args`.

Reader macros themselves are now kept in a separate table `&reader.reader-macros` from the "normal" reader dispatchers (in `&reader.reader-table`). Reader macro names are now *never mangled* as they are not actually symbols and never interact with the python side of things. This removes any confusion with `defreader`/`require`/etc. mangling names. As a side-effect, this does mean that e.g. `#r---x` and `#r___x` are distinct macros now.

Additionally, I moved `#^` out into a reader macro function so it's no longer a hardcoded special case.


Kodiologist:
Sounds excellent overall. From your summary, having not looked at any of the code yet, I'm only concerned about the mangling bit. Like, I see the motivation, but this is inconsistent with the treatment of regular macros, where you can say

```
=> (defmacro a-b [] 1)
=> (a_b)
1
```

and I think it's ultimately less confusing to users to know that macros (and reader macros), just like regular Python objects, can be referred to with premangled names—even though there isn't really the same need for mangling. Later this may make for a smoother transition to a more first-class treatment of macros, which I'd like to make happen before 1.0.

What do you think?

scauligi:
My argument for nonmangled reader macros is that conceptually, the reader deals with a raw character stream instead of a sequence of forms. We already make an effort to differentiate symbol macros and reader macros (eg in `require`) and so to me, this cements the fact that reader macros happen at a "lower" level than the rest of Hy.
This also still strikes me as the more consistent option: anything that is or will be a `hy.models.Symbol` gets mangled, but reader names never become `Symbol`s. As an example: there's no way to quote a reader form (eg `'#doesnt-quote`) in Hy, and there's no way to produce a reader form as the result of a symbol macro.

I think the perceived benefit of mangling reader macro names is outweighed by the technical complications of having mangling at the reader level; I remember it was confusing tracking down all the places we needed to mangle or not-remangle names while overhauling the reader, and as seen in #2306 we still didn't actually get it totally correct.

Incidentally, after this PR, the only use of `mangle` in the reader is to expose the reader as `hy.&reader`. I will probably make a PR to move even that to the compiler side of the API, so mangling will take place entirely outside of the reader layer.

scauligi:
The failure in 3.12-dev seems to be related to https://peps.python.org/pep-0695/ .
This looks to be a brand new typing feature that we'll need to deal with eventually, but for now it should be sufficient to just do the bare minimum to make python3.12 happy (which I can do in a separate PR).

Kodiologist:
> anything that is or will be a `hy.models.Symbol` gets mangled

This part isn't true. Consider for example `(print 'a-b)` and `(!= 'a-b 'a_b)`. Mangling happens as late as possible, during the transformation of symbols into Python identifiers, and that only ever happens for some symbols.

But overall I think you make some good points, and you've still worked with the reader a lot more than I have, so let's go ahead with this. I'll review the code soon.

Kodiologist:
In `if tag[0] == '#' and tag[1:]`, is `and tag[1:]` necessary?

Regarding `num_stars = len(stars)`, you might as well inline `len(stars)` in this commit since it's now used only once. Or you could even write  `"unpack-" + {"*": "iterable", "**": "mapping"}[stars]` or something if you prefer.

In `ident and ident in self.reader_macros`, is the `ident and` necessary? Also, it looks `tag` is now pointless is a variable and you can just use `ident`.

I would replace rather than delete `test-reader-macro` in `tests.native-tests.mangling`. You could do something like define one reader macro named `rm--x` and one named `rm__x`, with different effects.

Changes to the manual may be necessary. An item or two for NEWS is definitely necessary.

Kodiologist:
Rebased onto `master`.

scauligi:
> is `and tag[1:]` necessary?

It is, since the reader table also contains `"#": HyReader.tag_dispatch`.

> Or you could even write `"unpack-" + {"*": "iterable", "**": "mapping"}[stars]` or something if you prefer.

I do like that. Done!

> is the `ident and` necessary? 

It's not, as it turns out, since `getc()` should always return a character. I've refactored this (and added a test) so that `#` followed by whitespace now also raises `PrematureEndOfInput`.

> I would replace rather than delete `test-reader-macro` in `tests.native-tests.mangling`.

The equivalent test was actually moved to `test-reader-macros` in `native_tests/reader_macros.hy`!

> Also, it looks `tag` is now pointless as a variable and you can just use `ident`.

> An item or two for NEWS is definitely necessary.

:+1: x 2

> Changes to the manual may be necessary.

All the examples of `#*` etc already include a space, which I guess I should have expected ;)

The explanation of reader macros already talks about the names in terms of reading symbol names; I tried adding a little bit of text explaining how eg `#*args` dispatches `*args`, but there didn't seem to be a great place to put it (and honestly might just be more confusing than elucidating).

Kodiologist:
> > I would replace rather than delete `test-reader-macro` in `tests.native-tests.mangling`.
> 
> The equivalent test was actually moved to `test-reader-macros` in `native_tests/reader_macros.hy`!

I saw the new test; I just think that omitting consideration of reader macros from `tests.native-tests.mangling` looks like forgetfulness rather than singling them out for special treatment. If you're sure you'd rather not have a test of reader macros in the file, I'd say you should add a comment (in place of the old test) saying that non-mangling of reader macros is tested in `reader_macros.hy`.

All else looks good.

Kodiologist (approved):

Kodiologist:
Nice work.

PR #2447, opened 2023-05-27
Replace all uses of features that are deprecated in Python 3.12

Kodiologist:
There are still some `DeprecationWarning`s produced by pytest itself, but these should be fixed by https://github.com/pytest-dev/pytest/pull/10894.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,7 +17,7 @@ jobs:
           - name-prefix: 'win-'
             os: windows-latest
             python: 3.11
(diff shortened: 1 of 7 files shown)


scauligi (approved):
Looks good to me!

You should remove the `pkgutil` imports in `hy/macros.py` and `hy/errors.py` since they're no longer being used (and while you're in `hy/errors.py` you might as well remove the import of `functools.reduce`).

Kodiologist:
Sounds good. Done.

PR #2463, opened 2023-06-29
`try` cleanup

Kodiologist:
The most interesting, user-visible commit is the last one. The others are just various kinds of testing improvements.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -37,6 +37,8 @@ New Features
 * Python 3.12 is now supported.
 * `nonlocal` and `global` can now be called with no arguments, in which
   case they're no-ops.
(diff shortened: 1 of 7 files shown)


scauligi:
I don't have any particular problem with it, but I'm confused; have you run into any examples where you had a macro generating `try` blocks with `else` but no `except`?

Kodiologist:
I don't think so, although it's not too hard to imagine it arising in a situation with a macro that generates `except` clauses, and is supposed to do something special when no exception occurs.

Kodiologist:
My opinion is that the core macros should allow any degenerate usage so long as the semantics that would be desired are obvious.

PR #2466, opened 2023-07-06
Make Kodi officially the maintainer

Kodiologist:
I joined Hy's core team seven years ago. By various metrics (eligible pull requests reviewed, pull requests written, commits made, lines of code changed, issues closed, releases managed, user questions answered, and hours spent on Hy development), I've since become the biggest single contributor to Hy by a substantial margin (particularly in the period after Hy's busy first two years of life). In general, I've taken on a lot of responsibility. So, I'd say I'm the de-facto lead developer. Let's make it official.

I'm not proposing I should get any special privileges, which would be pretty meaningless anyway so long as @scauligi and I are the only active developers. This is really just for my resume. Perhaps if Hy development gets a lot busier (after the release of 1.0, one could imagine), then extra rules about how I could resolve disagreements or something could make sense. But for now, proposing such things seems premature.

@hylang/core Any objections?

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ Project
 * Community: Join us on [Github Discussions](https://github.com/hylang/hy/discussions)!
 * [Stack Overflow: The [hy] tag](https://stackoverflow.com/questions/tagged/hy)
 
+Hy's current maintainer is [Kodi Arfer](https://github.com/Kodiologist). He takes responsibility for answering user questions, which should primarily be asked on Stack Overflow or GitHub Discussions, but feel free to [poke him](http://arfer.net/elsewhere) if he's missed a question or you've found a serious security issue.
+
 ![Cuddles the Hacker](https://i.imgur.com/QbPMXTN.png)
 
 (fan art from the one and only [doctormo](http://doctormo.deviantart.com/art/Cuddles-the-Hacker-372184766))


paultag:
We've never had such a construct in the history of the project, and even when I was active, I'd only taken on the BDFL role as a "tiebreaker" as an escape to figure out governance. Hy has been unique in that the barrier to being accepted as part of the core team was (intentionally!) very low, and project governance relied on people to work together. Even as "BDFL" I was always listed, in order, with the rest of the team.

In the history of Hy, I've only ever really had to come in once (that I can remember, but absolutely less than 3 times even accounting for fudge) when there was an unsolvable issue. I'm happy to continue to do so. Is this "lead developer" position solving for something within the project? FWIW, I will happily endorse you putting whatever you want on your resume, even saying "I am the lead developer and maintainer", I just don't see why we need to have this be called out in documentation, creating a hierarchy where none previously existed.

Also since I haven't said it -- thank you for all your hard work maintaining Hy, it's very appreciated.

Kodiologist:
> Is this "lead developer" position solving for something within the project? 

No. I just want what I do to be recognized appropriately.

> FWIW, I will happily endorse you putting whatever you want on your resume, even saying "I am the lead developer and maintainer", I just don't see why we need to have this be called out in documentation, creating a hierarchy where none previously existed.

I think it would be disingenuous to say that I'm the lead developer when it's not official.

> creating a hierarchy where none previously existed.

Calling yourself a dictator sure sounds like a hierarchy to me. I know you've been referred to as the "BFDL", but honestly, I thought it was more of a joke than anything else, and that's why no reference to it is made in policy.

paultag:
I hear the frustration you're telegraphing, and I'm sorry to hear it.

By royal decree: The Lead Developer is the one doing the most work. We don't need it in docs. As of right now and as seen in the commit log, that's you, @Kodiologist - feel free to put it in your Resume :)

paultag (changes requested):
I don't believe this needs to be put in the docs.

Kodiologist:
Thanks, then. But first, I do want to give Sunjay at least a chance to object if he wants to object. And I mean… I get what you mean about wanting the contributor list to be nice and neutral, but where would this go if not the manual? I could imagine saying it in the README or http://hylang.org, but that seems if anything more presumptuous (because more visible, and more distant from the names of other authors) than putting it in this obscure corner of the manual, so, you tell me.

Kodiologist:
AUTHORS, maybe? That's more complete, and also not transcluded in the manual.

paultag:
@Kodiologist I understand -- but I also don't think it should be a vanity title in of itself. You can do that without any project action :) 

In general, I think with some class of documentation and/or role should come some class of responsibility. I think as-is, I don't see the change in project structure or responsibility from status quo, and would rather you use whatever title is best descriptive of your role in the project. FWIW if you asked me to write your title on your resume, "Lead Developer" would be my pick -- I just don't see the point in calling it out in the list. You do a fuckload of work. You've earned that title, I just don't think it belongs in the doc.

I think you do bring up an interesting point, though: if this was a "point of contact" type thing, where newcomers would know to hit you up (email, github) if they had a question and didn't know where to go, I think README is a great place for it. Likely mirrored into the pypi landing page too, perhaps even the docs; but I'd like for it to be a clearly defined role (my 2c: "I'm a catchall when there's no clear owner, hit me up with any questions"), and make sure you're ready and willing to field those classses of request. I'd love for such a thing to exist and I think you'd do a great job with it. I would maybe call that "Maintainer"? "Point of Contact"?

Thoughts?

Kodiologist:
To an extent our GitHub Discussions page covers that sort of thing, where people can ask questions that aren't issues nor are appropriate for Stack Overflow. It's probably best that Hy questions go there rather than my email. But I do at least make an attempt to answer every user question, yeah, presuming some other core person hasn't beaten me to the punch. "Maintainer" makes sense and I've seen a bunch of other projects use it. I guess I would be the logical person to email in a (so far, completely hypothetical) case that Hy has a grevious security bug that should be kept secret until it's patched. And your point that I should formally take on responsibility if I want a formal title is well-taken.

paultag:
I totally agree about the security thing, and maybe a better specific illustration of why such a thing is a good idea.

OK, my hot takes have been heard, I trust the project to work through this on its own and move forward as it needs to. :heart: - and seriously, thank you for all you do @Kodiologist.

Kodiologist:
♩♫ What can I say except you're welcome?

paultag (approved):

scauligi:
I don't really have much to add to the conversation at this point, so no objections here.

PR #2467, opened 2023-07-10
hy2py now uses `-m` for modules

scauligi:
Fixes a bug where `hy2py` was accidentally importing (and thus executing) the module/script being parsed.

As part of the bugfix, I also changed `hy2py` to explicitly use the `-m` flag for parsing modules, bringing it more in line with `hy`/`python` commandline flags.
It also now follows the same `sys.path` rules too, where using `-m` adds the current directory to `sys.path`, but passing a file directly adds the parent directory of the file instead.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,5 +1,14 @@
 .. default-role:: code
 
+Unreleased
+=============================
+
+Breaking Changes
+------------------------------
+* `hy2py` now requires `-m` to specify modules, and uses
+  the same `sys.path` rules as Python when parsing a module
+  vs a standalone script.
+
 0.27.0 (released 2023-07-06)
 =============================
 
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -8,6 +8,7 @@ import py_compile
 import re
 import runpy
 import sys
+import types
 from contextlib import nullcontext
 from pathlib import Path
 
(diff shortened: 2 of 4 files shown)


Kodiologist:
This kind of thing probably should've gone in the previous release. Whoops.

`hy2py` will still need to import and execute sometimes; for example, if you say `hy2py foo.hy`, and `foo.hy` requires a macro from `bar`, then `bar` must be imported (and hence executed). But I see your point that `foo.hy` itself shouldn't need to be executed in typical cases.

Perhaps `test_hy2py_compile_only` should create a new source file and test that, like some of the other tests in `test_bin.py`, instead of using something in `tests.resources`. Otherwise there's a chance that bytecode already exists for `tests.resources.hello_world` and the test passes with bytecode but not without it. Maybe I'm being too paranoid. I think `assert not re.search(r"^hello world$", output, re.M)` can probably be simplified to `assert "hello world" not in output`, though.

scauligi:
Yeah I'm not sure how I didn't run into this earlier, but I ran into it while trying to use `hy2py` on a half-finished script that didn't have all its classes defined yet.

> Perhaps `test_hy2py_compile_only` should create a new source file

That's probably a good idea to do anyway, that way I can also add a test for a file that would produce a runtime (but not compile-time) error.

> [...] can probably be simplified [...]

In this case it can't, since the output from `hy2py` will include the line `print("hello world")` 🙃 
But I can create a file that instead does something like `(print (+ "hello" " world"))` to make the check simpler.

Kodiologist:
> In this case it can't, since the output from `hy2py` will include the line `print("hello world")` 🙃 
> But I can create a file that instead does something like `(print (+ "hello" " world"))` to make the check simpler.

Oh, right. Very good.

In `filtered_hy_exceptions`, it looks like `module` can be inlined; just use `sys.modules[module_name]` for the second usage.

Do you agree that the section on `hy2py` in `docs/cli.rst` is still correct? Given that we don't want to be redundant with `hy2py --help`.

That should be all, I think.

scauligi:
> `module` can be inlined

Done!

> `docs/cli.rst`

Whooops I'll be honest, I completely forgot about the docs. Updating now.

Kodiologist (approved):

PR #2469, opened 2023-07-19
Treat `hy -i` as a flag, and allow running scripts provided through standard input

### LINKED ISSUE #1395: hy -i should support the debugger

### LINKED ISSUE #2413: Hy's `-i` is confusingly different from Python's
For example, it requires an argument. Contrast

```
$ python3 -i -c 'import sys; print(sys.argv)' myarg  
['-c', 'myarg']
>>>
```

with

```
$ hy -i -c '(import sys) (print sys.argv)' myarg 
  File "string-ddbfe46d29072725b61a3ee03c6abfefa0973acd", line 1, in <module>
NameError: name 'hyx_XhyphenHminusXc' is not defined
Hy 0.26.0 using CPython(main) 3.10.7 on Linux
=>
now exiting REPL...
```

It would be nice to allow `hy -i foo.py`, which might fall out naturally from a more Pythonic take on `-i`.

### LINKED ISSUE #2438: Execute scripts provided through standard input
In Python, we get a very clean output when we pass commands via stdin/shell redirection:

```sh
$ python <<< 'print(123)'
123
$
```

In Hy, it is a bit more cluttered:

```sh
$ hy <<< '(print 123)'
Hy 0.26.0 using CPython(main) 3.10.11 on Linux
=> 123
=>
now exiting REPL...
$
```

I would argue, that we should clean up the output to be equal to the one of python.

Maybe this can help with that:

https://docs.python.org/3/library/os.html#os.isatty

Kodiologist:
- Closes #1395
- Closes #2413
- Closes #2438

I've checked by hand that #1395 works now, but I haven't added a test for it or mentioned it in NEWS because I don't want to suggest that we really support `pdb`. That seems like a deep rabbit hole to go down.

### DIFF
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -49,28 +49,7 @@ def run_command(source, filename=None):
     return 0
 
 
-def run_icommand(source, **kwargs):
-    if Path(source).exists():
-        filename = source
-        set_path(source)
-        with open(source, "r", encoding="utf-8") as f:
-            source = f.read()
-    else:
-        filename = "<string>"
-
-    hr = REPL(**kwargs)
-    with filtered_hy_exceptions():
-        res = hr.runsource(source, filename=filename)
-
-    # If the command was prematurely ended, show an error (just like Python
-    # does).
-    if res:
-        hy_exc_handler(sys.last_type, sys.last_value, sys.last_traceback)
-
-    return hr.run()
-
-
-USAGE = "hy [-h | -v | -i CMD | -c CMD | -m MODULE | FILE | -] [ARG]..."
+USAGE = "hy [-h | -v | -i | -c CMD | -m MODULE | FILE | -] [ARG]..."
 VERSION = "hy " + hy.__version__
 EPILOG = """
 FILE
@@ -114,9 +93,8 @@ def cmdline_handler(scriptname, argv):
         ),
         dict(
             name=["-i"],
-            dest="icommand",
-            terminate=True,
-            help="program passed in as string, then stay in REPL",
+            action="store_true",
+            help="launch REPL after running script; forces a prompt even if stdin does not appear to be a terminal",
         ),
         dict(
             name=["-m"],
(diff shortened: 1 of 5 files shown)


hy/cmdline.py
```diff
@@ -293,8 +288,27 @@ def proc_opt(opt, arg=None, item=None, i=None):
             except HyLanguageError:
                 hy_exc_handler(*sys.exc_info())
                 sys.exit(1)
+    elif action == "just_repl":
+        pass
+    else:
+        assert False
```
scauligi, line 294:
```python
else:
    assert action == "just_repl"
```

hy/cmdline.py
```diff
@@ -293,8 +288,27 @@ def proc_opt(opt, arg=None, item=None, i=None):
...
 
-    return REPL(spy=options.get("spy"), output_fn=options.get("repl_output_fn")).run()
+    assert repl
```
scauligi, line 296: redundant? In any case, raising an AssertionError here seems no better to me than raising an AttributeError just a few more lines down, since we're not trying to catch either of them higher up anywhere.

On the other hand, if it's for semantically documenting that we're definitely setting up the REPL with this combination of options, then I would replace this with an explicit comment stating as such.

scauligi (approved):
> This decouples the tests in question from the behavior of the command-line interface, and it makes them run faster because they're all in-process now.

Oh heck yeah, `test_bin` was always annoying to me with how it held up the rest of the tests


I have some notes, but nothing major/blocking, so marking as approve 🚀

hy/cmdline.py
```diff
@@ -242,41 +220,58 @@ def proc_opt(opt, arg=None, item=None, i=None):
         print(VERSION)
         return 0
 
-    if "command" in options:
+    action, action_arg = (
```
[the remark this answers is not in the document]
Kodiologist: How's this?

hy/cmdline.py
Kodiologist: Done.

hy/cmdline.py
Kodiologist: Sure, I replaced it with a comment.

PR #2472, opened 2023-07-25
Fix `hy2py` on standard input

Kodiologist:
This was a regression from #2467. It looks like I foolishly didn't include a test when I originally wrote #1163.


scauligi (approved):
woopsie

I rarely if ever use hy2py with stdin, so this didn't even cross my radar.

PR #2477, opened 2023-07-29
Correct `root` in pattern macros with dotted head

### LINKED ISSUE #2475: Weirdness with result macros and `.` access
Invoking some of the result macros with an unnecessary attribute access somehow gets hy to make them async:

```python
# (defn normal [])
def normal():
    pass

# ((. defn) weird [])
async def weird():
    pass


# (with [normal])
_hy_anon_var_1 = None
with normal:
    _hy_anon_var_1 = None

# ((. with) [weird])
_hy_anon_var_2 = None
async with weird:
    _hy_anon_var_2 = None


# (fn/a [] (do 0 normal))
async def _hy_anon_var_3():
    0
    return normal

# ((. fn/a) [] (do 0 weird))
def _hy_anon_var_4():
    0
    return weird

# ((. fn) [] (do 0 weirdly-normal))
def _hy_anon_var_5():
    0
    return weirdly_normal
```

It's even more strange to me that `fn` has the _opposite_ behavior of `defn` and `with`.
There might be other examples as well, these are just the ones I stumbled across.

Kodiologist:
- Fixes #2475

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -24,6 +24,8 @@ Misc. Improvements
 Bug Fixes
 ------------------------------
 * Double quotes inside of bracketed f-strings are now properly handled.
+* Fixed incomplete recognition of macro calls with a unary dotted
+  head like `((. defn) f [])`.
 
 0.27.0 (released 2023-07-06)
 =============================
--- a/hy/macros.py
+++ b/hy/macros.py
@@ -52,13 +52,12 @@ def pattern_macro(names, pattern, shadow=None):
                     ).replace(hy_compiler.this)
 
                 expr = hy_compiler.this
-                root = unmangle(expr[0])
 
                 if py_version_required and sys.version_info < py_version_required:
                     raise hy_compiler._syntax_error(
                         expr,
                         "`{}` requires Python {} or later".format(
-                            root, ".".join(map(str, py_version_required))
+                            name, ".".join(map(str, py_version_required))
                         ),
                     )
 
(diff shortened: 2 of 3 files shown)


scauligi:
This should work for now, but I don't like that we now have one more special case check for dot-expressions, especially here where we're specifically only checking for "degenerate" dots.

scauligi:
I guess just to make it a bit clearer, I'm ok with merging this, but I think at some point we'll need to clear technical debt on all of the places that are explicitly checking for dot-forms (and I believe usually they're trying to handle symbol-or-dotted-symbol).

Kodiologist:
Is `(require builtins)` actually supported on purpose? I don't recall seeing it before, I don't think it's tested, and I don't think it's documented, so if this is the only thing you could break with it, I would be surprised. Furthermore, "moving the unwrapping into `macroexpand` or have the single-sym dot-expr somehow translate itself into a Symbol form directly" might make the current handling of `((. foo) …)` cleaner but won't help with `((. builtins foo) …)`. Perhaps I'm missing your point.

scauligi:
> what @scauligi proposed as a solution vs. what @scauligi brought up as a problem

yup you're 100% right, that's on me

an actual fix that works is to just pass `name` to `fn` instead of taking the root from `expr`; I don't think we need the actual expression root for anything.

> macros via `builtins`

My guess was that it was intended to be supported, since `macroexpand` always uses `builtins` as the root in its macro lookup chain as opposed to just having an internal dict of the core macros; you could imagine a scenario where someone wants to shadow or wrap one of the core macros but still needs to be able to call the original macro.

Kodiologist:
> an actual fix that works is to just pass `name` to `fn` instead of taking the root from `expr`

Oh yeah, so it does. I wonder why I bothered with `root` to start with. Perhaps because there was a different arrangement of decorators originally, in which the different `name`s shared a single generated function.

Does this look good to merge?

scauligi (approved):
yup looks good to me

PR #2479, opened 2023-07-29
Fix quotes inside bracketed f-strings

### LINKED ISSUE #2474: Exception when using quotes inside bracketed f-string
```hy
=> #[f["{0}"]f]
  File "stdin-75e2587a4a8f7962cbde4e5df3ca4c83fa031704", line 1
    #[f["{0}"]f]
         ^
hy.reader.exceptions.LexException: unterminated string literal (detected at line 1) (<string>, line 1)
```

Seems like an edge case in the reader, since `#[f["a{0}"a]f]` works fine, but removing either of the `a`s (or swapping either `a` with its neighboring `"`) brings up the error again

scauligi:
Fixes #2474.

maneuvering escapes through encodings and encodings through escapes

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -21,6 +21,10 @@ Misc. Improvements
 * Some syntax errors raised by core macros now have more informative
   messages.
 
+Bug Fixes
+------------------------------
+* Double quotes inside of bracketed f-strings are now properly handled.
+
 0.27.0 (released 2023-07-06)
 =============================
 
--- a/hy/reader/hy_reader.py
+++ b/hy/reader/hy_reader.py
@@ -1,5 +1,6 @@
 "Character reader for parsing Hy source."
 
+import codecs
 from itertools import islice
 
 import hy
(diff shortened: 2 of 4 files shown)


Kodiologist:
You can flatten the `else: if not …` into `elif: not …`.

Why use ISO-8859-1 (a.k.a. `latin1`) in `res.encode('latin1', errors='backslashreplace')` instead of UTF-8 or ASCII? I looked at the linked answer, but it doesn't say. If you really need ISO-8859-1 for some reason, I think it's clearer to call it that than "latin1".

Looks good otherwise.

scauligi:
The `latin1` thing is because because of a python bug (https://github.com/python/cpython/issues/65530); encoding to UTF-8 (or otherwise non-latin1) causes it to mess up non-ascii codepoints in the original string:

```python
>>> r'\N{angry face} \u00fc | ü 😠'.encode().decode('unicode_escape')
'😠 ü | Ã¼ ð\x9f\x98\xa0'
>>> r'\N{angry face} \u00fc | ü 😠'.encode('latin1', 'backslashreplace').decode('unicode_escape')
'😠 ü | ü 😠'
```

 I added a comment as well.

Kodiologist:
A Python bug approaching the age of 10. Go figure. Thanks.

PR #2493, opened 2023-08-13
Improve quasiquote testing and docs

### LINKED ISSUE #2252: Nested unquotes leave behind extra unquote forms
In light of #2251, consider

    (setv c '(f))
    ```[~a ~~b ~~~c]

This returns

    ``[~a ~~b ~~(f)]

but a more useful, and perhaps more logical, behavior would be

    ``[~a ~~b (f)]

It's hard to say exactly what should happen in these cases because we usually think of nested backticks with intervening unquotes. Nested backticks without an intervening unquote only come up when you're trying to return a backticked form for later expansion.

Kodiologist:
- Closes #2252

After a lot of [soul-searching](https://old.reddit.com/r/lisp/comments/15oc52i/how_does_nested_quasiquotation_work_how_should_it/) and adding a few more tests, I've decided that the way Hy does quasiquotation is correct, or at least, correct enough.

I was pointed to [a pretty thorough test suite of quasiquotation in Common Lisp](https://github.com/fare/fare-quasiquote/blob/master/quasiquote-test.lisp), but translating it to Hy looks intimidating, so I'm not going to take that on before I'm convinced that we need even more tests than what I've added here.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -792,42 +792,60 @@ base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
    beware that significant leading whitespace in embedded string literals will
    be removed.
 
-.. hy:macro:: (quasiquote [form])
+.. hy:macro:: (quasiquote [model])
+.. hy:macro:: (unquote [model])
+.. hy:macro:: (unquote-splice [model])
 
-   ``quasiquote`` allows you to quote a form, but also selectively evaluate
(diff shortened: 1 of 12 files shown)


scauligi (approved):
wow trying to work through the Bawden tests really bends my brain

Kodiologist:
In hindsight, `test-nested-quasiquote--var-from-above` and `test-nested-quasiquote--macro-writing-macro ` may be redundant.

Issue #2484, opened 2023-08-04
Macro documentation


wrobell:
As newcomer to Lisp and Hylang, I am finding Macros section in the documentation quite confusing.

This is https://docs.hylang.org/en/stable/macros.html.

After few weeks of dealing with Hylang and few other Lisp related things, if I may suggest, it would be great if there was

- quick introduction/example
- differences between usual Lisp notation and Hylang, i.e. `,` vs `~`
- link to a Lisp macro tutorial/book (i.e. On Lisp by Paul Graham is available for free online)
- link to the API

Not bad information is in the API section, but hard to find it there, especially on a smartphone.

Kodiologist:
Yeah, the chapter is long overdue for a rewrite. I'm procrastinating because I first want to be reasonably sure no additional changes to macros or reader macros are coming up.

Issue #1467, opened 2017-12-16
Support user introspection and modification of the currently defined macros


ekaschalk:
Help me understand `hy.macros.-hy-macros` initialization.

First run will print `hy.macros.-hy-macros` fine. Second and subsequent runs will print an empty dict.

Deleting the `foo.pyc` will reset the `-hy-macros` dict as expected.

Is this intended behavior? If intended - how do I test code referencing `-hy-macros`?

`pkg/foo.hy`

    (import hy.macros)

    (defn foo []
      (print hy.macros.-hy-macros))

`pkg/test_foo.hy`

    (import [pkg.foo [foo]])

    (defn test-foo []
      (foo)
      (doc ->))  ; Similarly, will print on first runthrough, error on second run

gilch:
Maybe related to #1324.

It's easy to force a compile at runtime though, just use `hy.eval`, e.g.
```Hy
(hy.eval '(print hy.macros._hy-macros))
```

ekaschalk:
Ok thanks a simple `(hy.eval '(import hy.macros))` in the module being tested does the trick.

Kodiologist:
I don't think `hy.macros._hy_macros` is intended for public use.

ekaschalk:
Is there a better way to access all defined macros? `doc` takes the same approach. It's for code introspection for Emacs, not sure if that qualifies as public use.

Kodiologist:
No, I don't think so. We've never worked on introspection of the macro system beyond Hy's own internal needs.

Kodiologist:
Better support for introspection of the macro system could replace `hy.reserved` and perhaps also `delmacro`.

PR #2503, opened 2023-08-26
Implement local `defmacro` and `require`

### LINKED ISSUE #900: Local macros
We should implement `macrolet`. That's what Common Lisp calls it, but Clojure may end up calling it `letmacro`, so we could call it that instead.

It might work by using `defmacro` named by gensym. Anywhere the macro binding symbol appears in the `letmacro` body (and in an appropriate position for macroexpansion) `letmacro` would replace it with that gensym. It should not replace the binding symbol if it appears in an inappropriate position, such as the head of a quoted Hy expression.

This approach might litter some gensyms, but without a proper `let` most uses of `gensym` will do this anyway.

Alternatively, perhaps it could be done better at the Python level, in a way similar to how `defmacro` was implemented in the first place.

Kodiologist:
- Closes #900

I haven't added support for local macros to other parts of the language (like `hy.eval`) in lieu of a more comprehensive approach to macros as first-class objects (#1467), which I'd like to do soon. Similarly, I haven't polished the documentation because I'd like to do that as part of a larger overhaul of, and content addition for, macro-related documentation (#2484).

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -5,6 +5,13 @@ Unreleased
 
 Breaking Changes
 ------------------------------
+
+* `defmacro` and `require` can now define macros locally instead of
+  only module-wide.
+
+  * `hy.eval`, `hy.macroexpand`, `doc`, and `delmacro` don't work with
+    local macros (yet).
+
 * When a macro is `require`\d from another module, that module is no
   longer implicitly included when checking for further macros in
   the expansion.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -198,6 +198,10 @@ base names, such that ``hy.core.macros.foo`` can be called as just ``foo``.
       => (infix (1 + 1))
       2
 
+   If ``defmacro`` appears in a function definition, a class definition, or a
+   comprehension other than :hy:func:`for` (such as :hy:func:`lfor`), the new
+   macro is defined locally rather than module-wide.
+
    .. note:: ``defmacro`` cannot use keyword arguments, because all values
              are passed to macros unevaluated. All arguments are passed
              positionally, but they can have default values::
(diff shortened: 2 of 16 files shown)


scauligi:
Awesome! This is rad.
I might need a day or two to fully go over everything, if that's cool.

Looking over it briefly, it seems like creating a new `local_macros` scope has a 1-to-1 correspondence with creating a new `ScopeFn`/`ScopeGen` scope; maybe this could be rolled into the existing scope tracking instead of a separate list in the compiler?

Kodiologist:
> I might need a few days to fully go over everything, if that's cool.

That's fine.

> maybe this could be rolled into the existing scope tracking instead of a separate list in the compiler?

I thought about it, but I wasn't confident about trying to combine them, because of differences like how macro names should be independent of `let`.

scauligi:
> macro names should be independent of `let`.

Hunh fair enough. In any case, working it through in my head, I think using the `hy.scoping` classes would end up being more code just to ultimately do the same thing you've already implemented.

scauligi:
Agghhh, really sorry @Kodiologist but I haven't had time to get around to this and I'm going to be on leave for the next week and a half.

From what I was able to look over, I wasn't able to find any buggy corner cases, so any review I did would probably have been for code smell / etc. anyways; I'm fine with this being merged.

The only other thing I have is that it might be worth calling out in the docs that `(require :readers)` will *not* work in this way even though normal `(require)` now does, since readers still must be top-level to be usable immediately.

Kodiologist:
Okay, thanks.

> The only other thing I have is that it might be worth calling out in the docs that `(require :readers)` will not work in this way even though normal `(require)` now does, since readers still must be top-level to be usable immediately.

I added under the manual entry for `require` that reader macros "can't be required locally", which should cover this.

PR #2510, opened 2023-09-15
Fix comments in the index and why Hy documents

wrobell:
This is to avoid the following warnings when building the documentation

    Explicit markup ends without a blank line; unexpected unindent.


Kodiologist:
When you have several small changes to make (like this one and the previous PR), I'd recommend grouping them into a single PR.

PR #2511, opened 2023-09-30
First-class global macros

Kodiologist:
This pull request partly fulfills #1467, by testing and documenting introspection that's possible via `_hy_macros`, and removing some functions and macros that are made redundant by these features. `_hy_macros` only provides access to global macros, so the missing part is support for local macros. That I expect to require implementing new features, so I'm leaving it for another PR.


scauligi:
Hold on, I'm confused. Are we considering `_hy_macros` as part of Hy's public API now? I thought it was just internal?
Also I see no reason to remove `doc` just yet, even if we are officially exposing `_hy_macros`, since using `doc` is much more ergonomic than pulling mangled keys out of `_hy_macros`.

Kodiologist:
> Are we considering `_hy_macros` as part of Hy's public API now? I thought it was just internal?

Right, it was internal, but by documenting it, I'm making it public. Making what we already have public seemed like a more straightforward way to provide these features than to add another interface.

> I see no reason to remove `doc` just yet, even if we are officially exposing `_hy_macros`, since using `doc` is much more ergonomic than pulling mangled keys out of `_hy_macros`.

It is, yes, but that seems more like something for Hyrule than something to keep as a core macro.

scauligi:
I disagree wrt `doc`, I feel like it's important enough that it still belongs as part of Hy's core; it's the analogue of Python's `help` being a builtin rather than a function imported from the stdlib or asking the user to `print(function.__doc__)`.

Otherwise, for other removals like `delmacro`, is the intention to have a set of macro-managing macros eventually in Hyrule?

Kodiologist:
> I disagree wrt `doc`, I feel like it's important enough that it still belongs as part of Hy's core; it's the analogue of Python's `help`

Thing is, Hy's analog of Python's `help` is `help`. All the actual logic of `doc` is finding the named macro. Once it has the macro, it just calls `help`. So if you think `(help (get _hy_macros (hy.mangle "macroname")))` is too inconvenient, what if we provided `(get-macro macroname)`? Then you could say `(help (get-macro macroname))`. And it would be useful for other things; you could also say `(del (get-macro macroname))`, for example.

> for other removals like delmacro, is the intention to have a set of macro-managing macros eventually in Hyrule?

It's the sort of thing one could add to Hyrule, but I have no motivation to do so. I don't see myself manipulating the macro system like this a huge amount.

scauligi:
My preference is still to keep `doc`, but I'm willing to settle for `(help (get-macro ...))`.
I do agree that regardless, `get-macro` seems like a good general utility to have.

Kodiologist:
Okay, I'll take a stab at adding `get-macro` to this PR.

Kodiologist:
`get-macro` turns out to be harder to get right than I expected, and I don't really want to work out all the subtleties right now, so I just undid the removal of `doc` from this PR. Look good?

scauligi (approved):
Yup, otherwise looks fine :+1:

Issue #2507, opened 2023-09-06
let, lambda and scope of bindings


wrobell:
I can reference a binding in a lambda defined under let
```
(setv test-let-a
  (let [start 0]
    (fn []
      (setv result (cut "abcdefg" start (+ start 4)))
      result)))

(test-let-a) ; "abcd"
```
But I cannot update it
```
(setv test-let-b
  (let [start 0]
    (fn []
      (setv result (cut "abcdefg" start (+ start 4)))
      (+= start 1)
      result)))

(test-let-b) ; UnboundLocalError: cannot access local variable 'start' where it is not associated with a value
```
Using `nonlocal` does not seem to help
```
(setv test-let-c
  (let [start 0]
    (fn []
      (nonlocal start)
      (setv result (cut "abcdefg" start (+ start 4)))
      result))) ; SyntaxError: no binding for nonlocal '_hy_let_start_78' found
```
I might misunderstand something, or a bug?

Documentation says
> Basic assignments (e.g. setv, +=) will update the local variable named by a let binding when they assign to a let-bound name.

Therefore `test-let-b` shall work?

Kodiologist:
I think this behavior is intentional. Consider this simplified case:

```hy
(let [x 0]
  (fn []
    (setv x 1)))
```

This compiles to:

```python
import hy
_hy_let_x_1 = 0

def _hy_anon_var_2():
    x = 1
```

because, per Python, assigning to a variable in a function, without an accompanying `nonlocal` or `global`, implies that you want to assign to a variable local to that function. If you add `(global x)` before `(setv x 1)`, then the assignment happens to the `let` variable as desired. `(nonlocal x)` won't work because `x` is global, not local to a containing function. But having to refer to `x` as global does feel strange considering that it's not meant to be accessible outside the `let`.

@scauligi What say you?

Kodiologist:
No, actually, `global` doesn't work, either. This

```
(let [x 0]
  (fn []
    (global x)
    (setv x 1)))
```

compiles to

```
import hy
_hy_let_x_1 = 0

def _hy_anon_var_2():
    global x
    x = 1
```

Something is definitely wrong.

Kodiologist:
Yes, because that's the first reference to the variable. You can see it in `hy2py`.

wrobell:
> Yes, because that's the first reference to the variable. You can see it in `hy2py`.

`test-let-a` references `start` as well, but it does not raise an error about unbounded variable.

[...]

PR #2513, opened 2023-09-30
Automatically promote `nonlocal` as needed

scauligi:
Promote `nonlocal` declarations to `global` ones if the "nearest" matching definition is in the global scope.

Also addresses #2507; can be considered a fix if we decide that the explicit-`nonlocal` formulation is the proper way to do things.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -36,6 +36,7 @@ New Features
 * `HyReader` now has an optional parameter to install existing
   reader macros from the calling module.
 * New pragma `warn-on-core-shadow`.
+* `nonlocal` now also works for globally defined names.
 
 Misc. Improvements
 ------------------------------
@@ -53,6 +54,7 @@ Bug Fixes
   install a new reader.
 * `require` now warns when you shadow a core macro, like `defmacro`
   already did.
+* `nonlocal` now works for top-level `let`-bound names.
 
 0.27.0 (released 2023-07-06)
 =============================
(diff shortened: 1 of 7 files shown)


Kodiologist:
Nice branch name. Although my favorite is still one I saw for a PR to Cataclysm DDA, `the-gods-outlawed-burnt-water-long-ago`.

Kodiologist:
Looks good overall. I agree that this closes #2507.

Let's update the documenation of `nonlocal` in `api.rst`.

Can you add a comment or docstring to `OuterVar` or `ResolveOuterVars` explaining that their purpose is to decide when to promote `nonlocal` to `global`?

Shouldn't `OuterVar.__init__` call `super().__init__`? Even if `ast.stmt.__init__` doesn't do anything right now, it might in the future. If `super().__init__` actually won't work with what you need to do here, it's worth a comment saying so.

I think `scope.parent is not None` can be just `scope.parent`.

```
for name in has:
    if name in undefined:
        undefined.remove(name)
```

This is better written `undefined = [name for name in undefined if name not in has]`.

Tests of `nonlocal` with `let` are probably better put in `tests/native_tests/let.hy`, next to the other `let`-with-`nonlocal` tests, than in your new test file about `nonlocal.hy`.

`ration-log` isn't used after it's assigned. It would be better to only return `health`.

scauligi:
Done! I also removed one of the new tests that wasn't actually using `nonlocal`, as it was redundant with tests already in `native_tests/let.hy`.

Kodiologist:
Thanks. I wouldn't remove mention of what `nonlocal` compiles to; just say that it can compile to `global` or `nonlocal` statements (with links to the Python keywords), and explain in what circumstances it will produce `global` instead of `nonlocal`.

scauligi:
Alright, updated the `nonlocal` doc :+1:

PR #2514, opened 2023-09-30
Optimize short-circuiting operators

### LINKED ISSUE #854: Optimize and/or
Depends on #824.
Related #842.

Even in the case that `and`/`or` contains a statement, it may have several expressions in a row that don't need to be converted to `if` branches. Hy could avoid redundant assignments by keeping those runs as expressions and only assigning the result to an anonymous variable.

For example:

``` Python
=> (print (and 1 2 3 (setv x 1) 4 5 6))
_hy_anon_var_1 = 1 and 2 and 3
if _hy_anon_var_1:
    x = 1
    _hy_anon_var_1 = x and 4 and 5 and 6
print(_hy_anon_var_1)
6
```

scauligi:
Fixes #854.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -34,6 +34,7 @@ Misc. Improvements
 ------------------------------
 * Some syntax errors raised by core macros now have more informative
   messages.
+* Logical operators now compile to simpler Python code in some cases.
 
 Bug Fixes
 ------------------------------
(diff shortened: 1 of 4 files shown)


Kodiologist:
Thanks, this was long overdue. I struggled to understand the code, so I made some small simplifications and renames and added comments. I also added a test of an AST feature that I realized was untested when I broke it while testing an in-progress form of my changes.

Kodiologist (approved):

PR #2517, opened 2023-10-07
Make `hy -i <filename>` skip shebang line

scauligi:
Also sets `sys.argv` as `python3 -i <filename>` does.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -55,6 +55,7 @@ Bug Fixes
 * `require` now warns when you shadow a core macro, like `defmacro`
   already did.
 * `nonlocal` now works for top-level `let`-bound names.
+* `hy -i` with a filename now skips shebang lines.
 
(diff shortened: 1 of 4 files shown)


Kodiologist:
We already have shebang removal elsewhere in the code. Can you use that, or does it have to be reimplemented here for some annoying reason?

For `test_icmd_shebang`, would you please write your test file in the code with the `tmp_path` fixture (as demonstrated by e.g. `test_require_doesnt_pollute_core`)? I think it's neater for tests to be self-contained that way.

scauligi:
Hmm, the other way I can do this is to set a flag on the REPL instance and have it pass `skip_shebang` to `read_many`, I'll see if that isn't too ugly to do.

> use the `tmp_path` fixture
Can do!

Kodiologist (approved):

PR #2518, opened 2023-10-07
Fix overzealous boolean compaction

scauligi:
Can't use `isinstance(..., BoolOp)` to check if the tracked expression is "our" boolop node or not, because the first operand might itself be a `BoolOp`.


Kodiologist (approved):
I hope I didn't introduce this regression while       I was rewriting all that code, whoops.

scauligi:
Nah, I'm pretty sure this was bugged in my original PR, I was trying to be a little too clever with checking the singleton vs boolop case by relying on `isinstance`.

PR #2519, opened 2023-10-15
Add `get-macro` and remove `doc`

Kodiologist:


scauligi:
For the record, I still would prefer to keep `doc`, but having `get-macro` lessens the blow.

The `get-macro` docstring could use a clarification that when using `:reader`, the name is given without the leading `#` (and also that it takes/can take a string). Otherwise looks good to me.

Kodiologist:
> the name is given without the leading #

This is true everywhere else, too, like in `defreader` and `require`. The hash sign isn't part of the name of the reader macro; it's just used to call it, like the parentheses of a regular macro call. So are you sure it's worth a mention here?

> and also that it takes/can take a string

That's already mentioned: "one argument, a symbol or string literal".

scauligi:
> leading #

I think it's still worth a mention for clarity (and imo should be mentioned in the doc for `require` too, eventually).

> string already mentioned

So it is! I missed that, my bad.

Kodiologist:
@scauligi

scauligi:
Woops sorry, yeah lgtm :shipit:

PR #2522, opened 2023-10-16
Fix parsing of `nanJ` etc.

### LINKED ISSUE #2521: Any capitalization of `NaN` and `Inf` are parsed as imaginary numbers with uppercase suffix `J`
This issue seems related to #1288 and #1294. Floating-point special values such as `NaN` and `Inf` are required to be capitalized exactly as such, but when used in imaginary number literals with _uppercase_ `J` suffix, any capitalization is allowed. In the same spirit as #1288, I believe this is unintended/undesired behavior.

```hy
NaNj ; valid (good)
NANj ; invalid (good)
Infj ; valid (good)
iNfj ; invalid (good)

NaNJ ; valid (good)
NANJ ; valid (should not be!)
InfJ ; valid (good)
iNfJ ; valid (should not be!)
```

Interestingly enough, this behavior only occurs in pure-imaginary literals; it doesn't surface in "compound" complex literals, which seem to have the correct behavior:

```hy
1+NaNj ; valid (good)
1+NANJ ; invalid (good)
```

Kodiologist:
Fixes #2521.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -50,6 +50,8 @@ Bug Fixes
 * Fixed incomplete recognition of macro calls with a unary dotted
   head like `((. defn) f [])`.
 * `~@ #*` now produces a syntax error instead of a nonsensical result.
+* Fixed parsing of infinite and NaN imaginary literals with an
+  uppercase "J".
 * Fixed `hy.eval` failing on `defreader` or `require` forms that
   install a new reader.
 * `require` now warns when you shadow a core macro, like `defmacro`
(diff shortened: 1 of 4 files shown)


scauligi (approved):
Oh nice, that was fast. :rocket:

Kodiologist:
In truth, I dislike that Python allows `3J` for `3j`, `1E2` for `1e2`, and `0X10` for `0x10`, but it's not worth trying to make Hy case-sensitive about these things.

PR #2524, opened 2023-10-26
Async generators no longer implicitly return final expressions

### LINKED ISSUE #2523: Hy produces an illegal implicit `return` in async generators
This bug looks almost the same as #691 

```lisp
(defn/a naysayer []
  (yield "nope"))
```

raises `SyntaxError: 'return' with value in async generator`. The output of `hy2py` is

```python
async def naysayer():
    return (yield 'nope')
```

The current workaround is to use `assert` 

```lisp
(defn/a naysayer []
  (yield "nope")
  (assert True))
```

scauligi:
Fixes #2523.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -60,7 +60,7 @@ Bug Fixes
   already did.
 * `nonlocal` now works for top-level `let`-bound names.
 * `hy -i` with a filename now skips shebang lines.
-* Async generators no longer try to return final expressions.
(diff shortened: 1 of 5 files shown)


Kodiologist:
Nice! Maybe add a comment or two to the new attributes of `ScopeFn` describing their purpose. A reader would naturally assume that they have something to do with scope tracking, but they really don't.

Kodiologist (approved):

PR #2526, opened 2023-11-06
Add `hy.R` and rename `hy.M` to `hy.I`

### LINKED ISSUE #2512: `hy.M` for macros
I like the one-shot import syntax `hy.M` a lot, but it would be nice to have an equivalent for macros, too. Perhaps we should use the name `hy.M` for that and rename the existing feature to `hy.I`, with the "I" standing for "import".

Kodiologist:
- Closes #2512

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -20,6 +20,7 @@ Breaking Changes
 * When a macro is `require`\d from another module, that module is no
   longer implicitly included when checking for further macros in
   the expansion.
+* `hy.M` has been renamed to `hy.I`.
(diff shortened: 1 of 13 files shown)


scauligi:
I think add a test to verify that `hy.R` doesn't modules/macros into scope, but otherwise looks good to me

Kodiologist:
Good idea. Done.

Kodiologist:
@scauligi How's this?

scauligi (approved):
lgtm 🚀

PR #2555, opened 2024-03-05
If the action is run_script_stdin with the repl option, can simply en…

a690700752:
…ter the repl without reading it.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -3,6 +3,18 @@
 Unreleased
 =============================
 
+Removals
+------------------------------
+* `(defn/a …)` is now `(defn :async …)`.
+* `(fn/a …)` is now `(fn :async …)`.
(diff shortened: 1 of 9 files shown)


Kodiologist:
What's the motivation here? All the other `if` branches deal with the REPL case explicitly.

a690700752:
When I use the -i parameter to force the launch of the REPL and send code in subprocess (using nvim conjure), it directly enters the repl.runsource block for execution and never enters repl.run(). In this case, the current environment is not the REPL environment, lacking the PS1, PS2 prompts, and some other REPL-specific environment settings, which contradicts the meaning of the -i parameter. On the other hand, other branches execute code first, such as from a file, before entering repl.run().

a690700752:
OKay, I will add tests and update it later.

Kodiologist:
@a690700752 What happened here?

PR #2557, opened 2024-03-10
Fix non-interactive with param -i in subprocess

a690700752:
continue #2555 .
When I use the -i parameter to force the launch of the REPL and send code in subprocess (using nvim conjure), it directly enters the repl.runsource block for execution and never enters repl.run(). In this case, the current environment is not the REPL environment, lacking the PS1, PS2 prompts, and some other REPL-specific environment settings, which contradicts the meaning of the -i parameter. On the other hand, other branches execute code first, such as from a file, before entering repl.run().

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -110,3 +110,4 @@
 * Andrey Vlasovskikh <andrey.vlasovskikh@gmail.com>
 * Joseph LaFreniere <joseph@lafreniere.xyz>
 * Daniel Tan <danieltanfh95@gmail.com>
+* Zhan Tang <tz59pk@gmail.com>
--- a/hy/cmdline.py
+++ b/hy/cmdline.py
@@ -262,10 +262,7 @@ def cmdline_handler(argv):
         return 0
     elif action == "run_script_stdin":
         sys.argv = argv
-        if repl:
-            source = sys.stdin
-            filename = 'stdin'
-        else:
+        if not repl:
             return run_command(sys.stdin.read(), filename="<stdin>")
     elif action == "run_script_file":
         sys.argv = argv
--- a/tests/test_bin.py
+++ b/tests/test_bin.py
@@ -79,6 +79,10 @@ def test_stdin():
     assert "RS" in out
     assert "TU" in out
 
+    # With it, the REPL variable must exists, like ps1
+    out, _ = run_cmd("hy -i", '(import sys) (bool (getattr sys "ps1" sys.flags.interactive))')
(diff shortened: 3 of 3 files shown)


Kodiologist:
The test is a bit wonky. Giving the `default` argument to `getattr` means that you're accepting either of `sys.ps1` or `sys.flags.interactive`. Shouldn't you be more specific as to which you expect, or simply test for both?

a690700752:
There just a tests result, below is the result. Maybe we can add comment to that test code.

| Environment                                 | IPython | PyCharm Console | Shell (run Python, paste code) | Jupyter Notebook | PyCharm Run | PyCharm Run (Emulate terminal) | Shell (run Python file) |
| ------------------------------------------- | ------- | --------------- | ------------------------------ | ---------------- | ----------- | ------------------------------ | ----------------------- |
| Expected Value                              | TRUE    | TRUE            | TRUE                           | TRUE             | FALSE       | FALSE                          | FALSE                   |
| hasattr(sys,'ps1')                          | TRUE    | TRUE            | TRUE                           | TRUE             | FALSE       | FALSE                          | FALSE                   |
| hasattr(sys,'ps2')                          | TRUE    | TRUE            | TRUE                           | TRUE             | FALSE       | FALSE                          | FALSE                   |
| sys.\_\_stdin\_\_.isatty()                  | TRUE    | FALSE           | TRUE                           | FALSE            | FALSE       | TRUE                           | TRUE                    |
| sys.\_\_stdout\_\_.isatty()                 | TRUE    | FALSE           | TRUE                           | TRUE             | FALSE       | TRUE                           | TRUE                    |
| hasattr(\_\_builtins\_\_,'\_\_IPYTHON\_\_') | TRUE    | FALSE           | FALSE                          | TRUE             | FALSE       | FALSE                          | FALSE                   |
| not hasattr(\_\_main\*\*, '\_\_file\_\_')   | TRUE    | FALSE           | TRUE                           | TRUE             | FALSE       | FALSE                          | TRUE                    |
| bool(sys.flags.interactive)                 | FALSE   | FALSE           | FALSE                          | FALSE            | FALSE       | FALSE                          | FALSE                   |

Kodiologist:
That's an impressive table you've compiled, but your test only checks `hy -i`, so that's what the assertion should be written for.

a690700752:
Ok, I'll continue finish tests.

Kodiologist:
How's this going?

Kodiologist:
Superseded by #2587.

PR #2565, opened 2024-04-09
Overhaul rendering of the manual

### LINKED ISSUE #2417: Stop using Read the Docs
Having ads in the manual is extraordinarily tacky. We should probably just host the web versions of Hy and Hyrule's manuals on Arfer.net, where I also host [the new Hylang.org](http://hylang.org). For simplicity, we can serve only the stable release of the manual. We would then just rebuild it as part of the release process.

Kodiologist:
- Closes #2417

Here are various changes, in combination with a commit I just pushed to [hyhomepage](https://github.com/hylang/hyhomepage), to clean up the manual and put it on hylang.org. You can see the result at http://hylang.org/hy/doc/doc-testing.  URLs that now look like

https://docs.hylang.org/en/stable/syntax.html

will instead be

https://hylang.org/hy/doc/v0.26.0/syntax

- I use version numbers insted of `stable` and `master` because [cool URIs don't change](https://www.w3.org/Provider/Style/URI). We'll no longer provide official web versions of pre-release versions of the Hy manual, and that should no longer be terribly necessary now that Hy isn't breaking backwards compatibility so frequently. I'll redirect docs.hylang.org to the latest release.
- The file extension (`.html`) is gone because it's an implementation detail.
- `en` is gone because full translations, if they're ever written (which I doubt), will probably be hosted under different URLs, as with Python itself.

If there are no objections, I'll make similar arrangements for Hyrule's manual, too.

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,0 +1,15 @@
+div.body, div.footer, code, div.code-block-caption,
+div.body h1, div.body h2, div.body h3, div.body h4, div.body h5, div.body h6,
+a.toc-backref
+   {color: black}
+
+div.body p, div.body dd, div.body li, div.body blockquote
+   {hyphens: manual}
+
+a
+   {color: navy}
+a:visited
+   {color: purple}
+
+code
+   {background-color: #dfd}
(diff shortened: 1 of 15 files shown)


goyalyashpal:
> * I use version numbers insted of `stable` and `master` because [cool URIs don't change](https://www.w3.org/Provider/Style/URI).

but stable and master do help to link to the latest versions of a documentation.

the expected is for the stable to automatically redirect to the link with latest version number

PR #2581, opened 2024-05-01
Simplify versioning

Kodiologist:
Notes to self for when this is merged:

- Edit the release instructions.
- Arrange Hyrule similarly.


tianon:
I've had some pains in packaging thanks to this setup, and I think this will help alleviate them! 😀🫡

PR #2586, opened 2024-05-10
Fixes for `help` and JIT imports

Kodiologist:


seteshstrong:
Looking forward to the fix on the help regression. Just beginning to learn hylang and the regression you've fixed makes doing so **very** painful.

Kodiologist:
Sorry about that.

PR #2604, opened 2024-09-15
Release Hy 1.0.0

Kodiologist:
That's right, folks. It's finally happening.

As you can see from NEWS, I have the actual release scheduled for 2024-09-22, which is this coming Sunday. I'll post announcements in various places when the time comes.

@hylang/core


paultag:
**hell yes**

Kodiologist:
@paultag I was thinking about the question of how long Hy has been in development. Your first commit is from 15 Dec 2012. Do you know if that's a roughly accurate estimate of when you started work on Hy, or had you been working on prototypes for a long time before you brought anything into Git?

paultag:
@Kodiologist Yeah, totally roughly right. I started working on it a week or two before (at most!) that commit -- I definitely prototyped it a few times (wrote it, scrapped it, rewrote it, scrapped it) before checking that in -- but it wasn't more than a few days off from that date. Definitely December

PR #2623, opened 2025-01-04
Added SVG version of Cuddles

rcarmo:
Hy there!

I needed an SVG version of Cuddles to use as a favicon, and thought I'd contribute it back.


Kodiologist:
Cool, thanks.

PR #2658, opened 2025-06-01
Allow chained assignment in `setv`

Kodiologist:
@scauligi I decided it was best not to add a core macro, which would add to our already large set of core macros, for this rarely used feature.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,5 +1,14 @@
 .. default-role:: code
 
+Unreleased
+======================================================================
+
+Supports Python 3.x – Python 3.y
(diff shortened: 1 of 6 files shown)


scauligi:
Fair enough!
For completeness, maybe add a test that uses multiple `:chain` assignments in the same `setv`?
And for the `hy2py` test, I think a slightly more precise test is chain-assigning an object (even literally `object()`), then asserting `chained1 is chained2 is chained3`.

scauligi (approved):
lgtm! :shipit:

PR #2689, opened 2026-03-17
Add double-discard `#__` macro

KevinMGranger:
You know what's even cooler than discarding a form? Discarding _two_ forms.

If it should be easy to discard a single argument to something, it should also be easy to discard a key-value pair (be it a dictionary literal or function keyword arguments).

IIUC, since a user-defined reader macro named `__` should shadow this, so there's no risk of this being a breaking change.

Edit: whoops, missed that adding one's self to the `AUTHORS` file should be a separate commit. As in, after it's merged, or just one after the implementation?

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -7,6 +7,10 @@ Unreleased
 
 Supports Python 3.x – Python 3.y
 
+New Features
+------------------------------
+* Added ``#__``. Similar to ``#_``, but discards the next *two* forms.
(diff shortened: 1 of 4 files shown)


Kodiologist:
This seems extremely niche, too much to be a core feature. You could submit it to Hyrule if you want.

KevinMGranger:
Will gladly do.

PR #2698, opened 2026-04-12
Templated strings

### LINKED ISSUE #2687: 3.14 Template Strings
I'd like to take a swing at adding [3.14's template strings](https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-template-string-literals) (`t"foo{bar}"`) to Hy.

However, for bracket strings, following the same rules for f-strings... it'd be a breaking change for anyone already using `#[t[` or `#[t-$SOMETHING[`!

Thoughts on how to resolve that? Maybe a `pragma` option to opt in? Alternate syntax specifically for bracketed template strings?

scauligi:
- Closes #2687

Python 3.14 introduced [t-strings](https://docs.python.org/3/library/stdtypes.html#stdtypes-tstrings) as a new literal type.

This PR allows parsing t-strings, as well as introduces a new pragma `bracketed-templates` to additionally allow bracketed t-strings (a la bracketed f-strings).

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -10,6 +10,8 @@ Supports Python 3.x – Python 3.y
 New Features
 ------------------------------
 * ``hyc`` now supports ``-q``/``--quiet`` to suppress progress messages.
+* Added support for t-strings from Python 3.14.
+* Added new pragma `bracketed-templates` to allow parsing `#[t[...]t]` and `#[t-<ident>[...]t-<ident>]` as template strings (analogous to bracketed f-strings).
 
 Bug Fixes
 ------------------------------
@@ -24,6 +26,7 @@ Bug Fixes
   and Complex models.
 * Fixed an importlib exception when using `hy.eval` on code that
   required other modules.
+* Invalid conversion chars in f-strings now properly raise a syntax error.
 
 1.2.0 ("Crackers and Snacks", released 2026-01-14)
 ======================================================================
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -7,12 +7,18 @@ Unreleased
 
 Supports Python 3.x – Python 3.y
 
+New Features
+------------------------------
+* Added support for t-strings from Python 3.14.
+* Added new pragma `bracketed-templates` to allow parsing `#[t[...]t]` and `#[t-<ident>[...]t-<ident>]` as template strings (analogous to bracketed f-strings).
+
 Bug Fixes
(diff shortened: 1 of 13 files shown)


Kodiologist:
Looks like you gotta gate the tests to Python 3.14.

Kodiologist:
I don't think you need `pytest_ignore_collect` for this. In an analogous way that I guard some files with `do-mac`, you can prefix your file with a reader macro like so:

```hy
(defreader r []
  (if hy.compat.PY3-14
    `(do ~@(.parse-forms-until &reader ""))
    (for [_ (.chars &reader True)])))

#r
```

scauligi:
> you can prefix your file with a reader macro like so

oh very cute, I like that

Kodiologist:
I force-pushed to group some logically related commits together.

I think you might be gating for Python 3.14 in Hy itself a bit too aggressively. There's no reason that the Hy syntax `t"foo"` needs to be illegal in Python 3.13, for example. The Hy-level syntax and models for t-strings don't require support from Python. The only place we should need to check the Python version is in the compiler functions, when the user tries to compile a Hy t-string into a Python-level t-string. As a bonus, you'll no longer need a reader macro to gate the tests, and some tests won't need to be gated.

By the way, does it make sense for t-strings to by represented as `hy.models.FString`s, instead of getting their own model type? That seems confusing.

It looks like `hy-repr` for t-strings isn't tested.

I think `(= (type t1) (type t2) Template)` would be better written with `is` instead of `=`.

Finally, of course, we'll need documentation.

scauligi:
Added docs and the tests for `hy.repr`, and removed the aggressive reader-side checking for python 3.14.

> By the way, does it make sense for t-strings to by represented as `hy.models.FString`s, instead of getting their own model type? That seems confusing.

Really the only difference between f-strings and t-strings is what they finally evaluate to, but up until they're actually compiled, f-strings and t-strings are pretty much exactly the same. Parsing t-strings directly into `FString`s saves a bunch of otherwise repeated code and/or kludgy type checks.

Kodiologist (approved):
There are a lot of small things I want to change, but I think the changes would be clearer as their own PR. So I'll take care of that after.

Rebase your updates (removing the Python version checks in the reader, and using `is` instead of `=`) back into the original commits you're updating. Then, go ahead and merge.

Kodiologist:
Thanks!

PR #2699, opened 2026-04-14
Add pattern matching to "primitive" models

### LINKED ISSUE #2694: Cannot use `match` with models that inherit from certain builtins
(I have confirmed this on 3.14 and for certain built-in types, I haven't tried other versions/types yet.)

[Python's `match` spec for class patterns](https://docs.python.org/3/reference/compound_stmts.html#class-patterns) has an interesting edge case:

If a class singly-inherits from the list of builtins mentioned in the above link, then you can pattern-match with a single positional arg as the only class pattern.

```python
class DirectChild(str):
    pass

class DirectChildChild(DirectChild):
    pass

match DirectChildChild("foo"):
    case DirectChildChild("foo"):
        assert True
    case other:
        print(f"no matches for {other!r}")
        assert False
```

However, as soon as you also inherit from another class, such as with `hy.models.String`/`Integer`/etc, it requires standard `__match_args__` behavior.

```python
import hy.models as hm

match hm.String("foo"):
    case hm.String("foo"):
        print("matched hm string foo?")
        assert False
    case "foo":
        print("matched string literal?")
        assert False
    case other:
        print(f"no matches for {other!r}")
        assert True
```

This can be accomplished one of two ways:
1. Put the primitive type first in the inheritance (e.g. `class String(str, Object)`).
2. With a property or attr that'll return an instance of the builtin type, and naming that in `__match_args__`:

```python
class Base:
    pass

class Str(Base, str):
    __match_args__ = ("_as_str",)

    @property
    def _as_str(self):
        return str(self)

    # OR

    def __new__(cls, s):
        s = str(s)
        value = super().__new__(cls, s)
        value._as_str = s
        return value

match Str("foo"):
    case Str("foo"):
        print("matched Str foo")
        assert True
    case other:
        print(f"no matches for {other!r}")
        assert False
```

I'd be happy to make a PR for this.

I know one can get the desired behavior with model patterns, but if one's making tools for hy in python, this has less ceremony for the simple cases.

_Note: it'd be nice to use a dunder method to hide the implementation detail (e.g. `__as_str`), but that won't work because of python name mangling._

_Another note: this doesn't happen with the `hy.models.Sequence` subclasses for some reason, despite `tuple` also being in the list in the link. I don't why._

KevinMGranger:
Closes #2694.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -111,3 +111,4 @@
 * Daniel Tan <danieltanfh95@gmail.com>
 * Zhan Tang <tz59pk@gmail.com>
 * Ati Sharma <ati@agalmic.ltd>
+* Kevin M Granger <py@kmg.fyi>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -13,6 +13,8 @@ Bug Fixes
 * Captured exception variables are now properly scoped.
 * Fixed bugs in `hy.model-patterns.whole` (and other combinators that
   use it, like `brackets`) in which tuples were not always produced.
+* Added pattern matching support to String, Bytes, Symbol, Integer, Float,
+  and Complex models.
 
 1.2.0 ("Crackers and Snacks", released 2026-01-14)
 ======================================================================
--- a/hy/models.py
+++ b/hy/models.py
@@ -149,6 +149,8 @@ class String(Object, str):
       ``#[[hello]]`` is the empty string.
     """
 
+    __match_args__ = ("_as_str",)
+
     def __new__(cls, s=None, brackets=None):
         value = super().__new__(cls, s)
         if brackets is not None and f"]{brackets}]" in value:
@@ -165,6 +167,10 @@ class String(Object, str):
     def __add__(self, other):
         return self.__class__(super().__add__(other))
 
+    @property
+    def _as_str(self):
+        return str(self)
+
 
 _wrappers[str] = String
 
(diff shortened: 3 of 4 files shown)


Kodiologist:
Why only the "primitive" ones?

You don't need to generate a separate test function for every assertion. Just put all the assertions in one function. Also, I don't get why you're checking `(assert (!= ~good-pat ~lit))`. That does not seem to the point.

KevinMGranger:
All of the sequence forms are already pattern-match compatible, nothing to add there.

(Although I haven't really thought about f-strings, because it doesn't really make sense to pattern match those IMO?)

The `(assert (!= ~good-pat ~lit))` is to make sure we didn't accidentally make it so models equal the equivalent values, but I guess we don't need to test that.

I can move it all into one function, but is there a downside to having them be separate functions? I usually try to keep my tests as separate as possible, so errors aren't covered up by other errors.

Kodiologist:
> The `(assert (!= ~good-pat ~lit))` is to make sure we didn't accidentally make it so models equal the equivalent values, but I guess we don't need to test that.

Yes, that's tested elsewhere and is separate from this feature.

> I can move it all into one function, but is there a downside to having them be separate functions?

Grouping assertions into functions is just a bit neater. Hy currently has 627 tests. It would have thousands (and the test files would be a lot longer) if every assertion was its own test. You can also make any macro or helper functions needed for similar assertions local to the test function.

Kodiologist (approved):
Looks good. Remove the added import (which is a no-op). Add yourself to AUTHORS in a separate commit. Then we should be good to go.

PR #2700, opened 2026-04-22
hyc: add -o, -O, and -q flags

antonhibl:
This PR adds three flags that expose parameters `py_compile.compile()` already supports but that `hyc` never surfaced at the CLI:

- `-o`/`--output` — write the `.pyc` to a specific path instead of the default `__pycache__` location. Only valid for a single input file.
- `-O`/`--optimize` — set the bytecode optimization level (0, 1, or 2), mirroring Python's own `-O`/`-OO` interpreter flags. Defaults to `-1` which inherits the calling interpreter's level.
- `-q`/`--quiet` — suppress the `Compiling … --> …` progress message on stderr, useful when driving `hyc` from a build script that handles its own output.

I tried to touch as little as possible; only `hyc_main()` in `cmdline.py` is changed, and all existing `hyc` invocations are unaffected. Tests are in `tests/test_hyc_flags.py` and the full suite passes. NEWS.rst updated, AUTHORS updated as a separate commit per `CONTRIBUTING.rst`.


Kodiologist:
Before I get into this, can I ask what use you've found for `hyc` in practice? I can't remember ever using it myself, and it feels like an artifact.

antonhibl:
@Kodiologist  My background is offensive security and I work in the offensive cybersecurity space professionally. This will be a bit long winded but why Hy specifically is interesting to me is its unique spot in terms of capabilities that many other languages don't have.

To elaborate, Hy occupies a unique position in the Python ecosystem from a deployment perspective. When I compile Hy source to `.pyc` and deploy it to a target system, a defensive analyst reversing that bytecode recovers Python, not the Hy code that I wrote. The source language is pretty much invisible post-compilation because Hy compiles to standard CPython bytecode that is indistinguishable from anything else targeting the same runtime. No other Python-targeting language gives you this cleanly that I have found; they either require shipping source, or they're obscure enough that the bytecode itself becomes a fingerprint to defensive analysts and IDS. Hy produces standard bytecode while the source layer remains private. As a security researcher, this is a technique worth studying and fleshing out in my opinion.

Beyond the compilation story, Hy's macro system adds some architecturally interesting techniques that Python itself can't replicate cleanly. A Hy macro that expands at compile time into a full loader or offensive tooling module that leaves no runtime trace of its origin; meaning, the expanded AST is what gets compiled, not the macro call itself. A defensive analyst will see  the flat expanded Python bytecode with no indication it was generated from a concise macro definition. There's no code generation framework to fingerprint, no template engine artifact, and no wrapper layer. The abstraction is entirely compile-time and disappears completely in the output. This feature is really just a side-effect of this being a Lisp language but I see it as something usable for my security research.

So, these are some of the reasons I want a clean, scriptable workflow from Hy source to deployable bytecode with proper control over output paths, optimization levels, and eventually recursive directory compilation and single-file archive deployment. The flags in this PR are the first step toward that. I have a few follow-on contributions I would like to propose down the line: fixing the `_could_be_hy_src` bug in `importer.py` that I found while working on this (#2701), recursive compilation for `hyc` to allow larger projects consisting of separate .hy files in different sub-dirs of the main project dir (this could eventually even promote Hy packages), and longer term a proper deployment archive story similar to what `python -m compileall` does for Python projects.

I want to be clear, I'm not trying to turn Hy into a malware toolkit. The capabilities I'm describing are consequences of how the language already works, not things I'm trying to add; these capabilities are inherent in the language's design. What I'm contributing is better tooling around a compilation and deployment workflow that already exists. The offensive security use case happens to make that workflow matter more to me, so I'm motivated to do it properly rather than just well enough. I'd like to help guide Hy into more features and a more complete language in the long run.

Also given the offensive operational nature of some of this context I'd rather not go too deep into specifics in a public thread about these offensive capabilities, but I am happy to continue over email if you're open to it.

Kodiologist:
I'm no security expert, but I think you've been too credulous about what Claude has told you regarding Hy's detectability. Hy leaves a lot of distinguishing bits in its generated code. Obscurity isn't a goal of the project, so this is unlikely to change. If you want to write malware that's hard to reverse-engineer, an interpreted language like Python seems a poor choice anyway.

At any, malware is not exactly an argument in favor of the continued existence of ~~`pyc`~~ `hyc`. I should've removed it before the release of Hy 1. Now that we're responsible for backwards compatibility, I'm loath to bump the major version to remove it. As a compromise, let's avoid major feature development on ~~`pyc`~~ `hyc`.

I'll take `--quiet`, as the sort of thing that probably should've been there from the beginning, but I'm turning down the rest of this. For testing, you only need `test_quiet_suppresses_stderr`, not another test for the long name, and it should go in `test_bin.py`. Please avoid LLM fluff like the second argument of `assert` and redundant instructions on how to run the tests.

antonhibl:
I'll own that the assert messages were verbose and ultimately unnecessary, I'll remove them and close the issue. That being said, just because I gave you a long winded explanation does not equate to LLM use. The technical reasoning behind this PR along with the use case(s) I described with my understanding of the compilation pipeline, and the decision to contribute it are mine. Dismissing the substance of the argument because of how it was written isn't really fair in my opinion.

I'd push back on that characterization that pyc is an artifact frankly. The fact that you haven't personally found a use for hyc doesn't mean the use case doesn't exist. Compiling Hy to bytecode for deployment without shipping source is a legitimate workflow that Python itself supports and documents. pyc isn't obscure by some fluke of the Python language devs. It is a legitimate part of CPython's distribution.

You also said you're not a security expert, which I respect, but that cuts both ways. I am, professionally. Whether Hy leaves artifacts in bytecode is a technical question worth actually looking into rather than just making assumptions in my opinion. I'm happy to have that conversation and do most of the work involved.

I am gonna work on fixing the tests with your recommendations. But I don't think calling it malware tooling is a fair name for what I described, and I'd like if you can engage with the actual use case I described rather than the worst faith reading of what I am saying. You asked for what uses I found for Hy, I gave you them; turning around and calling my uses AI slop and malware is disingenuous.

Kodiologist:
> just because I gave you a long winded explanation does not equate to LLM use.

Are you claiming that you didn't use an LLM to write this code?

antonhibl:
@Kodiologist The tests were written by a locally hosted model WhiteRabbitNeo 33B, hence the things wrong you mentioned. The actual additions to hyc and the additional parameters were written by myself.

antonhibl:
@Kodiologist updated with the requested changes, I would like to argue for the -o/--output flag as its really more of a build pipeline improvement for hyc but if you are totally against it then it is what it is, its easy to implement locally for those who want it I guess.

antonhibl:
hold on, I think my vim fucked the formatting up. give me a few.

antonhibl:
@Kodiologist alright fixed the auto-formatting, let me know how this looks now.

Kodiologist:
Use `run_cmd`. Then you don't need to check the return code explicitly.

Use `tests/resources/hello_world.hy` instead of making a new temporary file.

Use a "New Features" section header in ~~the README~~ NEWS.

Otherwise, looks good.

antonhibl:
@Kodiologist should be good now

Kodiologist (approved):
Thanks.

Kodiologist:
Why did you close this and #2702? This is pretty much ready to merge, and I didn't even properly review #2702 yet.

antonhibl:
damn, did not mean to close this. I just was making my fork private since it has some offsec tooling built into one of the branches now. hopefully it still allows you to merge it. @Kodiologist

Kodiologist:
No, GitHub doesn't allow that. It doesn't allow me to reopen your PRs, either.

antonhibl:
dang, I am happy to remake a branch on a fork and link it here, wasn't a huge amount of code. just annoying git auto-closed it.

Kodiologist:
Can you just reopen the PRs that already exist?

antonhibl:
give me a few, I'll get it fixed @Kodiologist

PR #2702, opened 2026-04-22
importer: fix _could_be_hy_src failing on non-existent paths

### LINKED ISSUE #2701: `py_compile.compile` may read Hy as Python
`_could_be_hy_src` in `importer.py` gates on `os.path.isfile(filename)` before checking the extension. This means that if you pass a path that doesn't exist on disk, like an override path for the `.pyc` header, the function returns `False` and the source falls through to CPython's parser, which then fails on Hy syntax.

How to reproduce:

```python
import py_compile
py_compile.compile("hello.hy", dfile="/some/override/path.hy")
```

```
File "/some/override/path.hy", line 1
    (defn greet [name] (print "hello"))
     ^^^^^^^^^^^^^^^^^
SyntaxError: invalid syntax
```

The `os.path.isfile` check makes sense in `_get_code_from_file` where the file is about to be opened and read. In `_hy_source_to_code` the data is already loaded and passed in so I would think only the extension matters. My proposed fix is to split the two concerns so `_hy_source_to_code` checks extension only, while `_get_code_from_file` keeps its existing behavior. I am happy to submit this as a PR but wanted to raise the issue first.

antonhibl:
Fixes #2701.

`_could_be_hy_src` gates on `os.path.isfile` which caused it to return `False` for valid Hy source paths that don't exist on disk, such as `dfile` override paths passed to `py_compile.compile`. This meant the source fell through to CPython's parser which then failed on Hy syntax.

My proposed fix removes the `isfile` check from `_could_be_hy_src` so it checks extension only, and moves the `isfile` check explicitly into `_get_code_from_file` where it is needed before opening the file from disk. All existing behavior is preserved and tests pass.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -112,3 +112,4 @@
 * Zhan Tang <tz59pk@gmail.com>
 * Ati Sharma <ati@agalmic.ltd>
 * Kevin M Granger <py@kmg.fyi>
+* Anton Hibl <antonhibl11@gmail.com>
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -10,6 +10,9 @@ Supports Python 3.x – Python 3.y
 Bug Fixes
 ------------------------------
 
+* Fixed `_could_be_hy_src` in `importer.py` incorrectly returning
+  `False` for valid Hy source paths that do not exist on disk, such
+  as ``dfile`` override paths passed to `py_compile.compile`; Closes #2701.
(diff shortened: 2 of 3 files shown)


antonhibl:
let me know what you think of this proposed fix whenever you get a chance, no rush I know open source devs have lives outside of maintaining projects. @Kodiologist

PR #2703, opened 2026-04-23
Recreating #2700

antonhibl:
Recreating #2700. The original PR was closed when I made my fork private for security reasons; I have offensive security tooling in the same repo I didn't want public. No option to reopen once that happens apparently, I tried to make it public to test but no luck. Changes are identical to the approved PR, sorry for the duplication @Kodiologist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -112,3 +112,4 @@
 * Zhan Tang <tz59pk@gmail.com>
 * Ati Sharma <ati@agalmic.ltd>
 * Kevin M Granger <py@kmg.fyi>
+* Anton Hibl <antonhibl11@gmail.com>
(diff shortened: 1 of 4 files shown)


Kodiologist (approved):

antonhibl:
to be clear, I tried to re-open those PRs and really didn't want to open another PR but git just doesn't seem to allow it even after I tried to make the original fork public again. Just want to say its my bad, I didn't realize it would nuke the PR when I made the fork private, its frustrating because the branches still exist there so the logic of git not being able to re-open it is nonsense but I digress. Regardless, my apologies @Kodiologist 

I am gonna go fix(do the same thing) for the other PR I made now.

PR #2704, opened 2026-04-23
Recreating #2702

### LINKED ISSUE #2701: `py_compile.compile` may read Hy as Python

antonhibl:
Recreating #2702. Original PR was closed when I made my fork private for security reasons with no option to reopen. Changes are identical. Fixes #2701.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -14,6 +14,9 @@ New Features
 Bug Fixes
 ------------------------------
 
+* Fixed `_could_be_hy_src` in `importer.py` incorrectly returning
+  `False` for valid Hy source paths that do not exist on disk, such
+  as ``dfile`` override paths passed to `py_compile.compile`; Closes #2701.
 * Captured exception variables are now properly scoped.
 * Fixed bugs in `hy.model-patterns.whole` (and other combinators that
   use it, like `brackets`) in which tuples were not always produced.
--- a/hy/importer.py
+++ b/hy/importer.py
@@ -107,7 +107,7 @@ _py_source_to_code = importlib.machinery.SourceFileLoader.source_to_code
 
 
 def _could_be_hy_src(filename):
-    return os.path.isfile(filename) and (
+    return (
         os.path.splitext(filename)[1]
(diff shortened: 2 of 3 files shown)


antonhibl:
Ok, all should be fixed, sorry again! @Kodiologist

Kodiologist:
You don't need to ping me. I look at every PR.

Kodiologist:
It's not clear to me that `isfile` is needed in the place where you added it, either. If `isfile(fname)` returns false there, the code will immediately try to `open(fname, "rb")`, which is unlikely to work.

Don't forget a test for the bug.

antonhibl:
newest commit should address your feedback. Let me know if you have any other change suggestions otherwise this should address the bug @Kodiologist

Kodiologist (approved):
Add a comment to the test with the URL of the issue being fixed.

antonhibl:
URL should be added, let me know if I missed anything else

Kodiologist:
Some of your commits have a weird author line (e.g., 13636d95c4ea75612d87c3b693eca934f8301b0c). You may want to check your Git configuration.

PR #2706, opened 2026-04-28
fix: handle None target_module in same-module require check (close #2695)

### LINKED ISSUE #2695: Exception from `importlib` when `eval`ing hy code that uses `require`

## Partial Cause

Jah-yee:
## Summary

Fixes `AttributeError: 'NoneType' object has no attribute 'startswith'` when calling `hy.eval()` with code that uses `require` inside `importlib.eval` contexts.

## Root Cause

When `hy.eval` is called from `importlib.eval` or similar contexts, `target_module` can be `None`, causing `importlib.util.find_spec(None)` to fail with AttributeError.

## Fix

In `hy/macros.py`'s `_same_modules` function, the internal `get_filename` helper now returns `None` when `target_module` is `None`. This allows the `except` block to catch the subsequent `FileNotFoundError` from `os.path.samefile(None, ...)` and correctly return `False` (cannot confirm same module → assume not same module).

## Test

218 passing ✅

Fixes #2695

### DIFF
--- a/hy/macros.py
+++ b/hy/macros.py
@@ -100,6 +100,8 @@ def _same_modules(source_module, target_module):
         return True
 
     def get_filename(module):
+        if module is None:
+            return None
         if inspect.ismodule(module):
             return inspect.getfile(module)
         elif (


Kodiologist:
This doesn't fix the linked issue, and an LLM-generated essay is a poor way to describe a 2-line change.

PR #2708, opened 2026-04-29
Fall back to builtins for target module in derive_target_module

### LINKED ISSUE #2695: Exception from `importlib` when `eval`ing hy code that uses `require`

KevinMGranger:
Fixes #2695 with _100% artisanal, hand-crafted commits made by humans_.

Since you (Kodiologist) identified the fix, and I just commited it, shall I set your name/email as the commit's author?

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -22,6 +22,8 @@ Bug Fixes
   use it, like `brackets`) in which tuples were not always produced.
 * Added pattern matching support to String, Bytes, Symbol, Integer, Float,
   and Complex models.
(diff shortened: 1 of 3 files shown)


Kodiologist:
It doesn't matter very much since it's a tiny change. Anyway, we still need a test and a NEWS line.

KevinMGranger:
Added a test and NEWS line.

Kodiologist:
Thanks. I added comments and a trailing newline.

PR #2710, opened 2026-05-09
Support Python 3.15

Kodiologist:
Unpacking in comprehensions isn't yet fully supported, but I intend to address that in a future PR.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         name-prefix: ['']
         os: [ubuntu-latest]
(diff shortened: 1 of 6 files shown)


scauligi:
Python 3.15 also introduces [lazy import syntax](https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-lazy-imports)! Which is something we could use for our own `__init__.py` (eventually)

scauligi (approved):

scauligi (changes requested):
~~woops comment for wrong PR~~

scauligi (approved):

PR #2711, opened 2026-05-10
More features for Python 3.15

Kodiologist:
This PR depends on #2710.

### DIFF
--- a/hy/core/result_macros.py
+++ b/hy/core/result_macros.py
@@ -2031,8 +2031,13 @@ def compile_require(compiler, expr, root, entries):
     return ret
 
 
-@pattern_macro("import", [many(module_name_pattern + maybe(importlike))])
-def compile_import(compiler, expr, root, entries):
+@pattern_macro("import", [maybe(keepsym(":lazy")), many(module_name_pattern + maybe(importlike))])
+def compile_import(compiler, expr, root, is_lazy, entries):
+
+    if not PY3_15 and is_lazy:
+        hy_compiler._syntax_error(is_lazy, "Lazy imports require Python 3.15 or later")
+    is_lazy = dict(is_lazy = True) if is_lazy else {}
+
     ret = Result()
 
     for entry in entries:
@@ -2061,6 +2066,7 @@ def compile_import(compiler, expr, root, entries):
         ret += node(
             expr,
             names = names,
+            **is_lazy,
             **({} if node is asty.Import else dict(
                 module = module_name
                     if module_name and module_name.strip(".")
--- a/hy/core/result_macros.py
+++ b/hy/core/result_macros.py
@@ -16,7 +16,7 @@ from itertools import dropwhile, zip_longest
 from funcparserlib.parser import finished, forward_decl, many, maybe, oneplus, some
 
 from hy import last_version
-from hy.compat import PY3_11, PY3_12
+from hy.compat import PY3_11, PY3_12, PY3_15
 from hy.compiler import Result, asty, mkexpr
 from hy.errors import HyEvalError, HyInternalError, HyTypeError
 from hy.macros import pattern_macro, require, require_reader, local_macro_name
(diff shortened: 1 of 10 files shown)


hy/core/result_macros.py
```diff
@@ -1990,8 +2014,13 @@ def compile_require(compiler, expr, root, entries):
     return ret
 
 
-@pattern_macro("import", [many(module_name_pattern + maybe(importlike))])
-def compile_import(compiler, expr, root, entries):
+@pattern_macro("import", [maybe(keepsym(":lazy")), many(module_name_pattern + maybe(importlike))])
+def compile_import(compiler, expr, root, is_lazy, entries):
+
+    if not PY3_15 and is_lazy:
+        hy_compiler._syntax_error(is_lazy, "Lazy imports require Python 3.11 or later")
```
scauligi: 3.15?

scauligi (approved):
man the yield-from shim is super neat

hy/core/result_macros.py
Kodiologist: You saw nothing.

Kodiologist:
Yeah, I thought I'd have to write another little loop in there when I remembered that `yield from` does exactly what's needed.

scauligi:
ah, found a discrepancy:

Python 3.15:
```
=> (defn :async foo [] (lfor :async x (ait) #* expr))
async def foo():
    return [*expr async for x in ait()]
```

Python 3.14:
```
=> (defn :async foo [] (lfor :async x (ait) #* expr))
async def foo():

    def _hy_anon_4():
        async for x in ait():
            yield from expr
    return [].__class__(_hy_anon_4())
SyntaxError: 'async for' outside async function
```

This bug probably generalizes to any time we create temp functions for async comprehensions, so could also go into a separate PR?

Kodiologist:
Yep, good find, but it looks to be unrelated. I can get this on `master`:

```
=> (defn :async foo [] (lfor :async x (ait) :do (print "hi") expr))
async def foo():

    def _hy_anon_1():
        async for x in ait():
            print('hi')
            yield expr
    return [].__class__(_hy_anon_1())
None
------------------------------
  File "/usr/lib/python3.13/codeop.py", line 118, in __call__
    codeob = compile(source, filename, symbol, flags, True)
  File "stdin-5a7b3e84fa34900a1264035767872b9171448af4", line 1
SyntaxError: 'async for' outside async function
```

scauligi:
oh, reading the PEP in further detail, looks like there is a difference between yield-from and an inner loop for some semantic edge cases, and so they're explicitly targeting the inner-loop version of the semantics
https://peps.python.org/pep-0798/#alternative-generator-expression-semantics

So for the backwards compat version of comprehension unpacking we'd need to do the same to accurately maintain the correct semantics

Kodiologist:
Well, phooey. Good catch. I guess I'll have to switch to a real loop, and test the edge case where it matters.

Kodiologist:
@scauligi Looks good? I changed the statementization logic in "Support unpacking in comprehension forms", and I added a test as the new commit "Test an edge case for comprehension unpacking".

tests/native_tests/comprehensions.hy
```diff
@@ -114,6 +113,77 @@
...
+
+(defn test-unpack-setx []
+  ; Adapted from https://peps.python.org/pep-0798/#interaction-with-assignment-expressions
+  (do-mac (lfor  statementize [False True]  `(do
+    (setv y None)
+    (setv g (gfor
+      i [0 2 4]
+      ~@(when statementize '[:do (print "hi")])
+      #* (setx y [i (+ i 1)])))
+    (assert (is y None))
```
scauligi: Can you change this to explicitly check that `y` is not bound at all? It might require needing to execute this in a hy.eval to step through top-level forms, but basically I want this to verify that the scoping rules between py>=315 and py<315 match

scauligi (approved):
but yeah other than that one change lgtm!

tests/native_tests/comprehensions.hy
Kodiologist: Done.

